PR #5, opened 2019-06-06
Set version number for initial release

bouweandela:


esmvalcore/_version.py
```diff
@@ -1,2 +1,2 @@
 """ESMValCore version."""
-__version__ = '2.0a2'
+__version__ = '2.0.0a0'
```
mattiarighi: I am bit confused. I thought we would keep the version numbering consistent between tool and core.

bouweandela: Yes, that was the plan, but I've thought some more about it, and it will be a lot of extra work, because that would mean that any time there is an update/release of one of the two, we will also have to make a new release of the the other, even if nothing changed there.

bouweandela: I think wecould probably keep the first two numbers in sync, but better not the last number.

mattiarighi: Sounds like a good solution.

PR #21, opened 2019-06-10
Development fx restructured

valeriupredoi:
Related issue #109
porting ESMValGroup/ESMValTool#1065 to ESMValCore/version2_development


bouweandela:
Can you please merge the `development` branch into this branch, so we can see what's new?

valeriupredoi:
> Can you please merge the `development` branch into this branch, so we can see what's new?

le done :beer:

esmvalcore/_data_finder.py
```diff
@@ -196,16 +193,16 @@ def get_rootpath(rootpath, project):
...
 
 
-def _find_input_dirs(variable, rootpath, drs, fx_var=None):
+def _find_input_dirs(variable, rootpath, drs):
     """Return a the full paths to input directories."""
     project = variable['project']
 
     root = get_rootpath(rootpath, project)
-    input_type = 'input_{}dir'.format('fx_' if fx_var else '')
+    input_type = 'input_dir'
```
bouweandela: it looks like input_type can be removed from the code now, since there is only `'input_dir'` left

esmvalcore/_data_finder.py
```diff
@@ -218,50 +215,35 @@ def _find_input_dirs(variable, rootpath, drs, fx_var=None):
...
     files = find_files(input_dirs, filenames_glob)
 
     return files
 
 
 def get_input_filelist(variable, rootpath, drs):
     """Return the full path to input files."""
+    # change ensemble to fixed r0i0p0 for fx variables
+    if variable['project'] == 'CMIP5'and variable['frequency'] == 'fx':
+        variable['ensemble'] = 'r0i0p0'
```
bouweandela: You're already setting the ensemble in _recipe.py, so I think there is no need to do it again here

bouweandela:
I think we should also get rid of the funky syntax for specifying fx variables. So instead of this (example from recipe_ocean_Landschuetzer2016.yml):
```yaml
diagnostics:
  diag_timeseries_surface_average_vs_OBS:
    variables:
      dpco2:
        mip: Omon
        fx_files: [areacello, ]
      spco2:
        mip: Omon
        fx_files: [areacello, ]
      fgco2:
        mip: Omon
        fx_files: [areacello, ]
```
we should have

```yaml
diagnostics:
  diag_timeseries_surface_average_vs_OBS:
    variables:
      dpco2:
        mip: Omon
      spco2:
        mip: Omon
      fgco2:
        mip: Omon
      areacello:
        mip: Omon
```

esmvalcore/_data_finder.py
valeriupredoi: ermm, I don't think that gets picked up for all the cases, lemme test

valeriupredoi:
yes, I like that and I initially wanted to do it this way but the problem is that the preprocessor masking with fx is way more frequent in a lot of recipes than just using the fx files/variables in the diagnostics only so we need that functionality more stringently - I think we should probably leave it as is now for now, since this way (this PR) mirrors the previous functionality (with the added funky syntax yes) and set this up for a change for v2.1. What say you @mattiarighi @bouweandela @ledm

valeriupredoi:
https://github.com/ESMValGroup/ESMValCore/pull/22 is actually doing more so might wanna look at this instead

bouweandela:
Hi V., I though that the simplification of the variable definition as described in in https://github.com/ESMValGroup/ESMValCore/pull/21#issuecomment-502135816 was the whole point of this pull request. If not, what problem is this actually solving? Is there an associated issue?

valeriupredoi:
@bouweandela (sorry man, I was travelling yesterday, am fully back in work now) - this PR does a few things:
- simplifies the code base by removing the need for dedicated fx file finding functions in `_data_finder.py` and dedicated structures in `config-developer.yml`
- implements the usage for CMIP6 fx files that now have mips that are not just `fx`
- allows for using the fx variable as a stand-alone standard variable
- carries over the functionality for CMIP5 fx files finding and usage

Note that to have the usage of fx variables and files in the preprocessor via `mask_landsea` etc **without** explicitly declaring the fx dictionary `fx_files: {'short_name': 'bla', 'mip': 'blabla'}` is not trivial to implement since `_recipe.py` treats variables on an independent base. #22 is stepping towards that

Alas, there is no issue for this PR but there is one for both this PR and #22 in https://github.com/ESMValGroup/ESMValCore/issues/46

valeriupredoi:
https://github.com/ESMValGroup/ESMValCore/issues/109

esmvalcore/_data_finder.py
bouweandela: Have you had a chance to do this yet?

esmvalcore/_data_finder.py
```diff
@@ -218,50 +214,34 @@ def _find_input_dirs(variable, rootpath, drs, fx_var=None):
...
+    filenames_glob = _get_filenames_glob(variable, drs)
     files = find_files(input_dirs, filenames_glob)
 
     return files
 
 
 def get_input_filelist(variable, rootpath, drs):
     """Return the full path to input files."""
+    # change ensemble to fixed r0i0p0 for fx variables
+    if variable['project'] == 'CMIP5'and variable['frequency'] == 'fx':
```
```suggestion
    if variable['project'] == 'CMIP5' and variable['frequency'] == 'fx':
```

esmvalcore/_recipe.py
```diff
@@ -361,6 +362,45 @@ def _get_default_settings(variable, config_user, derive=False):
     return settings
 
 
+def get_input_fx_filelist(variable, rootpath, drs):
```
bouweandela: I think this function should be removed altogether, it looks like it migrated here from _data_finder.py

bouweandela:
> yes, I like that and I initially wanted to do it this way but the problem is that the preprocessor masking with fx is way more frequent in a lot of recipes than just using the fx files/variables in the diagnostics only so we need that functionality more stringently - I think we should probably leave it as is now for now, since this way (this PR) mirrors the previous functionality (with the added funky syntax yes) and set this up for a change for v2.1. What say you

I would prefer to have this done in this pull request. Because now it looks like a cleanup operation that is only half done, you remove the `get_input_fx_filelist` function from _data_finder.py, only to re-introduce it in _recipe.py.

esmvalcore/_data_finder.py
valeriupredoi: whoops

esmvalcore/_data_finder.py
valeriupredoi: not yet :cry:

esmvalcore/_recipe.py
valeriupredoi: reply below in the main body of the issue

valeriupredoi:
the problem is that we are still keeping the `fx_files` dictionary attached to any variable that needs it or created by the request of fx masking in the preprocessor - you can not remove this function if the `fx_files` is still needed. This function merely creates that dictionary and populates it with key (fx variable) and values (fx files). What you are proposing is, as I said before, nice and straightforward but a nightmare to implement in real life. Have a look at #109 and tell me if you have a good and easy way to couple VAR with FXVAR :beer:

bouweandela:
Which preprocessor function requires a dictionary with fx files? The ones in esmvalcore/prerpocessor/_mask.py only require a list of files, which can simply be returned by `get_input_filelist`

bouweandela:
> Have a look at #109 and tell me if you have a good and easy way to couple VAR with FXVAR

That's another discussion. I've written an answer in the issue.

valeriupredoi:
> Which preprocessor function requires a dictionary with fx files? The ones in esmvalcore/preprocessor/_mask.py only require a list of files, which can simply be returned by `get_input_filelist`

`mask_landsea` and `mask_lanseaice` need specific fx variables (sftlf, sftof, sftgif) and yes, they can be lists of files but they need to be attached to the variable that needs masking which brings us back to the issue in #109

valeriupredoi:
> > Have a look at #109 and tell me if you have a good and easy way to couple VAR with FXVAR
> 
> That's another discussion. I've written an answer in the issue.

yes, I saw it and makes sense but that's not gonna be easy-peasy. Would you rather soend time on that or allow a kludge like this one here. For a nice functional way we should get the independent fx variables as regular variables implementation but note that now, as things stand in the trunk, CMIP6 fx variables dont work **at all** so we need something in the meantime. I am buried under documentation at the moment and I can restart this only next week. Also, have you seen #22 - that fixes all the problems bar removing the fx_files dict, which can probably be done quickly

bouweandela:
> removing the fx_files dict, which can probably be done quickly

If you could do that in this pull request would be great.

> Also, have you seen #22

I believe that pull request includes the changes in this pull request plus extra stuff, right? If so, I think this pull request should be finished before we start on the other one.

valeriupredoi:
> > removing the fx_files dict, which can probably be done quickly
> 
> If you could do that in this pull request would be great.

:camel: :beer:

valeriupredoi:
superseded for now by #170

PR #75, opened 2019-06-11
Do not fail if we get multiple cubes from file

jvegasbsc:
Changed the behaviour to try to select the final cube using the `short_name`. In those cases, ESMValTool will print a couple of warnings detailing the cubes found and the risks of let this kind of errors slip

Should reduce the frustration when using new datasets and the need for fixes, even if fixes can be developed to remove those warnings completely

Replaces ESMValGroup/ESMValTool#1134

### DIFF
--- a/esmvalcore/cmor/fix.py
+++ b/esmvalcore/cmor/fix.py
@@ -7,12 +7,15 @@ variables to be sure that all known errors are
 fixed.
 
 """
+import logging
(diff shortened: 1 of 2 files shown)


bouweandela:
@mattiarighi Would you have time to test and merge this when you're happy?

mattiarighi:
Testing now... :desktop_computer:

PR #77, opened 2019-06-12
Update README.md

bouweandela:


README.md
```diff
@@ -3,36 +3,55 @@
...
 ```
 
+## Installing from source
+If you need the very latest features, please see
+[CONTRIBUTING.md](https://github.com/ESMValGroup/ESMValCore/blob/development/CONTRIBUTING.md)
+for instructions on installing ESMValTool Core from source.
+
+## Using ESMValTool Core as a Python library
+The ESMValTool Core package provides various functions for:
+- Finding data in a directory structure typically used for CMIP data
```
mattiarighi:
```suggestion
- Finding and reading data in a directory structure typically used for CMIP data
```

README.md
```diff
@@ -3,36 +3,55 @@
...
 ## Running ESMValTool
-- Review `config-user.yml`. To customize for your system, create a copy, edit and use the command line option `-c` to instruct `esmvaltool` to use your custom configuration.
-- Optionally install the ESMValTool recipes and diagnostics
-- Run e.g. `esmvaltool -c ~/config-user.yml examples/recipe_python.yml
+- Review `config-user.yml`. To customize for your system, create a copy, edit
+  and use the command line option `-c` to instruct `esmvaltool` to use your
+  custom configuration file.
+- Install the [ESMValTool](https://github.com/ESMValGroup/ESMValTool)
+  to run [ESMValTool recipes and diagnostics](https://esmvaltool.readthedocs.io/en/latest/recipes/index.html)
+- Run e.g. `esmvaltool -c ~/config-user.yml examples/recipe_python.yml` after
```
mattiarighi:
```suggestion
- Run e.g. `esmvaltool -c ~/config-user.yml recipes/examples/recipe_python.yml` after
```

bouweandela: There is no need to add `recipes/`, in fact that will only work if you're in a certain location of the filesystem. `examples/recipe_python.yml` will work regardless of the current working directory

README.md
bouweandela: Do we provide functions for reading data? Which ones?

mattiarighi: The Iris ones.

bouweandela: So iris provides this functionality, right? Not us?

mattiarighi: Well, then we should also remove this sentence "ESMValTool preprocessor functions for e.g. regridding, vertical interpolation, ...", since we use Iris for that.

What I wanted to point out, is that the core is used for finding, reading and preprocess the data.
Reading, in our case, also includes more complex functionalities, like concatenating multiple files and extracting a time range.

bouweandela: > Well, then we should also remove this sentence "ESMValTool preprocessor functions for e.g. regridding, vertical interpolation, ...", since we use Iris for that.

I disagree, our preprocessor functions offer something extra compared to just iris. Indeed they use iris, but they are not only that.

> Reading, in our case, also includes more complex functionalities, like concatenating multiple files and extracting a time range.

I'll add those to the list of preprocessor functions, since that is what they are

mattiarighi: Ok.

mattiarighi:
Are the Codacy issues relevant?

bouweandela:
Not sure, I'll try to fix them

PR #78, opened 2019-06-12
CMOR fixes for CMIP6 models BCC-CSM2 and BCC-ESM1

axel-lauer:
Following the new structure after the split of the repository into ESMValCore / ESMValTool, these 2 CMIP6 fixes have been moved from ESMValTool/version2_emergent_constraints to this branch.


esmvalcore/cmor/_fixes/CMIP6/BCC_ESM1.py
```diff
@@ -0,0 +1,66 @@
+# pylint: disable=invalid-name, no-self-use, too-few-public-methods
+"""Fixes for BCC-ESM1."""
+import numpy as np
+import iris
+
+from ..fix import Fix
+
+
+class allvars(Fix):
```
bouweandela: This looks identical to the class defined in the file above, can you delete it from this file and just `from .BCC_CSM2_MR import allvars`?

esmvalcore/cmor/_fixes/CMIP6/BCC_CSM2_MR.py
```diff
@@ -0,0 +1,66 @@
...
+        Returns
+        -------
+        iris.cube.CubeList
+
+        """
+        for cube in cubes:
+            try:
+                old_time = cube.coord('time')
+            except iris.exceptions.CoordinateNotFoundError:
+                return cubes
```
bouweandela:
```suggestion
                continue
```

esmvalcore/cmor/_fixes/CMIP6/BCC_CSM2_MR.py
```diff
@@ -0,0 +1,66 @@
...
+
+            # if time variable is monotonic, there is nothing to do
+            if old_time.is_monotonic():
+                pass
```
bouweandela:
```suggestion
                continue
```

esmvalcore/cmor/_fixes/CMIP6/BCC_CSM2_MR.py
```diff
@@ -0,0 +1,66 @@
...
+            time_diff = (time_units.num2date(time_data[1]) -
+                         time_units.num2date(time_data[0]))
+            days = time_diff.days
+
+            for idx in idx_zeros:
+                if idx == 0:
+                    continue
+                correct_time = time_units.num2date(time_data[idx - 1])
+                if days <= 31 and days >= 28:  # assume monthly time steps
+                    new_time = \
```
bouweandela: Please try to avoid using backslashes, just make a long line and use `yapf -i` to get the correct formatting automatically.

esmvalcore/cmor/_fixes/CMIP6/BCC_CSM2_MR.py
bouweandela: It would be better to just continue, another cube in the list might contain a time coordinate

esmvalcore/cmor/_fixes/CMIP6/BCC_CSM2_MR.py
bouweandela: [`pass`](https://docs.python.org/3.7/tutorial/controlflow.html#pass-statements) does nothing, I think you mean [`continue`](https://docs.python.org/3.7/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops)

esmvalcore/cmor/_fixes/CMIP6/BCC_ESM1.py
axel-lauer: Changed as suggested. Will test this once the DKRZ machine is back on Monday.

esmvalcore/cmor/_fixes/CMIP6/BCC_CSM2_MR.py
axel-lauer: Changed as suggested.

esmvalcore/cmor/_fixes/CMIP6/BCC_CSM2_MR.py
axel-lauer: Changed to continue.

esmvalcore/cmor/_fixes/CMIP6/BCC_CSM2_MR.py
axel-lauer: Done.

axel-lauer:
Not sure how to implement a unit test for a fix. Any help welcome!

bouweandela:
Here is a good example for a unit test that fixes a coordinate:
https://github.com/ESMValGroup/ESMValCore/blob/125931cac0b2c386d21dc2baba65746323cefd3d/tests/integration/cmor/_fixes/CMIP5/test_CCSM4.py#L1-L33

Does that help? If not, can you provide some example data of the time coordinate before and after fixing? Preferably the data should not be much more than 10 points or so. Then I can put that into a similar test.

axel-lauer:
Fixes don't seem to be needed any more. Current BCC-CSM2 and BCC-ESM1 data do not have the problem of zeros in the time coordinate.

PR #80, opened 2019-06-12
Make fixes lazy

bouweandela:
Unfortunately there are no unit tests, so i cannot be sure that I did not mess anything up, but the fixes are fairly simple, so I think this should be fine.


ledm:
Can't see any reason why these changes could cause trouble. I think they should address some of the memory problems that @valeriupredoi spotted too.

zklaus:
Yeah, I think that's the main motivation. @ledm since I think they are pertaining to some models you use in your diagnostics, could you run one of those just to make sure?

mattiarighi:
I tested the fixes for `msftmyz` using `recipe_ocean_amoc.yml`.
For `thetao` in `MRI-GCM3` I had to create an ad-hoc recipe.

They both run fine and the preproc output is unchanged.

Issue #42, opened 2019-03-08
Do not realize data and use dask in derivation functions


mattiarighi:
As reported by @bouweandela, many derivation functions still realize the data and use `numpy` instead of `dask`. This is detrimental for the performance and should be changed.

Affected variables:
- [x] `amoc`
- [x] `gtfgco2` should not be needed anymore, there is a preprocessor function for this
- [x] `sm`
- [ ] `toz`

ledm:
Similar to this discussion, is there a way to switch off writing the derived variables to disk? It seems to slow everything down and shouldn't be necessary.

bouweandela:
You probably mean the input variables needed to derive a variable? In that case the answer is no.

ledm:
Is it not possible to load the cubes into dask arrays, instead of saving them?

In the case of the derivation of OHC, it loads a 4D variable and saves it exactly as it is. It basically copies 20GB of data into the working directory for each dataset before doing any calculations! All I want is a scalar field, it should be a few kb!

Furthermore, we only have 100GB space in our home directories on jasmin, this means that there's only space for a few models using this method. (I will move my working directory somewhere with more space, but this still doesn't seem like a great method!)

bouweandela:
> Is it not possible to load the cubes into dask arrays, instead of saving them?

Maybe in the future, but not at the moment. Do you feel like implementing this yourself?

> Furthermore, we only have 100GB space in our home directories on jasmin, this means that there's only space for a few models using this method. (I will move my working directory somewhere with more space, but this still doesn't seem like a great method!)

The Jasmin user guide recommends using a group workspace for storing large amounts of data: https://help.jasmin.ac.uk/article/176-storage, not your home directory. I started on pull request https://github.com/ESMValGroup/ESMValCore/pull/265. which will make it possible to store preprocessor and other temporary data on a special temporary file system, but this is not ready yet.

ledm:
Just a comment that  gtfgco2 may still be needed. I've commented on the merged PR here https://github.com/ESMValGroup/ESMValCore/pull/418#issuecomment-635412157

but happy to continue the discussion here if needed.

bouweandela:
Up-to-date overview and discussion in #2451.

PR #83, opened 2019-06-12
Lazy derivation of gtfgco2

bouweandela:
Partially address #42. Please check carefully if results are the same, because there are no unit tests and changes to the code are large.


bouweandela:
This derivation function is no longer needed, as this derived variable can be computed from `fgco2` using the preprocessor:
```
preprocessors:
  global_total:
    area_statistics:
      operator: sum
      fx_files: [areacello]
```

PR #86, opened 2019-06-13
Allow preprocessor to be defined at diagnostic

jvegasbsc:
Just a small usability improvement.

Instead of having to write

```
diagnostic:
  variables:
    tas:
       preprocessor: preproc1
    pr:
       preprocessor: preproc1
    sfcwind:
       preprocessor: preproc1
    tos:
       preprocessor: preproc2
```
now users will be able to do:
```
diagnostic:
  preprocessor: preproc1
  variables:
    tas:
    pr:
    sfcwind:
    tos:
       preprocessor: preproc2
```

### DIFF
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -836,13 +836,15 @@ class Recipe:
 
         for name, raw_diagnostic in raw_diagnostics.items():
             diagnostic = {}
+            default_preproc = raw_diagnostic.get('preprocessor', 'default')
             diagnostic['name'] = name
             diagnostic['preprocessor_output'] = \
                 self._initialize_preprocessor_output(
                     name,
                     raw_diagnostic.get('variables', {}),
                     raw_datasets +
-                    raw_diagnostic.get('additional_datasets', []))
+                    raw_diagnostic.get('additional_datasets', []),
+                    default_preproc)
             variable_names = tuple(raw_diagnostic.get('variables', {}))
             diagnostic['scripts'] = self._initialize_scripts(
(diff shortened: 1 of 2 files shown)


jvegasbsc:
It will simplify a lot the recipe for ESMValGroup/ESMValTool#210

bouweandela:
I'm not sure this is a good idea: I would prefer having one way of doing things, because the recipe format is already confusing to new users as it is and adding all kinds of shortcuts will make it even more difficult to understand.

jvegasbsc:
We are allowing it for a lot of other things. I concede that this can be confusing for new users, but I don know if it will be more confusing that being allowed to do that for some keywords and not for others

bouweandela:
What other things are you thinking of? As far as I'm aware it's only the `additional_datasets` keyword, I just made a proposal to get that feature removed here #92.

jvegasbsc:
You can do it for `mip`, `project` and potentially any other variable that we only use for path generation. The issue there is that the difference between those keywords and things like `preprocessor` or `additional_datasets` is clear for us, I don't know if this will be the case for pure users

bouweandela:
> You can do it for `mip`, `project` and potentially any other variable that we only use for path generation. 

None of those can be defined in the diagnostic section, they can be defined in a variable section or in a  dataset section. but that is because it get actual variables that you do the computations with, you need to make a cross product between the variable and dataset sections anyway.

jvegasbsc:
Ok. My mistake. 

If we are to dump the other case, we can dismiss also this

bouweandela:
> Ok. My mistake.

I guess that really shows how confusing our current format is ;-)

Issue #24, opened 2019-06-06
Zonal and meridional statistics preprocessor


ledm:
This is an extension of Issue #1117 and PR #1123.

The following jobs need to be completed for the documentation paper (Table 1).

- [x] Change `zonal_means` preprocesor function into two functions: `zonal_statistics` and `meridional_statistics`.
- [x] Rename the `mean_type` to `operator`, like in the new `area_statistics`
- [x] Create associated tests
- [x] Add the new preprocessors to the ocean example recipe
  - [x] Correct ocean example recipe with correct `fx_files`.
- [x] Move the `get_iris_analysis_operation` function in a new file, `_shared.py` in the preprocesor folder.
  - Is it sensible to put an artificial list to limit this function?
- [x] Use this new function in the `volume_statistics` function, and  add documentation to the header of that function.
- [x] Add support for fx_files, and use cell volume to weight the means.
- [x] Write tests which include fx_files for area_statistics, volume_statistics, zonal and meridional statistics.

ledm:
One thing I'm working on here is adding `fx_files` to the `recipe_ocean_example.yml` recipe - basically making it accurate! 

When ESMValTool is unable to locate `fx_files`, everything seems to continue as normal. @valeriupredoi, I would expect a fatal error to be given if an FX file was requested but not found. Is that not the case?

valeriupredoi:
if the masking is done in the preprocessor, if fx files are not found, the code defaults to applying Natural Earth masks instead

ledm:
It doesn't work like this in the `area_statistics` & `volume_statistics` functions. They too try to calculate the area or volume fields, but I feel that this is very inaccurate. We'd be better off if it fails when there's no fx_files. After all, area and volume descriptions are a CMIP5/6 requirement, right?

Should the error message be added in the `get_input_fx_filelist` function of `_data_finder.py` or in the `area_statistics` & `volume_statistics` functions?

ledm:
So i guess the problem I'm seeing right now is: Why can't ESMValTool find the FX files anymore? 

Another point is that none of the preprocessor tests include fx files. How can we write tests that include fx files?

ledm:
Help @valeriupredoi! None of my recipes are able to find the fx_files anymore on my local desktop!

Can anyone else test a recipe that requires an fx_file to be loaded? @schlunma - have you had this problem?

valeriupredoi:
[...]

PR #87, opened 2019-06-13
Refactor time operations

jvegasbsc:
Adding other operators (min, max, median) to time, seasonal and year statistics.

Required by ESMValGroup/ESMValTool#212

I keep the seasonal_mean but probably we should prefer seasonal_statistics from now on. How are we going to manage deprecation of preprocessor functions? I think we are going to have more name changes when generalizing some of them


bouweandela:
> I keep the seasonal_mean but probably we should prefer seasonal_statistics from now on. How are we going to manage deprecation of preprocessor functions? I think we are going to have more name changes when generalizing some of them

Up until now we do not deprecate stuff, but just remove it and update all the existing recipes. Diagnostic developers with open pull requests will have to update themselves. For now I think we can keep this way of working. A minor version increase will be required for ESMValTool Core when we do this kind of thing, once we have the first official release out (in alpha versions this kind of thing is still acceptable), I've pinned the ESMValTool installation so it does not go to the next minor version and the old version will keep working.

esmvalcore/preprocessor/_time.py
```diff
@@ -161,6 +161,33 @@ def get_time_weights(cube):
     time_weights = time_thickness * ones
     return time_weights
 
+def time_statistics(cube, operator):
```
bouweandela: I think it would be better to implement this function so it accepts a list of operators instead of just one.

ledm:
One of the tasks in issue #24 is to move the [`get_iris_analysis_operation`](https://github.com/ESMValGroup/ESMValCore/blob/development/esmvalcore/preprocessor/_area.py#L75) function from `_area.py` to a new file, `_shared.py` in the preprocesor folder. 

This function can "Determine the iris analysis operator from a string", [like this](https://github.com/ESMValGroup/ESMValCore/blob/development/esmvalcore/preprocessor/_area.py#L132). It would be good if the time_statistics functions also behaved like this.

esmvalcore/preprocessor/_time.py
jvegasbsc: By the kind of operators we are using, this will mean producing several cubes. Can we mange that at this preprocessor step? I will very glad to have the option to get seasonal mean and median at the same time but I don know if the preprocessor can handle that

jvegasbsc:
>One of the tasks in issue #24 is to move the get_iris_analysis_operation function from _area.py to a new file, _shared.py in the preprocesor folder.
>
>This function can "Determine the iris analysis operator from a string", like this. It would be good if the time_statistics functions also behaved like this.

I didn't know you already have that. I will do it

esmvalcore/preprocessor/_time.py
bouweandela: Yes, it works fine, this is also an option for the mulitmodel_statistics preprocessor function

jvegasbsc: Great! I will check the multimodel module and implement it following that guide

bouweandela: Thinking a bit more about it, it may not be trivial to get this working correctly with the provenance recording..

jvegasbsc: Yes, I felt a bit lost while looking to the code. Maybe we can keep this merge small(er) by making it to support only one operator and then add the others. 

I also added other things related to time operations that I need to simplify seaice drift and other metrics, and don't want to keep them waiting for this fix

esmvalcore/preprocessor/_shared.py
```diff
@@ -0,0 +1,72 @@
...
+    operator: str
+        A named operator.
+
+    Returns
+    -------
+        bool: True if operator support weights, False otherwise
+
+    """
+    operator = operator.lower()
+    if operator in ('mean',):
```
bouweandela: Maybe you could just write
```python
return operator in ('mean',)
```
that would be a bit shorter

esmvalcore/preprocessor/_time.py
```diff
@@ -231,6 +213,108 @@ def spans_three_months(time):
...
+    """
+    Compute daily statistics.
+
+    Chunks time in daily periods and computes statistics over them;
+
+    Parameters
+    ----------
+    cube: iris.cube.Cube
+        input cube.
+
```
bouweandela: Please mention the supported values for operator for all time statistics function in the docstrings

esmvalcore/preprocessor/_time.py
```diff
@@ -231,6 +213,108 @@ def spans_three_months(time):
...
+    """
+    Compute climatologies in the specified granularity.
+
+    Computes daily, monthly, seasonal or yearly climatologies
+
+    Parameters
+    ----------
+    cube: iris.cube.Cube
+        input cube.
+
```
bouweandela: Document all function arguments

bouweandela:
Do all of these time statistics functions support all input data frequencies (provided they are higher than the frequency of the statistic funtion of course)?

jvegasbsc:
> Do all of these time statistics functions support all input data frequencies (provided they are higher than the frequency of the statistic funtion of course)?

Yes, there is nothing that will make them fail for any strange frequency. As long as the time coordinate is correct, they will be applied correctly

esmvalcore/preprocessor/_shared.py
jvegasbsc: Done

esmvalcore/preprocessor/_time.py
jvegasbsc: Done

esmvalcore/preprocessor/_time.py
jvegasbsc: Done

esmvalcore/preprocessor/_time.py
```diff
@@ -319,19 +434,50 @@ def annual_mean(cube, decadal=False):
...
+    return cube.aggregated_by('year', operator)
+
 
-            def get_decade(coord, value):
-                """Callback function to get decades from cube."""
-                date = coord.units.num2date(value)
-                return date.year - date.year % 10
+def decadal_statistics(cube, operator='mean'):
+    """
+    Compute annual or decadal statistics.
```
bouweandela:
```suggestion
    Compute decadal statistics.
```

esmvalcore/preprocessor/_time.py
```diff
@@ -319,19 +434,50 @@ def annual_mean(cube, decadal=False):
...
 
-            iris.coord_categorisation.add_categorised_coord(
-                cube, 'decade', 'time', get_decade)
+    Note that this function does not weight the annual or decadal mean if
```
bouweandela:
```suggestion
    Note that this function does not weight the decadal mean if
```

esmvalcore/preprocessor/_time.py
```diff
@@ -319,19 +434,50 @@ def annual_mean(cube, decadal=False):
...
+    uneven time periods are present. Ie, all data inside the year/decade
```
bouweandela:
```suggestion
    uneven time periods are present. Ie, all data inside the decade
```

esmvalcore/preprocessor/_time.py
```diff
@@ -294,9 +406,9 @@ def regrid_time(cube, frequency):
     return cube
 
 
-def annual_mean(cube, decadal=False):
+def annual_statistics(cube, operator='mean'):
     """
-    Compute annual or decadal means.
+    Compute annual or decadal statistics.
```
bouweandela:
```suggestion
    Compute annual statistics.
```

esmvalcore/preprocessor/_time.py
```diff
@@ -294,9 +406,9 @@ def regrid_time(cube, frequency):
...
 
     Note that this function does not weight the annual or decadal mean if
```
bouweandela:
```suggestion
    Note that this function does not weight the annual mean if
```

esmvalcore/preprocessor/_time.py
```diff
@@ -294,9 +406,9 @@ def regrid_time(cube, frequency):
...
     uneven time periods are present. Ie, all data inside the year/decade
```
bouweandela:
```suggestion
    uneven time periods are present. Ie, all data inside the year
```

bouweandela:
Can you also make the associated pull request in the ESMValTool repository to fix any diagnostics that use the removed preprocessor functions?

jvegasbsc:
>Can you also make the associated pull request in the ESMValTool repository to fix any diagnostics that use the removed preprocessor functions?

Yes, of course. I wanted to have your approval on this because this is quite a substantial change

jvegasbsc:
There is no way to make a single commit with all suggestions?

bouweandela:
> There is no way to make a single commit with all suggestions?

I believe there a button for that, something like add suggestion to batch

> Yes, of course. I wanted to have your approval on this because this is quite a substantial change

It's fine with me if @mattiarighi agrees. My only comment would be that maybe this is quite a large number of time statistics functions that cannot be used in the same preprocessor anyway, so maybe it makes more sense to make it all one function and add one extra argument to select the type of time statistic that you want. But it's up to you.

jvegasbsc:
>  . My only comment would be that maybe this is quite a large number of time statistics functions that cannot be used in the same preprocessor anyway, so maybe it makes more sense to make it all one function and add one extra argument to select the type of time statistic that you want. But it's up to you.

I expect use cases with several ones: compute daily max, average over month or season and compute the climatology. That's the reason I did not consolidate them in only one. Also I expect that we will have to add other statistics in the future (like percentiles) that will make this cases more common

mattiarighi:
Could you also update the documentation?

jvegasbsc:
Doc updated. Working in the pull request to update the metrics.

I would like your opinion on the parameter `period` name for the `climate_statistics` function

jvegasbsc:
Pull request open. One question, in which repository should I update the documentation? It is still duplicated

bouweandela:
> in which repository should I update the documentation? It is still duplicated

Wait for #89 and https://github.com/ESMValGroup/ESMValTool/pull/1157 to be merged.

esmvalcore/preprocessor/_shared.py
```diff
@@ -0,0 +1,71 @@
...
+    ----------
+    operator: str
+        A named operator.
+
+    Returns
+    -------
+        bool: True if operator support weights, False otherwise
+
+    """
+    if operator.lower() in ('mean',):
```
bouweandela: this function can be slimplified to
```suggestion
    return operator.lower() in ('mean',)
```

tests/unit/preprocessor/_time/test_time.py
```diff
@@ -2,19 +2,24 @@
...
 import unittest
 
-import iris
-import iris.coord_categorisation
-import iris.coords
+
 import numpy as np
 import pytest
 from cf_units import Unit
+import iris
```
bouweandela: please keep imports sorted

jvegasbsc: Done

jvegasbsc:
Doc updated

jvegasbsc:
This should be merged together with ESMValGroup/ESMValTool#1189 to adapt the recipes

doc/esmvalcore/preprocessor.rst
```diff
@@ -537,24 +537,47 @@ see also :func:`esmvalcore.preprocessor.multi_model_statistics`.
...
-* ``extract_time``: Extract a time range from an Iris ``cube``.
-* ``extract_season``: Extract only the times that occur within a specific
-  season.
-* ``extract_month``: Extract only the times that occur within a specific month.
-* ``time_average``: Take the weighted average over the time dimension.
-* ``seasonal_mean``: Produces a mean for each season (DJF, MAM, JJA, SON)
-* ``annual_mean``: Produces an annual or decadal mean.
-* ``regrid_time``: Aligns the time axis of each dataset to have common time
+* extract_time_: Extract a time range from a cube.
+* extract_season: Extract only the times that occur within a specific season.
```
JaroCamphuijsen:
```suggestion
* extract_season : Extract only the times that occur within a specific season.
```

esmvalcore/preprocessor/_time.py
```diff
@@ -176,32 +180,71 @@ def get_time_weights(cube):
...
+
+
+def monthly_statistics(cube, operator='mean'):
     """
-    Compute time average.
+    Compute monthly statistics.
 
-    Get the time average over the entire cube. The average is weighted by the
-    bounds of the time coordinate.
+    Chunks time in monthly periods and computes means over them;
```
JaroCamphuijsen:
```suggestion
    Chunks time in monthly periods and computes statistics over them;
```

esmvalcore/preprocessor/_time.py
```diff
@@ -176,32 +180,71 @@ def get_time_weights(cube):
...
 
 
-# get the seasonal mean
-def seasonal_mean(cube):
+def seasonal_statistics(cube, operator='mean'):
     """
-    Compute seasonal means with MEAN.
+    Compute seasonal statistics.
 
     Chunks time in 3-month periods and computes means over them;
```
JaroCamphuijsen:
```suggestion
    Chunks time in 3-month periods and computes statistics over them;
```

esmvalcore/preprocessor/_time.py
```diff
@@ -246,6 +296,213 @@ def spans_three_months(time):
...
+
+    operator: str, optional
+        Select operator to apply.
+        Available operators: 'mean', 'median', 'std_dev', 'variance', 'min',
+        'max'
+
+    Returns
+    -------
+    iris.cube.Cube
+        Annual mean cube
```
JaroCamphuijsen:
```suggestion
        Annual statistics cube
```

esmvalcore/preprocessor/_time.py
```diff
@@ -246,6 +296,213 @@ def spans_three_months(time):
...
+    """
+    # time_weights = get_time_weights(cube)
```
JaroCamphuijsen:
```suggestion
```

esmvalcore/preprocessor/_time.py
```diff
@@ -246,6 +296,213 @@ def spans_three_months(time):
...
+
+    operator: str, optional
+        Select operator to apply.
+        Available operators: 'mean', 'median', 'std_dev', 'variance', 'min',
+        'max'
+
+    Returns
+    -------
+    iris.cube.Cube
+        Annual mean cube
```
JaroCamphuijsen:
```suggestion
        Annual statistics cube
```

esmvalcore/preprocessor/_time.py
```diff
@@ -246,6 +296,213 @@ def spans_three_months(time):
...
+    """
+    # time_weights = get_time_weights(cube)
```
JaroCamphuijsen:
```suggestion
```

esmvalcore/preprocessor/_time.py
```diff
@@ -246,6 +296,213 @@ def spans_three_months(time):
...
+        input cube.
+
+    operator: str, optional
+        Select operator to apply.
+        Available operators: 'mean', 'median', 'std_dev', 'variance', 'min',
+        'max'
+
+    period: str, optional
+        Period to compute the statistic over.
+        Available periods: 'full', 'season', 'seasonal', 'monthly', 'month',
```
JaroCamphuijsen: Maybe it would be better to just support a single name per period. This makes recipes easier to understand and reduces the amount of code.

tests/unit/preprocessor/_time/test_time.py
JaroCamphuijsen: From https://www.python.org/dev/peps/pep-0008/#imports:

> Imports should be grouped in the following order:
> 
> 1. Standard library imports.
> 2. Related third party imports.
> 3. Local application/library specific imports.
> 
> You should put a blank line between each group of imports.

esmvalcore/preprocessor/_time.py
```diff
@@ -246,6 +296,213 @@ def spans_three_months(time):
...
+            cube, 'decade', 'time', get_decade)
+
+    return cube.aggregated_by('decade', operator)
+
+
+def climate_statistics(cube, operator='mean', period='full'):
+    """
+    Compute climate statistics with the specified granularity.
+
+    Computes daily, monthly, seasonal or yearly statistics for the
```
JaroCamphuijsen: This sentence confuses me, below the parameters section states available periods are: full, season, month, day. yet this description states it is also possible to calculate yearly statistics. Also what does "for the full available period" mean?

JaroCamphuijsen:
Could you please address the codacy issues?

bouweandela:
Thanks @JaroCamphuijsen for helping out with code review!

Note that we need to make a new release of ESMValCore after this PR is merged and only then merge the associated PR in the ESMValTool repository.

esmvalcore/preprocessor/_time.py
jvegasbsc: I added the most common ways of naming the frequencies to avoid users searching for the correct spelling

esmvalcore/preprocessor/_time.py
jvegasbsc: Reworked, but is complicated to explain this while being concise. 

>Also what does "for the full available period" mean?
For all the selected data

esmvalcore/preprocessor/_time.py
```diff
@@ -246,6 +296,204 @@ def spans_three_months(time):
...
+
+    return cube.aggregated_by('decade', operator)
+
+
+def climate_statistics(cube, operator='mean', period='full'):
+    """
+    Compute climate statistics with the specified granularity.
+
+    Computes statistics for the whole dataset. It is possible to get them for
+    the full period or with the data griouped by day, month or season
```
JaroCamphuijsen:
```suggestion
    the full period or with the data grouped by day, month or season
```

bouweandela:
Computing the variance results in a crash, because this squares the units, and then the function `cmor_check_data` fails:
```
  File "/home/bandela/src/esmvalcore/esmvalcore/cmor/check.py", line 667, in cmor_check_data
    checker(cube).check_data()
  File "/home/bandela/src/esmvalcore/esmvalcore/cmor/check.py", line 160, in check_data
    self._cube.convert_units(units)
  File "/home/bandela/conda/envs/esmvaltool/lib/python3.7/site-packages/iris/cube.py", line 917, in convert_units
    new_data = self.units.convert(self.data, unit)
  File "/home/bandela/conda/envs/esmvaltool/lib/python3.7/site-packages/cf_units/__init__.py", line 1904, in convert
    (self, other))
ValueError: Unable to convert from 'Unit('K2')' to 'Unit('K')'.
```

esmvalcore/preprocessor/_time.py
```diff
@@ -26,7 +30,7 @@
     filterwarnings(
         'ignore',
         "Collapsing a non-contiguous coordinate. "
-        f"Metadata may not be fully descriptive for '{_coord}'.",
+        f"Metadata may not be fully descriptive for '{0}'.".format(_coord),
```
JaroCamphuijsen:
```suggestion
        "Metadata may not be fully descriptive for '{0}'.".format(_coord),
```

jvegasbsc:
> Computing the variance results in a crash, because this squares the units, and then the function cmor_check_data fails:

This is tricky. We want users to use the preprocessor as much as possible, but we can not add functions that affect the units... 

I think I will remove the variance and open an issue to discuss if we want this kind of functions to be available in the preprocessor and, if that is the case, when and how to apply the checks.

esmvalcore/preprocessor/_time.py
```diff
@@ -246,6 +296,204 @@ def spans_three_months(time):
...
+        if operator_accept_weights(operator):
+            time_weights = get_time_weights(cube)
+            cube = cube.collapsed(
+                'time', operator_method, weights=time_weights
+            )
+        else:
+            cube = cube.collapsed('time', operator_method)
+        return cube
+
+    if period in ['daily', 'day']:
```
bouweandela: It would be better to use the function `_get_period_coord` (defined below) here. Also this fails if the coordinate is already present on the cube.

jvegasbsc: Done!

jvegasbsc:
> Note that we need to make a new release of ESMValCore after this PR is merged and only then merge the associated PR in the ESMValTool repository.

Note that this pull request will make some parts of the paper outdated. Do not merge before we get the zenodo release for the paper

mattiarighi:
Yep, that's why I didn't... :smiley:

bouweandela:
@mattiarighi Zenodo archived versions with DOIs, including a new v2.0.0b1 release for the paper have now been created. Can we continue with testing/merging this?

mattiarighi:
Will do!

mattiarighi:
@ledm should also give his OK before I merge.

ledm:
I've checked out this branch and the related ESMValTool PR, https://github.com/ESMValGroup/ESMValTool/pull/1189.

I ran a couple of the edited recipes, I've also glanced through the code changes. So far, I can't see any reason to delay either PR. Everything looks great. Nice work @jvegasbsc!

PR #96, opened 2019-06-17
Add sispeed derivation

jvegasbsc:
Sispeed is a new variable in CMIP6 but can be computed in CMIP5 datasets using usi and vsi (siu and siv in CMIP6)

### DIFF
--- a/esmvalcore/preprocessor/_derive/sispeed.py
+++ b/esmvalcore/preprocessor/_derive/sispeed.py
@@ -1,6 +1,5 @@
 """Derivation of variable `sispeed`."""
 
-import numpy as np
 from iris import Constraint
 import iris.cube
 from iris.coords import DimCoord
--- a/esmvalcore/preprocessor/_derive/sispeed.py
+++ b/esmvalcore/preprocessor/_derive/sispeed.py
@@ -56,7 +56,4 @@ class DerivedVariable(DerivedVariableBase):
             )
 
         speed = (siu ** 2 + siv ** 2) ** 0.5
-        speed.short_name = 'sispeed'
-        speed.standard_name = 'sea_ice_speed'
-        speed.long_name = 'Sea-ice speed'
         return speed
--- a/esmvalcore/preprocessor/_derive/sispeed.py
+++ b/esmvalcore/preprocessor/_derive/sispeed.py
@@ -1,7 +1,6 @@
 """Derivation of variable `sispeed`."""
 
 from iris import Constraint
-import iris.cube
 from iris.coords import DimCoord
 
 from ._baseclass import DerivedVariableBase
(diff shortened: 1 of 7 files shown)


esmvalcore/preprocessor/_derive/sispeed.py
```diff
@@ -0,0 +1,62 @@
...
+        else:
+            siv.add_aux_coord(
+                siu.coord('latitude'), siu.coord_dims('latitude')
+            )
+            siv.add_aux_coord(
+                siu.coord('longitude'), siu.coord_dims('longitude')
+            )
+
+        speed = (siu ** 2 + siv ** 2) ** 0.5
+        speed.short_name = 'sispeed'
```
bouweandela: There is no need to do this here, this is always set by the main derive function:
https://github.com/ESMValGroup/ESMValCore/blob/125931cac0b2c386d21dc2baba65746323cefd3d/esmvalcore/preprocessor/_derive/__init__.py#L114-L117

jvegasbsc: Removed

bouweandela:
You were probably planning on doing this already, but can you add a unit test?

jvegasbsc:
Test added, with a second (but highly related variable)

Should I open a linked pull request in the tool repository for the modifications in the recipe to test the derivation?

esmvalcore/preprocessor/_derive/sispeed.py
```diff
@@ -0,0 +1,55 @@
...
+            Cube containing sea ice speed.
+
+        """
+        siu = cubes.extract_strict(Constraint(name='sea_ice_x_velocity'))
+
+        siv = cubes.extract_strict(Constraint(name='sea_ice_y_velocity'))
+        siv.remove_coord('latitude')
+        siv.remove_coord('longitude')
+
+        # Models usually store siu and siv in slightly different points
```
bouweandela: Maybe it would be nice to check if they are actually close, to avoid errors?

bouweandela:
> Should I open a linked pull request in the tool repository for the modifications in the recipe to test the derivation?

Yes, that would be nice.

esmvalcore/preprocessor/_derive/sispeed.py
jvegasbsc: Yes, will fix it

jvegasbsc: Ok. I have been thinking in this:

- I will try to do it without doing anything with the coordinates.

- If not, I will do as it is now after a warning, unless the `np.allcose ` for more than 5 degrees fail. Why so big? The real expected values depend on the mesh and the resolution, so it is not easy to get a small estimate that will work well for all cases. I hope the big value will help us if we have other issues with the grid, although more of those will be caught by the checker

PR #97, opened 2019-06-18
Add usi and vsi fix for GFDL

jvegasbsc:
Fixing bad standard_names.

A general question: this is something I can fix on the fly in the checker, but I am a bit worried that this may let pass errors related to bad renaming of files because of a copy command or something like that

I put @axel-lauer on this review because I want his opinion on how frequent these kind of errors are and if it will be worth to put the automatic fix or not


bouweandela:
From my point of view we could always apply the standard_name and long_name from the CMOR table automatically, so long as the short_name is correct. We've seen that there are often mistakes with these additional names, I guess that is because most people don't use them.

esmvalcore/cmor/_fixes/CMIP5/GFDL_ESM2G.py
```diff
@@ -37,7 +37,7 @@ def fix_metadata(self, cubes):
         return cubes
 
 
-class co2(Fix):
+class co2(Fix): # noqa
```
bouweandela: Unrelated question, would it be possible to just use `class Co2(Fix)` etc. everywhere? Variable names are not case sensitive anyway and it would avoid a lot of linter errors that cause confusion for new contributors. I would be happy to make a pull request if you think it's too much work but do not object otherwise.

jvegasbsc: Replied in #101

jvegasbsc:
>From my point of view we could always apply the standard_name and long_name from the CMOR table automatically, so long as the short_name is correct. We've seen that there are often mistakes with these additional names, I guess that is because most people don't use them.

I will create a pull request for that. Fix is still "needed", as I plan to make it show a warning

jvegasbsc:
Updated to new naming convention

Issue #31, opened 2019-06-11
Consider using the intake library


bouweandela:
It could be interesting to replace our home grown esmvaltool/_data_finder.py module by the much more advanced [`intake`](https://intake.readthedocs.io/en/latest/index.html) library and the associated [`intake-iris`](https://github.com/informatics-lab/intake-iris) plugin.

bouweandela:
Interesting project, but it doesn't support iris and custom DRS at the moment, https://intake-esm.readthedocs.io/en/latest/index.html

bouweandela:
Example using intake_iris:
```python
import intake
cat = intake.Catalog("catalog.yaml")
source = cat['CMIP5'].get(short_name='ta', mip='Amon', dataset='MPI-ESM-MR', exp='historical', ensemble='r1i1p1')
cubes = source.to_dask()
```
where catalog.yaml looks like
```yaml
metadata:
  version: 1

sources:

  CMIP5:
    description: ''
    driver: intake_iris.NetCDFSource
    args:
      urlpath: '/home/bandela/esmvaltool_input/{{short_name}}_{{mip}}_{{dataset}}_{{exp}}_{{ensemble}}*.nc'
    parameters:
      dataset:
        description: Dataset
        type: str
      short_name:
        description: Short name
        type: str
      mip:
        description: MIP table
        type: str
      exp:
        description: Experiment
        type: str
      ensemble:
        description: Ensemble
        type: str
```
I'm not sure if this is the right way to use it though. [Structured file paths](https://intake.readthedocs.io/en/latest/making-plugins.html?highlight=path_as_pattern#structured-file-paths) also look useful, but are not supported by the intake_iris plugin.

larsbuntemeyer:
hello, since this issue is still open, i just wanted to let you know (maybe as an update to this issue) that there is also the `intake-esm` plugin build on top of intake that will give you access to filepathes. I have a good experience using it at DKRZ, e.g., there you can get filepathes like this:

```python
import intake

# local catalog at DKRZ (updated daily)
url = '/work/ik1017/Catalogs/mistral-cmip6.json'

dataframe = intake.open_esm_datastore(url)

models = dataframe.search(experiment_id='historical',
                              table_id='6hrLev',
                              variable_id='ta',
                              source_id='MPI-ESM1-2-HR',
                              institution_id='MPI-M',
                              member_id='r10i1p1f1')

filepathes = list(models.df.sort_values(by=['time_range'])['path'])
```

[...]

PR #98, opened 2019-06-19
Enhance error when input files are not found

schlunma:
This addresses an issue of @ledm's presentation at the workshop: Right now, the tool simply outputs

```
No input files found for variable {...}
```

if it does not find input data. To get an idea of what went wrong, user's have to look at the debug log. Also, I think the attribute `filename` in the variable dictionary is very confusing, since it does only indicate the location of the file after preprocessing, which is definitely not useful here.

The main changes are in the following code snippet. The text of the actual error messages themselves is of course open for discussion.

https://github.com/ESMValGroup/ESMValCore/blob/b9d82b05913709d4b322c3684b75818cd0d2e152/esmvalcore/_recipe_checks.py#L95-L110


ledm:
Good work @schlunma. The PR runs as expected! I ran it and it outputs better error messages.

I was thinking that the message:
```
Looked for files matching ['msftmyz_Omon_CanESM25_historical_r1i1p1_*.nc'] in ['/users/modellers/ledm/workspace/ESMValToolTest/KIT_data/']
```
would be better if the filename and directory were combined. Ie:
```
Looked for files matching: '/users/modellers/ledm/workspace/ESMValToolTest/KIT_data/msftmyz_Omon_CanESM25_historical_r1i1p1_*.nc'
```

You could then copy and paste it into the command line and figure out whats the problem. Does that make sense?

schlunma:
> I was thinking that the message:
> 
> ```
> Looked for files matching ['msftmyz_Omon_CanESM25_historical_r1i1p1_*.nc'] in ['/users/modellers/ledm/workspace/ESMValToolTest/KIT_data/']
> ```
> 
> would be better if the filename and directory were combined. Ie:
> 
> ```
> Looked for files matching: '/users/modellers/ledm/workspace/ESMValToolTest/KIT_data/msftmyz_Omon_CanESM25_historical_r1i1p1_*.nc'
> ```
> 
> You could then copy and paste it into the command line and figure out whats the problem. Does that make sense?

Yeah, makes perfectly sense, this is way better :grin: I will change it right now.

esmvalcore/_recipe_checks.py
```diff
@@ -90,10 +91,25 @@ def variable(var, required_keys):
...
+def data_availability(input_files, var, dirnames, filenames):
     """Check if the required input data is available."""
     if not input_files:
-        raise RecipeError("No input files found for variable {}".format(var))
+        var.pop('filename', None)
+        logger.error("No input files found for variable %s", var)
+        if dirnames and filenames:
+            patterns = itertools.product(dirnames, filenames)
+            patterns = [os.path.join(d, f) for (d, f) in patterns]
+            logger.error("Looked for files matching\n%s", "\n".join(patterns))
```
ledm:
```suggestion
            logger.error("Looked for files matching: %s", "\n".join(patterns))
```

esmvalcore/_recipe_checks.py
```diff
@@ -90,10 +91,25 @@ def variable(var, required_keys):
...
+        elif dirnames and not filenames:
+            logger.error(
+                "Looked for files in %s, but did not find any file pattern "
+                "to match against", dirnames)
+        elif filenames and not dirnames:
+            logger.error(
+                "Looked for files matching %s, but did not find any existing "
+                "input directory", filenames)
+        logger.error("Set 'log_level' to 'debug' to get more information")
```
ledm: Is this comment necessary?

schlunma: Yes, I think so. For example. if `dirnames` is an empty list (which happens when the directory does not exist), the user gets no information about the directory ESMValTool was trying to find. Only the debug log shows this.

esmvalcore/_recipe_checks.py
schlunma: I agree that this is nicer if `patterns` only contains one element, but if it contains more it's not really pretty. I change it to respect both cases.

mattiarighi:
Can we merge this?

bouweandela:
No, I want to have a look at the `intake` library (see #31) before doing major work on data_finder.py.

valeriupredoi:
@bouweandela and meself had a discussion at the Workshop about `intake` and we both agreed that it'd be a good thing but not just yet since it'll involve a major overhaul of all the data finding and logging functionalities; so I say let's get this in since it'll be very helpful especially in conunction with the dryrun capability #307

valeriupredoi:
quite a few conflicts tho, wouldn't want to be the one fixing them :rofl:

valeriupredoi:
@schlunma would you be a peach and fix the conflicts pls, then I'll approve and it'll be good to go :beer:

schlunma:
Ahhh, now you were too fast, I wanted to add a test for the `Check.data_availability` feature :grimacing: I also haven't tested it yet, I'm really not sure if that broke anything :smile:

valeriupredoi:
should be fine - can you add the test in a separate PR? That makes more sense since this dealt with printing messages only :beer:

PR #104, opened 2019-06-19
added fix file for UKESM1-00-LL mrsos

valeriupredoi:
solves metadata problem for `standard_name`

### DIFF
--- a/esmvalcore/cmor/_fixes/CMIP6/UKESM1_0_LL.py
+++ b/esmvalcore/cmor/_fixes/CMIP6/UKESM1_0_LL.py
@@ -1,0 +1,28 @@
+# pylint: disable=invalid-name, no-self-use, too-few-public-methods
+"""Fixes for UKESM1-0-LL model."""
+from ..fix import Fix
+
+
+class mrsos(Fix):
+    """Fixes for mrsos."""
+
+    def fix_metadata(self, cube):
+        """
+        Fix metadata.
+
+        Fixes wrong standard name.
+        Original error: mrsos: standard_name should be
+        moisture_content_of_soil_layer,
+        not mass_content_of_water_in_soil_layer.
+
+        Parameters
+        ----------
+        cube: iris.cube.Cube
+
+        Returns
+        -------
+        iris.cube.Cube
+
+        """
+        cube.standard_name = 'moisture_content_of_soil_layer'
+        return cube
--- a/esmvalcore/cmor/_fixes/CMIP6/UKESM1_0_LL.py
+++ b/esmvalcore/cmor/_fixes/CMIP6/UKESM1_0_LL.py
@@ -6,7 +6,7 @@ from ..fix import Fix
 class mrsos(Fix):
     """Fixes for mrsos."""
 
-    def fix_metadata(self, cube):
+    def fix_metadata(self, cubes):
         """
         Fix metadata.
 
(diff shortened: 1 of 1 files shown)


esmvalcore/cmor/_fixes/CMIP6/UKESM1_0_LL.py
```diff
@@ -0,0 +1,29 @@
...
+        Parameters
+        ----------
+        cube: iris.cube.Cube
+
+        Returns
+        -------
+        iris.cube.Cube
+
+        """
+        for cube in cubes:
```
bouweandela: Maybe it would be a bit safer to do something like this:
```python
        for cube in cubes:
            if cube.var_name == 'mrsos':
                cube.standard_name = 'moisture_content_of_soil_layer'
```

jvegasbsc: If you know that it only have one cube, you can even use
```
cubes[0].standard_name = 'moisture_content_of_soil_layer'
```
Fixes are heavily coupled with data, so this kind of shortcuts should not be a problem

valeriupredoi: > Maybe it would be a bit safer to do something like this:
> 
> ```python
>         for cube in cubes:
>             if cube.var_name == 'mrsos':
>                 cube.standard_name = 'moisture_content_of_soil_layer'
> ```

the fix object is for `mrsos` only, man

valeriupredoi: > If you know that it only have one cube, you can even use
> 
> ```
> cubes[0].standard_name = 'moisture_content_of_soil_layer'
> ```
> 
> Fixes are heavily coupled with data, so this kind of shortcuts should not be a problem

I nearly blew to pieces earlier today before I figured out there are more than one cube (and that was not at all obvious) :grin

bouweandela: If you have coordinates that are misinterpreted as cubes, just assigning standard_name to everything is a bit flaky

jvegasbsc: You should only assign the name to one cube, If you are positive there is only one cube at that step, you can use my code, If you know there are more or have doubts, use Bouwe's

valeriupredoi: no, there are 2 cubes and if Im not assigning the correct `standard_name` to both and only to one, the other one will fail the cmor check.

jvegasbsc: Why do you have two cubes in the same file with the same data? Can you share with use the path to the offending file in Jasmin?

valeriupredoi: that's a good question @jvegasbsc - at load point it appears to be a single cube but at fix time there are two cubes:
```
c = iris.load_cube("/badc/cmip6/data/CMIP6/CMIP/MOHC/UKESM1-0-LL/historical/r1i1p1f2/Lmon/mrsos/gn/latest/mrsos_Lmon_UKESM1-0-LL_historical_r1i1p1f2_gn_195001-201412.nc")
```

bouweandela:
@valeriupredoi Any plans to finish this?

bouweandela:
ping @valeriupredoi

valeriupredoi:
hang on @bouweandela lemme test again w/o this

valeriupredoi:
not a problem anymore, tested on 2-Dec-2019 :beer:

PR #105, opened 2019-06-19
Added several fixes and derivations

schlunma:
This PR adds several derivations and fixes for CMIP5 and CMIP6 models (including unit tests).


schlunma:
@bouweandela Can you have a look at that, please?

esmvalcore/cmor/_fixes/shared.py
```diff
@@ -0,0 +1,25 @@
+"""Shared functions for fixes."""
+import iris
+import numpy as np
+from cf_units import Unit
+
+
+def add_height_coord(cube):
+    """Add height coordinate to cube."""
+    height_coord = iris.coords.AuxCoord(2.0,
```
bouweandela: Maybe you can specify the 2m as a function argument? Because there are also other height coordinates (e.g. 10m, 100m) defined in the cmor table, so it could be useful to make this function slightly more generic.

esmvalcore/cmor/_fixes/shared.py
```diff
@@ -0,0 +1,25 @@
...
+    cube.add_aux_coord(height_coord, ())
+    return cube
+
+
+def round_coordinates(cubes, decimals=5):
+    """Round all dimensional coordinates of all cubes."""
+    for cube in cubes:
+        for coord in cube.coords(dim_coords=True):
+            for attr in ('points', 'bounds'):
+                setattr(coord, attr, np.round(getattr(coord, attr), decimals))
```
bouweandela: It would be nicer to keep the data lazy:
```python
            coord.points = da.round(coord.core_points(), decimals))
            coord.bounds = da.round(coord.core_bounds(), decimals))
```

esmvalcore/cmor/_fixes/cmip6/cesm2_waccm.py
```diff
@@ -0,0 +1,23 @@
+"""Fixes for CESM2-WACCM model."""
+from ..fix import Fix
```
bouweandela: This file is identical to the cesm2 module above. You can avoid the code duplication by replacing the code here with something like:
```python
from .cesm2 import Tas as BaseTas

class Tas(BaseTas)
    """Fixes for tas."""
```

It would also be good to avoid the same code duplication in the tests.

tests/unit/preprocessor/_derive/test_rsntcs.py
```diff
@@ -0,0 +1,25 @@
...
+    rsutcs_cube = iris.cube.Cube([[5.0, -1.2], [0.8, -3.0]],
+                                 standard_name=rsutcs_name)
+    ta_cube = iris.cube.Cube([1.0], standard_name='air_temperature')
+    return iris.cube.CubeList([rsdt_cube, rsutcs_cube, ta_cube])
+
+
+def test_rsntcs_calculation(cubes):
+    derived_var = rsntcs.DerivedVariable()
+    out_cube = derived_var.calculate(cubes)
+    assert np.array_equal(out_cube.data, np.array([[-4.0, 3.2], [-0.8, 1.0]]))
```
bouweandela: Maybe you could use a function from [`numpy.testing`](https://docs.scipy.org/doc/numpy/reference/routines.testing.html) instead, I think it will give a nicer error message and you can take into account floating point inaccuracies if needed

esmvalcore/cmor/_fixes/shared.py
schlunma: Done

esmvalcore/cmor/_fixes/shared.py
schlunma: `da.round()` raises `AttributeError: 'numpy.ndarray' object has no attribute 'map_blocks'` in its current version.

esmvalcore/cmor/_fixes/cmip6/cesm2_waccm.py
schlunma: Done, but not for the tests.

tests/unit/preprocessor/_derive/test_rsntcs.py
schlunma: Done

esmvalcore/cmor/_fixes/shared.py
bouweandela: I guess the function expects a dask array, but in your test the data is not lazy to begin with. Maybe you can try
```python
coord.points = da.round(da.asarray(coord.core_points()), decimals)
```
to get it working with both dask and numpy arrays?

schlunma: Done

PR #106, opened 2019-06-20
Change varname to lat and lon for multidimensional grids

### LINKED ISSUE #36: Problem with native grids in `_check_dim_names` for CMIP6 data
The cmor checker complains as follows about some CMIP6 data:
```
Coordinate longitude has var name longitude instead of lon
 Coordinate latitude has var name latitude instead of lat
```
I think this is because it checks whether the `var_name` agrees with the `out_name` of the respective cmip6 table. Unfortunately, this seems not to be in line with the intentions of CMIP6, see [this discussion at the cmip6 tables](https://github.com/PCMDI/cmip6-cmor-tables/issues/215).

So requiring presence and `var_name` for coordinates is too much. I don't know how to solve this generally. Maybe a special exception for spatial coordinates is the best approach?

sloosvel:
Fixes #36


valeriupredoi:
looks good to me @sloosvel - can u pls remove the EC-Earth3 test case?

esmvalcore/cmor/check.py
```diff
@@ -257,12 +257,28 @@ def _check_dim_names(self):
...
                         coord = self._cube.coord(coordinate.standard_name)
-                        self.report_error(
-                            'Coordinate {0} has var name {1} instead of {2}',
-                            coordinate.name,
-                            coord.var_name,
-                            coordinate.out_name,
-                        )
+                        if self._cmor_var.table_type == 'CMIP6' and \
+                           coord.ndim > 1 and \
+                           coord.standard_name in ['latitude', 'longitude']:
```
mattiarighi: Are `latitude` and `longitude` the only possible coordinate names?
I remember there were other cases. @jvegasbsc ?

zklaus: Well, this is about `standard_name`s, and certainly `latitude` and `longitude` are the only correct ones here. According to pure CF 1.7 Sections [4.1](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#latitude-coordinate) and [4.2](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#longitude-coordinate) these standard names are optional, but CMIP 6 sharpens this to a strict requirement in the [model output requirements, p. 6, note 9](https://goo.gl/neswPr) in the present case of 2d coordinate variables.

esmvalcore/cmor/check.py
```diff
@@ -257,12 +257,28 @@ def _check_dim_names(self):
...
+                            self.report_warning(
```
bouweandela: This should not cause a warning: using CMIP6 data that follows the CMOR conventions will now result in a huge number of warnings. A debug level log message seems more appropriate. Also, I think this should only be applied if automatic fixes are enabled.

jvegasbsc: What we do if we do not fix it automatically? It is not incorrect by itself, just inconvenient

zklaus: Since it is really just a change for convenience' sake, with valid data before and after the change, how about just a `report_info` (which admittedly doesn't exist at the moment)?

jvegasbsc: > how about just a report_info (which admittedly doesn't exist at the moment)?

Better

bouweandela: I think debut would be more appropriate, because using info will result in a very large number of log messages, e.g. if you have tens of models with a few variables in a recipe, you will get a hundred or so identical messages printed to your screen.

zklaus: So not only adding a `report_info`, but also a `report_debug` and using the latter here? Sounds good to me.

jvegasbsc: That is done

jvegasbsc:
Updated to last version of development branch

ledm:
So this is approved now, can we merge it?

PR #107, opened 2019-06-20
Determine activities and institutes for CMIP6 loading CV table

### LINKED ISSUE #37: Determine `activity` from `exp`
In CMIP6 the `activity_id` that is part of the path is determined by the `experiment_id`. Hence, I suggest we determine it from that information in a similar way as we already find institutes from `source_id`s, thus making it optional to specify them explicitly in the recipe.

sloosvel:
Fixes #37


esmvalcore/cmor/table.py
```diff
@@ -73,6 +73,8 @@ def __init__(self, cmor_tables_path, default=None):
         cmor_tables_path = self._get_cmor_path(cmor_tables_path)
 
         self._cmor_folder = os.path.join(cmor_tables_path, 'Tables')
+        if 'obs4mips' not in cmor_tables_path:
```
bouweandela: This will not work if another project is created that also has no controlled vocabulary file available. I think it would be better to check if there is a controlled vocabulary file, e.g.
```
if Path(self._cmor_folder).glob('*_CV.json'):
```

esmvalcore/_config.py
```diff
@@ -172,13 +172,22 @@ def get_project_config(project):
 
 
 def get_institutes(variable):
-    """Return the institutes given the dataset name in CMIP5."""
+    """Return the institutes given the dataset name in CMIP5 and CMIP6."""
     dataset = variable['dataset']
     project = variable['project']
     logger.debug("Retrieving institutes for dataset %s", dataset)
+    if project in 'CMIP6':
```
bouweandela:
```suggestion
    if CMOR_TABLES[project].institutes:
```
would work better if we add more cmor tables later on

esmvalcore/_recipe.py
```diff
@@ -903,6 +903,8 @@ def _initialize_variables(self, raw_variable, raw_datasets):
             institute = get_institutes(variable)
             if institute:
                 variable['institute'] = institute
+            if 'CMIP6' in variable['project']:
```
bouweandela: I would make this a bit more general, e.g. similar to how institute is retrieved above: if you can find an activity in the projects cmor table, add it, if not leave it out.

esmvalcore/cmor/table.py
```diff
@@ -141,6 +143,23 @@ def _load_coordinates(self):
...
 
+    def _load_controlled_vocabulary(self):
+        self.activities = {}
+        self.institutes = {}
+        json_file = os.path.join(self._cmor_folder, 'CMIP6_CV.json')
+        with open(json_file) as inf:
+            table_data = json.loads(inf.read())
+
+            exps = table_data['CV']['experiment_id']
+            for exp_id in exps.keys():
```
bouweandela:
```suggestion
            for exp_id in exps:
```

esmvalcore/cmor/table.py
```diff
@@ -141,6 +143,23 @@ def _load_coordinates(self):
...
+                activity = exps[exp_id]['activity_id']
+                self.activities[exp_id] = activity
+
+            sources = table_data['CV']['source_id']
+            for source_id in sources.keys():
```
bouweandela:
```suggestion
            for source_id in sources:
```

esmvalcore/cmor/table.py
sloosvel: obs4mips has a CV file but its contents are different that the CMIP6 CV file. The thing here is that I only want to load it for CMIP6 because everything else that I do later with the activities and institutes would not work for obs4mips.

bouweandela: That is unfortunate. In that case, could you change the code to make it read the institutes and activities if the information is present in the correct format (i.e. as in CMIP6) and just leave it empty if it is not there or not in the correct format? It should be possible to add additional or different tables without modifying the code, as this is actually a feature that we support (you can specify a custom config-developer.yml file in your config-user.yml file, defining additional projects and tables).

esmvalcore/config-developer.yml
```diff
@@ -21,121 +21,12 @@
 CMIP6:
   input_dir:
     default: '/'
-    BADC: '[institute]/[dataset]/[exp]/[ensemble]/[mip]/[short_name]/[grid]/[latestversion]'
+    BADC: '[activity]/[institute]/[dataset]/[exp]/[ensemble]/[mip]/[short_name]/[grid]/[latestversion]'
     DKRZ: '[institute]/[dataset]/[exp]/[ensemble]/[mip]/[short_name]/[grid]/[latestversion]'
```
zklaus: DKRZ follows the same layout (which is, I believe, just the esgf layout), so you can also add `[activity]/` in this line.

Note that both changes require a change to the rootpath in people's `config-user.yml`. Should we advertise this somehow? Maybe prominently in a changelog or release notes or something?

esmvalcore/_config.py
```diff
@@ -172,13 +172,28 @@ def get_project_config(project):
...
+    except (KeyError, AttributeError):
+        pass
     return CFG.get(project, {}).get('institutes', {}).get(dataset, [])
 
 
+def get_activity(variable):
+    """Return the activity given the experiment name in CMIP6."""
+    project = variable['project']
+    try:
+        exp = variable['exp']
```
bouweandela: Note that `'exp'` can also be a list, e.g. `['historical', 'rcp85']` when using CMIP5 data. In that case the data from the two experiments will be concatenated.

jvegasbsc:
Can the people with access to DKRZ to check that the changes in the config-developer are ok for them? I have checked BADC and it works smoothly, allowing us to mix data from CMIP and HighResMIP in the same recipe

esmvalcore/_config.py
```diff
@@ -189,6 +189,11 @@ def get_activity(variable):
     try:
         exp = variable['exp']
         logger.debug("Retrieving activity_id for experiment %s", exp)
+        if isinstance(exp, list):
+            activities = []
+            for value in exp:
+                activities.append(CMOR_TABLES[project].activities[value])
+            return activities
```
zklaus: A bit more concise:
```
return [CMOR_TABLES[project].activities[value] for value in exp]
```

esmvalcore/_config.py
zklaus: This is now addressed in the following lines.

valeriupredoi:
this is very useful, totally missed it until @zklaus pointed me to it offline :beer:

mattiarighi:
@bouweandela can we merge this?
@schlunma needs that quite urgently for the automatisation of the results browser.

Thanks!

mattiarighi:
@schlunma just merge this once tested.

schlunma:
Also, once this is merged the users might need to update their CMIP6 `rootpath` in the `config-user.yml` file. We also need to update the symlink on DKRZ pointing to the CMIP6 archive to one directory above the current one.

jvegasbsc:
>Do we use an old version of the CMIP6 tables? @zklaus @jvegasbsc

Yes, we have #158 to solve this. BCC-CSM2-MR is in that pull request

mattiarighi:
@schlunma can you update and test again after #158?

mattiarighi:
Is there any easy workaround @jvegasbsc?
Otherwise I would suggest to merge this and wait for the next release of the tables.

zklaus:
> Otherwise I would suggest to merge this and wait for the next release of the tables.

I thing that's the right way. That source_id was added on 2019-07-29, a couple more have been added since, but there is currently no release of the tables that contains this information.

jvegasbsc:
> Is there any easy workaround @jvegasbsc?
The easiest one is to add the institute in the recipe. I think is not a good idea to keep an extra list of those missing in the tool, as it will be a nightmare to maintain

PR #108, opened 2019-06-20
Convert units of troubling attributes in time coordinate

### LINKED ISSUE #33: Changed time units break branch_time information
In CMIP6, some(many) experiments start off from initial conditions provided by other experiments. For deck and historical runs, there is a `piControl-spinup` that starts from observations and has arbitrary times that don't correspond to any real time, which is the parent to `piControl`, which is run with constant 1850 forcing, thus again having fictitious time.
From `piControl`, the `historical` simulations are started, different members of the `historical` ensemble starting from different points in the `piControl` simulation. Similar relationships exist for many, perhaps most, other experiments.

It is important to know to which time-span of the `piControl` experiment a given `historical` simulation corresponds to perform correct drift correction and comparisons. This information is recorded in the attributes `branch_time_in_parent`, `branch_time_in_child`, and `parent_time_units`.

By changing the time units we make this information invalid, even misleading. For example the `historical` runs of CESM2 suddenly claim to have started in the year 3797 instead of 1848.

Potential remedies would be one of

- Keep original time units
- Change the attributes
- Provide the information additionally

Note that the problem also exists for CMIP5 data, but in a weaker form since there only the branch time in the parent is recorded.

sloosvel:
Fixes #33


esmvalcore/cmor/check.py
```diff
@@ -402,13 +403,30 @@ def _check_time_coord(self):
...
                 cf_units.Unit(
-                    'days since 1950-1-1 00:00:00',
+                    'days since 1850-1-1 00:00:00',
                     calendar=coord.units.calendar))
             simplified_cal = self._simplify_calendar(coord.units.calendar)
             coord.units = cf_units.Unit(coord.units.origin, simplified_cal)
 
+            attrs = cube.attributes
+            if 'branch_time_in_child' in attrs:
+                attr = 'branch_time_in_child'
```
bouweandela: Maybe you could move this before the if statement, to avoid having `'branch_time_in_child'` twice

esmvalcore/cmor/check.py
```diff
@@ -402,13 +403,30 @@ def _check_time_coord(self):
...
+                value = attrs[attr]
+                cube.attributes[attr] = old_units.convert(value, coord.units)
```
bouweandela:
```suggestion
                attrs[attr] = old_units.convert(attrs[attr], coord.units)
```
would be a bit shorter

mattiarighi:
Let me please test it with at least one recipe first.

bouweandela:
Yes, I think @zklaus should also test it.

mattiarighi:
@sloosvel can you please resolve the conflicts?

@zklaus are you OK with this PR?

esmvalcore/cmor/check.py
```diff
@@ -402,13 +403,27 @@ def _check_time_coord(self):
...
             coord.convert_units(
                 cf_units.Unit(
-                    'days since 1950-1-1 00:00:00',
+                    'days since 1850-1-1 00:00:00',
                     calendar=coord.units.calendar))
             simplified_cal = self._simplify_calendar(coord.units.calendar)
             coord.units = cf_units.Unit(coord.units.origin, simplified_cal)
 
+            attrs = cube.attributes
+            attr = 'branch_time_in_child'
```
zklaus: `attr` is too generic and used for different (albeit related) purposes below.
I think it would be best to simply repeat the string literal the three times it is needed or choose a more meaningful name for the variable, like `branch_time_child = 'branch_time_in_child'`.

esmvalcore/cmor/check.py
```diff
@@ -402,13 +403,27 @@ def _check_time_coord(self):
...
+            if attr in attrs:
+                attrs[attr] = old_units.convert(attrs[attr], coord.units)
+
+            attr = 'parent_time_units'
```
zklaus: See comment above.

esmvalcore/cmor/check.py
```diff
@@ -402,13 +403,27 @@ def _check_time_coord(self):
...
+            if attr in attrs:
+                parent_units = cf_units.Unit(attrs[attr], simplified_cal)
+                attr = 'branch_time_in_parent'
```
zklaus: See previous two comments.

esmvalcore/cmor/check.py
```diff
@@ -402,13 +403,27 @@ def _check_time_coord(self):
...
+                if attr in attrs:
+                    attrs[attr] = parent_units.convert(attrs[attr],
+                                                       coord.units)
```
zklaus: The `parent_time_units` need to be converted to! At the moment the resulting file is wrong.

tests/unit/cmor/test_cmor_check.py
```diff
@@ -339,12 +339,18 @@ def test_bad_time(self):
...
+        self.cube.attributes['parent_time_units'] = 'days since ' \
+                                                    '1850-1-1 00:00:00'
+        self.cube.attributes['branch_time_in_parent'] = 0.
+        self.cube.attributes['branch_time_in_child'] = 0.
         self._check_cube()
         assert (self.cube.coord('time').units.origin ==
-                'days since 1950-1-1 00:00:00')
+                'days since 1850-1-1 00:00:00')
+        assert self.cube.attributes['branch_time_in_parent'] == 0.
+        assert self.cube.attributes['branch_time_in_child'] == 3652.
```
zklaus: Should also check `parent_time_units`.

zklaus:
Note that this only works for CMIP6, but I stand no chance for CMIP5, since there we only have `branch_time` which is in the parents units which are unknown.

So unless somebody has an idea how to fix that, I am ok with having this work for CMIP6 only.

mattiarighi:
> So unless somebody has an idea how to fix that, I am ok with having this work for CMIP6 only.

Which consequences could this have on the diagnostics side?

zklaus:
It would mean that certain forms of drift correction can only be done on CMIP6 data. To the best of my knowledge, at the moment we don't have such automatic drift correction and since I expect new developments will increasingly focus on CMIP6, I think that is where we will see it first.

Of course, in principle it is possible to do this also for CMIP5, but it is a bit more involved. One would have to find the parent dataset, that is look at the CMIP5 global attributes `parent_experiment_id` and `parent_experiment_rip` to figure out the parent netcdf file and look in there for the time units to understand the `branch_time` attribute of the child file. This seems to only make sense if the parent file is also in the recipe, but even then it brakes with the principle of doing these checks one file at a time.

mattiarighi:
Thanks for the explanation :+1:

esmvalcore/cmor/check.py
sloosvel: Ok, I wanted to avoid having long lines but if it's too confusing I'll rename it.

zklaus: You are absolutely right to avoid long lines! But this should not be at the expense of clarity. Rather prefer introducing line breaks in the argument lists of the function calls.

esmvalcore/cmor/check.py
```diff
@@ -436,13 +436,33 @@ def _check_time_coord(self):
...
+                attrs[branch_child] = old_units.convert(attrs[branch_child],
+                                                        coord.units)
+
+            parent_time = 'parent_time_units'
+            if parent_time in attrs:
+                parent_units = cf_units.Unit(attrs[parent_time],
+                                             simplified_cal)
+                attrs[parent_time] = cf_units.Unit(
+                    'days since 1850-1-1 00:00:00',
+                    calendar=coord.units.calendar)
```
zklaus: Did this work for you? I get
```
TypeError: illegal data type for attribute b'parent_time_units', must be one of dict_keys(['S1', 'i1', 'u1', 'i2', 'u2', 'i4', 'u4', 'i8', 'u8', 'f4', 'f8']), got O
```
from the netCDF4 library. I think the attribute needs to be only the string, ie
```
attrs[parent_time] = 'days since 1850-1-1 00:00:00'
```

sloosvel: That's true, sorry I was just checking if the tests were running fine instead of using actual data.

zklaus: Looking good now.

PR #110, opened 2019-06-20
Automatic fixes for standard and long name

jvegasbsc:
As discussed in #97, I added automatic fixes for both standard and long names. 

They both will show a warning if the fix is needed


esmvalcore/cmor/check.py
```diff
@@ -178,9 +178,35 @@ def _check_var_metadata(self):
...
+                if self.automatic_fixes:
+                    self.report_warning(
+                        'Long name for %s changed from %s to %s',
+                        self._cube.var_name,
+                        self._cube.long_name,
+                        self._cmor_var.long_name
+                    )
+                    self._cube.long_name = self._cmor_var.long_name
+                else:
+                    self.report_warning(
```
bouweandela:
```suggestion
                    self.report_error(
```
?

jvegasbsc: Long name are not that used for the purpose of getting the data. People usually rely on standard and var names. I just did not want to add an extra source of errors (specifically because we were not testing it before)

bouweandela: But we're fixing it automatically now, right? So it will never cause an error, unless you disable `fix_metadata` in the recipe.

mattiarighi:
Please resolve the conflicts.

PR #111, opened 2019-06-20
Add support for ERA5

bouweandela:
This pull request
- Adds an OBS6 (CMIP6 + custom) project
- Fixes an issue with the cmor table path, which caused the obs4mips to be not used


bascrezee:
Thanks for this @bouweandela . May I ask for prioritization of this PR? :) I am mainly interested in using the new frequencies (e.g. hourly) that become available as part of OBS6. This is functionality that our project needs right now and it would drastically simplify my workflow if this PR gets finished @hb326

jvegasbsc:
> Adds several variables to the custom table so they can be used with other (hourly) frequencies

This should not be needed. You should be able to just override the frequency in the dataset definiton, at least for not cmor_strict datasets. If that is not the case, it is a bug

bascrezee:
> > Adds several variables to the custom table so they can be used with other (hourly) frequencies
> 
> This should not be needed. You should be able to just override the frequency in the dataset definiton, at least for not cmor_strict datasets. If that is not the case, it is a bug

That is great news. I will try it out soon.

bouweandela:
>> Adds several variables to the custom table so they can be used with other (hourly) frequencies  

> This should not be needed. You should be able to just override the  frequency in the dataset definiton, at least for not cmor_strict  datasets. If that is not the case, it is a bug

@jvegasbsc What about the use of the mip in the filename? Because even if I can override the frequency for the cmor check, if I set the mip to e.g. Amon and the frequency to 1hr, the tool will find the monthly files instead of the hourly files I need.

jvegasbsc:
> @jvegasbsc What about the use of the mip in the filename? Because even if I can override the frequency for the cmor check, if I set the mip to e.g. Amon and the frequency to 1hr, the tool will find the monthly files instead of the hourly files I need.
Ok. Let's be clear:

- Custom tables do not belong to any mip.
- For using variables from the standard CMIP5/6 definitions in other frequencies we are adding pull request #156
- Frequencies can be overriden, and after #156 is merged any mip can be specified

bouweandela:
@mattiarighi @jvegasbsc Would you have time to have a look at this?

bouweandela:
Closing this because it grew into two separate issues.

- OBS6 table is added in #237
- Issue #235 is solved in #236

Issue #101, opened 2019-06-19
Use standard names for fix modules and classes


bouweandela:
At the moment, many of the Python modules and classes in `esmvalcore.cmor._fixes` have names that do not conform to PEP8. Would it be possible to change that? I would be happy to do it if everyone agrees that it's a good idea.

The advantage of using the PEP8 standard in this part of the code too is that:
1) we no longer get unnessary linter errors, which are confusing to new contributors
2) we do not run the risk of creating files with names that only differ in upper/lower case (we have a unit test to check this too, so we're not allowing it anyway), as this causes problems for people that use filesystems that are not case sensitive, see https://github.com/ESMValGroup/ESMValTool/issues/12

bouweandela:
@jvegasbsc Would this be OK with you? Does anyone else object?

jvegasbsc:
I have no trouble with that if we make it clear how we construct the names. We are already changing `-` for `_` in model names

- For models and we can just use lower() after the change I mentioned before. It will be quite straightforward
  
- For variables I see it more difficult, but as short names are always lowercase, I think we can let the fixes be capitalized in any way: (Tos, TOS, AllVars, Allvars) and just complicate a bit the code to get the class so it is case insensitive

PR #112, opened 2019-06-20
Change fixes naming convention

jvegasbsc:
As discussed in #101, I have changed the name convention for the fixes so codacy will throw less errors.

I want to remove all #noqa and similar and create tests for all fixes before merging


esmvalcore/cmor/_fixes/fix.py
```diff
@@ -134,18 +135,22 @@ def get_fixes(project, dataset, variable):
...
         variable = variable.replace('-', '_')
 
         fixes = []
         try:
             fixes_module = importlib.import_module(
                 'esmvalcore.cmor._fixes.{0}.{1}'.format(project, dataset))
+
+            classes = inspect.getmembers(fixes_module, inspect.isclass)
+            classes = dict((name.lower(), value) for name, value in classes)
+            print(classes)
```
bouweandela:
```suggestion
```

jvegasbsc:
I have added a test for `Fix.get_fixes` for each one we have currently implemented

PR #117, opened 2019-06-26
Increase version number to v2.0.0b0 and fix tests

bouweandela:


mattiarighi:
Didn't merge because I was not sure in which order you want to do that.
Just go ahead and merge as required.

PR #155, opened 2019-07-10
Create recipe templates for emac quicklooks

LisaBock:


LisaBock:
I removed _quicklook.py, the corresponding test and the test recipe templates from this branch as @bjoernbroetz will adress these changes in his branch  `dev_append`.

PR #156, opened 2019-07-11
Not cmor strict looks for variable in other tables

### LINKED ISSUE #160: Running with CMIP6 data and OBS data fails if using mip: AERmonZ
...since OBS data is using `cmor.table.CMIP5Info` that doesn't know of `AERmonZ` since that is not a CMIP5 mip

jvegasbsc:
For not CMOR-strict datasets (like OBS) now we are looking onto the custom tables. The idea is to look, before that, in all the other tables in the same request. Solves an issue that @valeriupredoi  has when using AERMonZ tables from CMIP6, but will also allow us to use observational datasets in frequencies not requested by CMIP5 easily

 Fixes #160


valeriupredoi:
ace, @jvegasbsc Mr CMOR :grin: I shall test today

valeriupredoi:
I tested with the recipe that was previously failing and now all works fine, cheers, dude! You got a failing test btw :grin:

jvegasbsc:
Fixed!

bouweandela:
How will this work if the same variable is present in multiple mips, but slightly different?

And what about frequency, if a variable is selected from a mip with a different frequency, won't the cmor check fail because the model data now has a different frequency than what is specified in the cmor table variable definition?

jvegasbsc:
>How will this work if the same variable is present in multiple mips, but slightly different?

It will return the first it encounters. Usually there will be very similar (if not the same, except by the frequency). As I am going through the values in the dict, this may be inconsistent between installations, so we will probably need to fix an order (alphabetical will be enough, I think).  

>And what about frequency, if a variable is selected from a mip with a different frequency, won't the cmor check fail because the model data now has a different frequency than what is specified in the cmor table variable definition?

Frequency is overwritten with the one defined in the mip requested, if exists.
```python
    mip_info = CMOR_TABLES[cmor_table].get_table(mip)
    if mip_info:
        table_entry.frequency = mip_info.frequency
```
Also, you can specify it in the dataset and will overwrite any info you get from the tables

I think this merge request will solve your need for custom variables in #111

jvegasbsc:
By the way, I am moving the code for frequency overwrite to the get variable method to be applied only when getting var info from other tables or the custom one

bouweandela:
>>    How will this work if the same variable is present in multiple mips, but slightly different?

> It will return the first it encounters. Usually there will be very similar (if not the same, except by the frequency). As I am going through the values in the dict, this may be inconsistent between installations, so we will probably need to fix an order (alphabetical will be enough, I think).

Wouldn't it be better to ask the user to specify exactly which table they want to use the variable from if it is defined in more than one table, e.g. with a new key called `source_mip` in the variable section of the recipe?

jvegasbsc:
> Wouldn't it be better to ask the user to specify exactly which table they want to use the variable from if it is defined in more than one table, e.g. with a new key called source_mip in the variable section of the recipe?

It can be done, but it may be a bit of an overkill. Will benefit only cases where the variable is defined multiple times and we are using one that is not one of those. I don't know if the extra complexity is worth it

valeriupredoi:
Agreed with Javi, it's also prone to user errors

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

bouweandela:
> It can be done, but it may be a bit of an overkill. Will benefit only cases where the variable is defined multiple times and we are using one that is not one of those. I don't know if the extra complexity is worth it

Ok, we can leave it until users start complaining.

PR #158, opened 2019-07-11
Update CMIP6 and obs4mips cmor tables to last release

### LINKED ISSUE #45: Outdated CMOR CMIP6 Tables
In the current **version2_development** branch (up to commit d72f6c3915dcc7a55a51b8abae00213a03349659) CMIP6 tables are quite old (about 11 months) and several changes occurred in the data request definition.

I guess it would be good to have a fresh version in the repository and possibly add also a note about the reference Dreq version (likely in cmip6 table README.md).

### LINKED ISSUE #150: CMIP6 tables are out of sync
There is something going on with the CMIP6 tables. The `esmvalcore/cmor/tables/CMIP6/README.md` file claims that it is version 01.00.30, but the actual tables are version 01.00.10, ie quite old and are indeed not suitable for current cmorized data.

For instance, the `standard_name` of (Amon, psl) is still `air_pressure_at_sea_level` where the correct one now is `air_pressure_at_mean_sea_level`.

jvegasbsc:
Updating tables to last release

https://github.com/PCMDI/cmip6-cmor-tables/releases/tag/6.5.29

Applied a fix to Efx tables (generic level missing). It is already on master, but I prefer to use tags for this so it is easier to know the exact version we use

Closes #150 and closes #45 

As a bonus, I added a better error message when trying to get a coordinate that is not defined


zklaus:
Thanks, I think this helps a lot!

There is a problem that we are slowly approaching: CMIP6 data will not follow a uniform `data_specs_version`. This matters because it means that for example data from different models will have different `standard_name`s. For example the two already published datasets
```
{dataset: GISS-E2-1-G, exp: historical, ensemble: r1i1p1f1, mip: Amon, short_name: psl, grid: gn}
{dataset: GFDL-CM4, exp: historical, ensemble: r1i1p1f1, mip: Amon, short_name: psl, grid: gr1}
```
use `standard_name` `air_pressure_at_sea_level` and `air_pressure_at_mean_sea_level`, respectively.

Both of them do so following the `data_specs_version` that they give in the file, namely `01.00.23` and `01.00.27`. But how do you compare this?

The oldest `data_specs_version` I have seen in CMIP6 data is `01.00.23`, but there is no guarantee that there isn't and older one and I have not checked what has changed since then.

I am not sure what, if anything, to do about it right now, but at some point we will get this kind of error.
Broadly speaking there seem to be three approaches:

- Do nothing
- Support multiple data spec versions
- Choose one data spec version and provide "fixes" to align other data with that.

What do you think?

esmvalcore/cmor/tables/cmip6/README.md
```diff
@@ -1,9 +1,9 @@
...
-## Data Request 01.00.30 (March 11, 2019)
+Note:
+----
+Only `CMOR` and `PrePARE` should use the `JSON` file of concatenated `CMIP6` Control Vocabulary [CMIP6_CV.json](https://github.com/PCMDI/cmip6-cmor-tables/blob/master/Tables/CMIP6_CV.json) found in this repository.  
 
-Source is https://github.com/PCMDI/cmip6-cmor-tables
-
-- branch: 01.00.30
-- commit: c9089bf17e81d5dcc41a839e8299cee69da890d3
+All other software should get the CV's directly from the [WCRP](https://github.com/WCRP-CMIP/CMIP6_CVs) repository.
```
zklaus: Only one comment here. We loose the information about which version is in here. Since `README.md` seems to be taken by the tables themselves now, maybe we should add a little file like `VERSION` or similar?

jvegasbsc:
> Do nothing

I think this is not possible
    
> Support multiple data spec versions

This will be a nightmare, not only for us but for users. They will find different CMIP6 specifications in their diagnostic code and will have to adapt to it. 
   
> Choose one data spec version and provide "fixes" to align other data with that.

I prefer this, with one condition: we choose last version released and keep updating. We should assume that changes are done for a reason. Most variables will be the same in all versions anyway, so I think the number of fixes will be manageable. We can add fixes at `PROJECT` level that are applied after `MODEL` fixes to convert data to the latest version of the request. 

In your example, we can add a `PROJECT` fix that changes the `standard_name` for `psl` to the new one automatically if it finds the old one.

zklaus:
Ok, @jvegasbsc I agree with that approach. I opened a new issue for that (#159) so that we can finish this PR.

Thanks!

esmvalcore/cmor/table.py
```diff
@@ -122,7 +122,14 @@ def _assign_dimensions(self, var, generic_levels):
                 coord.generic_level = True
                 coord.axis = 'Z'
             else:
-                coord = self.coords[dimension]
+                try:
+                    coord = self.coords[dimension]
+                except KeyError:
+                    logger.exception(
+                        'Can not find dimension %s for variable %s',
```
bouweandela: It would be nice if you could also mention the mip that the variable is defined in

bouweandela:
@valeriupredoi I believe you worked on this topic while @jvegasbsc was on parental leave. Can you also have a look at this pull request?

bouweandela:
There is also this code
https://github.com/ESMValGroup/ESMValCore/blob/d94e6a7ed6294f16bad24a3d72e4438fd77f46ee/esmvalcore/_recipe.py#L127-L129
which will not work anymore for CMIP6 tables. I'm not entirely sure why it is here anyway, but it seems somehow related to derived variables.

jvegasbsc:
> which will not work anymore for CMIP6 tables. 

Out of the box it will not work, but we are assigning a frequency to the mip by ourselves: the most common in the variables. 

> I'm not entirely sure why it is here anyway, but it seems somehow related to derived variables.

Yes, it is for assigning a frequency for the checks for custom and derived variables

jvegasbsc:
> It looks like frequency is now always defined in the variable and never in the table header. Could you simplify the code so it only reads the frequency from the variable section, i.e. remove the frequency related code from CMIP6Info._load_table and add it in VariableInfo.read_json?

There are other projects that are be using CMOR3 format for their tables (i.e. PRIMAVERA) I don't know if all of them made this change. 

By the way, we are using CMIP5 and CMIP6 as cmor types when we are really refering to CMOR 2 and CMOR 3. I think it is easier for users to match, but the naming is not entirely correct

bouweandela:
>>    I'm not entirely sure why it is here anyway, but it seems somehow related to derived variables.

>Yes, it is for assigning a frequency for the checks for custom and derived variables

Would it be possible to replace
https://github.com/ESMValGroup/ESMValCore/blob/d94e6a7ed6294f16bad24a3d72e4438fd77f46ee/esmvalcore/_recipe.py#L116-L120
https://github.com/ESMValGroup/ESMValCore/blob/d94e6a7ed6294f16bad24a3d72e4438fd77f46ee/esmvalcore/_recipe.py#L127-L129
by something like
```python
    cmor_strict = False if derive else 'default'
    table_entry = CMOR_TABLES[cmor_table].get_variable(mip, short_name, cmor_strict)
```
and change the table readers so this works (i.e. if 'default' is passed it should use whatever is configured in config-developer.yml)? I think it's confusing to change the content of the cmor variable definitions outside of the cmor table module and also using the `custom` or default table outside of the table module seems a bit hacky, because it's not a proper table.

bouweandela:
>>     which will not work anymore for CMIP6 tables.

> Out of the box it will not work, but we are assigning a frequency to the mip by ourselves: the most common in the variables.

I don't see any code that does this, can you point me to it?

jvegasbsc:
> I don't see any code that does this, can you point me to it?

Here: https://github.com/ESMValGroup/ESMValTool/pull/985/files 

I thought this was merged, but it seems it got lost in the split

zklaus:
Notice how that PR actually contains updated tables.

valeriupredoi:
yes, my old PR! And now @mattiarighi is busting my balls about it :grin:

valeriupredoi:
> Notice how that PR actually contains updated tables.

well, they were the latest at their time (the time of the PR), now they've become obsolete as well, I believe the latest CMOR version is 01.00.30 (if not 31beta)

valeriupredoi:
@jvegasbsc are you going to port https://github.com/ESMValGroup/ESMValTool/pull/985 in here?

jvegasbsc:
Yes, I will start working on it now

esmvalcore/cmor/table.py
jvegasbsc: I do not have that info in this method and seems quite ugly to pass extra variables just for error messages

jvegasbsc:
> > > I'm not entirely sure why it is here anyway, but it seems somehow related to derived variables.
> 
> > Yes, it is for assigning a frequency for the checks for custom and derived variables
> 
> Would it be possible to replace
> 
> https://github.com/ESMValGroup/ESMValCore/blob/d94e6a7ed6294f16bad24a3d72e4438fd77f46ee/esmvalcore/_recipe.py#L116-L120
> 
> https://github.com/ESMValGroup/ESMValCore/blob/d94e6a7ed6294f16bad24a3d72e4438fd77f46ee/esmvalcore/_recipe.py#L127-L129
> 
> by something like
> 
> ```python
>     cmor_strict = False if derive else 'default'
>     table_entry = CMOR_TABLES[cmor_table].get_variable(mip, short_name, cmor_strict)
> ```
> 
> and change the table readers so this works (i.e. if 'default' is passed it should use whatever is configured in config-developer.yml)? I think it's confusing to change the content of the cmor variable definitions outside of the cmor table module and also using the `custom` or default table outside of the table module seems a bit hacky, because it's not a proper table.

Will take a lokk

jvegasbsc:
To summarize:

- CMIP6 tables updated to last release
- obs4mips tables updated to master (commit hash added in VERSION file). They do not release very often
- Changes from ESMValGroup/ESMValTool#985 ported
- Updated code following @bouweandela suggestions

esmvalcore/cmor/table.py
```diff
@@ -304,7 +309,18 @@ def __init__(self, table_type, short_name):
 
         self._json_data = None
 
-    def read_json(self, json_data):
+    def copy(self):
+        """
+        Returns a shalow copy of VariableInfo
```
zklaus: Typo: shallow

esmvalcore/cmor/table.py
```diff
@@ -98,22 +92,26 @@ def _load_table(self, json_file):
...
-                    var.frequency = var_data['frequency']
-                else:
-                    var.frequency = table.frequency
-                var.read_json(var_data)
+                var.read_json(var_data, table.frequency)
                 self._assign_dimensions(var, generic_levels)
                 table[var_name] = var
+                self.var_to_freq[table.name][var_name] = var.frequency
+
+            if not table.frequency:
```
zklaus: Probably this should be `table.frequency is ''`.

esmvalcore/cmor/table.py
```diff
@@ -562,9 +585,16 @@ def get_variable(self, table, short_name):
 
         """
         var_info = self.tables.get(table, {}).get(short_name, None)
-        if not var_info and self.default:
-            return self.default.get_variable(table, short_name)
-        return var_info
+        if var_info:
```
zklaus: `if var_info is not None:`

esmvalcore/cmor/table.py
valeriupredoi: no, the check on the existence of `frequency` is needed here since it's either a valid value of `None` (from what I remember)

mattiarighi:
Please do not merge this until is tested with several recipes.
The last tables update gave us lots of headaches...

esmvalcore/cmor/table.py
zklaus: Sorry, @valeriupredoi, I don't understand. This is not a test for existence. It checks the truthiness of `table.frequency`. What do you mean? A valid value of `None` or what?

esmvalcore/cmor/check.py
```diff
@@ -64,7 +64,7 @@ def __init__(self,
         self._errors = list()
         self._warnings = list()
         self._cmor_var = var_info
-        if frequency is None:
+        if not frequency:
```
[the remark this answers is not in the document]
jvegasbsc: To include also frequency being an empty string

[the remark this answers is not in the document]
zklaus: I suggest the more explicit `frequency in [None, '']` then. This is less prone to future breakage and clearer to read.

Perhaps an even better way to address it would be to standardize on *one* singleton, either `None` or `''`, to mark absence consistently, but I am not sure how far reaching and feasible that would be.

zklaus:
Re Mattias concern: In principle, the headaches should be reduced this time round simply because of the split. No change to the master branch here impacts recipe development, so I wonder if we should not rather get this merged relatively quickly, but make sure that things work or at least that there is a clear way to adjust existing recipes, only before we produce the next (pre-)release?

Also, there are very few CMIP6 recipes in the public part of ESMValTool at the moment. Which recipes do you think should be checked?

zklaus:
> > make sure that things work or at least that there is a clear way to adjust existing recipes?
> 
> That's what I meant, just run a few recipes to make sure nothing gets broken.

Hah, what a sneaky way to quote :) The important bit for me was of course
>> only before a release.

> We should make sure that the development branch is always stable, independently of the official releases.

Sure, but what does stable mean? We will want to introduce changes that break existing recipes. That's ok and should not in and of itself delay a release of the core, nor should the core be considered unstable because of it. Rather the ESMValTool should be judicious in adjusting the dependency on a new version of the core.

> > Also, there are very few CMIP6 recipes in the public part of ESMValTool at the moment. Which recipes do you think should be checked?
> Some of the ones using `obs4mips` (I can do that with perfmetrics), since last time there were issues with this project, and maybe one or two with CMIP6.
> 
> So just give me your OK when this is ready for testing.
> It won't take long...

Great, that should definitely work for this PR. :+1:

mattiarighi:
> We will want to introduce changes that break existing recipes. 

Yes, but not at this stage.
A lot of people are working hard on their diagnostics to finalize the papers and we should avoid introducing changes that could break existing recipes and delay the diagnostic development.

But as I said, the usual testing procedure plus a couple of more recipes with CMIP6 models should be enough.

esmvalcore/cmor/check.py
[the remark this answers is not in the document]
jvegasbsc: From PEP8:

> For sequences, (strings, lists, tuples), use the fact that empty sequences are false.
Yes: if not seq:
     if seq:
No:  if len(seq):
     if not len(seq):

We are doing this several times across the whole tool

zklaus:
The flip side of course is that we surely won't want to release with `01.00.10` tables, so getting this in quickly will allow people to develop there recipes and diagnostics with a version of the tables that is close to the one used in all their data instead of trying to bend the data to a long out-of-date version of the standards.

But yeah, point taken, smooth development experience is important now.

esmvalcore/cmor/check.py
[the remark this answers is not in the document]
zklaus: The goal here is not to check the length of a sequence, but to compare with exactly two singletons. The fact that one of them is a sequence is purely coincidental, so imho the section you quote is not applicable here.

The risk I see for the future is twofold: At some point someone might see this code together with the `None` default argument a few lines above and decide to refactor it into `if frequency is None:`. The other thing that is a possible different value that evaluates to false, perhaps an empty list or some form of numerical zero in a new table?

Having said all that, I did approve this PR, so take it as a recommendation and feel free to ignore it if you really come down on the other side of the fence.

jvegasbsc:
Any update on this?

esmvalcore/cmor/table.py
valeriupredoi: it works like this, no need to double check

esmvalcore/cmor/table.py
```diff
@@ -297,7 +309,18 @@ def __init__(self, table_type, short_name):
...
+    def copy(self):
+        """
+        Returns a shalow copy of VariableInfo
+        Returns
+        -------
+        VariableInfo
+           Shallow copy of this object
+        """
+        from copy import copy
+        return copy(self)
```
valeriupredoi: do you really need to create a function that returns a copy, can just call `copy` on the fly

jvegasbsc: Just for convenience, in case we need something special later (or to avoid writing extra imports, I am that lazy)

valeriupredoi: +1 for the reply :grin:

zklaus:
> Any update on this?

Would like to see this go in. At the moment though, there are a bunch of open conversations that should be resolved one way or another first.

esmvalcore/cmor/table.py
zklaus: Any particular reason not to put the import at the top of the module?

esmvalcore/cmor/table.py
zklaus: There is no double check. What do you mean?

esmvalcore/cmor/table.py
jvegasbsc: Now all tables log the exact table where the error happens if an exception is raised and then raise again.
I think is better than logging that at this stage, as will benefit all possible errors fired when reading the tables

jvegasbsc: Now all tables log the exact table where the error happens if an exception is raised and then raise again.
I think is better than logging that at this stage, as will benefit all possible errors fired when reading the tables

jvegasbsc:
Ready to merge from my side

mattiarighi:
@schlunma can you test it with one of your cmip6 recipes?

esmvalcore/cmor/table.py
zklaus: @valeriupredoi care to answer and/or resolve (in the github sense) this conversation?

esmvalcore/cmor/table.py
zklaus: @valeriupredoi ditto?

valeriupredoi:
On my way to work, shall do first thing

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

esmvalcore/cmor/check.py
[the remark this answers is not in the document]
zklaus: @jvegasbsc guess you want to keep it that way. Resolving as a matter of procedure.

esmvalcore/cmor/table.py
jvegasbsc: Chanegd

Issue #151, opened 2019-07-09
Error while regridding sic for BCC model


nperezzanon:
Hi!
I am trying to regrid 'sic' variable to 1x1 in order to have a regular
grid with longitudes and latitudes, so the same script can work with
atmospheric or ice variables.

I did it in the past and it worked with BCC data. Now, I found the
following errors:

         Cell methods:
                 mean: time, {'target_grid': '1x1', 'scheme': 'nearest'})
        ...
          File
"/home/users/nperez/miniconda3/envs/esmvaltool/lib/python3.7/multiprocessing/pool.py",
line 657, in get
    raise self._value
        TypeError: 'NoneType' object is not subscriptable

If I remove the regrid option, I found different error

         Cell methods:
                  mean: time, {'start_longitude': 0, 'end_longitude':
360, 'start_latitude': 50, 'end_latitude': 90})
        ....

          File
"/home/users/nperez/miniconda3/envs/esmvaltool/lib/python3.7/multiprocessing/pool.py",
line 657, in get
            raise self._value
        ValueError: The truth value of an array with more than one
element is ambiguous. Use a.any() or a.all()

The data is on Jasmin. Using CNRM-CM5 data this error doesn't appear, so I keep testing with CNRM-CM5 data.

[main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/3373149/main_log_debug.txt)
[recipe_modes_of_variability.txt](https://github.com/ESMValGroup/ESMValCore/files/3373150/recipe_modes_of_variability.txt)

Cheers, Núria

bouweandela:
It looks like it's failing because the `latitude` and `longitude` coordinates do not have bounds:
```
$ ncdump -h /home/bandela/esmvaltool_input/sic_OImon_bcc-csm1-1_historical_r1i1p1_185001-201212.nc
netcdf sic_OImon_bcc-csm1-1_historical_r1i1p1_185001-201212 {
dimensions:
	time = UNLIMITED ; // (1956 currently)
	rlat = 232 ;
	rlon = 360 ;
	bnds = 2 ;
variables:
	double time(time) ;
		time:bounds = "time_bnds" ;
		time:units = "days since 1850-01-01" ;
		time:calendar = "noleap" ;
		time:axis = "T" ;
		time:long_name = "time" ;
		time:standard_name = "time" ;
	double time_bnds(time, bnds) ;
	double rlat(rlat) ;
[...]

PR #166, opened 2019-07-15
Added warning and automatic fix for no bounds

jvegasbsc:
As discussed in #151


bouweandela:
Can you please fix the merge conflict?

jvegasbsc:
Done!

mattiarighi:
I would suggest @nperezzanon to test it, since she reported the issue in #151.

mattiarighi:
@nperezzanon can you please test this so it can be merged?

bouweandela:
@mattiarighi Can you have a look at this please? Or maybe @axel-lauer can test, because this is a feature he requested?

mattiarighi:
@axel-lauer any use case for testing this?

zklaus:
Just got aware of this via @jhardenberg's mention in #470. Every time this is actually used it masks a serious data error. What's more, `guess_bounds` is unreliable and *will* lead to wrong bounds for example for staggered grids.

jvegasbsc:
>  Every time this is actually used it masks a serious data error. What's more, guess_bounds is unreliable and will lead to wrong bounds for example for staggered grids.

I agree with you that this can lead to errors, but:

- In regular grids, `guess_bounds` is usually a sound approximation and it does not work in any other case

- We are adding a warning in any case, so users should be aware that something may be not totally correct

- Don't think too much about our computations in staggered grids: i.e we are using the same areacello for our computations on thetao, uo, and vo and we know that it should not be the same, at least for EC-Earth and other NEMO-based models

zklaus:
* `guess_bounds` guesses that the points are in the middle of the cells. If the points are on the faces, edges, or corners of cells, as they can be in staggered, regular grids, it effectively shifts and/or distorts the cells.
* Nobody reads the warnings because you get an impenetrable wall of text
* I am not sure how that relates: In staggered grids the cells are the same also for points that live on faces, edges, and corners, consequently the cell areas *are* the same. Regardless, are you saying somebody is making an error without caring about it so nobody should care about it?

Regarding usability vs compliance, the case doesn't seem so clear. I absolutely think this can be fixed, so high usability from a user point of view, but imho really only on a case by case basis.

bouweandela:
Maybe we can revisit this after #374 is merged, because then we could only apply the guess bounds when using reduced cmor check strictness?

Issue #164, opened 2019-07-15
Create alias info for each dataset


jvegasbsc:
There is some code that ends in most of my diagnostics: the composition of an alias for each dataset. My idea is to add this to ESMValCore so it is passed as part of the metadata to the diagnostics so it is consistent between diagnostics.

How I plan to compose it?

`PROJECT_DATASET_EXP_ENSEMBLE' 

but if all datasets share any of the attributes it is omitted to simplify alias (no need to show dataset and project when we are trying to compare between experiments of the same model)

I also plan to make it work hierarchically: 
```yaml
- {project: CMIP5, model: EC-EARTH, exp: historical, ensemble: r1i1p1}
- {project: CMIP6, model: EC-EARTH, exp: historical, ensemble: r1i1p1f1}
```
will yield alias [CMIP5, CMIP6]

```yaml
- {project: CMIP5, model: EC-EARTH, exp: historical, ensemble: r1i1p1}
- {project: CMIP6, model: EC-EARTH, exp: historical, ensemble: r1i1p1f1}
- {project: CMIP6, model: EC-EARTH, exp: historical, ensemble: r2i1p1f1}
```
and this will yield alias [CMIP5, CMIP6_r1i1p1f1, CMIP6_r2i1p1f1}]

Users will be able to overwrite the alias, as with any other dataset info

Any suggestions?

jvegasbsc:
I forgot to ask you if we will want an option to always show the full alias in the config_user

PR #167, opened 2019-07-15
Automatic definition of alias

jvegasbsc:
Automatic composition of a unique alias for each dataset used by a diagnostic. Should help diagnostics developer, as it can be used as a dictionary key, plot label. Will also help with the compatibility of the diagnostics with different datasets and uses, i.e comparing different datasets or comparing different ensemble members of the same dataset can be done with the same code

Now it takes into account project, dataset, experiment, ensemble and version to compose the alias, but can be extended easily by adding keys to a tuple. Does not fail if any of those keys is missing.

If user adds a custom alias, it is respected but the dataset is taken into account for the purpose of getting the automatic for the others

See #164

### DIFF
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -787,7 +787,7 @@ class Recipe:
     """Recipe object."""
 
     info_keys = ('project', 'dataset', 'exp', 'ensemble', 'version')
-    """List of keys to be used to compose the alias, ordered by priority"""
+    """List of keys to be used to compose the alias, ordered by priority."""
 
     def __init__(self,
                  raw_recipe,
@@ -948,7 +948,8 @@ class Recipe:
         return preprocessor_output
 
     def _set_alias(self, preprocessor_output):
-        """Add unique alias for datasets
+        """
+        Add unique alias for datasets.
 
         Generates a unique alias for each dataset that will be shared by all
         variables. Tries to make it as small as possible to make it useful for
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -786,6 +786,8 @@ def _get_preprocessor_task(variables, profiles, config_user, task_name):
 class Recipe:
     """Recipe object."""
 
+    info_keys = ('project', 'dataset', 'exp', 'ensemble', 'version')
+
     def __init__(self,
                  raw_recipe,
                  config_user,
@@ -945,11 +947,10 @@ class Recipe:
         return preprocessor_output
 
     def _set_alias(self, preprocessor_output):
-        info_keys = ('project', 'dataset', 'exp', 'ensemble', 'version')
         datasets_info = set()
         for variable in preprocessor_output.values():
             for dataset in variable:
-                alias = tuple(dataset.get(key, None) for key in info_keys)
+                alias = tuple(dataset.get(key, None) for key in self.info_keys)
                 datasets_info.add(alias)
                 if 'alias' not in dataset:
                     dataset['alias'] = alias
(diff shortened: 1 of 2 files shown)


bouweandela:
I think functionality like this would be more at home in `esmvaltool.diag_scripts.shared`, e.g. imported from the file [esmvaltool/diag_scripts/shared/_base.py](https://github.com/ESMValGroup/ESMValTool/blob/version2_development/esmvaltool/diag_scripts/shared/_base.py). It seems useful with functions like `get_plot_filename` and `get_diagnostic_filename`, which are located in the same module.

At the workshop in June @zklaus asked if some or all of the functionality there could be made part of the esmvalcore package. While I think this is something we could do in the future, it would mean that the core package supports Python much better than other languages, because I think it's not a good idea to start adding code in many different languages to the core package: it should stay as simple as possible to keep it maintainable.

jvegasbsc:
> I think functionality like this would be more at home in esmvaltool.diag_scripts.shared, e.g. imported from the file esmvaltool/diag_scripts/shared/_base.py. It seems useful with functions like get_plot_filename and get_diagnostic_filename, which are located in the same module.

I also thought about this, but in that case it will be available only for python diagnostics. I think it make sense to compose the alias in the same way for all the languages. We are already adding some extra keys to the dataset info, so I think it makes sense to add this at this level.

esmvalcore/_recipe.py
```diff
@@ -940,8 +942,57 @@ def _initialize_preprocessor_output(self, diagnostic_name, raw_variables,
             preprocessor_output[variable_group] = \
                 self._initialize_variables(raw_variable, raw_datasets)
 
+        self._set_alias(preprocessor_output)
+
         return preprocessor_output
 
+    def _set_alias(self, preprocessor_output):
+        datasets_info = set()
```
bouweandela: Can you add a short docstring describing what the function does?

zklaus:
It would be nice to add some documentation about this feature to the diagnostic developer documentation. If I got the gist right, this is intended to provide

- a label for plots
- possibly a component of filenames for data, right?

jvegasbsc:
> It would be nice to add some documentation about this feature to the diagnostic developer documentation. If I got the gist right, this is intended to provide

Yes, I will do that

 >   a label for plots
  >  possibly a component of filenames for data, right?

Yes, that's the idea. Also a key for dictionaries, a quick way to get other variables from the same dataset...

zklaus:
Great! So maybe it will be a family of things? I mean, eg in filenames you do not want spaces, in labels you absolutely want them, right?

jvegasbsc:
> Great! So maybe it will be a family of things? I mean, eg in filenames you do not want spaces, in labels you absolutely want them, right?

I would create only one for simplicity. If you want spaces instead of underscores in your legends is just using `alias.replace('_', ' ')`

zklaus:
Oh, the spaces were only meant as an example. Though I have to admit that the other typical issues are maybe less critical for CMIP6 because the standards and/or esgf might already limit what can happen. But generally speaking, problems one would expect stem from a reduced character set for filenames (no unicode, no maths symbols, maybe no dots (.), ...) compared to what you would like to see in a label, and perhaps filename conventions. For the latter, in CMIP at least, underscore separates components of the filename whereas separation within one component is denoted by hyphens. Now, is the label one component or many?

jvegasbsc:
The alias is a single string. It is composed from (project, dataset, experiment, ensemble, version) to be the smallest unique string that can be composed from those keys. They are composed using underscores to separate the relevant key values

```yaml
        - {project: CMIP5, dataset EC-Earth, ensemble: r1i1p1}
        - {project: CMIP6, dataset EC-Earth, ensemble: r1i1p1f1}
```
        will generate alias 'CMIP5' and 'CMIP6'
```yaml
        - {project: CMIP5, dataset EC-Earth, experiment: historical}
        - {project: CMIP5, dataset MPI-ESM, experiment: piControl}
```
        will generate alias 'EC-Earth,' and 'MPI-ESM'

```yaml
        - {project: CMIP5, dataset EC-Earth, experiment: historical}
        - {project: CMIP5, dataset EC-Earth, experiment: piControl}
```
        will generate alias 'historical' and 'piControl'
```yaml
        - {project: CMIP5, dataset EC-Earth, experiment: historical}
        - {project: CMIP6, dataset EC-Earth, experiment: historical}
        - {project: CMIP5, dataset MPI-ESM, experiment: historical}
        - {project: CMIP6, dataset MPI-ESM experiment: historical}
```
        will generate alias 'CMIP5_EC-EARTH', 'CMIP6_EC-EARTH', 'CMIP5_MPI-ESM'
        and 'CMIP6_MPI-ESM'

```yaml
        - {project: CMIP5, dataset EC-Earth, experiment: historical}
```
        will generate alias 'EC-Earth'

Issue #145, opened 2019-07-05
[EMAC] infrastructure for monitoring


axel-lauer:
Create infrastructure needed to use the ESMValTool as a monitoring tool for running model simulations.

**Tasks**
- [x] extend user-config.yml with quicklook entry including -> PR #154 
  * output_dir: quicklook output directory (concatenated netCDF files + plots)
  * recipe_dir: directory containing recipe template(s)
  * recipes: list of recipe(s) to be processed
- [x] create template recipe(s) to be used for monitoring (e.g. atmos_dynamics, atmos_chem, forcings)
- [x] create function to fill templates for given dataset (i.e. simulation) and time period
- [x] logging capability for all years of a simulation per recipe already processed
- [x] add possibility to run ESMValTool in quicklook mode -> PR #169 
  - get simulation name (via command line)
  - get years to be processed (via command line)
  - create one recipe from recipe templates and fill with given simulation(s) and years
- [ ] feature: check available years in input directory, compare to list of years already processed for given simulation and recipe and automatically "fill the gaps", i.e. process all missing years
- [ ] feature: ignore non-existing variables in quicklook recipes and continue with other variables (**bonus feature**)

PR #169, opened 2019-07-16
[EMAC] Adjust commandline interface for quicklooks

bjoernbroetz:
This PR addresses the issue #145 by 
- adjusting the commandline interface of the esmvaltool command for running in quicklook-mode 
- adjusting the files formerly known as templates
- adjusting the `create_recipe` function to create a single quicklook recipe

The command works now as follows:
```
esmvaltool -c <CONFIG-USER.yml> --quicklook <SIMULATION-ID> --startyear 1990 --endyear 1992
```


bjoernbroetz:
If there are corrupted netcdf files lurking in the persistent quicklook folder the tool crashes by running into [this](https://github.com/SciTools/iris/issues/3312) iris bug that will be fixed soon.

> Do we really need the command line option --quicklook? I thought the quicklook mode is turned on/off in  the user-config.yml: active: true/false. What if there is a conflict between the config file and the command line options?

As discussed with @axel-lauer off-line the name of this option should be changed but maybe in a different PR. We need the info whether we run in *quicklook mode* or not *before* we read the config-file. A possible conflict is caught.

valeriupredoi:
infrastructure question: how come this PR doesn't need at least 1 approver for merging? @bjoernbroetz you ninja whizz :grin:

valeriupredoi:
an nvmnd, it's not in `version2_development` :man_facepalming:

mattiarighi:
Good idea.

bjoernbroetz:
Full test at DKRZ for historical run from 1850-1970 in steps of 2 years currently on the way. Lets wait with this PR this it passes.

schlunma:
I added an additional logger which just logs anomalies of the quicklook feature (especially errors in the concatenation). This log file can be specified in `config-user.yml`. New messages are appended to the end of it, that means the log file includes information from all running quicklook processes.

Unfortunately, testing this still failed because of the large numbers of threads. Using just one single diagnostic at the end of all variables works fine, so we should do that (right now, the quicklook features builds a recipe with 4 diagnostic blocks). However, I think this solution is also close to the thread limit (e.g. changing ancestors caused it to fail), so we should try to change this approach somehow to make it more stable.

bjoernbroetz:
> Unfortunately, testing this still failed because of the large numbers of threads. Using just one single diagnostic at the end of all variables works fine, so we should do that (right now, the quicklook features builds a recipe with 4 diagnostic blocks). However, I think this solution is also close to the thread limit (e.g. changing ancestors caused it to fail), so we should try to change this approach somehow to make it more stable.

@schlunma : The test I ran on Friday (1850-1970 in steps of 2 years) with this branch worked fine for the first cycle but failed with:
```
KeyError: 'Provenance record for /work/XXX/XXX/esmvaltool_output/persistent/work/glob_timeseries_TN_GHG_CO2/glob_timeseries/AIRC_NO_s_historical_global_timeseries.nc already exists.'
```

I'm fixing this now but the thread problem was solved for me after rearranging the diagnostic-blocks. Do you run your tests interactive of do you submit it. I submit it to SLURM, maybe this is a difference as we saw on Friday there is a limit per node.

bjoernbroetz:
Test is running fine with adjustment to the ESMValTool in https://github.com/ESMValGroup/ESMValTool/pull/1218
20 min WCT -> 15 years

schlunma:
> > Unfortunately, testing this still failed because of the large numbers of threads. Using just one single diagnostic at the end of all variables works fine, so we should do that (right now, the quicklook features builds a recipe with 4 diagnostic blocks). However, I think this solution is also close to the thread limit (e.g. changing ancestors caused it to fail), so we should try to change this approach somehow to make it more stable.
> 
> @schlunma : The test I ran on Friday (1850-1970 in steps of 2 years) with this branch worked fine for the first cycle but failed with:
> 
> ```
> KeyError: 'Provenance record for /work/XXX/XXX/esmvaltool_output/persistent/work/glob_timeseries_TN_GHG_CO2/glob_timeseries/AIRC_NO_s_historical_global_timeseries.nc already exists.'
> ```
> 
> I'm fixing this now but the thread problem was solved for me after rearranging the diagnostic-blocks. Do you run your tests interactive of do you submit it. I submit it to SLURM, maybe this is a difference as we saw on Friday there is a limit per node.

The `KeyError` should not appear anymore with the changes I made to the core today, can you remove your changes to the provenance logger and try again? This error should not be ignored, as it occurs when similar data (identical `short_name` and `dataset`) is read by the diagnostic. This means that also only one plot (instead of multiple ones) is created.

I requested a node for interactive work for my test, but I will try again with a regular submitted job.

bjoernbroetz:
> The `KeyError` should not appear anymore with the changes I made to the core today, can you remove your changes to the provenance logger and try again?

I reverted the changes and will check. Thanks for the logger feature! Particularly useful.

Issue #61, opened 2018-07-31
Apply cmor checks (and fixes) to fx-files


mattiarighi:
Going through the fx data in CMIP5, I've noticed that there are several issues with the mask variables `sftlf` and `sftof` which needs to be fixed (wrong units, wrong coordinates, wrong fillvalues, etc.), but this is currently not implemented, although some fixes are available (see for example [here](https://github.com/ESMValGroup/ESMValTool/blob/fc376aded0d2092b05e3d1f74132592f0a1dc2a6/esmvaltool/cmor/_fixes/CMIP5/EC_EARTH.py#L29)). 

This affects both, the use of fx variables in the diagnostics (via `fx-files` key in the variable dict) and the use of fx variables for the masking (via `mask_landsea` key in the preprocessor dict).

In principle, whenever a fx-variable is needed, it should be read, cmor-checked, (fixed) and saved in the `work_dir` / `preproc_dir`, as for any other variable. The corresponding path should be then made available to the diagnostics if the `fx-files` key is specified in the recipe (as it happens now with the original input fx-file).

zklaus:
So @valeriupredoi... does it work now? Can we build fixes for fx files?

valeriupredoi:
it do, if you ask for an fx variable just like any other variable in #170

zklaus:
I need it for a derived variable (#192), namely the fx variable volcello. Can I do that somehow?

valeriupredoi:
fixed by #170 and #230

PR #22, opened 2019-06-10
Development cmip6fx conventions with cmor fixes to fx vars

valeriupredoi:
Related issue #109 
ported ESMValGroup/ESMValTool#1037 to ESMValCore/version2_development

PR #170, opened 2019-07-17
Working FX handling

valeriupredoi:
OK chaps @bouweandela @zklaus @ledm here is the long promised implementation of fx vars handling: this is a sister PR of #21 but with much improved user interface and functionality. It does the following:

- allows for use of fx variables (CMIP5 and 6) as regular variables;
- removes any special file retrieval functionality for fx files; the regulat `get_input_filelist` is used; also removes the special entries for fx files/dirs in `config-developer`;
- removes the need for the user to specify the `fx_files` dictionary in the variable (see NOTE below);
- reverts a bunch of changes I made previously;

NOTE: 

- if a user wants to use fx files in the diagnostic, all they have to do is include the fx variables in the variables list; they will get the fx files at the end of preprocessing, with any preprocessor of their choice applied to them (and CMOR checked and fixed);
- if a user wants to use `mask_landsea` or `mask_landseaice`, all they have to do is to set these preproc steps in the preprocessor; no more need to specify the `fx_files` in variabl; the code will automatically grab the fx files for masking BUT it will apply the masks from those files WITHOUT doing any CMOR checks/fixes or any preprocessor on the fx files (basically what we had before). The next step will be to implement this functionality (allow preprocessing on fx files that are used in the preprocessor) but that is absolutely non-trival.


valeriupredoi:
...in any case, this is a big-ish PR and let's get this one in first before concerning ourselves with more enhancements :beer:

valeriupredoi:
I will close #21 without merge when this is approved

jvegasbsc:
Thanks V. I am currently having issues with a diagnostic because preprocessed data is just a region and areacello is the full globe, so this is coming at the perfect time for me

valeriupredoi:
hah, excellent! does it work well? you and @bouweandela can actually actively review it then :beer:

ledm:
Glad there's been some progress here, @valeriupredoi. What changes do we need to make to our recipes to take advantage of the new features here?

valeriupredoi:
I am on annual leave this week so my replies will be patchy (sorry in advance) but as for changes - if you want to use fx files all you have to do now is to list the fx variables in the recipe/diagnostic as any regular CMOR variable in the variables list (note that you need to specify the `mip` for CMIP6 since those have varied `mip`'s; also note that CMIP6 fx variables are not fully populated with data on ESGF) :beer:

ledm:
No worries... If I was on leave, I would not be checking emails! Please take your time replying!

So, what should a call look like? I'm trying to use `area_statistics` to calculate the area-weighted mean of an irregular grid. (Bascially, one of the simplest examples `recipe_ocean_example.py`, but with an irregular grid model (`MPI-ESM-LR` in CMIP5).

The recipe diagnostic looks like this:
```
  diag_timeseries_1:
    variables:
      tos: # Temperature ocean surface
        preprocessor: prep_timeseries_1
      areacello:
        preprocessor: prep_timeseries_1
        exp: historical
        mip: Omon
    scripts:
      Global_Ocean_Surface_mean_timeseries:
        script: ocean/diagnostic_timeseries.py
```
But of course, this doesn't work (otherwise I wouldn't be asking!). Any ideas what changes are needed to this recipe?

Lee

valeriupredoi:
why is it not working? Can you post the error message pls? Also no need to specify the `exp`

valeriupredoi:
and mip is not Omon - it is fx if you are using CMIP5 and Ofx if CMIP6; as I said, no need to specify the mip if CMIP5, only for CMIP6

ledm:
Here is the entire recipe:
```
# ESMValTool
# recipe_testing_PR170.yml
---
documentation:
  description: |
    Recipe to test new FX tools.

  authors:
    - demo_le

  maintainer:
    - demo_le

  references:
    - demora2018gmd

  projects:
    - ukesm


datasets:
  # Unstructured grids
  - {dataset: MPI-ESM-LR, project: CMIP5, exp: historical, ensemble: r1i1p1, start_year: 2001, end_year: 2003}

preprocessors:
  # --------------------------------------------------
  # Time series preprocessors
  # --------------------------------------------------
  prep_timeseries_1: # For 2D fields
    custom_order: true
    area_statistics:
      operator: mean
    multi_model_statistics:
      span: overlap
      statistics: [mean ]

diagnostics:
  # --------------------------------------------------
  # Time series diagnostics
  # --------------------------------------------------
  diag_timeseries_1:
    description: Global Ocean Surface mean timeseries
    variables:
      tos: # Temperature ocean surface
        preprocessor: prep_timeseries_1
        mip: Omon
      areacello:
        mip: fx
    scripts:
      Global_Ocean_Surface_mean_timeseries:
        script: ocean/diagnostic_timeseries.py

```
Using this recipe, the `area_statistics` tool still can not find the fx_files, so it splits out the error message:
```
2019-07-22 12:51:00,593 UTC [17830] ERROR   fx_file needed to calculate grid cell area for irregular grids.
```

Any ideas why the preprocessor function, `area_statistics` can't find the fx_file?

jvegasbsc:
I tested your example and I am having trouble even before, because the ensemble for fx is r0i0p0

ledm:
Okay, so the recipe is now:
```
datasets:
  # Unstructured grids
  - {dataset: MPI-ESM-LR, project: CMIP5, exp: historical,  start_year:
2001, end_year: 2003}

preprocessors:
  # --------------------------------------------------
  # Time series preprocessors
  # --------------------------------------------------
  prep_timeseries_1: # For 2D fields
    area_statistics:
      operator: mean
      fx_files: [areacello, ]

diagnostics:
  # --------------------------------------------------
  # Time series diagnostics
  # --------------------------------------------------
  diag_timeseries_1:
    variables:
      tos:
        preprocessor: prep_timeseries_1
        mip: Omon
        ensemble: r1i1p1
    scripts:
      Global_Ocean_Surface_mean_timeseries:
        script: ocean/diagnostic_timeseries.py
```

The main difference is that the fx_files is added to the preprocessor,
but it works!

I've also had to make a few changes to `preprocessor/_area.py` to get it
to work. These changes address previous errors that slipped into the
code base. Basically, the problems were that dask doesn't work using
numpy nomenclature.

(Also, github is down now?)





Please visit our new website at www.pml.ac.uk and follow us on Twitter @PlymouthMarine

We have updated our Privacy Notice which is available at http://www.pml.ac.uk/System-pages/Privacy_Notice

Plymouth Marine Laboratory (PML) is a company limited by guarantee registered in England & Wales, company number 4178503. Registered Charity No. 1091222. Registered Office: Prospect Place, The Hoe, Plymouth  PL1 3DH, UK. 

This message is private and confidential. If you have received this message in error, please notify the sender and remove it from your system. You are reminded that e-mail communications are not secure and may contain viruses; PML accepts no liability for any loss or damage which may be caused by viruses.

valeriupredoi:
right, so the `area_statistics` is the only bit that I didn't test with this PR, could you pls add those changes you made to this PR so it's fully functional? Cheers for doing the testing and making the code changes! The other options with `masklandsea` or `_landseaice` and getting just the fx vars work fine :beer:

ledm:
Okay, so, should I push the suggested changes?

The problem is that some of my changes in `_area.py` may no longer be using lazy arrays. Some previous changes switched numpy functions for dask functions, but the dask version of the functions don't use the same arguments (notably dask.array.tile).

zklaus:
Go ahead and push the changes. We will have a look and discuss what to change if anything. Bonus points for pushing it in few (to very few) neat, clear, self-contained commits that are easily discussed and dealt with one by one.

ledm:
Okay, it should be easy to see the changes, as they are all in the file [`_area.py`]( https://github.com/ESMValGroup/ESMValCore/pull/170/files#diff-997c54b672404a31a4d2a12bc72bbe02).

ledm:
Added a couple very minor changes to `_area.py` and `_volume.py`.

valeriupredoi:
cheers @ledm I will take care of the documentation next week; have people tested this PR fully? :beer:

ledm:
The regrid preprocessor is currently being applied to the fx_files. Is that expected?

For instance, in the recipe:
```
datasets:
  # Unstructured grids
  - {dataset: MPI-ESM-LR, project: CMIP5, exp: historical, mip: Omon, ensemble: r1i1p1, start_year: 2001, end_year: 2002}

preprocessors:
  prep_timeseries_regridtest:
    custom_order: true
    regrid:
      target_grid: 1x1
      scheme: linear
    area_statistics:
      operator: mean
      fx_files: [areacello, ]

diagnostics:
  diag_timeseries_1:
    variables:
      tos: # Temperature ocean surface
        preprocessor: prep_timeseries_regridtest
    scripts:
      Global_Ocean_Surface_mean_timeseries:
        script: ocean/diagnostic_timeseries.py
```
The `areacello` fx file is not regridded to the same grid as the model data. So the whole thing fails.

esmvalcore/preprocessor/_area.py
```diff
@@ -56,20 +57,24 @@ def extract_region(cube, start_longitude, end_longitude, start_latitude,
...
-    select_lons = start_longitude < lons < end_longitude
-    selection = select_lats & select_lons
-    data = da.ma.masked_where(~selection, cube.core_data())
-    return cube.copy(data)
+    mask = np.ma.array(cube.data).mask
+    mask += np.ma.masked_where(lats < start_latitude, lats).mask
+    mask += np.ma.masked_where(lats > end_latitude, lats).mask
+    mask += np.ma.masked_where(lons < start_longitude, lons).mask
+    mask += np.ma.masked_where(lons > end_longitude, lons).mask
+    cube.data = da.ma.masked_array(data=cube.data, mask=mask)
```
zklaus: This doesn't look right:

- `masked_where(condition, array).mask` should be the same as `condition`
- On boolean arrays `+` is logical or  (and would be better written as |)

The resulting mask is probably correct, but the way is overly complicated and unnecessarily excludes dask.
I'll see if I can get something akin to the old code working correctly with both dask and numpy. @ledm, what exactly was the problem with the old code?

esmvalcore/preprocessor/_area.py
```diff
@@ -156,19 +161,32 @@ def tile_grid_areas(cube, fx_files):
...
+        return None
+
+    if grid_areas.shape != cube.shape[-2:]:
+        raise ValueError('Fx area {} and dataset {} shapes do not match.'
+                         ''.format(grid_areas.shape, cube.shape))
+
+    if cube.ndim == grid_areas.ndim:
+        return grid_areas
+
+    # Use dash.array.stack to tile areacello.
```
zklaus: Typo: dash -> dask?

esmvalcore/preprocessor/_area.py
```diff
@@ -156,19 +161,32 @@ def tile_grid_areas(cube, fx_files):
                 continue
             logger.info('Attempting to load %s from file: %s', key, fx_file)
             fx_cube = iris.load_cube(fx_file)
-
             grid_areas = fx_cube.core_data()
```
zklaus: Perhaps it would be better to use `.lazy_data()` to guarantee a dask array compatible with the dask functions?

esmvalcore/preprocessor/_area.py
```diff
@@ -156,19 +161,32 @@ def tile_grid_areas(cube, fx_files):
...
+    elif cube.ndim == 4 and grid_areas.ndim == 3:
+        grida = [grid_areas for itr in range(cube.shape[0])]
+        grid_areas = da.stack(grida, axis=0)
+    elif cube.ndim == 3 and grid_areas.ndim == 2:
+        grida = [grid_areas for itr in range(cube.shape[0])]
+        grid_areas = da.stack(grida, axis=0)
+    else:
+        raise ValueError('Grid and dataset number of dimensions not '
+                         'recognised: {} and {}.'
+                         ''.format(cube.ndim, grid_areas.ndim))
```
zklaus: Could [`broadcast_to`](https://docs.dask.org/en/latest/array-api.html#dask.array.broadcast_to) be used to simplify this block of `if`s?

esmvalcore/preprocessor/_area.py
```diff
@@ -246,7 +264,7 @@ def area_statistics(cube, operator, fx_files=None):
     if operator == 'mean':
         return cube.collapsed(coord_names,
                               operation,
-                              weights=grid_areas)
+                              weights=np.array(grid_areas))
```
zklaus: Is this explicitly to turn a dask array into a numpy array? Maybe `grid_areas.compute()` or `np.asarray(grid_areas)` would be more appropriate.

esmvalcore/preprocessor/_volume.py
```diff
@@ -260,12 +260,13 @@ def volume_statistics(
...
                     cube[time_itr, z_itr].data.mask,
                     grid_volume[time_itr, z_itr]).sum()
-
             except AttributeError:
                 # ####
                 # No mask in the cube data.
                 layer_vol = grid_volume.sum()
             depth_volume.append(layer_vol)
+
+        column = np.ma.masked_invalid(column)
```
zklaus: What is the relation of this to this PR? It seems rather unrelated to the fx stuff. @ledm, maybe you want to comment?

valeriupredoi:
@ledm if you read my top comment carefully you will find the answer:

> if a user wants to use mask_landsea or mask_landseaice, all they have to do is to set these preproc steps in the preprocessor; no more need to specify the fx_files in variabl; the code will automatically grab the fx files for masking BUT it will apply the masks from those files WITHOUT doing any CMOR checks/fixes or any preprocessor on the fx files (basically what we had before). The next step will be to implement this functionality (allow preprocessing on fx files that are used in the preprocessor) but that is absolutely non-trival.

So yes, any of the preprocessing functions that need fx files (mask_landsea, mask_landseaice, area_statistics, derive) will use those as they are found at the ESGF node and not preprocessed in any way; if you need to use fx files in diagnostic then you can specify the fx variable in the diag's variable list and allow for any preprocessor to be applied to your fx variables

esmvalcore/_recipe.py
```diff
@@ -6,13 +6,14 @@
 from copy import deepcopy
 
 import yaml
+
```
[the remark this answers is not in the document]
valeriupredoi: see below for next comment

esmvalcore/_recipe.py
```diff
@@ -6,13 +6,14 @@
...
 from netCDF4 import Dataset
 
-from . import __version__
 from . import _recipe_checks as check
+from . import __version__
```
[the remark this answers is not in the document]
valeriupredoi: just pushed the `isort`-ed version, we are using `isort` since otherwise @bouweandela will send assassins after us :grin:

esmvalcore/_recipe.py
```diff
@@ -361,6 +362,51 @@ def _get_default_settings(variable, config_user, derive=False):
...
+    if var['project'] == 'CMIP5':
+        fx_var = _add_fxvar_keys({'short_name': fx_varname, 'mip': 'fx'}, var)
+    elif var['project'] == 'CMIP6':
+        if fx_varname == 'sftlf':
+            fx_var = _add_fxvar_keys({'short_name': fx_varname, 'mip': 'fx'},
+                                     var)
+        elif fx_varname == 'sftof':
+            fx_var = _add_fxvar_keys({'short_name': fx_varname, 'mip': 'Ofx'},
+                                     var)
+        # TODO check this is the correct mip
```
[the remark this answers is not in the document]
valeriupredoi: what list? the one that was mapping fx variable to realm? That's gone man, we have CMOR do it automatically now; but the mip is needed for CMIP6

esmvalcore/_recipe.py
```diff
@@ -361,6 +362,51 @@ def _get_default_settings(variable, config_user, derive=False):
...
+        elif fx_varname == 'sftgif':
+            fx_var = _add_fxvar_keys({'short_name': fx_varname, 'mip': 'Efx'},
+                                     var)
```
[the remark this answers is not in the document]
valeriupredoi: cheers dude! I changed the mip to fx and changed the TODO as well

esmvalcore/_recipe.py
```diff
@@ -632,23 +665,54 @@ def _get_preprocessor_products(variables, profile, order, ancestor_products,
...
+        'extract_season',
+        'extract_month',
+        'annual_mean',
+        'seasonal_mean',
+        'time_average',
+        'regrid_time',
+    ]
+    for temporal_preproc in temporal_preprocs:
+        if temporal_preproc in profile:
+            fx_profile[temporal_preproc] = False
```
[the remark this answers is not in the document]
valeriupredoi: yes, removing individual preproc steps is an option as well but I'd like to have the same shape of `profile` so the user will see the actual setting to `False` for the temporal ones if using debug mode

esmvalcore/_recipe.py
```diff
@@ -632,23 +665,54 @@ def _get_preprocessor_products(variables, profile, order, ancestor_products,
...
-    products = _get_preprocessor_products(
-        variables=variables,
-        profile=profile,
-        order=order,
-        ancestor_products=ancestor_products,
-        config_user=config_user,
-    )
+
+    if variables[0]['frequency'] != 'fx':
+        products = list(_get_preprocessor_products(
```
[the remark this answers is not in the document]
valeriupredoi: yep, good question, it's dead now :grin:

esmvalcore/_recipe.py
```diff
@@ -632,23 +665,54 @@ def _get_preprocessor_products(variables, profile, order, ancestor_products,
...
+            variables=variables,
+            profile=profile,
+            order=order,
+            ancestor_products=ancestor_products,
+            config_user=config_user,
+        ))
+    else:
+        fx_profile = _remove_temporal_preprocs(profile)
+        # not do time extraction; this is a default preproc
+        fx_profile['extract_time'] = False
```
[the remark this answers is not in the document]
valeriupredoi: no, that's a default preproc that needs to be removed upstream

esmvalcore/_recipe.py
```diff
@@ -632,23 +665,54 @@ def _get_preprocessor_products(variables, profile, order, ancestor_products,
...
+        products = list(_get_preprocessor_products(
+            variables=variables,
+            profile=fx_profile,
+            order=order,
+            ancestor_products=None,
+            config_user=config_user,
+        ))
```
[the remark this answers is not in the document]
valeriupredoi: good call! I just simplified it :beer:

esmvalcore/config-developer.yml
```diff
@@ -148,26 +148,7 @@ CMIP5:
...
-    'areacello': 'Omon'
-    'basin': 'Omon'
-    'deptho': 'Omon'
-    'mrsofc': 'Lmon'
-    'orog': 'Amon'
-    'rootd': 'Lmon'
-    'sftgif': 'Lmon'
-    'sftlf': 'Amon'
-    'sftof': 'Omon'
-    'volcello': 'Omon'
```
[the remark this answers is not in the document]
valeriupredoi: this list is removed completely, the cmor info is populated automatically now

esmvalcore/config-developer.yml
```diff
@@ -241,11 +222,6 @@ OBS:
   input_file:
     default: '[project]_[dataset]_[type]_[version]_[mip]_[short_name]_*.nc'
     BSC: '[short_name]_*.nc'
-    RCAST: '[short_name]_[mip]_[type]_[dataset]_*.nc'
```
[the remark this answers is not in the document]
valeriupredoi: probably - but it seems to be quite obsolete - @mattiarighi what is RCAST?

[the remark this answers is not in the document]
mattiarighi: It has been added recently by @LisaBock, see #168.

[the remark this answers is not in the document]
valeriupredoi: ok I put it back, sorry I removed it, I usually shoot first and ask questions after :grin:

esmvalcore/_recipe.py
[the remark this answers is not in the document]
zklaus: I am happy that both you and isort agree with my suggestions :)

esmvalcore/_recipe.py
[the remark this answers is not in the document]
zklaus: I am not sure I follow your reasoning. To me a user will be equally unsurprised (or surprised) by a preprocessor not showing up or it being set to false. But it seems a matter of taste and this is ok.

esmvalcore/config-developer.yml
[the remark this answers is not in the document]
zklaus: This kind of makes sense, but I didn't find where. Could you point me in the right direction?

esmvalcore/_recipe.py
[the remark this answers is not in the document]
zklaus: Sorry, I don't follow. What does [CMOR](https://cmor.llnl.gov/) have to do with this? And why is, eg, `areacello` handled automatically, but `sftof` not?

esmvalcore/_recipe.py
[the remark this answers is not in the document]
zklaus: :+1: Is it possible to specify a mip explicitly if I as a user want to do that in the recipe?

esmvalcore/_recipe.py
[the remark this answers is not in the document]
zklaus: What do you mean by upstream? Putting this in `_remove_temporal_preprocs` seems to produce the exact same state, no?

esmvalcore/config-developer.yml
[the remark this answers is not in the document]
zklaus: That's an ICBM.

ledm:
> @ledm if you read my top comment carefully you will find the answer:
> 
> > if a user wants to use mask_landsea or mask_landseaice, all they have to do is to set these preproc steps in the preprocessor; no more need to specify the fx_files in variabl; the code will automatically grab the fx files for masking BUT it will apply the masks from those files WITHOUT doing any CMOR checks/fixes or any preprocessor on the fx files (basically what we had before). The next step will be to implement this functionality (allow preprocessing on fx files that are used in the preprocessor) but that is absolutely non-trival.
> 
> So yes, any of the preprocessing functions that need fx files (mask_landsea, mask_landseaice, area_statistics, derive) will use those as they are found at the ESGF node and not preprocessed in any way; if you need to use fx files in diagnostic then you can specify the fx variable in the diag's variable list and allow for any preprocessor to be applied to your fx variables

Hi V, welcome back! Hope you had a good week off!

Unfortunately, this still doesn't address the problem. In this case, the fx_files need to be preprocessed for the rest of the preprocessor to work! I don't want them in the diagnostic stage. In the example below, the `areacello` data needs to be regrided before it can be used by the `area_statistics` function. 

Can you let me know what I need to do to this recipe in order for it to work?

```
datasets:
  # Unstructured grids
  - {dataset: MPI-ESM-LR, project: CMIP5, exp: historical, mip: Omon, ensemble: r1i1p1, start_year: 2001, end_year: 2002}

preprocessors:
  prep_timeseries_regridtest:
    custom_order: true
    regrid:
      target_grid: 1x1
      scheme: linear
    area_statistics:
      operator: mean
      fx_files: [areacello, ]

diagnostics:
  diag_timeseries_1:
    variables:
      tos: # Temperature ocean surface
        preprocessor: prep_timeseries_regridtest
    scripts:
      Global_Ocean_Surface_mean_timeseries:
        script: ocean/diagnostic_timeseries.py
```

zklaus:
I realize this is a rather special case, but it seems to me that for `areacella/o` most likely the best way is to recompute them with `iris.analysis.cartography.area_weights`. For regular grids this is likely to be more accurate than any regridding.

@ledm, could you please have a look at my comments on `_area.py` and `_volume.py`?

esmvalcore/preprocessor/_area.py
ledm:
```
select_lats = start_latitude < lats < end_latitude
```
is not a valid dask command and doesn't work. We should replace these lines with something compatible with dask. I just re-added the old numpy lines because they worked.

esmvalcore/preprocessor/_area.py
ledm:
```suggestion
    # Use dask.array.stack to tile areacello.
```

esmvalcore/preprocessor/_area.py
zklaus: Then this seems to be a bug in `extract_region` that has nothing to do with this fx file stuff, correct?
I suggest to

- Undo the changes here
- Open a `bug` issue
- Write a test that shows the problem
- Write a proper fix, maybe the more explicit `start_latitude < lats & lats < end_latitude` works.

We should not mix up completely independent things, less when the PR is as complicated as this one.

zklaus: PS: I am happy to help. Just let me know how you want to proceed and what part you would like to do yourself :)

esmvalcore/preprocessor/_volume.py
ledm: This is a issue that arose now that we finally have working FX files. It is related to the issue, promise!

esmvalcore/preprocessor/_area.py
ledm: I can confirm that @zklaus's suggestion to use compute works:

```suggestion
                              weights=grid_areas.compute())
```
Presumably it's better to keep things native in dask than to cast into a numpy array.

zklaus: `.compute()` *does* give you a numpy array. What it *doesn't* do is go through the numpy.array constructor which probably is a bit more performant. If completely sticking with dask arrays is an option, then of course a simple
```
weights=grid_areas)
```
is the best solution.

esmvalcore/preprocessor/_volume.py
zklaus: I'd ask you to be a bit more specific, but really this whole function is in need of a rewrite, so until then knock yourself out.

esmvalcore/preprocessor/_area.py
ledm: Yep, the stack trace is:
```
2019-07-30 10:35:35,079 UTC [4633] ERROR   Program terminated abnormally, see stack trace below for more information
  File "/users/modellers/ledm/workspace/ESMValTool/ESMValCore/esmvalcore/_main.py", line 228, in run
    conf = main(args)
  File "/users/modellers/ledm/workspace/ESMValTool/ESMValCore/esmvalcore/_main.py", line 156, in main
  File "/users/modellers/ledm/workspace/ESMValTool/ESMValCore/esmvalcore/_main.py", line 206, in process_recipe
    recipe.run()
  File "/users/modellers/ledm/workspace/ESMValTool/ESMValCore/esmvalcore/_recipe.py", line 1126, in run
    self.tasks, max_parallel_tasks=self._cfg['max_parallel_tasks'])
  File "/users/modellers/ledm/workspace/ESMValTool/ESMValCore/esmvalcore/_task.py", line 594, in run_tasks
    _run_tasks_sequential(tasks)
  File "/users/modellers/ledm/workspace/ESMValTool/ESMValCore/esmvalcore/_task.py", line 605, in _run_tasks_sequential
    task.run()
  File "/users/modellers/ledm/workspace/ESMValTool/ESMValCore/esmvalcore/_task.py", line 224, in run
    input_files.extend(task.run())
  File "/users/modellers/ledm/workspace/ESMValTool/ESMValCore/esmvalcore/_task.py", line 227, in run
  File "/users/modellers/ledm/workspace/ESMValTool/ESMValCore/esmvalcore/preprocessor/__init__.py", line 394, in _run
  File "/users/modellers/ledm/workspace/ESMValTool/ESMValCore/esmvalcore/preprocessor/__init__.py", line 261, in apply
  File "/users/modellers/ledm/workspace/ESMValTool/ESMValCore/esmvalcore/preprocessor/__init__.py", line 206, in preprocess
    result.append(_run_preproc_function(function, item, settings))
  File "/users/modellers/ledm/workspace/ESMValTool/ESMValCore/esmvalcore/preprocessor/__init__.py", line 189, in _run_preproc_function
  File "/users/modellers/ledm/workspace/ESMValTool/ESMValCore/esmvalcore/preprocessor/_area.py", line 267, in area_statistics
    weights=grid_areas) 
  File "/users/modellers/ledm/workspace/ESMValToolTest/miniconda/minicondaLee/envs/esmvaltool2/lib/python3.7/site-packages/iris/cube.py", line 3277, in collapsed
    weights = kwargs["weights"].view()
TypeError: view() missing 1 required positional argument: 'dtype'
```

Bascially, iris can't accept a dask array as a weights field.

zklaus: Ok. I think this can be considered a dask bug and I filed an issue [here](https://github.com/dask/dask/issues/5175). 
Anyway, since we know the weights array to be 2d only, I think it's fine to go with the numpy array. `.compute()` would be my favorite method to get there in this case.

esmvalcore/_recipe.py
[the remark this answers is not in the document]
valeriupredoi: that could be easily implemented as argument to the preproc masking function but I think keeping things automated is much better in this case since we know that either sftlf or sftof or sftgif are needed; I reckon this should be addressed in another PR if really needed

esmvalcore/_recipe.py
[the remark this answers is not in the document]
valeriupredoi: https://github.com/ESMValGroup/ESMValCore/blob/bd55ef6a665ff74ee92d9b06b77c5605f0dae75f/esmvalcore/_recipe.py#L97 is the function that does it - all it needs is the variable dict with correct mip

esmvalcore/_recipe.py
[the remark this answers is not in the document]
valeriupredoi: should be but last time I did it it ddin't work because `profile` gets populated with the default options only at the very last step before creating products, I meant downstream not upstream, my bad

esmvalcore/config-developer.yml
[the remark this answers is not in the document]
valeriupredoi: https://github.com/ESMValGroup/ESMValCore/blob/bd55ef6a665ff74ee92d9b06b77c5605f0dae75f/esmvalcore/_recipe.py#L97

valeriupredoi:
OK guys - am back (cheers Lee, had a good boozing week last week :grin: ) but only to leave today on hols again until next Monday. Let's wrap this up - please make any changes you need to while I'm gone again but please don't implement any major features since this needs to go in trunk soon otherwise we don't have any handling of fx vars and files. Lee, as I told you a couple times above in comments - there is no way at the moment to apply preprocessing on the fx vars that go in the three masking functions (that is not easy to implement because those fx vars should be treated as separate variables)

esmvalcore/preprocessor/_area.py
ledm: Okay, so this line fails when `grid_areas` is still an numpy array (because numpy arrays do have a `compute` method), we need to add a line which casts numpy array `grid_areas` to a dask array.

valeriupredoi:
OK guys, finally back to this, am gonna go ahead and as per @zklaus suggestion close the other two PRs that are dealing with the FX stuff - https://github.com/ESMValGroup/ESMValCore/pull/22 and #21

valeriupredoi:
are people happy with the functionality here or is there still work to be done? I think we need to get this one in trunk asap :beer:

esmvalcore/preprocessor/_area.py
valeriupredoi: daesh :grin:

esmvalcore/_recipe.py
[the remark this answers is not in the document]
bouweandela: I think we could add a list of mips instead of a single mip, that would work so long as there is only a single file

esmvalcore/_recipe.py
```diff
@@ -632,16 +664,40 @@ def _get_preprocessor_products(variables, profile, order, ancestor_products,
...
                                   ancestor_tasks=None):
-    """Create preprocessor tasks for a set of datasets."""
+    """Create preprocessor tasks for a set of datasets w/ special case fx."""
     if ancestor_tasks is None:
         ancestor_tasks = []
     order = _extract_preprocessor_order(profile)
     ancestor_products = [p for task in ancestor_tasks for p in task.products]
+
+    if variables[0]['frequency'] == 'fx':
+        profile = _remove_temporal_preprocs(profile)
```
bouweandela: There should be no need for this function: these preprocessor functions are not in the default preprocessor and a user should simply not specify in the recipe them for a preprocessor that preprocesses fx files.

valeriupredoi:
> Hi V.,
> 
> Nice job, but can you please move the changes to the preprocessor functions (changes in esmvalcore/preprocessor/_area.py and esmvalcore/preprocessor/_volume.py) to a separate pull request?

cheers B-man, but these are @ledm 's changes and don't want to take credit for them :grin: I also think it's best if he moved them coz it's his stuff :beer:

esmvalcore/_recipe.py
valeriupredoi: err..yeah but you know how users are, a tiny slip in the recipe and all goes kaboom and then they start opening issues on gitHub, better safe than sorry :grin:

bouweandela: Let's wait for them to do that. If we get many issues, we can always add extra code to generate a nice error message. I would like to keep the amount of code we have to a minimum, because _recipe.py is already pretty complicated.

ledm:
For now, just nuke those two files. 

I don't think any of the other changes in _area or _volume are affected by the FX handling.

esmvalcore/_recipe.py
valeriupredoi: nah man, come on, not including code that's needed just because it introduces extra lines of code is not a solid reason - we can make a dedicated module outside `_recipe.py` or something for this stuff (altho it's kinda overkill), users will go on and apply various preprocessors to the `fx` variables without knowing that some steps are not allowed and then they'll complain left right and center. This is a needed safeguard and @ledm as a user can veto on this too. I believe the easier for users the less headaches we get :beer:

valeriupredoi: plus - this allows the user to recycle the preprocessor X without having to rewrite it as preprocessor Y

bouweandela: Silently ignoring wrong settings in a recipe is not a good idea, it leads to messy recipes that are hard to understand for users. Adding a check with a user friendly error message is a good idea, you can add code for checks in esmvalcore/_recipe_checks.py

valeriupredoi: yeah that's actually a good idea! I'll plug it there then :beer: But you see, thos settings are ignored not only because their execution leads to code error but also because they don't affact the data - since there is no time axis in the fx data doing temporal analysis on it results in nothing, so it's not really a silent ignoring

valeriupredoi: @bouweandela recipe check/`RecipeError` implemented and that function is now gonzo :beer:

valeriupredoi:
OK I believe this one (bar its documentation that I will be taking care of in a separate PR) is ready to roll :roller_coaster:

bouweandela:
>  OK I believe this one (bar its documentation that I will be taking care of in a separate PR) is ready to roll

Note that changes in esmvalcore/preprocessor/_area.py and esmvalcore/preprocessor/_volume.py still need to be reverted.

Also, to complete the new fx file syntax for recipes, an associated PR in the ESMValTool repository to change the recipes that use the old fx files as a list `fx_files: [sftlf, mrsofc]` in their variable section. I think this code can then also be removed: https://github.com/ESMValGroup/ESMValCore/blob/9f1c1e489496328dfb97d82a58d34a9660cb1b1a/esmvalcore/preprocessor/_io.py#L242-L245

valeriupredoi:
> > OK I believe this one (bar its documentation that I will be taking care of in a separate PR) is ready to roll
> 
> Note that changes in esmvalcore/preprocessor/_area.py and esmvalcore/preprocessor/_volume.py still need to be reverted.
> 
> Also, to complete the new fx file syntax for recipes, an associated PR in the ESMValTool repository to change the recipes that use the old fx files as a list `fx_files: [sftlf, mrsofc]` in their variable section. I think this code can then also be removed:
> 
> https://github.com/ESMValGroup/ESMValCore/blob/9f1c1e489496328dfb97d82a58d34a9660cb1b1a/esmvalcore/preprocessor/_io.py#L242-L245

good call, have totally missed that bit of code, now removed. Unfortunately I am not sure what bits in `_area` and `_volume` need reversion, @ledm 's changes -> can you @bouweandela or @ledm do that pls :beer:

valeriupredoi:
I will have a stab at the change of recipes now...

bouweandela:
> Unfortunately I am not sure what bits in _area and _volume need reversion

All of them, those files shouldn't need to be changed when changing data_finder

valeriupredoi:
reverted area and volume to the `development` versions and moved those changes in a new PR https://github.com/ESMValGroup/ESMValCore/pull/214

valeriupredoi:
PS ignore #213 - that branch used this one as parent (and is closed and died and went to Valhalla already)

valeriupredoi:
OK so I have seen two relatively major issues that affect the functionality of this implementation, mind you both solvable but at the moment not in the prettiest way:

- OBS data has no fx files for variables: this can be easily sorted out by explicutly running with `--skip-nonexistent` but I reckon there should be a more robust, user-independent treatment -> the question is, what if some OBS's actually have fx variables?
- diagnostics that use brute force listing of files from `metadata.yml` and find the files -> variables (like https://github.com/ESMValGroup/ESMValTool/blob/version2_development/esmvaltool/diag_scripts/ocean/diagnostic_timeseries.py ) and they assume there is *always* a time coordinate will not work -> I patched that one to do a check on `frequency != 'fx'` and only then to include the cube but I don't like it and I reckon an `exclude_variables` should be a more elegant solution

What do you think?

valeriupredoi:
OK we don't have to check the fx vars data files existence since it's OK to have an empty list

bouweandela:
> OBS data has no fx files for variables: this can be easily sorted out by explicutly running with --skip-nonexistent but I reckon there should be a more robust, user-independent treatment -> the question is, what if some OBS's actually have fx variables?

If a (fx or other) variable is not available for some dataset, you should not put that dataset in the recipe for that variable.

valeriupredoi:
> > OBS data has no fx files for variables: this can be easily sorted out by explicutly running with --skip-nonexistent but I reckon there should be a more robust, user-independent treatment -> the question is, what if some OBS's actually have fx variables?
> 
> If a (fx or other) variable is not available for some dataset, you should not put that dataset in the recipe for that variable.

not totally practical, man, since the user needs the obs files to run their diagnostic; if the diagnostic uses the top-level data request via `datasets` or `additional_datasets` dict then one is stuck with this issue when the fx files are not available; but rest assured - the recipe checks don't fail for that no more since the diagnostic *should* work without fx files since they get replaced by NE masks if they are not available, it's just the `fx_files` dict that can be empty

valeriupredoi:
stuff needing fx is piling up, what's the review status, folks?

esmvalcore/_recipe.py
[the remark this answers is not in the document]
zklaus: Unfortunately, usually the choice is to have it either in one {,O,E,...}fx table or in both {,O,E,...}mon and dec tables. But let's not worry about it for now and just get on with the dominant fx case.

zklaus:
Let's move on with this, it definitely seems to be a big improvement over the status quo.


To comment on the last two comments from @bouweandela and @valeriupredoi regarding the absence of fx files for obs data, I think the underlying issue is that the fx files are not really the primary objects.
For example for masks, that would be, well, the mask and that is generally already available in the data, probably more so for obs data since that is a bit more specific, where model data sometimes wants different masks for different applications. Likewise cell areas and other geometric properties are usually known for obs data where applicable, but complicated grids invented for numerical efficiency make it necessary to deliver this in an extra vehicle for model data.
All in all we might contemplate at some point what the actual first class objects are here and then have fx files simply as one of several possible sources for them.

But until someone has a lot of time to invest in that, I think this PR gets us a long way, so well done, @valeriupredoi!

mattiarighi:
I will test once I'm done with the paper resubmission, so hopefully beginning of next week.

valeriupredoi:
cheers guys! Let's get this in and then there's still plenty of work to be done with it. Note to @zklaus to answer a query from https://github.com/ESMValGroup/ESMValCore/issues/61 -> we currently don't have a way to pass the derived variable the fx variable as a derivation member, if we can remove the requirement `settings['derive']['fx_files']` and replace it with the fx variable as a derivation variable then CMOR checks will be applied on the fx data as on any other variable data that's needed in the derivation

valeriupredoi:
PS that shouldn't be hard to be done as soon as this PR gets in, but I'd like to get this in so I dont have to parent that PR from an unmerged PR (that change depends on the code changes here)

valeriupredoi:
brilliant, cheers @mattiarighi

valeriupredoi:
@bouweandela you done revieweing, man? :beer:

mattiarighi:
Can we finalize this?
It is needed in #888 and #1310.

mattiarighi:
@valeriupredoi can you have a look at which recipes need to be updated after this?

valeriupredoi:
@mattiarighi there are two important side-PR's related to this:
https://github.com/ESMValGroup/ESMValTool/pull/1272
https://github.com/ESMValGroup/ESMValCore/pull/230
the only reason I am not merging is I am waiting for a final OK from @bouweandela (@zklaus and you have already OK-ed it, cheers)

bouweandela:
I think that it might be good to have ESMValGroup/ESMValTool#1272 also ready before merging this, otherwise we will end up in a situation where things don't work if you install both ESMValCore and ESMValTool from the latest development branch for a longer period of time.

valeriupredoi:
Agreed! The autoassess unreviewed/prob not yet ready to include in the
release recipes need fixing in the Tool PR, I'll do that today, in the
meantime @mattiarighi maybe's got time to test with the 'official' recipes
and then we good to go?

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

mattiarighi:
Ok, I'll test the `recipe_ncl.yml` example and report back in a few minutes...

mattiarighi:
`recipe_ncl.yml` stops already during the reading phase:
```
2019-09-13 08:37:58,061 UTC [32515] ERROR   Program terminated abnormally, see stack trace below for more information
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_main.py", line 228, in run
    conf = main(args)
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_main.py", line 156, in main
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_main.py", line 202, in process_recipe
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_recipe.py", line 70, in read_recipe_file
    raw_recipe, config_user, initialize_tasks, recipe_file=filename)
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_recipe.py", line 842, in __init__
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_recipe.py", line 1166, in initialize_tasks
    task_name=task_name)
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_recipe.py", line 788, in _get_preprocessor_task
    _add_cmor_info(variable)
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_recipe.py", line 121, in _add_cmor_info
    format(cmor_table, short_name, mip))
esmvalcore._recipe_checks.RecipeError: Unable to load CMOR table 'CMIP5' for variable 'sftlf' with mip 'Amon'
2019-09-13 08:37:58,124 UTC [32515] INFO    If you suspect this is a bug or need help, please open an issue on https://github.com/ESMValGroup/ESMValTool/issues and attach the run/recipe_*.yml and run/main_log_debug.txt files from the output directory.
```

mattiarighi:
I've fixed that (mip was wrong), but now I have another issue:
```
2019-09-13 08:44:02,938 UTC [32346] ERROR   Failed to run concatenate([], {})
2019-09-13 08:44:03,381 UTC [32218] ERROR   Program terminated abnormally, see stack trace below for more information
"""
  File "/mnt/lustre01/pf/b/b309057/SOFTWARE/miniconda3/envs/esmvaltool/lib/python3.7/multiprocessing/pool.py", line 121, in worker
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_task.py", line 671, in _run_task
    output_files = task.run()
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_task.py", line 227, in run
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/preprocessor/__init__.py", line 403, in _run
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/preprocessor/__init__.py", line 270, in apply
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/preprocessor/__init__.py", line 212, in preprocess
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/preprocessor/__init__.py", line 198, in _run_preproc_function
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/preprocessor/_io.py", line 91, in concatenate
    cube = iris.cube.CubeList(cubes).concatenate_cube()
  File "/mnt/lustre01/pf/b/b309057/SOFTWARE/miniconda3/envs/esmvaltool/lib/python3.7/site-packages/iris/cube.py", line 495, in concatenate_cube
    raise ValueError("can't concatenate an empty CubeList")
ValueError: can't concatenate an empty CubeList
"""

  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_main.py", line 228, in run
    conf = main(args)
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_main.py", line 156, in main
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_main.py", line 206, in process_recipe
    recipe.run()
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_recipe.py", line 1216, in run
    self.tasks, max_parallel_tasks=self._cfg['max_parallel_tasks'])
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_task.py", line 596, in run_tasks
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_task.py", line 641, in _run_tasks_parallel
    task.output_files, updated_products = result.get()
  File "/mnt/lustre01/pf/b/b309057/SOFTWARE/miniconda3/envs/esmvaltool/lib/python3.7/multiprocessing/pool.py", line 657, in get
    raise self._value
ValueError: can't concatenate an empty CubeList
```

mattiarighi:
Apparently it does not find the `sftlf` files for NCEP and ERA-Interim...

valeriupredoi:
Mattia, did you use the new syntax in recipe for sftlf as a stand alone
variable? Post that bit pls

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

mattiarighi:
There was a bug in `config-developer.yml` for `OBS`: I fixed it but this introduces another problem.

The `OBS` data were named like:
```
[project]_[dataset]_[type]_[version]_[mip]_[short_name]_*.nc
```
which works for standard variables, e.g. `tas`:
```
OBS_ERA-Interim_reanaly_1_Amon_ta_200001-200012.nc
```
but not for `fx` variables, due to the extra `_`:
```
OBS_ERA-Interim_reanaly_1_fx_sftlf.nc
```

So, I changed `config-developer.yml` and removed the extra `_`:
```
[project]_[dataset]_[type]_[version]_[mip]_[short_name]*.nc
```
But this now generates another problem, since not only `ta` is found, but all variables whose short name starts with `ta`, e.g.:
```
OBS_ERA-Interim_reanaly_1_Amon_tauu_200001-200012.nc
OBS_ERA-Interim_reanaly_1_Amon_tas_200001-200012.nc
OBS_ERA-Interim_reanaly_1_Amon_ta_200001-200012.nc
OBS_ERA-Interim_reanaly_1_Amon_tauv_200001-200012.nc
```

Ideas?

mattiarighi:
This was not an issue before, since we had 2 different entries for `fx` and non-`fx` variables in `config-developer.yml`.

How did you handle this for CMIP5? I guess you have a similar issue there, don't you?

mattiarighi:
Looking at our current observational data pool, it looks like only ERA-Interim is affected by this naming conflict, since the other datasets we have do not include any `fx` variable.
So we could also think about renaming the file in `OBS` to something more convenient.

valeriupredoi:
@mattiarighi the change in `config-developer` you made is not a bug but a feature, we have not yet implemented its functionality in `_data_finder`. The problem is that you will get an error when there are no fx var files (like for OBS) and you are asking just for those in a diagnostic. Lemme have a look, just to confirm, can you pls point me to the recipe you are using?

mattiarighi:
> @mattiarighi the change in config-developer you made is not a bug but a feature, we have not yet implemented its functionality in _data_finder. The problem is that you will get an error when there are no fx var files (like for OBS) and you are asking just for those in a diagnostic.

The problem occurs before the diagnostic script is called.
Changing the definition of `input_file` of `OBS` by removing the `_`, the `fx` file is correctly found for ERA-Interim, but we are no longer able to distinguish, e.g., `ta` from `tas`.

mattiarighi:
> Indeed, looking at config-developer.yml this seems to be done by some institutes already now.

How can that work?
This naming convention is adopted by all our cmorizers at the moment, so if you generate the `OBS` data using the cmorizers you have to stick to such naming convention.

mattiarighi:
Changing the naming convention is an option, but it would be quite some work adapting all cmorizers.

valeriupredoi:
as I said, the `_` should not be removed right now from `OBS/input_file` in config-devel since we have not implemented its functionality for OBS file finding! I am having a test with `recipe_ncl` as we speak and will see what's happening :beer:

valeriupredoi:
again, @mattiarighi can you pls post the recipe you are running

mattiarighi:
It's `recipes/example/recipe_ncl.yml` in `version2_development_align_fx_PR170CORE`.

mattiarighi:
> as I said, the _ should not be removed right now from OBS/input_file in config-devel since we have not implemented its functionality for OBS file finding!

I don't get your point. Isn't the file finder supposed to work with any setting given in `config-developer.yml`?

I understood that this PR is just removing the differentiation between `input_file` and `input_fx_file` and threating both cases as a standard variable.

valeriupredoi:
yes, it's supposed to work with any setting given in config-devel that *works*; in this case passing a formatter:
```
default: '[project]_[dataset]_[type]_[version]_[mip]_[short_name]*.nc'
```
will look for `[short_name]*` and find anything that contains `ta` e.g. `ta`, `tas`, `tauu`, `tacrap`...
Obviously not what you want :grin: 
I have not fixed this in this PR because it's beyond the scope of this PR, or at least I thought so, since I thought there are no fx files for OBS data

valeriupredoi:
> I understood that this PR is just removing the differentiation between input_file and input_fx_file and threating both cases as a standard variable.

That is correct, but it looks like there are caveats, like the OBS situation above. That's why I wanted you guys to test as much as possible :grin: This case now needs to be fixed, I'll do it now

mattiarighi:
Or we simply rename the `fx` file for ERA-Interim to something else, since it is the only fx variable in OBS at the moment.

valeriupredoi:
another aspect I just noticed, that will probably have to be fixed for a more general case is, when providing the `mip` in both `datasets` and `variable`, the mip from `variable` should overwrite the one from `datasets`, which is not the case now

valeriupredoi:
> Or we simply rename the `fx` file for ERA-Interim to something else, since it is the only fx variable in OBS at the moment.

probably much better since we would want to keep the data finding generality rather than start introducing special cases

mattiarighi:
> another aspect I just noticed, that will probably have to be fixed for a more general case is, when providing the mip in both datasets and variable, the mip from variable should overwrite the one from datasets, which is not the case now

Is this happening since this PR or was it the case before as well?

mattiarighi:
> probably much better since we would want to keep the data finding generality rather than start introducing special cases

A quick workaround would be to rename the ERA-Interim fx file like this:
```
OBS_ERA-Interim_reanaly_1_fx_sftlf_.nc
```
until we found a better solution (like @zklaus' one).

zklaus:
As to the feasibility, it seems for the ncl cmorizers there typically is a line towards the end with
```
      fout = output_dir_path + \
        str_join((/"OBS", OBSNAME, TYPE, VERSION, \
                   MIP(vv), VAR(vv), DATESTR/), "_") + ".nc"
```
All that seems to be needed is to remove the `"OBS"` part and move the `VAR` part to the front such that you end up with
```
      fout = output_dir_path + \
        str_join((/VAR(vv), OBSNAME, TYPE, VERSION, \
                   MIP(vv), DATESTR/), "_") + ".nc"
```

valeriupredoi:
> > another aspect I just noticed, that will probably have to be fixed for a more general case is, when providing the mip in both datasets and variable, the mip from variable should overwrite the one from datasets, which is not the case now
> 
> Is this happening since this PR or was it the case before as well?

I haven't done any changes about this in this PR so I am assuming this is from before

mattiarighi:
> All that seems to be needed is to remove the "OBS" part and move the VAR part to the front such that you end up with

Yes, changing the scripts is just about one or two lines, but the whole data pool would need to be renamed on DKRZ and Jasmin, and on any other machine where OBS data are stored.
Given that a lot of people are working to finalize the diagnostic papers, this is probably not the best moment for such a renaming action.

So I would go for the quick workaround for ERA-Interim at the moment.

zklaus:
Well, is it better after the release of 2.0? I think if (and only if) we decide that it is a good idea and should happen in principle we can start with it. Since both old and new filenames are very well defined and will stay in the same directory, we can easily rename them and put a link with the old name in place. If some institute want's to hold back they can also have there own line in `config-developer.yml`.

mattiarighi:
> Well, is it better after the release of 2.0? 

No, but at least we should wait until the papers are submitted.

> Since both old and new filenames are very well defined and will stay in the same directory, we can easily rename them and put a link with the old name in place. If some institute want's to hold back they can also have there own line in config-developer.yml.

Sure, this is not a big issue for the developers. 
I just want to avoid adding complications to the users at this particular stage.

What about changing this during the workshop in October?

valeriupredoi:
#d2898db fixes the problem with the mip, that should have been there eons ago @bouweandela :grin:

mattiarighi:
> d2898db fixes the problem with the mip, that should have been there eons ago @bouweandela

This is a pretty big change, we should test some recipes...

valeriupredoi:
> > [d2898db](https://github.com/ESMValGroup/ESMValCore/commit/d2898db4ce85f7cb0a01811e4a3ab48f936942fa) fixes the problem with the mip, that should have been there eons ago @bouweandela
> 
> This is a pretty big change, we should test some recipes...

it's a logical change (if you asked me) since there is a solid reason why there'd be a mip under select variables

Now you can run the recipe but it will fail at concatenate as before, since it's not finding sftlf data for any of the OBS files (ERA-Interim coz crappy naming formatter and the others coz there's none). I don't know how to handle the missing fx var data - on the one hand we can just put back in the recipe check (if there's no file, exit) but on the other hand we could survive without fx var data since the user may write a diagnostic where they can replace them with NE masks or something

mattiarighi:
> it's a logical change (if you asked me) since there is a solid reason why there'd be a mip under select variables

Yes, but the recipes were designed without this feature, so to be on the safe side we need to test them.

> Now you can run the recipe but it will fail at concatenate as before, since it's not finding sftlf data for any of the OBS files (ERA-Interim coz crappy naming formatter and the others coz there's none). I don't know how to handle the missing fx var data - on the one hand we can just put back in the recipe check (if there's no file, exit) but on the other hand we could survive without fx var data since the user may write a diagnostic where they can replace them with NE masks or something

I think it's fine to get an error if the variable is not there, for the reason you mentioned:

> on the other hand we could survive without fx var data since the user may write a diagnostic where they can replace them with NE masks or something

Let me rename the ERA-Interim fx file and test again.

mattiarighi:
> Let me rename the ERA-Interim fx file and test again.

Now it works until the diagnostic, but that's an issue for #1272, reporting there...

esmvalcore/_recipe.py
```diff
@@ -445,7 +476,10 @@ def _get_input_files(variable, config_user):
                 '\n'.join(input_files))
     if (not config_user.get('skip-nonexistent')
             or variable['dataset'] == variable.get('reference_dataset')):
-        check.data_availability(input_files, variable)
+        if variable['frequency'] != 'fx':
```
valeriupredoi: @mattiarighi @bouweandela so shall we put back in the checks for fx files availability?

valeriupredoi:
well, it works because it found at least a single OBS fx file, otherwise you'd get the concatenation error again -> shalee we put back the checks as in my comment?

esmvalcore/_recipe.py
mattiarighi: Yes, I would put it back.

valeriupredoi:
actually since this happened https://github.com/ESMValGroup/ESMValCore/pull/230 we really *need* to put back in the checks on fx var files availability

bouweandela:
A simple solution to this problem https://github.com/ESMValGroup/ESMValCore/commit/d2898db4ce85f7cb0a01811e4a3ab48f936942fa#issuecomment-531160887 could be to replace the square brackets with other brackets, e.g. `{}`, so we can use the square brackets in globs and then the pattern could look like:
```
{project}_{dataset}_{type}_{version}_{mip}_{short_name}[_.]*nc
```

valeriupredoi:
good call @bouweandela we should put this in a new PR :beer:

esmvalcore/_recipe.py
valeriupredoi: it is le put back :grin:

valeriupredoi:
OK @mattiarighi what's the verdikt? :+1: or :-1: ?

mattiarighi:
@bouweandela dismissed his review, so you have to wait for his approval...

valeriupredoi:
@bouweandela ye olde gnashgab, what say you? :beer:

esmvalcore/_recipe.py
```diff
@@ -910,6 +943,9 @@ def _initialize_variables(self, raw_variable, raw_datasets):
             variable = deepcopy(raw_variable)
             variable.update(dataset)
             variable['recipe_dataset_index'] = index
+            # preserve the MIP from RAW in variable object
+            if 'mip' in raw_variable:
```
bouweandela: I'm not too keen on this, it makes the behaviour more difficult to understand. Now it's use whatever is in variable and overwrite with dataset. This adds an exception that can be confusing and is not needed.

bouweandela:
If you don't want to solve https://github.com/ESMValGroup/ESMValCore/commit/d2898db4ce85f7cb0a01811e4a3ab48f936942fa#issuecomment-531160887 in this pull request, can you at least make an issue for it?

valeriupredoi:
> If you don't want to solve [d2898db#issuecomment-531160887](https://github.com/ESMValGroup/ESMValCore/commit/d2898db4ce85f7cb0a01811e4a3ab48f936942fa#issuecomment-531160887) in this pull request, can you at least make an issue for it?

yes, I will PR the feathers out of it in a different PR, way too much stuff in this one

esmvalcore/_recipe.py
valeriupredoi: moreover for CMIP6 when fx mips differ so much pretty much eveything mip related will have to stay in variable and not in datasets

mattiarighi: Most of the users are used to the current behaviour and changing it without obvious advantages could be dangerous at this stage.

bouweandela: By far the most logical place for the mip is with the variable, because it defines the variable: i.e. project/mip/variable is how variables are defined in the CMOR tables.

>  so even if they have 99 variables with the same mip and only one with a different mip they'll have to put the mip under each of the variables

Yes, but they can use yaml anchors to avoid repeating themselves.

valeriupredoi: ok but then variables that dont have the mip from the datasets will not work without this

valeriupredoi: "most logical place for the mip is with the variable..." -> exactly, but if the mip is in `datasets` that'll get reset and we're back in the dungeons

bouweandela: It has worked fine so far

valeriupredoi: it doesn't, man! look at your code :grin:

mattiarighi: Both approaches have pros and cons, the big advantage of the current one is that everybody is used to it.

valeriupredoi:
```
datasets:
  - {dataset: bcc-csm1-1,  project: CMIP5, mip: Amon, exp: historical,  ensemble: r1i1p1,  start_year: 2000,  end_year: 2002}

diagnostics:
  example:
    description: Example diagnostic
    themes:
      - phys
    realms:
      - atmos
    variables:
      ta:
        preprocessor: preprocessor_1
        reference_dataset: ERA-Interim
      sftlf:
        mip: fx
```
will automatically assign `Amon` to `sftlf`

valeriupredoi: all I'm saying is that if the user wants to declare the mip under their variable, automatic assignment from `datasets` should be disabled, it's a pro option or whatevs you wanna call it, it doesn't turn off the previous functionality that they're used to. It's like an override button

mattiarighi:
> yes, I will PR the feathers out of it in a different PR, way too much stuff in this one

I opened issue #250 so that we won't forget.

esmvalcore/_recipe.py
mattiarighi: This should be the last open issue on this PR.
Any further thoughts?
To stay on the safe side, I would suggest **not** to change this behaviour now.

valeriupredoi:
> > yes, I will PR the feathers out of it in a different PR, way too much stuff in this one
> 
> I opened issue #250 so that we won't forget.

cheers @mattiarighi :beer:

esmvalcore/_recipe.py
valeriupredoi: OK so if yous are so much against this then we'll have to make sure we include a very **clear** instruction in the documentation about how to handle different mip's from `datasets`/`variable` -- I don't agree with you but then again, as long as it works in one way or another then I'm happy(ish) :beer:

mattiarighi: Let's revert the change for now. We can introduce it in a separate PR later.

mattiarighi: @valeriupredoi can you revert this change as discussed? We need to finalize this.

valeriupredoi:
Can you do it pls. Am travelling to Paris today and dont have my laptop
with me :beer:

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

valeriupredoi:
Soz, man, going to the CRESCENDO meeting so I actually wanna listen to the
talks, hence no laptop :grin:

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

esmvalcore/_recipe.py
mattiarighi: Change reverted.

mattiarighi:
Done. Waiting for final approval...

valeriupredoi:
Boss!

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

mattiarighi:
@valeriupredoi please close the related issues.

valeriupredoi:
Wohoo, I will, on Friday when I is back. Cheers, man!

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

Issue #173, opened 2019-07-17
CMORCheck appears to ignore non-monotonic time coordinate


evertrol:
Using `CMORCheck` on a file / cube with a non-monotonic time coordinate does not result in an error.
Iris does note the problem, issues a warning, and changes the dimension coordinate to an auxiliary coordinate. Either this auxiliary coordinate is ignored by `CMORCheck().check_metadata()/check_data()`, or it's simply not checked for monotonicity.

Below is a near-minimal example.

----

Python code:

```python
import sys
import iris
from esmvalcore.cmor.check import CMORCheck
from esmvalcore.cmor.table import CMIP5Info

info = CMIP5Info('cmip5-cmor-tables')
varinfo = info.get_variable('Amon', 'tas')
cube = iris.load(sys.argv[1])[0]
print(cube.coord('time').points)
checker = CMORCheck(cube, varinfo,  fail_on_error=False, automatic_fixes=False)
checker.check_data()
checker.check_metadata()
```

Output (ignore the errors about the latitude & longitude unit; see issue #172 ):

```text
/Users/evert/Library/Python/3.7/lib/python/site-packages/iris/fileformats/_pyke_rules/compiled_krb/fc_rules_cf_fc.py:2076: UserWarning: Failed to create 'time' dimension coordinate: The points array must be strictly monotonic.
Gracefully creating 'time' auxiliary coordinate instead.
  error=e_msg))
[ 15.5  45.5  74.5 105.  135.5 166.  196.5 227.5 258.  288.5 319.  349.5
  15.5  45.   74.5 105.  135.5 166.  196.5 227.5 258.  288.5 319.  349.5
  15.5  45.   74.5 105.  135.5 166.  196.5 227.5 258.  288.5 319.  349.5]
Traceback (most recent call last):
  File "testcase.py", line 13, in <module>
    checker.check_metadata()
  File "/usr/local/lib/python3.7/site-packages/esmvalcore/cmor/check.py", line 104, in check_metadata
    self.report_errors()
  File "/usr/local/lib/python3.7/site-packages/esmvalcore/cmor/check.py", line 122, in report_errors
    raise CMORCheckError(msg)
esmvalcore.cmor.check.CMORCheckError: There were errors in variable tas:
lon: units should be degrees_east, not degrees
 lat: units should be degrees_north, not degrees
in cube:
air_temperature / (K)               (-- : 36; latitude: 2; longitude: 3)
     Dimension coordinates:
[...]

PR #175, opened 2019-07-18
Check time is monotonic

jvegasbsc:
As a bonus, we were limiting time checking to DimCoords (to avoid issues when having time, time_counter, forecast_time). Now if we don't find a DimCoord we look for AuxCoords also

Fixes issue #173


jvegasbsc:
By the way, the main issue here was not checking the auxiliary coordinates. Monotonicity is also ensured by the frequency check. Anyway, it is better to have it as a separate check, as it can help point to the underlying problem

Issue #127, opened 2019-02-13
Fix iris (>2) warnings


schlunma:
Since we will switch to iris 2 very soon now (at least I hope so :grin:), it would be nice to address warnings which still appear in the new version. I will list some of them which regularly appear in my runs, but this is definitely not a complete list. Anyone is invited to add further items.

- [x]  `/iris/coords.py:1355: UserWarning: Collapsing a non-contiguous coordinate. Metadata may not be fully descriptive for 'year'.`
This warning appears when the `time` coordinate is collapsed. The reason are the temporal `AuxCoords` `year`, `day_of_year`, etc. which do not have bounds (which makes sense), so it would be best to disable this warning for those certain coordinates. It also appears when any other coordinate without bounds is collapsed.

- [x] `/numpy/ma/core.py:3174: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use arr[tuple(seq)] instead of arr[seq]. In the future this will be interpreted as an array index, arr[np.array(seq)], which will result either in an error or a different result.`
This appears somewhere in the iris code (I do not know exactly where) and should probably be fixed by iris itself. *edit* This one will be fixed in the next Iris release: https://github.com/SciTools/iris/pull/3212#issuecomment-510095803

- [x] `iris/fileformats/cf.py:798: UserWarning: Missing CF-netCDF measure variable 'areacella', referenced by netCDF variable 'tas'`
Warnings like this appear very often in the preprocessor. Most likely it's because of the `cell_measures` attribute in the raw cubes (e.g. `tas:cell_measures = "area: areacella"`). Would it break something if we removed this attribute?

- [ ] `iris/__init__.py:237: IrisDeprecation: setting the 'Future' property 'cell_datetime_objects' is deprecated and will be removed in a future release. Please remove code that sets this property.`
This should be solved by removing the appearances of this property in the code (in `esmvaltool/cmor/check.py` and `esmvaltool/diag_scripts/autoassess/stratosphere/strat_metrics_1.py`).

- [ ] `iris/analysis/cartography.py:394: UserWarning: Using DEFAULT_SPHERICAL_EARTH_RADIUS.`
This appears when grid cell areas are calculated by `iris.analysis.cartography.area_weights`. The warning does not appear if the Earth radius is specified in the cube. Alternatively, we could use the fx_files to calculate areas.

Not iris warnings, but still appear very often:

[...]

PR #177, opened 2019-07-19
Ignore known iris warnings

bouweandela:
This will reduce the number of warnings coming from iris
- Ignore warnings about missing cell measures
- Ignore warnings about missing time bounds where not needed

See #127


mattiarighi:
Please update the to-do list in #127.

PR #179, opened 2019-07-22
Add fixes for tos (sea surface temperature) for some cmip5 models

zklaus:
This adds fixes for tos (temperature ocean surface, sst) for a couple of cmip5 models.
One thing that surprised me was that our common
```
metadata = cube.metadata
do_something(cube)
cube.metadata = metadata
```
trope did not work for me when adding a scalar to a cube. I had to add a `deepcopy(metadata)` to avoid the loss of important metadata.

Maybe in these cases it is better to work on the data directly like
```
cube.data = cube.core_data() + 273.15
```
so?

### DIFF
--- a/esmvalcore/cmor/_fixes/cmip5/ec_earth.py
+++ b/esmvalcore/cmor/_fixes/cmip5/ec_earth.py
@@ -1,5 +1,6 @@
-
 """Fixes for EC-Earth model."""
+from dask import array as da
+
 from ..fix import Fix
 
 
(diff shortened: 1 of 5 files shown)


mattiarighi:
Looks fine to me, but I haven't tested.
Maybe @ledm can have a look, since `tos` is used in his `recipe_ocean_example.yml`

zklaus:
@ledm is on leave for a couple of weeks. This impacts the paper in the sense that ESMValGroup/ESMValTool#1183 needs it for some models. Maybe it's ok to just wait until everybody is back (I think @bouweandela is also gone at the moment)? Otherwise I could deactivate those models, we can proceed with the other PR, and re-integrate these models later.

@mattiarighi, please advise.

mattiarighi:
If you tested it and it works as expected we can merge it.
It will not break other functionalities.

zklaus:
Perfect. Yes, I tested all of these changes since I needed them to make the other PR work.

bouweandela:
> One thing that surprised me was that our common ... trope did not work for me when adding a scalar to a cube. I had to add a deepcopy(metadata) to avoid the loss of important metadata.

Yes, V also noticed that at some point

> Maybe in these cases it is better to work on the data directly like

Yes, I think so, it is more clear what it going on that way

bouweandela:
>     Maybe in these cases it is better to work on the data directly like

Thinking about it some more, in this case it would probably be most safe to just correct the units and then convert to what is expected by the CMOR table:
```python
cube.units = 'celsius'
cube.convert_units('K')
```

PR #182, opened 2019-07-24
Making ESMValTool available for CORDEX (Regional) Data

aperezpredictia:
First of all, I decided to create the esmvalcore/cmor/_fixes/CORDEX/Cordex_boundaries_guessing.py  file. It was created because the CORDEX data has not boundaries for rlat and rlon, so they were guessed with an Iris functionality. This file is really similar to the fix given by "bcc_csm1_1.py", but with the guessing difference.

Secondly, I added the CORDEX project into the "config-developer.yml" file as defined at the head of the file where a template is given. I also changed the input default directory for some of the other projects in order to get simplicity, but it is not necessary. Moreover, I have created the rootpath in the "config-user.yml"

Third, I have created an if statement before performing the regrid to add the coordinate system of the target grid (ERA5 dataset) before entering into Iris functions. That's because Iris does not let to make a regrid between two cubes whose coordinates systems are not given. In the case of the CORDEX data, they are given. However, the ERA5 cubes had not any coord_system associated to their coordinate variables. You can see this mandatory need in the __call__ function for the iris/analysis/_regrid.py

```
        src_x_coord, src_y_coord = get_xy_dim_coords(src)
        grid_x_coord, grid_y_coord = self._tgt_grid
        src_cs = src_x_coord.coord_system
        grid_cs = grid_x_coord.coord_system

        if src_cs is None and grid_cs is None:
            if not (src_x_coord.is_compatible(grid_x_coord) and
                    src_y_coord.is_compatible(grid_y_coord)):
                raise ValueError("The rectilinear grid coordinates of the "
                                 "given cube and target grid have no "
                                 "coordinate system but they do not have "
                                 "matching coordinate metadata.")
        elif src_cs is None or grid_cs is None:
            raise ValueError("The rectilinear grid coordinates of the given "
                             "cube and target grid must either both have "
                             "coordinate systems or both have no coordinate "
                             "system but with matching coordinate metadata.")
```
That's why I changed it from 'None' to 'GeogCS' with the ellipsoid features given by the WGS84.

In the regrid method of the ESMValTool, you have created a function that decided to use Iris or ESMPy depending on the regrid 'scheme' given in the recipe. When I used ESMPy, it did not recognize the temperature values given in the dataset, so I decided to change it to Iris. For changing it, I undo a step that some of the developer did before at the very beginning of preprocessor/_regrid_esmpy.py where some lines are commented in and commented out.
In the same directory, I changed the function is_lon_circular() due the fact that I did not have boundaries in the longitude coordinate. Therefore, I imported the fix file mentioned before, in order to guess the rlat and rlon boundaries and produce with the function, the longitude and latitude boundaries, making the function is_lon_circular() to work. Now, instead of giving only the longitude, we also have to give the function a cube.

If you have any doubts, recommendation on how to do it more clear, or concerns... Let me know!
Thank you in advance.

### DIFF
--- a/esmvalcore/cmor/_fixes/CORDEX/Cordex_boundaries_guessing.py
+++ b/esmvalcore/cmor/_fixes/CORDEX/Cordex_boundaries_guessing.py
@@ -6,7 +6,7 @@ from scipy.ndimage import map_coordinates
 from ..fix import Fix
 
 
-class Tos(Fix):
+class CordexFix(Fix):
     """Fixes for tos."""
 
     def fix_data(self, cube):
@@ -31,8 +31,8 @@ class Tos(Fix):
             rlat, np.arange(len(rlat)), k=1)
         rlon_to_idx = InterpolatedUnivariateSpline(
             rlon, np.arange(len(rlon)), k=1)
-        rlat_idx_bnds = rlat_to_idx(cube.coord('grid_latitude').bounds())
-        rlon_idx_bnds = rlon_to_idx(cube.coord('grid_longitude').bounds())
+        rlat_idx_bnds = rlat_to_idx(cube.coord('grid_latitude')._guess_bounds())
+        rlon_idx_bnds = rlon_to_idx(cube.coord('grid_longitude')._guess_bounds())
 
         # Calculate latitude/longitude vertices by interpolation
         lat_vertices = []
--- a/esmvalcore/cmor/_fixes/cmip5/bcc_csm1_1.py
+++ b/esmvalcore/cmor/_fixes/cmip5/bcc_csm1_1.py
@@ -31,8 +31,8 @@ class Tos(Fix):
             rlat, np.arange(len(rlat)), k=1)
         rlon_to_idx = InterpolatedUnivariateSpline(
             rlon, np.arange(len(rlon)), k=1)
-        rlat_idx_bnds = rlat_to_idx(cube.coord('grid_latitude').bounds)
-        rlon_idx_bnds = rlon_to_idx(cube.coord('grid_longitude').bounds)
+        rlat_idx_bnds = rlat_to_idx(cube.coord('grid_latitude')._guess_bounds())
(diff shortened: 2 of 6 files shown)


zklaus:
Nice work! This pull request is rather large and deals with some separate things. We like to keep things as self-contained as possible. This has several advantageous: For example, the discussion in the pull request can focus on one topic and doesn't get confusing by having several threads intermingled, and clear changes can get merged fast and are not held back by unrelated more contested changes.

For this reason I suggest to split up this pull request into three pull requests:

- One to just add the project in config-developer.yml and config-user.yml
- One that adds the coordinate fix
- One that deals with regridding.

In fact, I already prepared the corresponding three branches. Since with current permissions I cannot write to your repository and I did not want to just take your work to the main repository, I created these branches in my own fork.
Please have a look at zklaus/ESMValCore@382f38510ebf3f7d58892d0d3505dba35e86bb3f, zklaus/ESMValCore@e69781f9f8266d5c899458b631410cf0ba9d9068, and zklaus/ESMValCore@bc2ae2ada5e29446fe0065ed554ed2da1ef40e45.

You can either take those commits into your own repository and open the corresponding PRs, or invite me to your repository so that I can push them myself.

How do you want to proceed?

aperezpredictia:
@zklaus I have invited you to my repository.

zklaus:
Perfect. I have accepted the invitation and pushed the three branches. Now you should be able to open the corresponding PRs. Since the branches depend on each other, I would start with the CORDEX_add_project PR and then take it from there.

aperezpredictia:
@zklaus I have created the first one in #183 . Should I create the others or wait till the first is accepted?

zklaus:
I think you can already create them. Then we can close this one and have the full history preserved.

aperezpredictia:
> I think you can already create them. Then we can close this one and have the full history preserved.

DONE

zklaus:
Perfect! Let's continue the discussion there.

PR #184, opened 2019-07-24
Add general CORDEX coordinate boundary fix

aperezpredictia:
## Description

CORDEX has not rlat / rlon / lat / lon boundaries, so we have to use the function already created for the bcc but guessing the boundaries of rlat / rlon in order to interpolate later them to lat / lon.

This PR is part of #182 , which is being separated into parts.

In the documentation, we should write something like '''Fixes for cordex-eur11-44''' instead of '''Fixes for bcc-csm1-1.'''

Closes #issue_number

Link to documentation:

***

## [Before you get started](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#getting-started)

- [ ] [☝ Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do

## [Checklist](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#checklist-for-pull-requests)

It is the responsibility of the author to make sure the pull request is ready to review. The icons indicate whether the item will be subject to the [🛠 Technical][1] or [🧪 Scientific][2] review.

[1]: https://docs.esmvaltool.org/en/latest/community/review.html#technical-review
[2]: https://docs.esmvaltool.org/en/latest/community/review.html#scientific-review

- [ ] [🧪][2] The new functionality is [relevant and scientifically sound](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#scientific-relevance)
- [ ] [🛠][1] This pull request has a [descriptive title and labels](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-title-and-label)
- [ ] [🛠][1] Code is written according to the [code quality guidelines](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#code-quality)
- [ ] [🧪][2] and [🛠][1] [Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available
- [ ] [🛠][1] [Unit tests](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#tests) have been added
- [ ] [🛠][1] Changes are [backward compatible](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#backward-compatibility)
- [ ] [🛠][1] Any changed [dependencies have been added or removed](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#dependencies) correctly
[...]

PR #185, opened 2019-07-24
Adapt regridding to CORDEX needs

aperezpredictia:
First, we added the coordinate system to those coordinates that have not one to avoid iris error ( better explained in #182 ).

Second, I commented out and commented in the command lines in the inverse order in order to make ESMValTool to use Iris.

Third, I opened a try/except clause in the is_lon_circular() function because lon.bounds does not exist unless you use the fix given in #184 .

This pull request is a small part of #182


zklaus:
This PR mixes fixes in data files with some changes to regridding that imho are likely to be unnecessary once we address other aspects of CORDEX data properly.

We will have a closer look at this after the first step of making CORDEX data available, which is addressed in another PR.

PR #192, opened 2019-07-26
Add ocean heat content (ohc) derived variable

zklaus:
This adds a new derived variable for the ocean heat content per cell.
It's the core part of ESMValGroup/ESMValTool#922.


zklaus:
Will do.

zklaus:
@bouweandela ok, I think this is ready and has a test now. Could you have another look?

PR #194, opened 2019-07-31
Added fix for wrong parent_time_units attribute for UKESM1-0-LL

schlunma:
Some CMIP6 set the `parent_time_units` attribute like this:

```
:parent_time_units = "days since 1850-01-01-00-00-00" ;
```

This can't be read by `cf_units`. This PR fixes this issue.


zklaus:
Good catch, @schlunma , this needs fixing! However, since this is really an invalid format, I wonder if we should rather rely on model fixes (and of course on telling the modellers)?

schlunma:
I first found this issue for `HadGEM3-GC31-LL` and wrote a fix. Right after that, it appeared for `UKESM1-0-LL` and I decided to fix it directly in the check function. If only these two models are affected, fixes would probably be best. If it is necessary for more, I think the general approach is better.

zklaus:
I guess the main concern is to make sure we notice it and let the modellers know since for CMIP6 this data still can and absolutely should be fixed on the esgf. But maybe a big warning in the general code could do the trick?

sloosvel:
I have been using HadGEM-GC31-LL data that has the right parent_time_unit format, so maybe this only affects some variables.

But I think that the general approach goes more in line with the request that was made during the workshop to relax the checks to make the tool not stop every time it finds an issue with the standards.

sloosvel:
I've also just had issues with models that have the parent_time_unit set to "no parent" so maybe this case could also considered added in this fix.

zklaus:
> I've also just had issues with models that have the parent_time_unit set to "no parent" so maybe this case could also considered added in this fix.

Actually I think this should be considered a bug in our original code to modify these attributes. Sorry I didn't catch it at the time. But this should be dealt with by re-reading [CMIP6_global_attributes_filenames_CVs](http://goo.gl/v1drZl) and including the correct detection of the absence of a parent experiment together with the appropriate (non-)conversion there.

schlunma:
Because of the discussion and #209 I changed this PR to only include the `parent_time_units` fix for UKESM1-0-LL.

valeriupredoi:
just hit this bug today myself, I asked the boffins who did the CMORization what's what with that `:parent_time_units = "days since 1850-01-01-00-00-00"`

bouweandela:
Is this still needed, given #231?

schlunma:
> Is this still needed, given #231?

I think with #231 the exception is replaced by a warning (not tested, though), so this fix is not necessary anymore to load the model. However, we added a [similar fix](https://github.com/ESMValGroup/ESMValCore/blob/development/esmvalcore/cmor/_fixes/cmip6/hadgem3_gc31_ll.py) for HadGEM3-GC31-LL, so why not for UKESM1-0-LL?

Issue #26, opened 2019-05-22
Cube concatenation failure for ocean thetao IPSL data


valeriupredoi:
**Describe the bug**
While running @tillku ocean heat recipe with IPSL CMIP6 data I encountered this problem:
Failure to concatenate cubeList:

```
In [4]: cubeList = iris.load("/badc/cmip6/data/CMIP6/CMIP/IPSL/IPSL-CM6A-LR/historical/r1i1p1f1/Omon/thetao/gn/latest/thetao_Omon_IPSL-CM6A-LR_his
   ...: torical_r1i1p1f1_gn_195001-201412.nc")
In [5]: cubeList                                                                                                                                  
Out[5]: 
[<iris 'Cube' of cell_area / (m2) (-- : 332; -- : 362)>,
<iris 'Cube' of sea_water_potential_temperature / (degC) (time: 780; Vertical T levels: 75; -- : 332; -- : 362)>]
In [10]: cubeList[0].var_name                                                                                                                     
Out[10]: 'area'

In [11]: cubeList[1].var_name                                                                                                                     
Out[11]: 'thetao'
```

It looks like the `thetao` variable is packaged in two separate cubes - one for the variable itself and one for the cell measures. This is not a problem for eg UKESM1 data that comes correctly cmorized eg
```
<iris 'Cube' of sea_water_potential_temperature / (degC) (time: 600; depth: 75; cell index along second dimension: 330; cell index along first dimension: 360)>
```

I am not sure what the best approach would be here: the IPSL data that I am looking at is already on ESGF so a rerun/replace of the data production will happen for a later version but not soon, so we'll have to account for this in our code before (and if they actually fix it).

**Full trace**
```
2019-05-22 12:43:55,612 UTC [14724] ERROR   Failed to run fix_metadata([<iris 'Cube' of cell_area / (m2) (-- : 332; -- : 362)>, <iris 'Cube' of sea_water_potential_temperature / (degC) (time: 780; Vertical T levels: 75; -- : 332; -- : 362)>], {'project': 'CMIP6', 'dataset': 'IPSL-CM6A-LR', 'short_name': 'thetao', 'cmor_table': 'CMIP6', 'mip': 'Omon', 'frequency': 'mon'})
2019-05-22 12:43:56,366 UTC [46729] ERROR   Program terminated abnormally, see stack trace below for more information
multiprocessing.pool.RemoteTraceback: 
"""
Traceback (most recent call last):
[...]

PR #203, opened 2019-08-12
Add ipsl fix for cmip6

jvegasbsc:
From  #26 

Replaces ESMValGroup/ESMValTool#1100


jvegasbsc:
It also removes lots of duplicated coded I introduced by mistake

jvegasbsc:
No, is just a duplication of a folder that got committed by error

Issue #94, opened 2019-06-17
extract_region preprocessor no longer works on irregular grids


ledm:
Hi,

we recently changed the irregular grid section of the `extract_region` preprocessor in `preprocessor/_area.py` from this:

```
    # irregular grids
    lats = cube.coord('latitude').points
    lons = cube.coord('longitude').points
    mask = np.ma.array(cube.data).mask
    mask += np.ma.masked_where(lats < start_latitude, lats).mask
    mask += np.ma.masked_where(lats > end_latitude, lats).mask
    mask += np.ma.masked_where(lons > start_longitude, lons).mask
    mask += np.ma.masked_where(lons > end_longitude, lons).mask
    cube.data = np.ma.masked_where(mask, cube.data)
    return cube
```

to this:
```
    # irregular grids
    lats = cube.coord('latitude').points
    lons = cube.coord('longitude').points
    select_lats = start_latitude < lats < end_latitude
    select_lons = start_longitude < lons < end_longitude
    selection = select_lats & select_lons
    data = da.ma.masked_where(~selection, cube.core_data())
    return cube.copy(data)
```

The new code doesn't work for irregular grids. It produces the error message:
```
  File "/users/modellers/ledm/workspace/ESMValToolTest/ESMValTool/esmvalcore/preprocessor/_area.py", line 82, in extract_region
    select_lats = start_latitude < lats < end_latitude
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

```
Any suggestions on how to fix it and keep it lazy? For now, I've reverted my local code back to the previous version. (but I'm not going to push it!)

The reason that we didn't catch the problem at the time is because the ocean recipes that use this were designed to demonstrate the diagnostics, not test the preprocessor. I'm changing recipe_ocean_example.yml to include an irregular grid CMIP5 model for future testing.

bouweandela:
Hi Lee,

Maybe we can solve this issue together in the following way:
1) you create a new branch and draft pull request in which you implement a unit test that demonstrates that it is broken
2) I fix the function so it passes the unit test
Do you think that could work?

ledm:
I should be okay to write most of a test - once I get the new split code installed! 

To write this test, I'll need to create an iris cube with an irregular grid - I've never had to do this before. Is there an example somewhere of how you create an irregular gridded cube in iris, @bouweandela 
 or @bjlittle?

ledm:
[...]

PR #204, opened 2019-08-14
Make 2d and 1d region extraction consistent

jvegasbsc:
There were two issues with consistency between regular and irregular grids region extraction

- Regular was being extracted if any point in the cell belonged to the region, 2D only if the center is inside
- Regular was adding points in the boundaries case,  2D was not

Now the behaviour is changed so we always use the center and boundaries are included

Detected by @mwjury

### DIFF
--- a/esmvalcore/preprocessor/_area.py
+++ b/esmvalcore/preprocessor/_area.py
@@ -59,14 +59,16 @@ def extract_region(cube, start_longitude, end_longitude, start_latitude,
     if cube.coord('latitude').ndim == 1:
         region_subset = cube.intersection(
             longitude=(start_longitude, end_longitude),
-            latitude=(start_latitude, end_latitude))
+            latitude=(start_latitude, end_latitude),
+            ignore_bounds=True,
+        )
         region_subset = region_subset.intersection(longitude=(0., 360.))
         return region_subset
     # irregular grids
     lats = cube.coord('latitude').points
     lons = cube.coord('longitude').points
-    select_lats = start_latitude < lats < end_latitude
-    select_lons = start_longitude < lons < end_longitude
+    select_lats = start_latitude <= lats <= end_latitude
+    select_lons = start_longitude <= lons <= end_longitude
     selection = select_lats & select_lons
     data = da.ma.masked_where(~selection, cube.core_data())
     return cube.copy(data)


esmvalcore/preprocessor/_area.py
```diff
@@ -59,14 +59,16 @@ def extract_region(cube, start_longitude, end_longitude, start_latitude,
     if cube.coord('latitude').ndim == 1:
         region_subset = cube.intersection(
             longitude=(start_longitude, end_longitude),
-            latitude=(start_latitude, end_latitude))
+            latitude=(start_latitude, end_latitude),
+            ignore_bounds=True,
+        )
```
valeriupredoi: it'd be useful if you added a comment that `ignore_bounds` defaults to `False` - I had to look up the iris documentation to be 100% sure the default is false

jvegasbsc: Ok

mattiarighi:
@zklaus are you OK with this?

zklaus:
Hm, couple of points:

- I am not sure if we should really only consider the centers or also the bounds. Intuitively I would go for the bounds, but since this is somewhat a matter of application, I think it should be reflected in the doc string and perhaps be made an option.
- In a slightly different context Lee stumbled upon a problem with `a < b < c` style comparison with dask which seems to accept only simple comparison. I'll try to dig that up and see if it is a problem here.
- For me the biggest issue with this function has always been that for irregular grids it only masks. Of course, masking is necessary, but to reduce the amount of data (in some cases drastically), it would be nice to also use `intersection` to restrict to the covering rectangle.

bouweandela:
> In a slightly different context Lee stumbled upon a problem with a < b < c style comparison with dask which seems to accept only simple comparison. I'll try to dig that up and see if it is a problem here.

Yes, `extract_region` is broken for irregular grids, see https://github.com/ESMValGroup/ESMValCore/issues/94. It got broken because there are no unit tests for this code and the persons who tested the pull request that broke it accidentally did not use data with an irregular grid -> unit tests should be added and the code fixed. However, i think that's may be beyond the scope of this pull request.

zklaus:
> However, i think that's may be beyond the scope of this pull request.

Hm, fair enough. I guess my minimal request for this PR would be to document the center vs bounds issue in the doc string, the rest seems ok.

@jvegasbsc, if you feel on the other hand that there is not much point in homogenizing two functionalities where one of them isn't working anyway, the ideas above together with the fix in #94 are an option. :smiling_imp:

bouweandela:
I just made pull request #243, that hopefully fixes #94.

PR #205, opened 2019-08-14
Always check monotonicity and direction

sloosvel:
A condition was preventing the automatic fix of the coordinates' monotonicity and direction.


jvegasbsc:
Thanks, Saskia!

valeriupredoi:
cheers @sloosvel :beer: also, @jvegasbsc for introducing and then removing all those duplicates :rofl:

PR #207, opened 2019-08-19
Allow concatenation of diferent ensembles

jvegasbsc:
@mwjury encountered a dataset in which historical and projection use different ensemble names. For example: historical is 'r1i1850p1' and rcp85 is 'r1i2010p1' (I don remember the exact ones, but they include the initial year). 

The only thing that prevents the current version to do this is the recipe schema, that forces the ensemble to be a string. I just updated it to allow it to be a string or a list of strings, as we allow with the exp key


esmvalcore/config-developer.yml
```diff
@@ -21,8 +21,8 @@
 CMIP6:
   input_dir:
     default: '/'
-    BADC: '[institute]/[dataset]/[exp]/[ensemble]/[mip]/[short_name]/[grid]/[latestversion]'
-    DKRZ: '[institute]/[dataset]/[exp]/[ensemble]/[mip]/[short_name]/[grid]/[latestversion]'
+    BADC: '[activity]/[institute]/[dataset]/[exp]/[ensemble]/[mip]/[short_name]/[grid]/[latestversion]'
```
bouweandela: These changes look unrelated to this PR

jvegasbsc: Yes, the bad habit of commit -a bit me

jvegasbsc:
Done! By the way, I was not able to found any mention to the concatenation of experiments, so I also added it

mattiarighi:
Please close the related issue, if there is one.

jvegasbsc:
There is no one, I directly created the pull request

PR #208, opened 2019-08-22
Installing esmvalcore on OSX

valeriupredoi:
installing using @bouweandela newly created `noarch` package


valeriupredoi:
ok @bouweandela updated the instructions and added warning that these are not continuously tested via CircleCI - let's get this in :beer:

bouweandela:
Unfortunately I cannot test this myself either, because I do not have a Mac. Maybe you can ask someone with a Mac to review? If this installation instructions are identical to the linux instructions, wouldn't it be better to just refer to those instead of copy/pasting them?

valeriupredoi:
let's get this and https://github.com/ESMValGroup/ESMValTool/pull/1257 in and close https://github.com/ESMValGroup/ESMValTool/issues/1253 - testing is not possible on a regular basis but I have access to a Mac and can do that on a say 2-monthly basis; it's better than having no mention at all about Mac's -> @mattiarighi and @zklaus pls comment/accept :beer:

valeriupredoi:
also @bjoernbroetz

CONTRIBUTING.md
```diff
@@ -21,6 +21,22 @@ To install in development mode, follow these instructions.
 -   Test that your installation was succesful by running `esmvaltool -h`.
 -   If you log into a cluster or other device via `ssh` and your origin machine sends the `locale` environment via the `ssh` connection, make sure the environment is set correctly, specifically `LANG` and `LC_ALL` are set correctly (for GB English UTF-8 encoding these variables must be set to `en_GB.UTF-8`; you can set them by adding `export LANG=en_GB.UTF-8` and `export LC_ALL=en_GB.UTF-8` in your origin or login machines' `.profile`)
 
+## Getting started on MacOS OSX-64
+To install on a MacOS you will have to perform the following installation procedure:
+
+**WARNING**
+The Mac OSX installation is not being continuously tested via CircleCI! If you notice any problems please report them on GitHub.
+
+  - Download and install Anaconda [for OSX](https://repo.anaconda.com/archive/Anaconda3-2019.07-MacOSX-x86_64.sh)(install by e.g. `bash Anaconda3-2019.07-MacOSX-x86_64.sh` or whatever the installer is called);
```
zklaus: I would prefer to link to the distribution page (https://www.anaconda.com/distribution/) instead of the direct link to the package. This is more likely to stay useful long-term and allows the user to see any notes from anaconda.

Also, I personally prefer miniconda (https://docs.conda.io/en/latest/miniconda.html). Functionally equivalent, but a lot fewer packages that we don't need.

CONTRIBUTING.md
```diff
@@ -21,6 +21,22 @@ To install in development mode, follow these instructions.
...
+  - Answer yes when, at the end of the installation, Anaconda asks you to initialize the installation in your `.bash_profile`;
```
zklaus: If we go with miniconda, this needs to changed here as well. Maybe `conda` is neutral?

CONTRIBUTING.md
```diff
@@ -21,6 +21,22 @@ To install in development mode, follow these instructions.
...
+  - Clone the ESMValTool Core Git repository: `git clone git@github.com:ESMValGroup/ESMValCore`
```
zklaus: Is git by default available on osx? Otherwise, a quick `conda install git` might be a good idea, but we would need a pristine installation to know for sure.

Also, we should probably add ".git" to the end of the url, ie
`git clone git@github.com:ESMValGroup/ESMValCore.git`

CONTRIBUTING.md
```diff
@@ -21,6 +21,22 @@ To install in development mode, follow these instructions.
...
+  - Go to the source code directory: `cd ESMValCore`
+  - Use the `environment.yml` file to create the conda viartual environment, as explained above for a Linux installation: `conda env create -n esmvaltool -f environment.yml`
```
zklaus: Typo in viartual.

CONTRIBUTING.md
```diff
@@ -21,6 +21,22 @@ To install in development mode, follow these instructions.
...
+  - Activate the esmvaltool environment: `conda activate esmvaltool`
+  - Install in development mode: `pip install -e '.[develop]'`. If you are installing behind a proxy that does not trust the usual pip-urls you can declare them with the option `--trusted-host`, e.g. `pip install --trusted-host=pypi.python.org --trusted-host=pypi.org --trusted-host=files.pythonhosted.org -e .[develop]`
+  - Test that your installation was succesful by running `esmvaltool -h`.
+  - Note that the installation instructions are identical to the Linux installation due to the package being architecture agnostic (specified by the `noarch` label).
```
zklaus: Is this note really useful? After all, we just read through a separate set of instructions and at any rate, who will care?

CONTRIBUTING.md
valeriupredoi: it's neutral, it'll be the same prompt

CONTRIBUTING.md
valeriupredoi: yeah good point dude, I changed it

valeriupredoi:
cheers K-man, I plugged in the recommendations - about Anaconda/Miniconda - I like Miniconda better too but the other instructions are all Anaconda-based so if we change, we should probably change for all the instructions - something we should probably do but not in this wee PR :beer:

CONTRIBUTING.md
zklaus: I would change "Anaconda" in line 31 to "conda", but oh well.

zklaus:
Fair enough about the miniconda/anaconda thing. What about an explicit mention of `ESMValTool.git`?

valeriupredoi:
> Fair enough about the miniconda/anaconda thing. What about an explicit mention of `ESMValTool.git`?

eugh, I gotta go change the `CONTRIBUTING.md` on the ESMValTool side, there is a PR there for exactly this and need to fish it out :fish:

zklaus:
Alright, then maybe let's just finish this and open a new issue with a link to the ESMValTool one to update here once that is done. If you agree with this, I'll have one last pass and give my thumbs up.

valeriupredoi:
dis https://github.com/ESMValGroup/ESMValTool/pull/1257

bouweandela:
Looking through this installation instructions, I'm wondering why we need Mac OS specific instructions at all? They appear to be identical to the non-os specific instructions?

valeriupredoi:
Because we need to tell the Mac users to get the specific OSX ana/miniconda and tell them if they encounter problems they should report them. Plus, it's always nice for the new user to see clearly that we support Macs especially since Mac users are always seeing themselves as more special, so it's good to acknowledge their 'oh boy, look, I am a Mac user' status :grin:

bouweandela:
Maybe we can simply update the link to the conda installation page so it points here?
https://conda.io/projects/conda/en/latest/user-guide/install/index.html
and add a note that we officially only support linux, but users have reported succesfully installing on mac and linux subsystem for windows too?

I wouid really prefer not to duplicate all the installation instructions, because the contributing guide is already intimidatingly long and if we update something we will need to carefully update it in all places and it will quickly become a mess.

valeriupredoi:
Bouwe, duplicating instructions sets should not be a factor here - support for users and detailed instructions are imho paramount. Moving the instructions to the user guide might be a good idea tho, but being shy on detailed instructions is something we shouldn't do :beer:

zklaus:
On the other hand every bit of information is a bit that needs maintaining. Code rot happens to documentation too.

bouweandela:
I've already spoken to several people who didn't read the documentation because it looks intimidatingly long..

valeriupredoi:
> On the other hand every bit of information is a bit that needs maintaining. Code rot happens to documentation too.

Absolutely! That's why I said I can regularily test this once every month or two

> I've already spoken to several people who didn't read the documentation because it looks intimidatingly long..

Sorry, Bouwe man, but those people need to read through and figure out what they need to do, starting to remove things from the documentation just because it's long and it intimidates technophobes is not something we should do.

I'll let you guys decide where to put this - no preferences in my case, but I strongly support a documentation that is comprehensive rather than one that's short and omits useful information :beer:

valeriupredoi:
an example: https://www.amazon.co.uk/Sherman-Medium-Tank-Crew-Manual/dp/1935700812/ref=pd_aw_sbs_14_1/261-9087780-3328831?_encoding=UTF8&pd_rd_i=1935700812&pd_rd_r=c26dee79-ac50-4360-959d-3885d30511af&pd_rd_w=hisI4&pd_rd_wg=u3xm4&pf_rd_p=15216146-4434-47f5-99d9-56240c85bf4d&pf_rd_r=N7G4ZWYK7AHG1DCJSVQ8&psc=1&refRID=N7G4ZWYK7AHG1DCJSVQ8 - 118 pages to learn how to operate a tank, I think we're not too far away from the complexity of a tank in our case :grin:

valeriupredoi:
since @bvreede has just told us on the ESMValTool she's gonna be looking into OSX deployment, I added her here :+1: - this is ancient stuff, but it may still be useful

bvreede:
Thanks a bunch @valeriupredoi -- there is a lot of history to be aware of :)

bouweandela:
@valeriupredoi Can this be closed? i think OSX support now works, right?

valeriupredoi:
indeed it is :+1: https://github.com/ESMValGroup/ESMValCore/actions/runs/1080622626

valeriupredoi:
cheers to @bvreede - our resident OSX firefighter :fire_engine:

PR #209, opened 2019-08-23
Fix hadgem3 parent_time_units error

jvegasbsc:
I have detected that at least some HadGEM-GC31-LL files have a bad format in parent time units.

This pull request add a fix for it. It also contains #106, so it should be added after it


esmvalcore/cmor/check.py
```diff
@@ -283,12 +298,28 @@ def _check_dim_names(self):
...
+                            coord.var_name = coordinate.out_name
+                        else:
+                            self.report_error(
+                                'Coordinate {0} has var name {1}'
+                                'Coordinate {0} has var name {1} '
+                                'instead of {2}',
+                                coordinate.name,
+                                coord.var_name,
+                                coordinate.out_name,
+                                )
```
valeriupredoi: codacy complaining about the indentation for the `)` - not sure why, maybe you have some blank space or smth?

sloosvel:
I think that #194 is already taking care of that

ledm:
I think that the CMIP6 UKESM model data also has the same faults.

PR #210, opened 2019-08-26
Use glob in datafinder

jvegasbsc:
A small change to allow users to use wildcard characters in path definiton (we already allow it on files).

I added it because of a quirk our observations convention has: variable folder can just be the variable name  or it can have appended the original frequency used for the computation:

Example:
- psl monthly means for ERA5 if directly provided by ECMF: `/esarchive/recon/ecmwf/era5/daily_mean/psl`

- psl monthly means for ERA5 if computed from hourly outputs provided by ECMF: `/esarchive/recon/ecmwf/era5/daily_mean/psl_f1h`

Both will be found with the new code if folder path is defined `whatever\[short_name]*` but I will need to add a extra tag for the user to provide with the old version

As a bonus, things like PRIMAVERA workspaces in jasmin (which are primavera1 to primavera5) can be add now as only one 'primavera?'


valeriupredoi:
not a big fan of `glob.glob` man, why not use `os.walk` :beer:

jvegasbsc:
Because `os.walk` does not support wildcards and that is the reason to use `glob`. Also, I do not want to walk the whole tree

valeriupredoi:
ok, the "I do not walk the whole tree" sold me :grin:

mattiarighi:
Can this be extended to also address #250?

jvegasbsc:
> Can this be extended to also address #250?
Yes, will do it

Another question is how to make both work with #206

jvegasbsc:
Done!

esmvalcore/config-developer.yml
```diff
@@ -127,63 +127,63 @@ CMIP5:
...
-    BSC: '[type]/[institute.lower]/[dataset.lower]/[freq_folder]/[short_name][freq_base]'
-    RCAST: '[dataset]'
+    default: 'Tier{tier}/{dataset}'
+    BSC: '{type}/{institute.lower}/{dataset.lower}/{freq_folder}/{short_name}{freq_base}'
+    RCAST: '{dataset}'
   input_file:
-    default: '[project]_[dataset]_[type]_[version]_[mip]_[short_name]_*.nc'
-    BSC: '[short_name]_*.nc'
-    RCAST: '[short_name]_[mip]_[type]_[dataset]_*.nc'
+    default: '{project}_{dataset}_{type}_{version}_{mip}_{short_name}_*.nc'
```
mattiarighi:
```suggestion
    default: '{project}_{dataset}_{type}_{version}_{mip}_{short_name}[_]*.nc'
```
This would make the `_` optional, right?

mattiarighi: See discussion in #170

bouweandela: Yes, it should be 
```yaml
    default: '{project}_{dataset}_{type}_{version}_{mip}_{short_name}[._]*nc'
```
so it matches files with and without dates in the name, but not `short_name`s starting with the same characters, e.g. `ta` and `tas`.

esmvalcore/_data_finder.py
```diff
@@ -115,7 +114,7 @@ def _replace_tags(path, variable, fx_var=None):
                            "your recipe entry".format(tag, variable))
 
         paths = _replace_tag(paths, original_tag, replacewith)
-
+    print(paths)
```
```suggestion
```

mattiarighi:
We should test this on several machines/drs'.
I added @valeriupredoi and @bascrezee as reviewers.

valeriupredoi:
Ok Ill test it out myself asap

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

mattiarighi:
`recipe_check_obs` is a good test case for this PR.

valeriupredoi:
hold merging, this breaks the data finding for fx variables. I am taking a look now

valeriupredoi:
so why was `config-developer.yml` not aligned from `development`?

mattiarighi:
There were some conflicts which I fixed, maybe I overlooked something.

mattiarighi:
But now you reverted all the changes of this PR, we use `{` instead of `[`...

valeriupredoi:
yeah I just saw that, need to revert the revert and fix the issues manually

valeriupredoi:
nobody move an inch, am making the changes now

bascrezee:
Testing is ongoing here.

bascrezee:
> Testing is ongoing here.

What is the meaning of this? I do not specify any preprocessor, but it does call a 'default' preprocessor. This seems not to occur for OBS.
`INFO    Creating preprocessor 'default' task for variable 'tas'`

valeriupredoi:
> > Testing is ongoing here.
> 
> What is the meaning of this? I do not specify any preprocessor, but it does call a 'default' preprocessor. This seems not to occur for OBS.
> `INFO Creating preprocessor 'default' task for variable 'tas'`

any time you don't specify a preprocessor name for a variable, it will create the `default` one -> this is per variable and not per dataset; the `default` preprocessor will contain only the basic preproc steps: load, cmor check, fix, extract time etc (note that extract time does not happen for default prprocs for fx varibales that are time-invariant)

PR #211, opened 2019-08-27
Better error messages

jvegasbsc:
Improved error management in some parts of the CMOR module. New messages implemented and fixed a couple that were not working as intended

### DIFF
--- a/esmvalcore/cmor/table.py
+++ b/esmvalcore/cmor/table.py
@@ -84,7 +84,7 @@ class CMIP6Info(object):
             try:
                 self._load_table(json_file)
             except Exception:
-                msg = "Exception raised when loading %s" % json_file
+                msg = f"Exception raised when loading {json_file}"
                 # Logger may not be ready at this stage
                 if logger.handlers:
                     logger.error(msg)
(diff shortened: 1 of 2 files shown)


esmvalcore/cmor/table.py
```diff
@@ -81,15 +81,28 @@ def __init__(self, cmor_tables_path, default=None):
         for json_file in glob.glob(os.path.join(self._cmor_folder, '*.json')):
             if 'CV_test' in json_file or 'grids' in json_file:
                 continue
-            self._load_table(json_file)
+            try:
+                self._load_table(json_file)
+            except Exception:
+                msg = "Exception raised when loading %s" % json_file
```
bouweandela:
```suggestion
                msg = f"Exception raised when loading {json_file}"
```
It would be nice to start using f-strings everywhere, they are more readable and more concise. If you don't like them, you can also use `.format()`, but I wouldn't start using % formatting (except in logger function calls), because we're not using that anywhere so it gives the code an inconsistent appearance

jvegasbsc: I didn't know they existed! They are way better than the old ones.

PR #216, opened 2019-08-29
Fix regrid

jvegasbsc:
I added a regrid from irregular grid to another irregular grid (from the U points to the T points in NEMO terms) in the computation of the ocean heat transport derived variable.

It worked well but for some auxiliary coordinates that were not added to the new cube. this pull request fix this


jvegasbsc:
Maybe the easiest way to test is to use the branch in #217 and derive ohuo

bouweandela:
@zklaus Could you please have a look at this?

PR #218, opened 2019-08-30
Add detrend preprocessor function

jvegasbsc:
Add a new preprocessor function that performs a linear or constant detrend using `scipy.signal`.

### DIFF
--- a/doc/esmvalcore/preprocessor.rst
+++ b/doc/esmvalcore/preprocessor.rst
@@ -811,7 +811,7 @@ Note that this function uses the expensive ``interpolate`` method from
 
 See also :func:`esmvalcore.preprocessor.extract_trajectory`.
 
-.. _detrend
+.. _detrend:
 
 Detrend
 =======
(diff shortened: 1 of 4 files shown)


jvegasbsc:
> maybe we should create a dedicated package for stats operations (in the future);

Good ide

> a side question: how is detrending different from obtaining the least squares errors? It sounds to me they are the same fish; also it's worth specifying this method is only the linear one

In this case, what the method I'm adding is doing is to compute the linear trend of the data along a given axis and the substracting that trend. In the case of the constant is susbstracting the mean, so in fact is computing a climatology. 

So suppose we are using time (as it is the defauilt): if 'constant' it is just computing anomalies, if 'linear' it is removing the first order tendency of the data (way useful if you are using a variable that is heavily affected by climate change, like ice concentration or temperature)

valeriupredoi:
> > maybe we should create a dedicated package for stats operations (in the future);
> 
> Good ide
> 
> > a side question: how is detrending different from obtaining the least squares errors? It sounds to me they are the same fish; also it's worth specifying this method is only the linear one
> 
> In this case, what the method I'm adding is doing is to compute the linear trend of the data along a given axis and the substracting that trend. In the case of the constant is susbstracting the mean, so in fact is computing a climatology.
> 
> So suppose we are using time (as it is the defauilt): if 'constant' it is just computing anomalies, if 'linear' it is removing the first order tendency of the data (way useful if you are using a variable that is heavily affected by climate change, like ice concentration or temperature)

excellent! you sold me :grin: Put this explanation in the documentation, man, this way you tell the scientists what exactly this is useful for (tailored to their needs) :beer:

Issue #44, opened 2019-03-07
Extract odd-shaped (irregular) regions


LisaBock:
I would like to extract regions in the preprocessor of the ESMValTool which are not rectangular but as they are defined in the IPCC AR5 or AR6. For example to create the panels for a figure like this:

![image](https://user-images.githubusercontent.com/25476740/53959556-38bad400-40e4-11e9-9ccd-501bff7c4295.png)

Is there a possibility to include this in the preprocessor functions? 

At the moment I do not have the specific region definitions but I could share them when I received them.

mattiarighi:
@LisaBock is this solved? If yes, please close.

valeriupredoi:
closing since it looks to be solved, please re-open if need be :beer:

PR #220, opened 2019-09-02
Add extract_shape preprocessor function

### LINKED ISSUE #233: Add extract_shape preprocessor function
It would be useful to have a preprocessor function `extract_shape` which can select a region based on a shapefile.

An example implementation is already available in recipe shapeselect.

Some discussion on this feature can also be found here https://github.com/ESMValGroup/ESMValCore/issues/44

bouweandela:
Closes #233


bouweandela:
Related to issue https://github.com/ESMValGroup/ESMValCore/issues/44

valeriupredoi:
note that `shapely` is a dependency of `cartopy` https://scitools.org.uk/cartopy/docs/latest/installing.html so we shouldn't explicitly install it on top of the conda dependency solver+installer

BTW very cool stuff! :beer:

esmvalcore/preprocessor/_area.py
```diff
@@ -295,3 +302,86 @@ def extract_named_regions(cube, regions):
     constraints = iris.Constraint(region=lambda r: r in regions)
     cube = cube.extract(constraint=constraints)
     return cube
+
+
+def _clip_geometries(cube, geometries):
+    """Clip cube to only include the geometries and padding."""
```
valeriupredoi: a bit confused as to why to include `geometries` as a functional argument when it's actually used only in the main function `extract_shape`

esmvalcore/preprocessor/_area.py
```diff
@@ -295,3 +302,86 @@ def extract_named_regions(cube, regions):
...
+    ----------
+    cube: iris.cube.Cube
+       input cube.
+    shapefile: str
+        A shapefile defining the region(s) to extract.
+    method: str
+        Select all points contained by the shape ('contains') or
+        select a single representative point ('representative').
+    clip: bool
+        Clip the resulting cube ('true') or not ('false').
```
valeriupredoi: can we think of any used case when `clip` could be False?

esmvalcore/preprocessor/_area.py
```diff
@@ -295,3 +302,86 @@ def extract_named_regions(cube, regions):
...
+                    np.stack((lon.flat, lat.flat), axis=1))
+                nearest_point = shapely.ops.nearest_points(
+                    points, representative_point)[0]
+                nearest_lon, nearest_lat = nearest_point.coords[0]
+                select = (lon == nearest_lon) & (lat == nearest_lat)
+            selection |= select
+
+        selection = da.broadcast_to(selection, cube.shape)
+        cube.data = da.ma.masked_where(~selection, cube.core_data())
+        return cube
```
valeriupredoi: it'd probably useful to add a reference in the `cube.metadata` something like `contains whatever region only`

doc/esmvalcore/preprocessor.rst
```diff
@@ -770,6 +771,26 @@ matches the named regions against the requested string.
...
+also be clipped to a rectangular region around the shape. This function takes
+the following arguments:
+
+* ``shapefile``: the shapefile containing the geometry of the region to be
+  extracted
+* ``method``: the method to select the region, selecting either all points
+  contained by the shape or one representative point
+* ``clip``: whether to mask the original data set or to clip to a minimal
+  rectangular region containing the shape
+
```
valeriupredoi: this is confusing: clip means restricting the data to a rough rectangle around the desired shape to be selected; this is always useful since data is shrunk to a smaller, yet sufficiently large shape to be able to more efficiently extract the desired shape. Maybe we should phrase it this way, and also, see my previous comment, when will we *not* clip?

valeriupredoi:
good stuff! couple questions:

- is the `shapely` that comes from `cartopy` not good enough so we specify it as a clear dependency?
-  the `shapefile` argument value in the preprocessor step should probably be checked for and if it's missing the code should throw a user friendly error; also the itehrity of the path should be checked (`path.exists` or `isfile`) :beer:

esmvalcore/preprocessor/_area.py
bouweandela: Yes, if you prefer keeping your data global

esmvalcore/preprocessor/_area.py
bouweandela: I'm not sure if that is needed, `extract_region` does not do that either

valeriupredoi: yeah by extension that's be useful there too

esmvalcore/preprocessor/_area.py
valeriupredoi: wut? if you extract a region you obviously don't want your data global :grin:

valeriupredoi: as in, all is masked anyway so your region of interest is the only bit you care for and multimodel operations will be uselelss anyway since you need to unmask, which is redundant

esmvalcore/preprocessor/_area.py
bouweandela: Implementing this is also not trivial, because there is no descriptive name available. There is just a shapefile.

esmvalcore/preprocessor/_area.py
bouweandela: That's why the default is to clip, but if you're selecting a region that is most of the globe, I think it makes sense to keep the entire dataset.

bouweandela: We're unable to clip irregular grids too, so if you have datasets of both kinds that you want to use with your diagnostic, you probably do not want to clip either.

bouweandela:
> is the shapely that comes from cartopy not good enough so we specify it as a clear dependency?

No, you need to list all dependencies that you're directly importing. Imagine cartopy would drop shapely as a dependency, then that would break our list of dependencies.

bouweandela:
> the shapefile argument value in the preprocessor step should probably be checked for and if it's missing the code should throw a user friendly error; also the integrity of the path should be checked (path.exists or isfile)

Good point, will try to implement that

valeriupredoi:
Very good point about cartopy/shapely dependency Bouwe. Yeah all cool with
the clipping func too, just don't want the n00b user not to apply it when
they should in fact :beer:

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

esmvalcore/preprocessor/_area.py
JaroCamphuijsen: It is used in line 317 to get the bounds for the clipping box. I suppose we could also just pass geometries.bounds, since it's the only thing we need.

esmvalcore/preprocessor/_area.py
JaroCamphuijsen: Maybe adding a generic "masked irregular shape" to the metadata would work so it can be discerned from a cube of the same size (for example obtained with `extract_region`) without looking at the data.

bouweandela: I've thought about this some more and I think it's not a good idea. We have provenance for recording what happened to exactly to the data and it does not make sense to start recording some of it also as cube attributes.

bouweandela:
This is ready to be merged from my point of view.

mattiarighi:
Would it be possible to add an example / test recipe?

bouweandela:
Example recipe added in https://github.com/ESMValGroup/ESMValTool/pull/1318

mattiarighi:
Thanks!

Issue #223, opened 2019-09-04
Changed handling of metadata in iris breaks fixes


zklaus:
For quite a few fixes we use a pattern of
```
metadata = cube.metadata
do_something(cube)
cube.metadata = metadata
```
This was done to preserve metadata, but doesn't work like this anymore.
The issue is briefly discussed in #179.
The following fixes seem to be affected:

- [ ] cmip5/bnu_esm.py
- [ ] cmip5/canesm2.py
- [ ] cmip5/cesm1_bgc.py
- [ ] cmip5/cnrm_cm5.py
- [ ] cmip5/ec_earth.py
- [ ] cmip5/fio_esm.py
- [ ] cmip5/gfdl_cm2p1.py
- [ ] cmip5/gfdl_cm3.py
- [ ] cmip5/gfdl_esm2g.py
- [ ] cmip5/gfdl_esm2m.py
- [ ] cmip5/inmcm4.py
- [ ] cmip5/miroc5.py
- [ ] cmip5/miroc_esm_chem.py
- [ ] cmip5/miroc_esm.py
- [ ] cmip5/mpi_esm_lr.py

jvegasbsc:
Those fixes are one of the first parts of the code we added to v2.0. 

I would like to take a look at all of them, not only because of this issue but also because probably we will have more elegant and efficient solutions for a bunch of them.

PR #225, opened 2019-09-04
Convert fix_files to fix_metadata

jvegasbsc:
While reading #223 I realized that most of the fixes we were doing using the expensive `fix_files` method probably can now be implemented in fix_metadata

I took a look at the 3 (I thought there were more) and all can be replaced!! This pull request address that

PS: I think they are not needed now, as `standard_name` and `long_name` are now part of the attribute we automatically fix. I keep them anyway to avoid annoying the users with warnings.


mattiarighi:
Maybe @ledm and @schlunma could briefly test this, since I guess those fixes affect some of their recipes (variables `o2` and `nbp`).

mattiarighi:
Can you please fix the conflicts?

PR #231, opened 2019-09-05
Check parent_time_unit format

sloosvel:
After several issues with the parent_time_unit format (#194, #209) due to #108 , the checker now considers the non-conversion of the time attributes if there is no parent experiment.   If there is a parent experiment but the format of the parent_time_unit cannot be read, an error message appears.

I tried fixing the no parent case as suggested in [here]( https://github.com/ESMValGroup/ESMValCore/pull/194#issuecomment-518556998) , but maybe this solution is simpler.

### DIFF
--- a/esmvalcore/cmor/check.py
+++ b/esmvalcore/cmor/check.py
@@ -475,21 +475,29 @@ class CMORCheck():
             coord.units = cf_units.Unit(coord.units.origin, simplified_cal)
 
             attrs = self._cube.attributes
-            branch_child = 'branch_time_in_child'
-            if branch_child in attrs:
-                attrs[branch_child] = old_units.convert(attrs[branch_child],
-                                                        coord.units)
+            try:
+                parent_time = 'parent_time_units'
+                if parent_time in attrs:
+                    parent_units = cf_units.Unit(attrs[parent_time],
+                                                 simplified_cal)
+                    attrs[parent_time] = 'days since 1850-1-1 00:00:00'
 
-            parent_time = 'parent_time_units'
-            if parent_time in attrs:
-                parent_units = cf_units.Unit(attrs[parent_time],
-                                             simplified_cal)
-                attrs[parent_time] = 'days since 1850-1-1 00:00:00'
+                    branch_parent = 'branch_time_in_parent'
+                    if branch_parent in attrs:
+                        attrs[branch_parent] = parent_units.convert(
+                            attrs[branch_parent], coord.units)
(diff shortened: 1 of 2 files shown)


valeriupredoi:
the cmorizing team of UKESM1 (CDDS) were not aware of this issue (reported it yesterday), we'll expect a lot of CMIP6 models to suffer from it since the CDDS guys are actually on top of their game and this slipped through the cracks even for them

zklaus:
Looking through the badc it seems most models avoided that particular pitfall.

esmvalcore/cmor/check.py
```diff
@@ -475,21 +475,29 @@ def _check_time_coord(self):
...
+                    branch_parent = 'branch_time_in_parent'
+                    if branch_parent in attrs:
+                        attrs[branch_parent] = parent_units.convert(
+                            attrs[branch_parent], coord.units)
+
+                branch_child = 'branch_time_in_child'
+                if branch_child in attrs:
+                    attrs[branch_child] = old_units.convert(
+                        attrs[branch_child], coord.units)
+            except ValueError:
```
bouweandela: For `try`/`except` clauses, limit the `try` clause to the absolute minimum amount of code necessary, this avoids masking bugs. See [PEP8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations)

valeriupredoi:
can we try get this merged sooner than later pls, people are starting to write fixes for this issue == they need this to work. Cheers :beer:

sloosvel:
But fixes like #194 and #274 are still going to be needed even if this gets merged. This only avoids an error message when the units are set to "no parent", which is a valid value and should not make the checks fail. In #194 a general approach was proposed,  but #209 already got merged and deals with the bad format issue by adding the corresponding fix per model.

valeriupredoi:
aha cheers for clarifying @sloosvel - good point! :beer:

sloosvel:
@jvegasbsc pointed out that a warning message would be better than an error

valeriupredoi:
dis ready for merge? Ppl are using UKESM1 and complaining abt it :beer:

schlunma:
> dis ready for merge? Ppl are using UKESM1 and complaining abt it beer

For UKESM1 this should be fixed by #194. Are ppl using the development branch of the core or the release? :grin:

valeriupredoi:
the one their grandma told them to use :grin: I assume the release which is old as their grandma

PR #243, opened 2019-09-11
Fix bug in extract_region for irregular grids

### LINKED ISSUE #94: extract_region preprocessor no longer works on irregular grids

bouweandela:
Closes #94

### DIFF
--- a/tests/unit/preprocessor/_area/test_area.py
+++ b/tests/unit/preprocessor/_area/test_area.py
@@ -4,6 +4,7 @@ import unittest
 
 import iris
 import numpy as np
+import pytest
 from cf_units import Unit
 
 import tests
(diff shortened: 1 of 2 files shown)


ledm:
> Hi Lee,
> 
> Good point about the negative longitudes, I'll see what I can do. Looking at the code you posted here [#94 (comment)](https://github.com/ESMValGroup/ESMValCore/issues/94#issue-456883767), I doubt that it worked before.

I can assure you that this was working before in the numpy version, but I suspect that it did not work in the first dask version.

tests/unit/preprocessor/_area/test_area.py
```diff
@@ -146,5 +147,78 @@ def test_extract_named_region(self):
...
+        data,
+        var_name='tos',
+        units='K',
+        dim_coords_and_dims=dim_coord_spec,
+        aux_coords_and_dims=aux_coord_spec,
+    )
+    return cube
+
+
+def test_extract_region_irregular(irregular_grid_cube):
```
ledm: We should also have a test which uses extracts negative longitudes.

bouweandela: Implemented

bouweandela:
@ledm I ran you recipe with the latest changes and the plot now looks like this: 
diag_CMIP5_MPI-ESM-LR_Omon_historical_r1i1p1_tos_prep_map_4_diag_map_4_2001_2003_map_0

ledm:
Great work, I can confirm that I'm seeing the same result here too! Nice work @bouweandela!

PR #245, opened 2019-09-12
Clip irregular grids

bouweandela:
@zklaus commented on `extract_region`: For me the biggest issue with this function has always been that for irregular grids it only masks. Of course, masking is necessary, but to reduce the amount of data (in some cases drastically), it would be nice to also use intersection to restrict to the covering rectangle.

PR #243 should be tested/merged before this one.


bouweandela:
Yes, it would be if it works correctly. Did you test? I'm not sure if it's still working as expected after I solved the merge conflict.

stefsmeets:
Hi @bouweandela , I just had a look at this PR, and it seems to work as advertised. What would be needed to get it across the finish line?

bouweandela:
Thanks for looking into it @stefsmeets and @valeriupredoi! My plan was to move this functionality to iris instead at some point, but I haven't had the time to do it. If it works as expected, we could merge this though, and move to iris later. I think that the only thing remaining to be done would be to update the documentation [here](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/preprocessor.html#extract-region).

stefsmeets (approved):
Works as expected 👍

stefsmeets:
From my understanding the function now works the same for regular and irregular grids now that it no longer masks the outside area. So I updated the documentation accordingly.

bouweandela:
@valeriupredoi or @zklaus Could one of you please do a final check and merge when/if you're happy with the results? I feel I should not be merging this myself.

esmvalcore/preprocessor/_area.py
```diff
@@ -63,16 +63,29 @@ def extract_region(cube, start_longitude, end_longitude, start_latitude,
...
+    """Extract a region from a cube on an irregular grid."""
     # Convert longitudes to valid range
     if start_longitude != 360.:
         start_longitude %= 360.
     if end_longitude != 360.:
         end_longitude %= 360.
 
+    # Select coordinates inside the region
+    lats = cube.coord('latitude').points
+    lons = cube.coord('longitude').points
```
zklaus, line 88: It's probably easy to account for negative longitudes here, right? That would help with #1141. What do you think about
```suggestion
    lons = (cube.coord('longitude').points + 360.) % 360. 
```

esmvalcore/preprocessor/_area.py
```diff
@@ -84,8 +97,19 @@ def extract_region(cube, start_longitude, end_longitude, start_latitude,
...
-    cube.data = da.ma.masked_where(~selection, cube.core_data())
+
+    # Crop the selection, but keep rectangular shape
+    i_range, j_range = selection.nonzero()
+    if i_range.size == 0:
+        raise ValueError("No data points available in selected region")
+    i_min, i_max = i_range.min(), i_range.max()
+    j_min, j_max = j_range.min(), j_range.max()
+    i_slice, j_slice = slice(i_min, i_max + 1), slice(j_min, j_max + 1)
+    cube = cube[..., i_slice, j_slice]
```
zklaus, line 108: Maybe use `cube.coord_dims` to determine the correct axes?

zklaus (changes requested):
Looks good. Still a few comments inline plus:
- It's probably good to point out in the documentation what happens with irregular grids.
- What's this mysterious `masked_cube_lonlat` function mentioned in the docstring? I couldn't find it.

esmvalcore/preprocessor/_area.py
bouweandela: To keep preprocessor functions simple, we assume that the [data follows the CMOR table conventions in all preprocessor functions](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#design-considerations), i.e. longitudes are inside the range 0 to 360 degrees. A fix like this is done already as part of the automatic fixes here: https://github.com/ESMValGroup/ESMValCore/blob/933b37b225a89aeed84029d8c83105407ed7ae5f/esmvalcore/cmor/check.py#L656

zklaus: This is not an actual requirement of the CMOR tables for irregular grids. The fix is wrong and will be removed.

zklaus: Besides, we need to take other sources like CORDEX into account.

esmvalcore/preprocessor/_area.py
bouweandela: The CMOR conventions specify that the last two axes are latitude and longitude, right?

bouweandela:
> What's this mysterious masked_cube_lonlat function mentioned in the docstring? I couldn't find it.

Looking at the git history, maybe @ledm could explain that :laugh: I'll remove the line.

esmvalcore/preprocessor/_area.py
zklaus: Yeah, I think that's right. I was unsure with land use types etc. but a quick check seems to confirm it.

esmvalcore/preprocessor/_area.py
stefsmeets: Hi @zklaus, this PR does not change the current behaviour of the code in master: https://github.com/ESMValGroup/ESMValCore/blob/933b37b225a89aeed84029d8c83105407ed7ae5f/esmvalcore/preprocessor/_area.py#L69
If you think this is a bug, could you make a separate issue to discuss an appropriate fix so we do not delay this PR any further?

bouweandela: @zklaus Thanks for explaining, I changed it as suggested. If anything else needs changing to get this in: I'll be on holiday until next week Friday, so please feel free to make the changes yourself.

stefsmeets:
Hi @zklaus If you are happy with the changes, could you approve this PR so that it can be merged?

zklaus:
I think the comment about the documentation has not been addressed yet. A small modification to the docstring would be enough imho.

stefsmeets:
> 
> 
> I think the comment about the documentation has not been addressed yet. A small modification to the docstring would be enough imho.

I did attempt to update the docstring here: https://github.com/ESMValGroup/ESMValCore/pull/245/commits/841cbd6bc273e60bfd401cae660bec3f5e709129. Is there something you are still missing?

zklaus:
> > I think the comment about the documentation has not been addressed yet. A small modification to the docstring would be enough imho.
> 
> I did attempt to update the docstring here: [841cbd6](https://github.com/ESMValGroup/ESMValCore/commit/841cbd6bc273e60bfd401cae660bec3f5e709129). Is there something you are still missing?

Well, that commit was before my comment. What I am looking for is something along the lines of "If the grid is irregular, the returned region is the bounding box in the original coordinates with the deselected area inside the box masked.", only better formulated ;)

stefsmeets:
Thanks @zklaus , I thought that the description covered both cases, but I guess you wanted to have it explicit. I think you worded it pretty well, can you have a look at my attempt to improve it and let me know what you think? 😅

https://esmvaltool--245.org.readthedocs.build/projects/ESMValCore/en/245/recipe/preprocessor.html#extract-region

zklaus (approved):

PR #246, opened 2019-09-12
Mask glaciated

bascrezee:
A first attempt. It works for my purpose, but I had to change the function `get_geometry_from_shape` to return all geometries, and not just the first one. I do not see why only the first geometry is taken. But if this is for good reason, I have to adjust this PR to reflect a distinction between the different shape files. In a quick check I noted no difference when applying `mask out land` or `mask out sea` for my branch and the `development` branch. But of course, this does not guarantee that this is the case for all data sets. 

Some fine tuning needs to be done with regard to performance. As a rule of thumb I noted that each geometry (polygon) takes about 1 second of processing time for my input data (CMIP5 for testing), I therefore at the moment only decided to take the 10 largest.

*edit* one of the automated test fails, so it does seem to change something... I think I need your help here @valeriupredoi .


bascrezee:
Some more benchmarking with the following dataset (0.25 deg res):
```
datasets:
  - {dataset: CDS-SATELLITE-SOIL-MOISTURE, type: sat, project: OBS, mip: Lmon, version: COMBINED-TCDR-v201812.0.0, tier: 3, start_year: 2017,  end_year: 2018}
```

|     | 10 polygons | 100 polygons |   |   |
|:---:|:-----------:|:------------:|---|---|
| 10m |     4.9s    |      35s     |   |   |
| 50m |     5.3s    |      34s     |   |   |
|     |             |              |   |   |

Outcome, reducing resolution from 10m to 50m does not impact speed. It is really the number of polygons that counts. My question stays, why is only the first polygon selected here: 

https://github.com/ESMValGroup/ESMValCore/blob/493aff327dd8134153296d7925d32982af02289f/esmvalcore/preprocessor/_mask.py#L227-L232

@valeriupredoi or @bouweandela  any idea? To me it seems as if a part of the polygons from the NE masks gets lost, which would be a bug. Or maybe this behaviour was intended, but not documented. See below the total number of polygons given a certain preprocessor setting.

```
    mask_landsea:
      mask_out: land # 10 polygons
    mask_glaciated:
      mask_out: glaciated # 377 polygons
    mask_landsea:
      mask_out: sea # 1 polygons
```

valeriupredoi:
Bas, what are those polygons you are talking about? That bit of code you
are pointing to selects the most basic geometry from all the possible ones,
if I remember correctly it is the averaged one over the given resolution,
can't remember what the other geometries were, all bunch of particular
cases I think

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

bascrezee:
> Bas, what are those polygons you are talking about? 

I think our shapefiles contain only polygons and no other geometries (like e.g. points), that's why I like to refer to them as polygons.

Here an example for `ne_10m_land.shp`: 
```
In [11]: %paste                                                                                    
reader = shpreader.Reader('esmvalcore/preprocessor/ne_masks/ne_10m_land.shp')
list(reader.geometries())

## -- End pasted text --
Out[11]: 
[<shapely.geometry.multipolygon.MultiPolygon at 0x7fdc2fc81a58>,
 <shapely.geometry.multipolygon.MultiPolygon at 0x7fdc2fc81940>,
 <shapely.geometry.multipolygon.MultiPolygon at 0x7fdc1339c2e8>,
 <shapely.geometry.multipolygon.MultiPolygon at 0x7fdc2fc81908>,
 <shapely.geometry.multipolygon.MultiPolygon at 0x7fdc2fc8bf60>,
 <shapely.geometry.multipolygon.MultiPolygon at 0x7fdc2fc69fd0>,
 <shapely.geometry.multipolygon.MultiPolygon at 0x7fdc5d6a4358>,
 <shapely.geometry.multipolygon.MultiPolygon at 0x7fdc1339c518>,
 <shapely.geometry.multipolygon.MultiPolygon at 0x7fdc1339c278>,
 <shapely.geometry.polygon.Polygon at 0x7fdc2fcace48>]
```


> That bit of code you are pointing to selects the most basic geometry from all the possible ones, if I remember correctly it is the averaged one over the given resolution, can't remember what the other geometries were, all bunch of particular cases I think.

OK, then it seems like for my mask (glaciated areas) the data is organized differently, since taking the first polygon has absolutely no effect. Will try to further understand these shape files.

valeriupredoi:
I looked at the NE site, it seems that what they mean by polygons is the
map itself, and different polygons are different maps - ie they say polygon
0 for ocean masks represents bathymetry 0 so sea level, depth = 0. You
should look what polygon 0 means for you, sorry I'd do it but am in a
meeting

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

bascrezee:
> I looked at the NE site, it seems that what they mean by polygons is the map itself, and different polygons are different maps - ie they say polygon 0 for ocean masks represents bathymetry 0 so sea level, depth = 0. You should look what polygon 0 means for you, sorry I'd do it but am in a meeting 

Yes, I am exploring the data. Polygon 0 is not meaningful for `glaciated_areas`, it is just some random glaciated area in the long list. I found out that when using `reader.records()` one gets a list with access to attributes for each multipolygon. I think I will be able to figure it out, thanks for your help! I do think the code would need some restructuring, to account for the different organization of the shp-files from NE.

valeriupredoi:
For ocean and land masks we looked at the tests and the output looked good
:grin: Nikolay @nkoldunov provided info on various stuff, I had no prior
knowledge of these things. It'd be good to look into the 'best' masks for
ocean, land and ice, but for now they work at first order level

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

valeriupredoi:
Sorry, am on via email/phone, got Nikolay's tag wrong @koldunovn :beer:

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

bascrezee:
@valeriupredoi : I think a first step that needs to be taken, is to restructure the code, such that the functions are general for any given NE mask. As you can see in this draft PR, I had to make changes to some functions that were implemented already. I am afraid to break backward compatibility, so I think it is best if I close this PR and that you/we start rewriting the existing functions in such a way that new masks can be added later without touching any of the existing masks. Please look at the changes in this draft PR and let me know if you get my point :)

esmvalcore/preprocessor/_mask.py
```diff
@@ -224,12 +247,12 @@ def mask_landseaice(cube, fx_files, mask_out):
...
 
 
-def _get_geometry_from_shp(shapefilename):
+def _get_geometries_from_shp(shapefilename):
     """Get the mask geometry out from a shapefile."""
     reader = shpreader.Reader(shapefilename)
-    # Index 0 grabs the lowest resolution mask (no zoom)
-    main_geom = [contour for contour in reader.geometries()][0]
-    return main_geom
+    geometries = [contour for contour in reader.geometries()]
```
valeriupredoi: @bascrezee try using `fiona`:
```
with fiona.open(shapefilename) as geometries:
    for item in geometries:
        shape = shapely.geometry.shape(item['geometry'])
```
see example in `preprocessor/_area.py/extract_shape()`
Also be aware of the issue that (most of the times) the shapefiles go from -180 to +180, see solution here https://github.com/ESMValGroup/ESMValCore/pull/319

bascrezee: Thanks for the reviewing, can you also have a look at my comment, above. Basically I suggest to cancel this PR first, and restructure the code before I add this new mask again.

bascrezee: I think also when switching from `cartopy` to `fiona` for reading the shapefiles, this should be implemented before. I simply followed examples for the other masks.

bascrezee: No need to switch, since `cartopy` actually uses `fiona`? https://github.com/SciTools/cartopy/blob/0e5eef86afe3502bfc41b84f71ba2bf48e2812bd/lib/cartopy/io/shapereader.py#L54

bascrezee: >  I think it is best if I close this PR and that you/we start rewriting the existing functions in such a way that new masks can be added later without touching any of the existing masks. Please look at the changes in this draft PR and let me know if you get my point :)

ping @valeriupredoi

bascrezee:
@valeriupredoi Can you please have a look at this and let me know what you think? Another option is that I do the suggested changes (an initial PR should just address the restructuring without any additional functionality). @BenMGeo and me need this functionality for our project.

valeriupredoi:
@bascrezee having a looksee now :beer:

esmvalcore/preprocessor/_mask.py
```diff
@@ -270,18 +293,19 @@ def _mask_with_shp(cube, shapefilename):
...
+        elif cube.ndim == 3:
+            mask[:] = shp_vect.contains(region, x_p_180, y_p_90)
+        elif cube.ndim == 4:
+            mask[:, :] = shp_vect.contains(region, x_p_180, y_p_90)
+
+        # Then apply the mask
+        if isinstance(cube.data, np.ma.MaskedArray):
+            cube.data.mask |= mask
+        else:
+            cube.data = np.ma.masked_array(cube.data, mask)
```
valeriupredoi: so is this implementation working for your glaciated mask? I am a bit concerned by the looping over the 10 regions since for the land/sea all one needs is the 0th (first) region. Do these regions stack up to be the full, most accurate mask?

bascrezee: Hi @valeriupredoi , thanks for looking into this. Instead of going into the details of the glaciated mask, I think the masking should be rewritten to make the code more modular. My point is that the selection of only the first region is not something general, and is different per NE mask. Therefore, I suggest to not proceed with this PR, but instead go for a new PR where we restructure the code, such that one can add new NE masks, without having to touch the existing the code.

valeriupredoi: OKay the current implementation (that has been around for about 2 years) relies on the existence of *only* the NE land and NE ocean masks (in the masks aux dir) - I suggest, for ease of change and preserve the current functionality w/o the need to retest to keep this current implementation, only adding functionality for the glaciated mask (and the mask itself in the masks aux dir). What do you think? :beer:

esmvalcore/preprocessor/_mask.py
valeriupredoi: @bascrezee have a look at #389 - you can plug in all the regions you want and not change the existinmg mask landsea :beer:

bascrezee:
Closing this since superseded by https://github.com/ESMValGroup/ESMValCore/pull/392

PR #249, opened 2019-09-16
Add R and Julia installation checks

jvegasbsc:
Just as we do with NCL, if there is an R or Julia script in the recipe we check if the executables are available. Checks are simpler because we are not requesting any minimal version.

This should avoid the frustration of going through a lengthy preprocessor to found that you don't have the required interpreters installed.

Any rumours that I want to add this because I skipped the R installation and weeks later tried to run an R recipe are wrong.


bouweandela:
This has been implemented in #528, code here: https://github.com/ESMValGroup/ESMValCore/blob/735f4160da4d2ae113dd87f93f25cb38d943c656/esmvalcore/_task.py#L304-L312

PR #256, opened 2019-09-17
fix ec_earth: missing_height

mathause:
* In the ETH repository we have some older EC-EARTH files that do not include the height coordinate. This fixes the problem for tas.
* Also some files included `long_name` for time while some did not.
* Only applies the fixes if `height`/ `long_name` is missing.

@ruthlorenz

### DIFF
--- a/esmvalcore/cmor/_fixes/cmip5/ec_earth.py
+++ b/esmvalcore/cmor/_fixes/cmip5/ec_earth.py
@@ -94,8 +94,7 @@ class Tas(Fix):
         """
 
         for cube in cubes:
-            coord_names = [coord.var_name for coord in cube.coords()]
-            if 'height' not in coord_names:
+            if not cube.coords(var_name='height'):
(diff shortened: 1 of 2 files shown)


mathause:
ping

esmvalcore/cmor/_fixes/cmip5/ec_earth.py
```diff
@@ -72,3 +73,32 @@ def fix_data(self, cube):
...
+
+        Returns
+        -------
+        iris.cube.CubeList
+
+        """
+
+        for cube in cubes:
+            coord_names = [coord.var_name for coord in cube.coords()]
+            if 'height' not in coord_names:
```
jvegasbsc: This two lines can be simplified:

```python
if not cube.coords(var_name='height')

mathause: Thanks! Fixed.

PR #257, opened 2019-09-18
Retrieve properly the activity for exps linked to multiple ids 

### LINKED ISSUE #248: Bug in activity_id determination for some cmip6 experiments
In PR #107 (addressing issue #37), we introduced the automatic determination of `activity_id` from `experiment_id`. This works nicely in most cases, but a few experiments have more than one activity associated with them. In that case, all activities are listed in the CV field separated by spaces, but following [the official DRS standard (bottom of page 17)](http://goo.gl/v1drZl), only the first one is used in the path.

As a fix, the field coming from the CV should be `split` on `" "` and only the first entry be taken. This should also work in the normal case of only one activity.

The list of afflicted experiments is
- dcppC-forecast-addPinatubo
- esm-1pct-brch-1000PgC
- esm-1pct-brch-2000PgC
- esm-1pct-brch-750PgC
- esm-1pctCO2
- esm-bell-1000PgC
- esm-bell-2000PgC
- esm-bell-750PgC
- land-hist
- piClim-aer
- piClim-control
- ssp370

sloosvel:
Fixes #248

### DIFF
--- a/esmvalcore/_config.py
+++ b/esmvalcore/_config.py
@@ -190,8 +190,8 @@ def get_activity(variable):
         exp = variable['exp']
         logger.debug("Retrieving activity_id for experiment %s", exp)
         if isinstance(exp, list):
-            return [CMOR_TABLES[project].activities[value] for value in exp]
-        return CMOR_TABLES[project].activities[exp]
+            return [CMOR_TABLES[project].activities[value][0] for value in exp]
+        return CMOR_TABLES[project].activities[exp][0]
     except (KeyError, AttributeError):
         return None
 
--- a/esmvalcore/cmor/table.py
+++ b/esmvalcore/cmor/table.py
@@ -191,7 +191,7 @@ class CMIP6Info(object):
                 try:
                     exps = table_data['CV']['experiment_id']
                     for exp_id in exps:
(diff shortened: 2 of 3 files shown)


zklaus:
I couldn't test yet, but just to clarify, @jvegasbsc: There are two ways lists could enter here:

* One experiment has two activities associated. In this case this PR correctly uses only the first one.
* One dataset spans two experiments, given as a list in `{dataset: ..., exp: [historical, ssp585]}`. Is this second case considered?

sloosvel:
>     * One dataset spans two experiments, given as a list in `{dataset: ..., exp: [historical, ssp585]}`. In this case there will also be two activities, one for every experiment. Is this second case considered?

Yes, this is also considered. If the experiment is a list, the code loops over experiments. And for each experiment, if it is associated to multiple activities it only considers the first one.

The test needed to be fixed because if there is only one experiment, what gets returned is just the value of the activity, instead of list with one value. But if there are multiple exps, the activity is returned as a list as well. At least that's what I get when I try it.

mattiarighi:
@axel-lauer does this solve the problem you mentioned last week?

PR #258, opened 2019-09-18
Rename pull request template

JaroCamphuijsen:
Rename the template and move a directory up, since the pull request template doesn't show up and since there is only one template for the core pull request.


JaroCamphuijsen:
According to https://stackoverflow.com/questions/52139192/github-pull-requests-template-not-showing renaming the file like this should work. It is suggested that the expected multiple pull request templates behavior is not yet working on github.

PR #268, opened 2019-09-20
Clarify docs sidebar

Peter9192:
Changed "Diagnostics API" to "ESMValTool API"

also see: https://github.com/ESMValGroup/ESMValTool/pull/1330


Peter9192:
After a discussion we decided that the original name was better. Perhaps in the introduction it should be clarified that ESMValTool consists of a core and diagnostics, but that can be changed in the ESMValTool docs.

PR #270, opened 2019-09-24
Global Total Integrated primary production derived variable

ledm:
This PR brings in the Global Total Integrated primary production derived variable. This variable has appeared in the upcoming ESMValTool Paper - part 2, so it would be nice to get it out there before the publication gets to discussions format. 

This is linked with the ESMValTool PR https://github.com/ESMValGroup/ESMValTool/pull/1331, which uses this derived field.

### DIFF
--- a/esmvalcore/preprocessor/_derive/intppgt.py
+++ b/esmvalcore/preprocessor/_derive/intppgt.py
@@ -1,0 +1,89 @@
+"""Derivation of variable `intppgt`."""
+import iris
+import numpy as np
+
+from ._baseclass import DerivedVariableBase
+
+
+def calculate_total_flux(intpp_cube, cube_area):
+    """
+    Calculate the area of unmasked cube cells.
+
+    Requires a cube with two spacial dimensions. (no depth coordinate).
+
+    Parameters
+    ----------
+    cube: iris.cube.Cube
+        Data Cube
+    cube_area: iris.cube.Cube
+        Cell area Cube
+
+    Returns
+    -------
+    numpy.array:
+        An numpy array containing the total flux of CO2.
+    """
+    data = []
+    times = intpp_cube.coord('time')
+
+    intpp_cube.data = np.ma.array(intpp_cube.data)
+    for time_itr in np.arange(len(times.points)):
+
+        total_flux = intpp_cube[time_itr].data * cube_area.data
+
+        total_flux = np.ma.masked_where(intpp_cube[time_itr].data.mask,
+                                        total_flux)
+        data.append(total_flux.sum())
+
+    ######
+    # Create a small dummy output array
+    data = np.array(data)
+    return data
+
+
+class DerivedVariable(DerivedVariableBase):
+    """Derivation of variable `intppgt`."""
+
+    # Required variables
+    required = [
+        {
+            'short_name': 'intpp',
+            # 'mip': 'Omon',
+            'fx_files': [
+                'areacello',
+            ],
+        },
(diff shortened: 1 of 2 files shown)


esmvalcore/preprocessor/_multimodel.py
```diff
@@ -107,15 +107,22 @@ def _put_in_cube(template_cube, cube_data, statistic, t_axis):
             t_axis,
             standard_name='time',
             units=template_cube.coord('time').units)
-    lats = template_cube.coord('latitude')
-    lons = template_cube.coord('longitude')
+
```
bouweandela: These changes seem unrelated to deriving the variable? Would it make sense to put this in a separate pull request?

esmvalcore/preprocessor/_derive/intppgt.py
```diff
@@ -0,0 +1,89 @@
...
+    @staticmethod
+    def calculate(cubes):
+        """Compute longwave cloud radiative effect."""
+        intpp_cube = cubes.extract_strict(
+            iris.Constraint(name='net_primary_mole_productivity_of_carbon_by_'
+                                 'phytoplankton'))
+
+        try:
+            cube_area = cubes.extract_strict(iris.Constraint(name='cell_area'))
+        except iris.exceptions.ConstraintMismatchError:
```
bouweandela: If this happens, `cube_area` will be undefined and the code below will fail, so I think it is not a good idea to let this `pass`

esmvalcore/preprocessor/_derive/intppgt.py
```diff
@@ -0,0 +1,89 @@
...
+
+        total_flux = calculate_total_flux(intpp_cube, cube_area)
+
+        # Dummy result cube
+        result = intpp_cube.collapsed(
+            ['latitude', 'longitude'],
+            iris.analysis.MEAN,
+        )
+
+        # TODO: Load seconds per year from model calendar.
```
bouweandela: Do your to do?

esmvalcore/preprocessor/_derive/intppgt.py
```diff
@@ -0,0 +1,89 @@
...
+    Returns
+    -------
+    numpy.array:
+        An numpy array containing the total flux of CO2.
+    """
+    data = []
+    times = intpp_cube.coord('time')
+
+    intpp_cube.data = np.ma.array(intpp_cube.data)
+    for time_itr in np.arange(len(times.points)):
```
bouweandela: Would it be possible to use `iris.analysis.SUM` to aggregate over the required dimensions?

mattiarighi:
@ledm can you finalize this PR so that I can test it?

znicholls (commented):
Following the logic of https://github.com/ESMValGroup/ESMValCore/pull/331, is this now redundant? This could be specified in a recipe using a combination of fx weighting, the area information and the sum preprocessor?

This PR can thus be closed?

znicholls:
The fact that the derived variable's units lose the per m2 is related to the comment here https://github.com/ESMValGroup/ESMValCore/pull/1574#issuecomment-1123428679

In other words, the fact that a sum weighted by area does not include units (i.e. by multiplying the units by area ** 2) is an ongoing issue (issue now open at #1613 to track)

schlunma:
I think this can now be done with the following preprocessor:

```yaml
area_statistics:
  operator: sum
```

(including correct unit handling).

Please re-open if necessary.

PR #276, opened 2019-09-25
Support Windows Subsystem for Linux

jvegasbsc:
I am using Ubuntu running on the WSL to run ESMValTool. It works mostly fine, but the resource logger was failing because the `io_counters` are not available.

I adapted the code to work even if they are not available and now it works well

It is not real Windows support, but at least it will be easier for Windows users to run ESMValTool

### DIFF
--- a/esmvalcore/_task.py
+++ b/esmvalcore/_task.py
@@ -54,6 +54,12 @@ def _get_resource_usage(process, start_time, children=True):
     gigabyte = float(2**30)
     precision = [1, 1, None, 1, None, 3, 3]
     cache = {}
+    try:
+        process.io_counters()
+    except AttributeError:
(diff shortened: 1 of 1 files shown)


valeriupredoi:
this is useful not only for Windows, seen Solaris hitting the same issue :beer:

bouweandela:
@valeriupredoi Could you please test if it works for you on solaris?

valeriupredoi:
> @valeriupredoi Could you please test if it works for you on solaris?

if only I could grab hold of a Solaris machine now. Can't do bud, assume it works. I don't think any of our customers would use Solaris, if they did then they won't need ESMValTool, they've probably already written their own evaluation pipeline in Assembler :rofl:

PR #279, opened 2019-09-26
Change badc cmip5 drs version `latest` to `[latestversion]` (closes #278)

### LINKED ISSUE #278: version part of CMIP5 drs on badc
The drs for CMIP5 on badc/jasmin contains a literal `latest` for the version part. This mostly works because there is always a `latest` link pointing to the most recent version. However, for some datasets there are several versions and the latest only contains a few variables where the bulk of the data comes from earlier versions.

An extreme example is
```
/badc/cmip5/data/cmip5/output1/FIO/FIO-ESM/historical/mon/ocean/Omon/r1i1p1
```
where in the latest version there is *only* `so`, all other variables being available in the earlier version.

Can we replace `latest` with `[latestversion]` as in most of the other drs?

zklaus:
Closes #278

### DIFF
--- a/esmvalcore/config-developer.yml
+++ b/esmvalcore/config-developer.yml
@@ -33,7 +33,7 @@ CMIP5:
   cmor_strict: true
   input_dir:
     default: '/'
-    BADC: '[institute]/[dataset]/[exp]/[frequency]/[modeling_realm]/[mip]/[ensemble]/latest/[short_name]'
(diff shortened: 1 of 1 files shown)


valeriupredoi:
note that (even though it may be a little too late to implement) the actual, latest files are always in `files` eg `/badc/cmip5/data/cmip5/output1/FIO/FIO-ESM/historical/mon/ocean/Omon/r1i1p1/files/`

Issue #54, opened 2018-11-12
Preprocessing overlapping data not implemented yet


alasdairhunter:
Hi,
I am having an issue with the preprocessing when trying to run a diagnostic which uses two variables from a single model:
https://github.com/ESMValGroup/ESMValTool/blob/MAGIC_BSC/esmvaltool/recipes/recipe_diurnal_temperature_index_wp7.yml
   
  An unexpected problem prevented concatenation.
  Expected only a single cube, found 2.

I guess this is just a minor error in our .yml file. Do you have an example of any other diagnostics which load two (or more) variables from a single model for use as input for a single R script (or Python script).

mattiarighi:
#280 seems to have solved this issue as well.

I tried reading the following dataset (see also @bjoernbroetz's [comment above](https://github.com/ESMValGroup/ESMValCore/issues/54#issuecomment-437858210)):
```
  - {dataset: IPSL-CM5A-LR, project: CMIP5, exp: historical, mip: day, ensemble: r1i1p1, start_year: 1998, end_year: 2003}
```
which I would expect to give problems, since the input directory contains overlapping files:
```
/mnt/lustre02/work/bd0854/DATA/ESMValTool2/CMIP5_DKRZ/IPSL/IPSL-CM5A-LR/historical/day/atmos/day/r1i1p1/v20110915/tasmin/tasmin_day_IPSL-CM5A-LR_historical_r1i1p1_19500101-19991231.nc
/mnt/lustre02/work/bd0854/DATA/ESMValTool2/CMIP5_DKRZ/IPSL/IPSL-CM5A-LR/historical/day/atmos/day/r1i1p1/v20110915/tasmin/tasmin_day_IPSL-CM5A-LR_historical_r1i1p1_19500101-20051231.nc
```
but it went through without errors.

However, from the log I cannot understand which of the two was actually used.

@valeriupredoi ?

valeriupredoi:
did you read the debug log? It's all there @mattiarighi (like in the Da Vinci Code :grin: )

valeriupredoi:
yay! :tada:

PR #280, opened 2019-09-26
Attempt concatenation when cubes have overlapping data

### LINKED ISSUE #349: recipe: using exp: [historical, ssp126] with overlapping data
Theoretically it is possible to combine experiments in the recipe as `{exp: [historical, ssp585], ...}`. Practically this does not work as many models have an overlap (e.g. historical going further than 2014) and then this technique fails (see e.g this [recipe](https://github.com/ESMValGroup/ESMValTool/blob/version2_development/esmvaltool/recipes/recipe_cox18nature.yml#L51-L52)). Is there a way to achieve this? Can a per-experiment `end_year` or `start_year` be defined?

valeriupredoi:
Better concatenation for **two** cubes that have time overlapping data and identical metadata - solves https://github.com/ESMValGroup/ESMValTool/issues/1329 and the olde issue with overlapping CMIP data (ie 1850-2000 and 1850-2300 both files in the same repo)

BTW whoever implemented the `exp: list` functionality **should have tested* properly or added info to documentation, at the moment it's completely useless :beer:

**Tasks**

-   [x] Add unit tests

* * *

Closes #349 and possibly (untested) #54


jvegasbsc:
I was the one that added the documentation, I will do my penitence adding tests for this feature

valeriupredoi:
hahaha no worries @jvegasbsc am almost finished writing tests :beer:

jvegasbsc:
It is already done

jvegasbsc:
I have to add #259 to make the tests pass, as I was using 0:00 hours in the time

valeriupredoi:
sweeet! I was just about to commit, but not no more, cheers Javi :beer:

valeriupredoi:
@jvegasbsc I fixed your fix, man :grin:
The way was implemented was not working unfortunately, now it's all very schoolboy explicit for all possible cases. I removed an older test since we can do concatenation of overlapping cubes now; all your tests work; I also tested with real life data :beer:

esmvalcore/preprocessor/_io.py
```diff
@@ -274,3 +289,55 @@ def _write_ncl_metadata(output_dir, metadata):
     write_ncl_settings(info, filename)
 
     return filename
+
+
+def _concatenate_two_overlapping_cubes(cubes):
+    """Concatenate time-overlapping cubes (two cubes only)."""
+    # get time end points
+    time_1 = cubes[0].coord('time')
```
mathause: maybe rename to `time_0`, `time_1`, etc. so it is in line with the indices.

valeriupredoi:
@mathause cheers muchly for testing!! good to hear you are getting the (mostly) desired output, yeah, I'll have a look at the changes next week, off to pub now :beer:

esmvalcore/preprocessor/_io.py
valeriupredoi: meh, it's cube1, cube2 etc :grin:

mathause:
Could not get to work right now, have to come back on monday...

valeriupredoi:
@schlunma remined me of #54 that this PR should solve as well :beer:

mattiarighi:
Please consider merging #210 first, since it changes a bit the way input files are handled.

valeriupredoi:
> Please consider merging #210 first, since it changes a bit the way input files are handled.

@mattiarighi it shouldn't matter, this one here only deals with the concatenation in `_io.py` after files have been found

bouweandela:
Does this close #54?

jvegasbsc:
> Does this close #54?
Not really, as we are only working on the case of two overlapping cubes, mostly for concatenation of experiments.

Some of the specific problems listed (in particular, cases like `tasmax_day_IPSL-CM5A-LR_historical_r1i1p1_19500101-19991231.nc`
`tasmax_day_IPSL-CM5A-LR_historical_r1i1p1_19500101-20051231.nc`) will be solved but mos of them no

valeriupredoi:
note however that the concatenation function will work with any case of overlapping data, so it's not just for different experiments :beer:

valeriupredoi:
#210 merged, let's get this baby in @bouweandela @mattiarighi @schlunma :beer:

esmvalcore/preprocessor/_io.py
```diff
@@ -90,6 +91,16 @@ def concatenate(cubes):
...
         return concatenated[0]
+    if len(concatenated) == 2:
+        try:
+            concatenated[0].coord('time')
+            concatenated[1].coord('time')
+        except iris.exceptions.CoordinateNotFoundError:
+            pass
+        else:
+            concatenated = _concatenate_two_overlapping_cubes(concatenated)
+            if len(concatenated) == 1:
```
bouweandela: these two lines are duplicated, if you move the `if len(concatenated) == 2` block before the `if len(concatenated) == 1` statement above, they are not needed.

valeriupredoi: no, there are two possible cases stemming from two different concatenation processes: the first if check - `if len(concatenated) == 1: return concatenated[0]` comes from simple concatenation and if that returns false then we check on the length to be exactly 2 - `if len(concatenated) == 2: try: except: else: if len(concatenated) == 1: return concatenated[0]` note that the seemingly repetive check on the lenght = 1 comes from the `_concatenate_two_overlapping_cubes`

bouweandela: There is really no need to check `if len(concatenated) == 1` two times, can you have another look please?

tests/integration/preprocessor/_io/test_concatenate.py
```diff
@@ -14,33 +14,41 @@ class TestConcatenate(unittest.TestCase):
...
-        self.assertTrue((concatenated.coord('coord').points == np.array(
+        self.assertTrue((concatenated.coord('time').points == np.array(
             [1, 2, 3, 4, 5, 6])).all())
 
-    def test_fail_with_duplicates(self):
-        """Test exception raised if two cubes are overlapping."""
-        self.raw_cubes.append(self.raw_cubes[0].copy())
-        with self.assertRaises(ValueError):
-            _io.concatenate(self.raw_cubes)
+    def test_concatenate_with_overlap(self):
```
bouweandela: You still need to test for failing if there is more than 2 cubes

valeriupredoi:
@bouweandela cheers for the review, dude. It seems to me that you are not quite fully understanding the functionality of this implementation - yes, we need to test twice for the length of cubeList since once is from the actual standard concatenation (and the smart concatenation can only be performed if that returns 2 cubes) and the second time is for the actual result of the smart concatenation - if it's one cube we have bingo! if not we let the function fail and throw exception - this comment also takes care of your second comment :beer:

tests/integration/preprocessor/_io/test_concatenate.py
valeriupredoi: that is taken care by the older test

valeriupredoi:
OK documentation needs to be addressed indeed :beer:

tests/integration/preprocessor/_io/test_concatenate.py
bouweandela: which you removed in this pull request..

valeriupredoi: dang! did I? :man_facepalming:

valeriupredoi:
Mr @bouweandela you are to find a myriad of tests now :grin: Where do you think we should put the Concatenation info in the doc? I didn't seem to create and deciated section called Concatenation when I rehauled the preprocessor documentation :beer:

esmvalcore/preprocessor/_io.py
```diff
@@ -269,3 +280,63 @@ def _write_ncl_metadata(output_dir, metadata):
...
     return filename
+
+
+def _concatenate_two_overlapping_cubes(cubes):
+    """Concatenate time-overlapping cubes (two cubes only)."""
+    # we arrange [cube1, cube2] so that cube1.start <= cube2.start
+    if cubes[0].coord('time').points[0] <= cubes[1].coord('time').points[0]:
+        cubes = [cubes[0], cubes[1]]
+    else:
+        cubes = [cubes[1], cubes[0]]
```
JaroCamphuijsen: @valeriupredoi This case is not covered in the tests yet.

esmvalcore/preprocessor/_io.py
```diff
@@ -269,3 +280,63 @@ def _write_ncl_metadata(output_dir, metadata):
...
+    data_start_2 = time_2.cell(0).point
+    data_end_1 = time_1.cell(-1).point
+    data_end_2 = time_2.cell(-1).point
+
+    # case 1: both cubes start at the same time -> return longer cube
+    if data_start_1 == data_start_2:
+        if data_end_1 < data_end_2:
+            cubes = [cubes[1]]
+        else:
+            cubes = [cubes[0]]
```
JaroCamphuijsen: @valeriupredoi This case is not covered in the tests yet.

doc/esmvalcore/datafinder.rst
```diff
@@ -257,3 +257,22 @@ Since observational data are organized in Tiers depending on their level of
...
+
+Oftentimes data retrieving results in assembling a continuous data stream from
+multiple files or even, multiple experiments. The internal mechanism through which
+the assembly is done is via cube concatenation. One peculiarity of iris concatenation
+(see `iris cube concatenation <https://scitools.org.uk/iris/docs/latest/userguide/merge_and_concat.html>`_)
+is that it doesn't allow for concatenating time-overlapping cubes; this case is rather
+frequent with data from models overlapping in time, and is accounted for by a function that performs a
+flexible concatenation between two cubes, depending on the particular setup:
+
+* cubes overlap time: resulting cube is made up of the overlapping data plus left and
```
bouweandela: Can you explain here how you choose which data to use? This matters in case the overlapping data is not the same.

esmvalcore/preprocessor/_io.py
valeriupredoi: good call, mate! Now it has full coverage :beer:

doc/esmvalcore/datafinder.rst
valeriupredoi: just did, good call, man!

valeriupredoi:
really need an approve here folks @mattiarighi @ledm @schlunma and anyone else pls

esmvalcore/preprocessor/_io.py
```diff
@@ -269,3 +281,63 @@ def _write_ncl_metadata(output_dir, metadata):
...
+
+    # case 2: cube1 starts before cube2
+    else:
+        # find time overlap, if any
+        start_overlap = next(
+            (time_1.units.num2date(t) for t in time_1.points
+             if t in time_2.points),
+            None
+        )
+        # case 0: no overlap (new iris implementaion does allow
```
bouweandela: This seems to conflict with `# case 2.3: there is no overlap: return original cubes` defined on line 341. Also, I think this check is not needed here, because a discontinuous time coordinate is something that should be found by the cmor check.

tests/integration/preprocessor/_io/test_concatenate.py
```diff
@@ -14,34 +16,174 @@ class TestConcatenate(unittest.TestCase):
...
+        self.raw_cubes.append(
+            Cube([33., 55.],
+                 var_name='sample',
+                 dim_coords_and_dims=((time_coord, 0), )
+                 )
+        )
+        with self.assertRaises(TypeError):
             _io.concatenate(self.raw_cubes)
 
+    def test_fail_on_units_concatenate_with_overlap(self):
```
bouweandela: Why should this fail? Can't you can just convert all the time coordinates to the same units?

valeriupredoi: Yes, but this is a catch so that in case there are situations that are not picked up by cmor.fix

tests/integration/preprocessor/_io/test_concatenate.py
```diff
@@ -14,34 +16,174 @@ class TestConcatenate(unittest.TestCase):
...
+            [1., 2., 5., 7., 100.], var_name='time', standard_name='time',
+            units='days since 1950-01-01')
+        cube2 = Cube([33., 44., 55., 77., 1000.],
+                     var_name='sample',
+                     dim_coords_and_dims=((time_coord_2, 0), ))
+        cubes_ordered = [cube2, cube1]
+        concatenated = _io.concatenate(cubes_ordered)
+        self.assertTrue(np.allclose(
+            concatenated.coord('time').points,
+            np.array([1., 2., 5., 7., 100.])
```
bouweandela: It looks like the time point at 1.5 as defined on line 112 is simply discarded, I'm not sure this is the kind of behaviour that users would find intuitive

esmvalcore/preprocessor/_io.py
valeriupredoi: Yeah that line 341 is not actually a case per-se (need to remove the comment) - it's safer to raise the error if no overlap than to allow the concatenation to happen and then have it found by cmor.check when looking for time discontinuities

tests/integration/preprocessor/_io/test_concatenate.py
valeriupredoi: Yeh well, it's part of the plan :grin:

tests/integration/preprocessor/_io/test_concatenate.py
```diff
@@ -14,34 +16,174 @@ class TestConcatenate(unittest.TestCase):
...
+        ))
+
+    def test_concatenate_with_overlap_2(self):
+        """Test a more generic case."""
+        self._add_cube([65., 75.], [3., 200.])
+        self._add_cube([65., 75.], [1000., 7000.])
+        concatenated = _io.concatenate(self.raw_cubes)
+        self.assertTrue(np.allclose(
+            concatenated.coord('time').points,
+            np.array([1., 2., 3., 4., 5., 6., 1000., 7000.])
```
bouweandela: Where did the time point 200 as defined on line 57 go?

valeriupredoi: Thing is, the overlap can not be one single border point, so the [3, 200] cube is not counted in

esmvalcore/preprocessor/_io.py
bouweandela: I'm not sure, because in the usual case no overlap is expected. E.g. if concatenation fails because there are problems related to coordinates other than the time coordinate, you will get this error, which is really confusing..

valeriupredoi: Ermm, that is true, but then again, having no overlap in time is still a valid issue with the cubes that must be reported...

valeriupredoi: Oh actually, yeah, you're right! in a standard case there is no overlap and the lack of it will be reported - we need to remove that error, maybe make it a logger debug or warning?

tests/integration/preprocessor/_io/test_concatenate.py
bouweandela: At the very least, I think it would be good to add some warning log messages telling the user that some data is ignored. 

If there is overlapping data, I think it would be nice to have a debug log message telling the user what data from what cube you're using.

valeriupredoi: Yeah that'd be very sensible, but it's a bit of a headache. I'll see what I can do after Xmas :grin: I still think we should get this in sopner than later because it's quite needed, maybe we can open an issue and a new PR for these logger mssges after merge?

bouweandela: I would prefer to have it working properly before merging. Silently ignoring data makes the tool really hard to understand.

esmvalcore/preprocessor/_io.py
bouweandela: I think it's enough to simply remove the check and exception here

valeriupredoi:
OK @bouweandela debug messages added - now it should be very clear what data is being used and how :beer: Can we go ahead pls :medal_military:

esmvalcore/preprocessor/_io.py
```diff
@@ -269,3 +281,88 @@ def _write_ncl_metadata(output_dir, metadata):
...
+    time_2 = cubes[1].coord('time')
+    data_start_1 = time_1.cell(0).point
+    data_start_2 = time_2.cell(0).point
+    data_end_1 = time_1.cell(-1).point
+    data_end_2 = time_2.cell(-1).point
+
+    # case 1: both cubes start at the same time -> return longer cube
+    if data_start_1 == data_start_2:
+        if data_end_1 < data_end_2:
+            logger.debug(f"Both cubes start at the same time but cube %s "
```
bouweandela:
```suggestion
            logger.debug("Both cubes start at the same time but cube %s "
```
no need to make an f-string if you're using it as a normal string

esmvalcore/preprocessor/_io.py
```diff
@@ -269,3 +281,88 @@ def _write_ncl_metadata(output_dir, metadata):
...
+        # find time overlap, if any
+        start_overlap = next(
+            (time_1.units.num2date(t) for t in time_1.points
+             if t in time_2.points),
+            None
+        )
+        # case 0: no overlap (new iris implementaion does allow
+        # concatenation of cubes with no overlap)
+        if not start_overlap:
+            raise ValueError(f"Attempting to concatenate cubes that are"
```
bouweandela:
```suggestion
            logger.debug("Unable to concatenate non-overlapping cubes\n%s\nand\n%s"
            return cubes
```
because of this https://github.com/ESMValGroup/ESMValCore/pull/280/files#r360837753

bouweandela:
@valeriupredoi I made some minor changes, can you check if you're still happy with the result?

esmvalcore/preprocessor/_io.py
```diff
@@ -305,52 +307,56 @@ def _concatenate_overlapping_cubes(cubes):
 
     # case 1: both cubes start at the same time -> return longer cube
     if data_start_1 == data_start_2:
-        if data_end_1 < data_end_2:
-            logger.debug("Both cubes start at the same time but cube %s "
-                         f"ends before %s", cubes[0], cubes[1])
-            logger.debug(f"Cube %s contains all needed data "
-                         f"so using it fully", cubes[1])
+        if data_end_1 <= data_end_2:
```
valeriupredoi: good catch!

valeriupredoi:
@mattiarighi you wanna test pls mate? :beer:

Issue #284, opened 2019-10-01
CMIP6 cmor table error ssp370


ledm:
There's an error in the CMOR table for the experiment SSP3 7.0. 

Looking at the file [`esmvalcore/cmor/tables/cmip6/Tables/CMIP6_CV.json`, line 8414](https://github.com/ESMValGroup/ESMValCore/blob/8a762f026e4df44929f8f0503bbc35251d90a55b/esmvalcore/cmor/tables/cmip6/Tables/CMIP6_CV.json#L8414):
```
            "ssp370":{
                "activity_id":[
                    "ScenarioMIP AerChemMIP"
                ],
```

When ESMValCore looks for the ssp370 experiemnt on BADC, it tries to set the `[activity]` section to `ScenarioMIP AerChemMIP`, which of course doesn't exist. 

The error message is: 
```
2019-10-01 12:46:05,861 UTC [62192] DEBUG   Skipping non-existent
 /badc/cmip6/data/CMIP6/ScenarioMIP AerChemMIP/MOHC/UKESM1-0-LL
/ssp370/r1i1p1f2/Amon/tas/gn/[latestversion]
```

I've made a local fix for this, but I don't think that we're supposed to edit the CMOR tables. Do we need to raise it with them? Alternatively, should we instead fix `_recipe.py` so that it can deal with spaces in the CMOR table entries?

valeriupredoi:
this should definitely be raised with the CMOR folk since the `activity_id` is a list of mip's and this is clearly a malformation; we can put a check and split the string by space, but this will mask any other possible issues

ledm:
Created CMOR issue here: https://github.com/PCMDI/cmip6-cmor-tables/issues/264

taylor13:
@ledm @ValerioLembo:  According to the first entry in the first table in https://goo.gl/v1drZl, which defines the global attributes, ``activity_id`` should be a list of space-separated activities, so I think CMIP6_CV.json is correct.  Do you agree?

ledm:
Hi @taylor13, thanks for the quick reply. That is indeed correct. It looks like the problem is at our end.

@valeriupredoi, @matti, @bouweandela, I guess we need to change `_recipy.py` such that it can deal with space characters in `activity`. However, looking at @taylor13's document, we also need to mange spaces in the following CMIP data attributes:
- `activity`
- `external_variables`
- `realm`
- `source`
Are we handling this metadata correctly at the moment?

[...]

PR #285, opened 2019-10-01
Fix CMOR table.

ledm:
PR associated with issue #284. 

Not sure that we want to merge this, just wanted to show the changes that were needed to get it running again.

### DIFF
--- a/esmvalcore/cmor/tables/cmip6/Tables/CMIP6_CV.json
+++ b/esmvalcore/cmor/tables/cmip6/Tables/CMIP6_CV.json
@@ -8411,7 +8411,7 @@
             },
             "ssp370":{
                 "activity_id":[
-                    "ScenarioMIP AerChemMIP"
+                    "ScenarioMIP"
                 ],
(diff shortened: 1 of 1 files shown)


valeriupredoi:
let's not go about changing CMOR tables, especially the Controlled Vocabularies; instead, alert the CMOR folk. Also, you 100% sure the `activity_id` is not `["ScenarioMIP", "AerChemMIP"]` (ie both but correctly formatted? I tried looking for multiple activity_id's but couldn't really, so I might be totally blabbering, but double checking didnt kill anyone yet)

ledm:
Created CMOR issue here: https://github.com/PCMDI/cmip6-cmor-tables/issues/264

valeriupredoi:
great! I reckon it's safe to close this PR but keep the issue open until the CMOR issue gets fixed and we'll update the CMOR tables version (btw are we actually at the latest CMOR version?)

ledm:
We're not at the most recent version, as the bug was on a different line in their recent version. 

Fine with me to close this PR.

PR #288, opened 2019-10-02
Added vegfrac as a derived variable

ledm:
This PR adds vegfrac as a derived variable. It's a really simple variable (1 - bare soil fraction). Surprised that it didn't exist already. It would be good if some land surface experts could weigh in on this though, as it's fairly basic.

### DIFF
--- a/esmvalcore/preprocessor/_derive/vegfrac.py
+++ b/esmvalcore/preprocessor/_derive/vegfrac.py
@@ -18,5 +18,5 @@ class DerivedVariable(DerivedVariableBase):
         baresoilFrac_cube = cubes.extract_strict(
             iris.Constraint(name='area_fraction'))
 
-        baresoilFrac_cube.data = 1. - baresoilFrac_cube.data
+        baresoilfrac_cube.data = 1. - baresoilfrac_cube.data
         return baresoilFrac_cube
--- a/esmvalcore/preprocessor/_derive/vegfrac.py
+++ b/esmvalcore/preprocessor/_derive/vegfrac.py
@@ -18,5 +18,5 @@ class DerivedVariable(DerivedVariableBase):
         baresoilfrac_cube = cubes.extract_strict(
             iris.Constraint(name='area_fraction'))
 
-        baresoilfrac_cube.data = 1. - baresoilfrac_cube.data
+        baresoilfrac_cube = -1. * baresoilfrac_cube + 1.
         return baresoilfrac_cube
--- a/esmvalcore/preprocessor/_derive/vegfrac.py
+++ b/esmvalcore/preprocessor/_derive/vegfrac.py
@@ -1,0 +1,22 @@
+"""Derivation of variable `vegFrac`."""
+
+from ._baseclass import DerivedVariableBase
+import iris
+
+
+class DerivedVariable(DerivedVariableBase):
+    """Derivation of variable `vegFrac`."""
+
+    # Required variables
+    required = [{'short_name': 'baresoilFrac', }]
+
+    @staticmethod
+    def calculate(cubes):
+        """Compute vegetation fraction from bare soil fraction.
+
+        """
+        baresoilFrac_cube = cubes.extract_strict(
+            iris.Constraint(name='area_fraction'))
+
+        baresoilFrac_cube.data = 1. - baresoilFrac_cube.data
+        return baresoilFrac_cube
--- a/esmvalcore/preprocessor/_derive/vegfrac.py
+++ b/esmvalcore/preprocessor/_derive/vegfrac.py
@@ -18,5 +18,5 @@ class DerivedVariable(DerivedVariableBase):
         baresoilfrac_cube = cubes.extract_strict(
             iris.Constraint(name='area_fraction'))
 
-        baresoilfrac_cube = -1. * baresoilfrac_cube + 1.
+        baresoilfrac_cube.data = 1. - baresoilfrac_cube.core_data()
         return baresoilfrac_cube
--- a/esmvalcore/preprocessor/_derive/vegfrac.py
+++ b/esmvalcore/preprocessor/_derive/vegfrac.py
@@ -1,7 +1,7 @@
 """Derivation of variable `vegFrac`."""
 
-from ._baseclass import DerivedVariableBase
 import iris
+from ._baseclass import DerivedVariableBase
 
 
 class DerivedVariable(DerivedVariableBase):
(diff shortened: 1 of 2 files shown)


esmvalcore/preprocessor/_derive/vegfrac.py
```diff
@@ -0,0 +1,22 @@
+"""Derivation of variable `vegFrac`."""
+
+from ._baseclass import DerivedVariableBase
+import iris
```
valeriupredoi: reverse import order

ledm:
Any reason not to merge this today? It's only a very simple diagnostic with two small files.

mattiarighi:
Is this implemented _without_ realizing the data?
See the discussion in #42.

If yes, it can be merged.

mattiarighi:
Can you also check if the Codacy issue can be addressed?

esmvalcore/preprocessor/_derive/vegfrac.py
```diff
@@ -0,0 +1,22 @@
...
+
+    # Required variables
+    required = [{'short_name': 'baresoilFrac', }]
+
+    @staticmethod
+    def calculate(cubes):
+        """Compute vegetation fraction from bare soil fraction.
+
+        """
+        baresoilFrac_cube = cubes.extract_strict(
```
ledm:
```suggestion
        baresoilfrac_cube = cubes.extract_strict(
```

esmvalcore/preprocessor/_derive/vegfrac.py
```diff
@@ -0,0 +1,22 @@
...
+            iris.Constraint(name='area_fraction'))
+
+        baresoilFrac_cube.data = 1. - baresoilFrac_cube.data
```
ledm:
```suggestion
        baresoilfrac_cube.data = 1. - baresoilfrac_cube.data
```

esmvalcore/preprocessor/_derive/vegfrac.py
```diff
@@ -0,0 +1,22 @@
...
+        return baresoilFrac_cube
```
ledm:
```suggestion
        return baresoilfrac_cube
```

esmvalcore/preprocessor/_derive/vegfrac.py
```diff
@@ -0,0 +1,22 @@
...
+
+    @staticmethod
+    def calculate(cubes):
+        """Compute vegetation fraction from bare soil fraction.
+
+        """
+        baresoilfrac_cube = cubes.extract_strict(
+            iris.Constraint(name='area_fraction'))
+
+        baresoilfrac_cube.data = 1. - baresoilfrac_cube.data
```
ledm: @mattiarighi this is the only like that manipulates the data in this derivation. I suspect that something this simple shouldn't realise the data, right?

mattiarighi: Not sure. @schlunma what do you think?

ledm: Looking at point 6.3 from https://scitools.org.uk/iris/docs/latest/userguide/real_and_lazy_data.html

I would guess that this is not lazy. However, if we change it to this suggestion, it should be okay.
```suggestion
        baresoilfrac_cube= 1. - baresoilfrac_cube
```

valeriupredoi: and that returns a dask array, so you can just do `baresoilfrac_cube.core_data() = 1. - baresoilfrac_cube.core_data()` and to realize it, `baresoilfrac_cube.compute()`

ledm: So you're suggesting:
```suggestion
        baresoilfrac_cube.data() = 1. - baresoilfrac_cube.data()
        baresoilfrac_cube.compute()
```

mattiarighi: Looking at other derivation functions (e.g. [rlns](https://github.com/ESMValGroup/ESMValCore/blob/development/esmvalcore/preprocessor/_derive/rlns.py)), it looks like you simply need:
```
baresoilfrac_cube = 1. - baresoilfrac_cube
```

mattiarighi: Ah, but that's not supported, crap.

mattiarighi: In [_shared.py](https://github.com/ESMValGroup/ESMValCore/blob/development/esmvalcore/preprocessor/_derive/_shared.py) there are similar operations.

valeriupredoi: sorry my line is `baresoilfrac_cube.data = 1. - baresoilfrac_cube.core_data()` - note that calling `.data` here keeps things lazy and the operation is done using the dask array `core_data()`

ledm: I found that that doesn't work, but this weird hack does seem to work:
```
        baresoilfrac_cube = -1. * baresoilfrac_cube + 1.
```

esmvalcore/preprocessor/_derive/vegfrac.py
```diff
@@ -18,5 +18,5 @@ def calculate(cubes):
         baresoilfrac_cube = cubes.extract_strict(
             iris.Constraint(name='area_fraction'))
 
-        baresoilfrac_cube.data = 1. - baresoilfrac_cube.data
+        baresoilfrac_cube = -1. * baresoilfrac_cube + 1.
```
valeriupredoi: come on dude, use `cube.data = 1. - cube.core_data()`, this is hilarious :grin:

esmvalcore/preprocessor/_derive/vegfrac.py
valeriupredoi: hahaha, it works of course, this is a hilarious implementation man :grin:

valeriupredoi: @mattiarighi yeah, also not the way to do it :imp:

ledm:
I don't understand why
```
1. - cube
```
doesn't work, but
```
-1*cube +1.
```
does.

mattiarighi:
But this is still realizing the data, right?

ledm:
The version that is in there now is not realising the data. I checked with a `has_lazy_data()` which outputted to `True`.

valeriupredoi:
Yes, it doesnt realize the data coz now you are assigning to function a
lazy array that keeps its lazy nature through the subtraction op. In other
words, cube.data = something lazy will not realize the data. @ledm your
previous 1 - cube not working should be raised with the iris guys, since
scalar x cube works, so scalar +/- cube should work too; note that cube +/-
scalar works, as you pointed out

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

Issue #294, opened 2019-10-07
Failure of applying masks to OBS datasets


bascrezee:
When trying to apply `mask_landsea` to an Observational dataset, it fails. 

```
datasets:
  - {dataset: CDS-SATELLITE-SOIL-MOISTURE, type: sat, project: OBS, mip: Lmon, version: COMBINED-TCDR-v201812.0.0, tier: 3, start_year: 2017,  end_year: 2018}

preprocessors:
  pp_test_mask:
    mask_landsea:
      mask_out: sea 
      always_use_ne_mask: True
```

Why is `_get_correct_fx_file` called for OBS datasets? This shouldn't happen, right? 

```
  File "/ESMValCore/esmvalcore/_recipe.py", line 873, in __init__
    self.tasks = self.initialize_tasks() if initialize_tasks else None
  File "/ESMValCore/esmvalcore/_recipe.py", line 1220, in initialize_tasks
    task_name=task_name,
  File "/ESMValCore/esmvalcore/_recipe.py", line 844, in _get_preprocessor_task
    name=task_name,
  File "/ESMValCore/esmvalcore/_recipe.py", line 712, in _get_single_preprocessor_task
    config_user=config_user,
  File "/ESMValCore/esmvalcore/_recipe.py", line 662, in _get_preprocessor_products
    config_user=config_user)
  File "/ESMValCore/esmvalcore/_recipe.py", line 429, in _update_fx_settings
    'sftlf': _get_correct_fx_file(variable, 'sftlf', config_user),
  File "/ESMValCore/esmvalcore/_recipe.py", line 402, in _get_correct_fx_file
    fx_files = get_input_filelist(variable=fx_var,
UnboundLocalError: local variable 'fx_var' referenced before assignment
2019-10-07 11:55:20,619 UTC [21865] INFO    If you suspect this is a bug or need help, please open an issue on https://github.com/ESMValGroup/ESMValTool/issues and attach the run/recipe_*.yml and run/main_log_debug.txt files from the output directory.
```

valeriupredoi:
a bug! But not really an accidental one since *some* OBS's actually have fx files associated (@mattiarighi pointed me to ERA-Interim I believe) - this is something that we've not actually agreed on - ie what to do when fx data files are not found. @bouweandela had the very Dutch opinion to stop the code and tell the user there are none :grin: My take is to let it go in certain cases (like this one here)

bascrezee:
> a bug! 
[...]

PR #295, opened 2019-10-07
exit right out when trying to get fx files for OBS

valeriupredoi:
Addresses the issue with OBS datasets not having fx files (well, at least most of them) and reported in #294 

**CAVEAT** this is a guillotine treatment that is not sensitive to the OBS datasets that may have fx files and should also be always used with  `always_use_ne_mask: True` as arg in the masking preprocessor

Up for discussion :floppy_disk:

### DIFF
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -386,6 +386,10 @@ def _add_fxvar_keys(fx_var_dict, variable):
 def _get_correct_fx_file(variable, fx_varname, config_user):
     """Wrapper to standard file getter to recover the correct fx file."""
     var = dict(variable)
+    # return empty if using OBS
+    # TODO what with OBS data that actually have fx files? Are there any?
+    if var['project'] == 'OBS':
+        return None
     if var['project'] == 'CMIP5':
         fx_var = _add_fxvar_keys({'short_name': fx_varname, 'mip': 'fx'}, var)
     elif var['project'] == 'CMIP6':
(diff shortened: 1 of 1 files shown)


valeriupredoi:
@mattiarighi @hb326 @rswamina you guys have a ton of experience with OBSes - which ones have fx data? If none then we can safely leave the cut-throat approach from here, if otherwise we need to plug in something more fancy :beer:

esmvalcore/_recipe.py
```diff
@@ -386,6 +386,10 @@ def _add_fxvar_keys(fx_var_dict, variable):
 def _get_correct_fx_file(variable, fx_varname, config_user):
     """Wrapper to standard file getter to recover the correct fx file."""
     var = dict(variable)
+    # return empty if using OBS
+    # TODO what with OBS data that actually have fx files? Are there any?
+    if var['project'] == 'OBS':
```
bascrezee:
```suggestion
    if var['project'] in ['OBS', 'OBS6']:
```

mattiarighi:
At the moment only `ERA-Interim` and `NSIDC-0116`.
But other ones are coming, e.g. [#1042](https://github.com/ESMValGroup/ESMValTool/pull/1042).

valeriupredoi:
> At the moment only `ERA-Interim` and `NSIDC-0116`.
> But other ones are coming, e.g. [#1042](https://github.com/ESMValGroup/ESMValTool/pull/1042).

damn! thought I might get out easy :grin: OK we need a dedicated issue for that then

valeriupredoi:
let's plug this in for now so we have a somewhat decent functionality and then concern about the others later

valeriupredoi:
#297 supersedes this so am gonna close w/o merge

Issue #299, opened 2019-10-08
Add option to standardize to anomalies preprocessor function


bascrezee:
See PR referenced below.

PR #300, opened 2019-10-08
Add option to standardize to anomalies preprocessor function

bascrezee:
I added a new preprocessor function, making use of two other existing preprocessor functions. This is the description I provided in the docstring: 

    This function standardizes the input data. It calculates the anomalies
    (x-mean(x)) and divides them through the standard deviation where both
    anomalies and standard deviation are calculated over the specified period.

As far as I could see it is not possible to achieve this behaviour by combining existing preprocessor functions in the recipe, therefore I needed to add this as a new preprocessor function.

**Tasks**

-   [x] https://github.com/ESMValGroup/ESMValCore/issues/299 
-   [x] Add unit tests
-   [x] Public functions should have a numpy-style docstring so they appear properly in the [API documentation]
-  [x] If writing a new/modified preprocessor function, please update the [documentation](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/esmvalcore/preprocessor.html)
-   [x] If you make backward incompatible changes to the recipe format, make a new pull request in the [ESMValTool repository](https://github.com/ESMValGroup/ESMValTool) and add the link below --> there are no backward incompatible changes, as far as I can see

Closes https://github.com/ESMValGroup/ESMValCore/issues/299


bascrezee:
Can I have a quick review (by @jvegasbsc or @valeriupredoi or ... ?)  before proceeding with writing a unit test and adding it to the documentation?

jvegasbsc:
Looks nice, only that I think it will be better to have it as an option in the anomalies preprocessor.

esmvalcore/preprocessor/_time.py
```diff
@@ -459,6 +459,52 @@ def anomalies(cube, period):
     return cube
 
 
+def standardize(cube, period='full'):
+    """Standardize the input data with the specified granularity.
+
+    This function standardizes the input data. It calculates the anomalies
+    (x-mean(x)) and divides them through the standard deviation where both
+    anomalies and standard deviation are calculated over the specified period.
```
valeriupredoi: divides them by the standard deviation

esmvalcore/preprocessor/_time.py
```diff
@@ -459,6 +459,52 @@ def anomalies(cube, period):
...
+    """
+    cube_anomalies = anomalies(cube, period)
+    cube_stddev = climate_statistics(cube, operator='std_dev', period=period)
+
+    if period == 'full':
+        cube.data = cube_anomalies.data / cube_stddev.data
+    elif period in ['season', 'seasonal', 'monthly',
+                    'month', 'mon', 'daily', 'day']:
+        ratio = [i / j for i, j in
+                 zip(cube_anomalies.shape, cube_stddev.shape)]
```
valeriupredoi: this will fail if `cube_anomalies.shape == cube_stddev.shape`, can you put a check and raise an exception pls?

esmvalcore/preprocessor/_time.py
```diff
@@ -459,6 +459,52 @@ def anomalies(cube, period):
...
+        # This will raise an error if the length of the cube their time axes
+        # are not multiples of each other or if other shapes are not equal
+        if not all([ratio[0] % 1 == 0] + [i == 1 for i in ratio[1:]]):
+            raise ValueError("Can not safely apply preprocessor to this data")
```
valeriupredoi: yeps but pls say why in the exception

bascrezee:
> nice one! a couple minor comments. Also, a bit confusing to me (as a non-scientist) as to why monthly, month, mon -> maybe detail this in the func docstring, defo needed in the documentation

This is equally confusing to me, my function inherits this from the other functions. Maybe @jvegasbsc can explain this?

bascrezee:
> Looks nice, only that I think it will be better to have it as an option in the anomalies preprocessor.

Yes, I agree. That would be an option as well. What do others think?

mattiarighi:
I also prefer extending existing preprocessors rather than creating new ones.

esmvalcore/preprocessor/_time.py
bascrezee: why exactly? I don't see why this would fail. I think it would fail only if `cube_anomalies.ndim != cube_stddev.ndim` is that the check that you would like to see?

valeriupredoi: sorry, fail *silently* (I ate the most important word) ie zip will work fine but if the dims are not equal, it will return truncated results that will not be detectable until probably at the last func line, if at all, and even there when you are tiling, it may tile on a wrong tile grid; statistically it's important to have equal dimensionality and any silent fail will introduce a cloaked bug

jvegasbsc:
> This is equally confusing to me, my function inherits this from the other functions. Maybe @jvegasbsc can explain this?

To support all common abreviatures used for monthly data

esmvalcore/preprocessor/_time.py
valeriupredoi: aka :dagger:

bascrezee: still not clear to me, can you check what you wrote originally (3h ago)? I see your point that an extra check here is needed, but I don't think you wrote originally (`shape`) what you intended (`dims`?) to write ?

valeriupredoi: sorry Bas, this fell swiftly though the cracks. Yes, so what I meant was if `cube_anomalies.shape != cube_stddev.shape` but the case to be reckoned with is if the dims are different (could they be?) ie `cube_anomalies.shape = (3, 4, 5)` and `cube_stddev.shape = (3, 4)` - in that case the ratios will be computed only for the common pairs `(i, j)` that `zip` returns

bascrezee: Ok, now I see! Didn't expect the zipping to work when unequal lengths are passed, good to know ! I will built in a check before.

bascrezee:
@jvegasbsc  I just included it as an option in the existing anomaly function. Could you test it?

esmvalcore/preprocessor/_time.py
```diff
@@ -456,6 +461,26 @@ def anomalies(cube, period):
...
+        # This will raise an error if the length of the cube their time axes
+        # are not multiples of each other or if other shapes are not equal
+        if not all([ratio[0] % 1 == 0] + [i == 1 for i in ratio[1:]]):
+            raise ValueError(
+                "Cannot safely apply preprocessor to this dataset,\
+                since the full time period of this dataset is not\
+                a multiple of the period {0}".format(period)
+            )
+        reps = tuple([int(i) for i in ratio])
+        cube.data = cube.data / np.tile(cube_stddev.data, reps)
```
bouweandela:
```suggestion
        cube.data = cube.core_data() / da.tile(cube_stddev.core_data(), reps)
```
Please try to keep the data lazy, maybe using something like the above?

esmvalcore/preprocessor/_time.py
```diff
@@ -456,6 +461,26 @@ def anomalies(cube, period):
         data[indexes] = data[indexes] - ref[cube_coord.points[i]]
 
     cube = cube.copy(data)
+
+    # Standardize the results if requested
+    if standardize:
+        cube_stddev = climate_statistics(cube,
+                                         operator='std_dev',
+                                         period=period)
+        assert cube.ndim == cube_stddev.ndim
```
bouweandela: If you think this will ever happen, please raise an exception with a nice error message instead of using `assert`

esmvalcore/preprocessor/_time.py
```diff
@@ -456,6 +461,26 @@ def anomalies(cube, period):
...
+        ratio = [i / j for i, j in
+                 zip(cube.shape, cube_stddev.shape)]
+        # This will raise an error if the length of the cube their time axes
+        # are not multiples of each other or if other shapes are not equal
+        if not all([ratio[0] % 1 == 0] + [i == 1 for i in ratio[1:]]):
+            raise ValueError(
+                "Cannot safely apply preprocessor to this dataset,\
```
bouweandela:
```suggestion
                "Cannot safely apply preprocessor to this dataset, "
```

esmvalcore/preprocessor/_time.py
```diff
@@ -456,6 +461,26 @@ def anomalies(cube, period):
...
+                since the full time period of this dataset is not\
```
bouweandela:
```suggestion
                "since the full time period of this dataset is not "
```

esmvalcore/preprocessor/_time.py
```diff
@@ -456,6 +461,26 @@ def anomalies(cube, period):
...
+                a multiple of the period {0}".format(period)
```
bouweandela:
```suggestion
                f"a multiple of the period '{period}'"
```

esmvalcore/preprocessor/_time.py
bascrezee: Tile is unfortunately not available in `dask`. [Here](https://stackoverflow.com/questions/38054970/what-is-the-dask-equivalent-of-numpy-tile) it is suggested to use concatenate, but it would be slightly more complicated in our case since `reps` is a tuple. Can we assume that the time axis always corresponds to index 0 ? In that case I can think of an implementation.

esmvalcore/preprocessor/_time.py
bascrezee: No I do not expect this to happen, so will take it out.

bascrezee: Or wait, this was actually a review suggestion by @valeriupredoi before... I'll go for an error message in that case :)

esmvalcore/preprocessor/_time.py
bascrezee: This is cool stuff. :)

esmvalcore/preprocessor/_time.py
bouweandela: Can't you simply use [`da.broadcast_to`](https://docs.dask.org/en/latest/array-api.html#dask.array.broadcast_to)?

bouweandela: I.e. `cube.data = cube.core_data() / da.broadcast_to(cube_stddev.core_data(), cube.shape)`

bascrezee: Actually broadcasting often fails here, so that is not a solution. But given the two checks above (which itself can be simplified a bit more, will think about this)  I am actually quite confident that concatenating along the first axis with given repetitions is safe. :)

esmvalcore/preprocessor/_time.py
```diff
@@ -442,20 +446,41 @@ def anomalies(cube, period):
...
+    if standardize:
+        cube_stddev = climate_statistics(cube,
+                                         operator='std_dev',
+                                         period=period)
+        if not cube.ndim == cube_stddev.ndim:
+            raise ValueError(
+                "Inconsistent dimensions between anomaly cube and "
+                f"standard deviation cube: {cube.ndim};{cube_stddev.ndim}"
+            )
+        ratio = [i / j for i, j in
```
bouweandela: You're only using the first item of `ratio`, so I think there is no need to compute the other items.

Note that you do not have to assume that the first dimension is the time dimension. You can also do something like
```python
tdim = cube.coord_dims('time')[0]
reps = cube.shape[tdim] / cube_stddev.shape[tdim]
weights = da.concatenate([cube_stddev.core_data() for _ in range(ratio)], axis=tdim)
cube.data = cube.core_data() / weights

**Comment by bascrezee** [comment]
Hi Bouwe, I updated the documentation. For the tests I am not sure if I can do that, I had a look into  `test_anomalies` but it is not so clear for me how to extend the different options given that we now add the option to `standardize`. Should all the possible combinations be covered? I would be interested to learn this, but have limited time until the end of this year. So it is maybe better if someone else can implement the tests for now. If someone has the time, that would be great.

File: doc/esmvalcore/preprocessor.rst [hunk 10]
```diff
@@ -818,11 +821,14 @@ Examples:
             anomalies:
                 period: month
 
-    * Anomalies from the full period climatology:
+    * Standardized anomalies from the full period climatology:
 
         .. code-block:: yaml
 
             anomalies:
+                standardize: True
```
bouweandela:
```suggestion
                standardize: true
```

bouweandela:
I have added the required test cases, but I'm not sure what the values should be. Could you add those? I have added a #TODO item where numpy arrays with the expected results should be added.

bascrezee:
Thanks Bouwe, working on it.

bascrezee:
How should I run the unit tests? I tried `python ./tests/unit/preprocessor/_time/test_time.py` which outputs: 
```
[ .... several warnings .... ]

----------------------------------------------------------------------
Ran 45 tests in 1.770s

OK
```

However, this does not call the function `test_anomalies` (checked by inserting `assert False` in the function).

bouweandela:
Use `pytest tests/unit/preprocessor/_time/test_time.py`

bascrezee:
I added a unit test for only one case, at the moment, as a proof of concept. I look forward to some feedback, maybe from @bouweandela or @jvegasbsc ? The strategy I took is to calculate the expected outcome by using `numpy` functions. Unfortunately this will get rather complicated for the other cases. Also, I noted, that actually there IS quite a difference in the outcome (see the `tolerances` that I set). I didn't find out yet why there are these differences, it is not related to the weighting along the time dimension, I checked that.

I do think that trying to put more functionality into one function (as suggested by @mattiarighi and @jvegasbsc above) in general renders unit testing way more complex (since the options that need to be covered equal the `product` between input arguments). Probably something to keep in mind in the future.

jvegasbsc:
> I added a unit test for only one case, at the moment, as a proof of concept. I look forward to some feedback, maybe from @bouweandela or @jvegasbsc ?

I modified it to simplify it a bit and corrected a couple of flake8 issues. Anyway, my suggestion will be to create a new method for testing the standardized case. I think it will be also a good idea to generate a new data cube that will make easier to know at a glance what the result should be

bouweandela:
> since the options that need to be covered equal the `product` between input arguments

Indeed this is only feasible if the number of possible input values is small. Parametrizing with pytest helps a bit, but not when the list of options/possible values grows very large. Usually people try to test at the very least every code path, have a look at the coverage report in test-reports/coverage_html/index.html and make sure there is no code that is not executed during a test (marked red).

bascrezee:
Just a small update. I got time again to work on this :) I did a first implementation solely for the case period='full'. Interestingly, the differences between `numpy` result and `esmvalcore preprocessor` result are pretty large (~0.0012 absolute difference). I checked and the difference can not be explained by weighting of the time axis (since all weights are 1 for this test case). I am now trying to track this further down.

valeriupredoi:
there is no weighting since `stdev` does not support weights computation - it seems to me that the difference comes from `np.stdev()` vs `cube.collapsed(axis, stdev)` no? :beer:

bascrezee:
> there is no weighting since `stdev` does not support weights computation - it seems to me that the difference comes from `np.stdev()` vs `cube.collapsed(axis, stdev)` no?

By default `Iris` uses `ddof=1` in the `STD_DEV` aggregator, whereas `numpy` uses `ddof=0` by default. After fixing this maybe `assert_allclose` can be replaced by `assert_equal`. :smile:

valeriupredoi:
here try this with a say, temperature cube:
```
import iris
import numpy as np

c = iris.load_cube("t1.nc")
# shape (1800, 96, 102)
c1 = c.collapsed('time', iris.analysis.STD_DEV)
c2 = np.std(c.data, axis=0, keepdims=True)[0, ...]
print(np.mean(c1.data - c2))
```
I get a mean delta of `10^-3` which is consistent with what you notice

tests/unit/preprocessor/_time/test_time.py
```diff
@@ -905,82 +905,87 @@ def make_map_data(number_years=2):
...
             expected = np.ma.masked_invalid(expected_stdanomalies)
-            assert_allclose(
+            assert_array_equal(
                 result.data,
-                expected,
-                # Note the pretty large tolerance, which is needed since
-                # there actually IS quite a difference --> WHY?
-                rtol=1.2e-5,
-                atol=0.0012
+                expected
```
valeriupredoi: you can still use `assert_allcolse` with an `rtol = 1e-5` like in https://numpy.org/doc/1.18/reference/generated/numpy.testing.assert_allclose.html

bascrezee: yes, but we don't need it anymore, not for this case at least.  It is better to use `assert_array_equal`, see latest commits. :)  I think this is nice, imagine in the future Iris decides (for whatever reason) to change `ddof` to 0, this test will complain :) and it should, since something is actually changing the results that would go unnoticed otherwise.

valeriupredoi: actually, it's better to use `allclose` with a say, `1e-5` tolerance since if iris changes and the test fails it means that the results differ by more than 1e-5 which is pretty bad (1e-4 for a temperature filed is already kinda bad, that means variations of order 0.01K per point which is edging on dangerous)

valeriupredoi: anyways, I might sound like Dr Pedanticus :grin:

bascrezee: I don't get your point, the testing as I currently implemented is more strict than what you are suggesting.

valeriupredoi: dont worry about it - all good as is now - my point was if the tolerances change for the iris operation this test as is now will fail, but if the abs diffs are still within 1e-5 that's fine, hence using `allclose` will still let the test pass. But it's fine, since if it fails, that'll prompt us to re evaluate the new differences anyway and see what's happened :+1:

bascrezee: > if it fails, that'll prompt us to re evaluate the new differences anyway and see what's happened
That was exactly my intention, this test should really flag when something changes. If we would have a very clear view on tolerable errors, we could implement that, but I would feel uncomfortable setting a threshold here.

doc/esmvalcore/preprocessor.rst
```diff
@@ -864,14 +864,17 @@ See also :func:`esmvalcore.preprocessor.climate_statistics`.
...
+anomalies from the full, seasonal, monthly and daily climatologies. Optionally
+standardized anomalies can be calculated.
 
 Parameters:
     * period: define the granularity of the climatology to use:
       full period, seasonal, monthly or daily.
       Available periods: 'full', 'season', 'seasonal', 'monthly', 'month',
       'mon', 'daily', 'day'. Default is 'full'
 
+    * standardize: if True calculate standardized anomalies (default: False)
```
bouweandela:
```suggestion
    * standardize: if true calculate standardized anomalies (default: false)
```
is the preffered yaml name for true and false

esmvalcore/preprocessor/_time.py
```diff
@@ -432,33 +432,63 @@ def anomalies(cube, period):
...
+            )
+        tdim = cube.coord_dims('time')[0]
+        reps = cube.shape[tdim] / cube_stddev.shape[tdim]
+        if not reps % 1 == 0:
+            raise ValueError(
+                "Cannot safely apply preprocessor to this dataset, "
+                "since the full time period of this dataset is not "
+                f"a multiple of the period '{period}'"
+            )
+        cube.data = cube.core_data() / \
```
bouweandela: Please avoid using backslashes, maybe you could do something like (or let `yapf` figure out the best formatting):
```python
      cube.data = cube.core_data() / da.concatenate(
            [cube_stddev.core_data() for _ in range(int(reps))], axis=tdim)
```

esmvalcore/preprocessor/_time.py
```diff
@@ -432,33 +432,63 @@ def anomalies(cube, period):
...
         data[indexes] = data[indexes] - ref[cube_coord.points[i]]
 
     cube = cube.copy(data)
+
+    # standardize the results if requested
+    if standardize:
+        cube_stddev = climate_statistics(cube,
+                                         operator='std_dev',
+                                         period=period)
+        if not cube.ndim == cube_stddev.ndim:
```
bouweandela: Does this ever happen?

esmvalcore/preprocessor/_time.py
```diff
@@ -432,33 +432,63 @@ def anomalies(cube, period):
...
     """
     reference = climate_statistics(cube, period=period)
     if period in ['full']:
-        return cube - reference
+        cube = cube - reference
+        if standardize:
+            cube_stddev = climate_statistics(
+                cube, operator='std_dev', period=period)
+            return cube / cube_stddev
+        else:
```
bouweandela: This `else` can be removed

bascrezee: Are you sure? One of the checkers also told me to remove it, but I think we need it. The code below line 452 should not be executed in the case that `period='full'`. `_get_period_coord` will raise ValueError, since `full` is not accepted as argument to `period`.

esmvalcore/preprocessor/_time.py
bascrezee: I hope not, I just wanted to make sure. But maybe it is better to do these checks in the unit tests if needed? I will take it out here.

esmvalcore/preprocessor/_time.py
bouweandela: The linter is right, becaues if `standardize` evaluates to `True`, you will leave the function on line 450. If it's false, it will continue. Maybe it would be more clear to write
```python
    if period == 'full':
        cube = cube - reference
        if standardize:
            cube_stddev = climate_statistics(
                cube, operator='std_dev', period=period)
            cube = cube / cube_stddev
        return cube

bascrezee: Yes! Now I get it. It suggests removing solely the `else` not the `else`-block. Your suggestion is much clearer I will go with that.

bascrezee:
Can this please be merged? @mattiarighi ?

PR #307, opened 2019-10-11
Development dry run

valeriupredoi:
OK chaps this is the start of https://github.com/ESMValGroup/ESMValTool/issues/1365 Currently it does pretty much all it needs to do:

- run the tool with `--dry-run` flag and it will run through the recipe and run a `default` preprocessor, without stopping where data is missing;
- it will tell you where data is missing via `MISSING DATA` in logger;
- it runs the cmor checks and fixes and **doesn't** output the fixed files and then it exits without doing anything else (no diagnostic) - writing to disk is switched off completely bar the `main_log` and `main_log_debug` and `resources` (what's in `run`) files;
- works seamlessly with derived variables and fx ones too;

It is pretty much the standard esmvaltool workflow but with all other stuffs than basics removed, and obviously no diagnostics.

CAVEAT: if there are issues with the cmor checks (either data or metadata) it will stop there, just as esmvaltool prod run would do; I need @jvegasbsc 's advice how to allow for just logging those errors and letting the flow on, I suggest using the same `dry_check` flag that gets passed through `_recipe.py` and allows for the dry run, but that means changing a bit the error handling in `cmor/` - what says you? :beer:


valeriupredoi:
BTW guys - this, a rather complicated workflow rerouting, in +42/-21 lines -> hell, note of a great robust codebase we have - cheers to yous :beer: See yous in Allemania next week, pub time now :beer:

valeriupredoi:
ok this one works now without saving and proper without running (or trying to run) any diagnostic, ready for review chaps :beer:

bascrezee:
This works great. It significantly speedens up testing of many datasets, since they do not need to be written to disk. I ran the test recipe that I created a few weeks ago containing about 350 dataset entries (CMIP5/CMIP6/OBS). This runs in about 2 minutes, compared to about 1 hour without the dry run option.

valeriupredoi:
@bouweandela did you see what @bascrezee said? Giddyup and review this pls, man :beer:

bascrezee:
Some test results from my side.

Following dataset / variable: 
```
    - {dataset: HadGEM3-GC31-LL, project: CMIP6, grid: gn, exp: historical, ensemble: r1i1p1f3}

      alb:
        <<: *variable_settings
        mip: Amon
        derive: true
        force_derivation: false
```

Gives an error in the variable `rsds` which is used to derive variable `alb`. Error does not occur in development branch.

```
  File "ESMValCore/esmvalcore/_recipe.py", line 1306, in run
    max_parallel_tasks=self._cfg['max_parallel_tasks'])
  File "ESMValCore/esmvalcore/_task.py", line 599, in run_tasks
  File "ESMValCore/esmvalcore/_task.py", line 645, in _run_tasks_parallel
    _copy_results(task, running[task])
  File "ESMValCore/esmvalcore/_task.py", line 666, in _copy_results
    task.output_files, updated_products = future.get()
  File "/crezees/conda/envs/esmvaltool-public/lib/python3.7/multiprocessing/pool.py", line 657, in get
    raise self._value
ValueError: Provenance of PreprocessorFile: /PROJECTS/C3S/datadir/esmval2output/recipe_albedolandcover_20191024_073924/preproc/albedolandcover/alb_derive_input_rsds/CMIP6_HadGEM3-GC31-LL_Amon_historical_r1i1p1f3_rsds_2000-2004.nc already initialized
2019-10-24 07:39:30,125 UTC [22897] INFO    If you suspect this is a bug or need help, please open an issue on https://github.com/ESMValGroup/ESMValTool/issues and attach the run/recipe_*.yml and run/main_log_debug.txt files from the output directory.
```

valeriupredoi:
cheers @bascrezee - most helpful, I think I know why, lemme have a hack at it :grin:

valeriupredoi:
@bascrezee I fixed the missing data cases and the derivation issues, can you pls give it another test at your convenience, man Cheers a bunch :beer:

valeriupredoi:
ok Bas I think I know what's the problem - pls remove `fx_files: [areacella]` from variable, we've changed the handling of fx variables in #170 . I ran a comprehensive test with missing partial data in the derived var and all goes through well :beer:

valeriupredoi:
I ran your `recipe_albedolandcover.yml` and went through ok, can you pls give me the full recipe you are having issues with?

valeriupredoi:
BLAST! I finally managed to reproduce the bugger (A) :grin: I was using `max_parallel_tasks: 1` all this time, phew :grin: OK so to be able to run smoothly in `--dry-run` one must use `max_parallel_tasks: 1` so there are no miscommunications between provenance pipes; I could change the behavior of parallel tasking but I think it's gonna be tricky to make sure we don't mess things up in actual production mode. Cheers for sticking with me on this one, Bas! :beer:

bascrezee:
I can confirm that it runs smoothly when running with `max_parallel_tasks: 1`.

valeriupredoi:
boss, @bascrezee :beer: @bouweandela @mattiarighi you guys wanna have a review pliz?

valeriupredoi:
can I pls pls ask either @bouweandela and/or @schlunma have a look and review this, @bascrezee has tested it and it worked fine for him -> it'd be good to get this in asap since it actually helps with the CMOR checks tiered running :beer:

bouweandela:
Actually, I think you would need the option to only CMOR check but not raise an exception, as you're planning to implement in https://github.com/ESMValGroup/ESMValCore/pull/374, to get this pull request working. Therefore I think it would make more sense to complete https://github.com/ESMValGroup/ESMValCore/pull/374 first and then have a look at this more extended case. Does that make sense?

valeriupredoi:
these things are separate things because there is a lot of flexibility in #374 and the cases should not be mixed. Note that `raise_exception` is used for both this PR and #374 so in theory we can wait for that one first BUT - the way I implemented it in 374 does allow for a nice merge and this PR should not be held in a pattern flightpath since this something we need sooner than later :beer:

valeriupredoi:
this PR has already waited too long in the pipes, man, and it's something that is very useful, so I'd say crack on with this and then 374 plss

esmvalcore/preprocessor/_io.py
```diff
@@ -103,7 +103,8 @@ def concatenate(cubes):
     raise ValueError('Can not concatenate cubes.')
 
 
-def save(cubes, filename, optimize_access='', compress=False, **kwargs):
+def save(cubes, filename, optimize_access='',
+         compress=False, dryrun=False, **kwargs):
```
JaroCamphuijsen: @valeriupredoi The `dryrun` option is not used in this function nor mentioned in the docstring.

valeriupredoi: indeed, good catch - Ill have to add documentation for it! I need the arg there because it's passed by `_recipe.py` and used in the `PreprocessorFile` class from `__init__`

valeriupredoi:
still waiting on #374 - giddyup guys! :beer:

stefsmeets:
Hi @valeriupredoi , I just had a look at this PR, and I'm wondering if it is still relevant with #917 merged. I merged master, and although I'm not super confident I got everything right (there were a lot of new code generating merge conflicts), I did get it to work... although I'm not sure what kind of output to expect 😅.

valeriupredoi:
cheers muchly @stefsmeets :beer: You shouldn't get any output since the purpose of this functionality is not to do any analysis but rather to alert the user to missing data, issues with data (CMOR hiccups) etc - I am not sure of the future of this PR though, @bouweandela was not keen to get it in last time I spoke to him but @axel-lauer was - I am on the fence myself since now we can throttle the CMOR checks and we get better error messages via #917 - what do peeps think?

stefsmeets:
That's why I was asking if it is still relevant. Personally, I like the idea because it would take away some of my own frustrations with missing data. Having looked at the code however, the changes are quite convoluted, and it does not make the existing code easier to understand. I'm questioning whether this is the right implementation.

jvegasbsc:
I think #917 is good enough for when data is missing. I also think we still need a way to check the data without running the full recipe. 

About the implementation, it will be possible to just  add a `dry_run` parameter to the `task._run` function so we just modify the
`PreprocessingTask._run` implementation to just run the critical preprocessor tasks (load and checking basically, maybe area / level selection) if that flag is set. For the `DiagnosticTask` it will be just `pass`.

I have some doubts about how this will interact with the multimodel, though, mostly because I have not a clear idea about how it is implemented

bouweandela:
The implementation here indeed looks a bit too complicated. All that would be needed now to just run the cmor checks is some way to disable saving the preprocessor output to file, right? We already have a nice report for missing data and a way to disable running diagnostics.

valeriupredoi:
I reckon Bouwe is right, Javi - all we need is a run through the data to find the missing ones and detect CMOR issues (depending on the chosen level) not a full run of the preprocessor - that takes time and memory and if we don't save any data what's the point? Gonna have a look at it now :beer:

valeriupredoi:
out of curiosity, do we have a separate report just for missing data? I don't think so, and the stdout is buried in all that screen output (or log) that, even if not in debug mode, is still a lot to sift through

valeriupredoi:
sadly this has been sidelined for too long and now it's obsolete like a phone booth. Closing it :cry:

PR #319, opened 2019-10-16
fixed the shape extraction to account for wraparound shapefile coords

valeriupredoi:

**Tasks**

-   [x] Add unit tests

* * *


valeriupredoi:
so I removed the optionality of masking irregular grids with shapefiles since I know it's dodgy and @koldunovn supported this when we implemented the shapefile masking in `_mask.py`; you can try put it back in but I don't vouch for its robustness; but wraparound of coords is mandatory :beer:

valeriupredoi:
I will actually add a check and see if the shapefiles are indeed (-180, 180) degs (I currently have no idea how to find that info, so if any of yous do know pls holler back here :beer: )

valeriupredoi:
I found a way to exmine the shapefile:

```
reader = shpreader.Reader(shapefilename)  # read the file
geometries = [contour for contour in reader.geometries()]  # extract geometries
geometries = sorted(geometries, key=lambda x: x.area, reverse=True)  # sort geometries by area
for geometry in geometries:
    print(gemoetry.bounds)
```

will return a tuple of (minLAT, minLON, maxLAT, maxLON) eg (-179.99999999999997, -85.19218750000005, 180.0, 89.99999999999994)

Bam! We use this :grin:

valeriupredoi:
OK chaps I have introduced the check on coords from shapefiles with the option to wraparaound if -180, 180 and re-added the test I previously deleted, my apologies - pls examine :beer:

valeriupredoi:
ping @bouweandela :ping_pong:

esmvalcore/preprocessor/_area.py
```diff
@@ -315,6 +315,25 @@ def _select_representative_point(shape, lon, lat):
...
+
+    if not cmor_coords:
+        # Wrap around longitude coordinate to match data
+        lon_180 = np.where(lon >= 180., lon - 360., lon)
+        # the NE mask has no points at x = -180 and y = +/-90
+        # so we will fool it and apply the mask at (-179, -89, 89) instead
+        lon = np.where(lon_180 == -180., lon_180 + 1., lon_180)
+        lat_0 = np.where(lat == -90., lat + 1., lat)
+        lat = np.where(lat_0 == 90., lat_0 - 1., lat_0)
+
```
bouweandela: This is not needed for general shapefiles, maybe it only applies to natural earth masks? I think it should be removed.

valeriupredoi:
It is needed for all the NE masks I have seen so far and can't remove it
since that will introduce a vertical dark line of NaN's

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

esmvalcore/preprocessor/_area.py
```diff
@@ -315,6 +315,25 @@ def _select_representative_point(shape, lon, lat):
...
+def _correct_coords_from_shapefile(cube, cmor_coords):
+    """Get correct lat and lon from shapefile."""
+    lon = cube.coord(axis='X').points
+    lat = cube.coord(axis='Y').points
+    if cube.coord(axis='X').ndim < 2:
+        lon, lat = np.meshgrid(lon, lat, copy=False)
+
+    if not cmor_coords:
+        # Wrap around longitude coordinate to match data
+        lon_180 = np.where(lon >= 180., lon - 360., lon)
```
bouweandela:
```suggestion
        lon[lon >= 180.] -= 360.
```
would be a bit more concise

valeriupredoi:
Cheers @bouweandela will do :beer:

valeriupredoi:
OK @bouweandela added a couple u-tests :beer:

valeriupredoi:
okido this is ready to go :beer:

bouweandela:
Could you also have a look at the other comments from the previous review?

esmvalcore/preprocessor/_area.py
valeriupredoi: it's needed for NE shapefiles (yes, specifically NE as far as Ive seen) otherwise you end up with a white delta function or stripe in the middle of the map

valeriupredoi:
ready when you ready Le Bouwe @bouweandela :beer:

valeriupredoi:
we need a Flying Dutchman to approve this - I think we have one -> @bouweandela :netherlands:

esmvalcore/preprocessor/_area.py
```diff
@@ -349,6 +349,26 @@ def _select_representative_point(shape, lon, lat):
...
+    lon = cube.coord(axis='X').points
+    lat = cube.coord(axis='Y').points
+    if cube.coord(axis='X').ndim < 2:
+        lon, lat = np.meshgrid(lon, lat, copy=False)
+
+    if not cmor_coords:
+        # Wrap around longitude coordinate to match data
+        lon = lon.copy()  # ValueError: assignment destination is read-only
+        lon[lon >= 180.] -= 360.
+        # the NE mask has no points at x = -180 and y = +/-90
```
bouweandela: The function `extract_shape` should work with any shapefile, so this very specific natural earth shapefile stuff should not be here, see also my earlier comment: https://github.com/ESMValGroup/ESMValCore/pull/319#discussion_r360447490

valeriupredoi: correct! I added a check see if we have NE file or not :grin:

valeriupredoi: it **has** to be there, unless we edit the NE file in advance and add data via some extrapolation procedure, but we dont do that; otherwise we end up with a white 1-deg wide tear in the middle of the blue Atlantic

esmvalcore/preprocessor/_area.py
```diff
@@ -349,6 +349,27 @@ def _select_representative_point(shape, lon, lat):
...
+    lon = cube.coord(axis='X').points
+    lat = cube.coord(axis='Y').points
+    if cube.coord(axis='X').ndim < 2:
+        lon, lat = np.meshgrid(lon, lat, copy=False)
+
+    if not cmor_coords:
+        # Wrap around longitude coordinate to match data
+        lon = lon.copy()  # ValueError: assignment destination is read-only
+        lon[lon >= 180.] -= 360.
+        if natural_earth_file:
```
JaroCamphuijsen: I don't see why you need to apply this to ne files. The ne files I got from https://www.naturalearthdata.com/downloads/ just run between the normal extends [-180, 180, -90, 90] so there is no need to wrap around the coordinates. And if some of them do wrap around, then the normal function would just work. 
I don't think there is anything different in natural earth shape files from other shape files, what I can see from them. I would propose to take this piece out.

esmvalcore/preprocessor/_area.py
JaroCamphuijsen: Can you share the recipe or diagnostic or add a test that produces this 1-deg wide tear? I would like to reproduce your problem.

esmvalcore/preprocessor/_area.py
```diff
@@ -464,18 +489,21 @@ def extract_shape(cube,
...
 
-        lon = cube.coord(axis='X').points
-        lat = cube.coord(axis='Y').points
-        if cube.coord(axis='X').ndim == 1 and cube.coord(axis='Y').ndim == 1:
-            lon, lat = np.meshgrid(lon.flat, lat.flat, copy=False)
+        cmor_coords = True
+        natural_earth_file = False
+        if geometries.bounds[0] < 0:
+            cmor_coords = False
+        if "ne_" in str(shapefile):
```
JaroCamphuijsen: I think this is not the best way to test for the fact whether the file does not work with the normal wrap around function even if it only goes for natural earth files (a shape file called coastline_10m.shp would also be picked up). Better would be to check for missing edge coordinates if that is the problem.

JaroCamphuijsen: Also could you add a test for the case that you want to solve here?

bouweandela:
Thanks @JaroCamphuijsen for helping out with reviewing!

esmvalcore/preprocessor/_area.py
valeriupredoi: no, this is actually needed since w/o it we end up with misalignment: our current ne masks need that, also @LisaBock had issues with selecting and masking continents that crossed the prime meridian via ne mask files, also see https://github.com/ESMValGroup/ESMValTool/issues/590

valeriupredoi: I don't get why you and @bouweandela are both so reluctant to this - it's checking if it needs wraparound and if not it's not doing it, even your docstring says `Note that this function does not work for shapes crossing the
    prime meridian or poles` so that will have to be removed once this gets in since it will work for regions crossing the prime meridian :grin:

esmvalcore/preprocessor/_area.py
valeriupredoi: good point! but unfortunately I am not sure how else to check if it's a Natural Earth file - do you have any clues, maybe an attribute in the metadata or smth? I am not that familiar with NE files :beer:

esmvalcore/preprocessor/_area.py
valeriupredoi: run with any of our current shapefiles in `ne_masks` and take the bit of code that eliminates the tear. Come on guys, this is old stuff that's been reviwed and tested aeons ago

esmvalcore/preprocessor/_area.py
JaroCamphuijsen: I am not reluctant about the wraparound, that's AWESOME! However the exception for ne files stings a bit. I just don't see how they are different from other shapefiles. If there is an issue, it is probably to also catch it in other possible cases, but for that we need to understand what the issue is.

valeriupredoi: yeah absolutely! I am not familiar with shapefiles and the only ones Ive seen so far are these NE masks, which all needed this correction. The `ne_` string test is clearly dodgy but at the moment I don't know any better :grin:

JaroCamphuijsen: then it's maybe better if we check for the availability of coordinates at these points in the shapefile, Instead of using the dodgy `ne_` check.

bouweandela:
Hi V,

Very good suggestion to try out with the natural earth shapefiles included in the ESMValCore. I tried running with `esmvalcore/preprocessor/ne_masks/ne_10m_ocean.shp` and immediately encountered two problems:
- the argument `crop: true` does not work
- the row of pixels that represents the pole is not masked, I'm not sure what the cause is (is it because the points are inside the shape or is it something with the function?) It would be good to understand why this happens.

valeriupredoi:
cheers @bouweandela - it seems we need to run some sort of pre-checks on these darn files; the ocean 10m one is overkill anyway coz we use the 50m one. This is a hairy problem, I'll have to think of a way to approach it. Let me see if I can do smth abt it tomorrow, am dealing with other stuffs ESMValCore-related today :beer:

valeriupredoi:
@bouweandela I found out why your cropping is failing: https://github.com/ESMValGroup/ESMValCore/issues/625 - an rightly so, that will fail for any shapefile that has the whole earth as region

valeriupredoi:
OK so I dug a bit in the NE shapefiles and I found out that they actually have metadata, access it via `geometries.meta` attribute, the `geometries.meta['crs_wkt']` tells us what coordinate system is using, and it proves out to be WGS 84 and Prime Meridian is Greenwich with specs [here](https://epsg.io/4326) so that is ineed -180 to 180 and -90 to 90 and that there actually is a standard for it!

Also, the darn thing contains layer upon layer of polygon contours and to get the actual lat-lon points you need to go through the m all like:
```python
    for geometry in geometries:
        for coord in geometry["geometry"]["coordinates"]:
            for coord_i in coord:
                for i in coord_i:
                    all_lats.append(i[0])
                    all_lons.append(i[1])
    print(min(all_lats), min(all_lons))
    print(max(all_lats), max(all_lons))
```
which mins/maxs are
```bash
-179.99999999999991 -85.22193775799991
180.0 90.00000000000006
```
which correspond to the bounds, which are accessible via `*geometries.bounds` or via the metadata. Now, the change in the code says:
```python
    if not cmor_coords:
        # Wrap around longitude coordinate to match data
        lon = lon.copy()  # ValueError: assignment destination is read-only
        lon[lon >= 180.] -= 360.
        if natural_earth_file:
            # the NE mask has no points at x = -180 and y = +/-90
            # so we will fool it and apply the mask at (-179, -89, 89) instead
            lon = np.where(lon == -180., lon + 1., lon)
            lat_0 = np.where(lat == -90., lat + 1., lat)
            lat = np.where(lat_0 == 90., lat_0 - 1., lat_0)
```
which is perfectly determined by the bounds - indeed it's -180 to 180 and is missing points at -180 and -90 so the fix in code is perfectly needs. Only thing that's dodgy is the if-conditional argument, but using the bounds instead will generalize it to all files! Wow, now I can write a Masters thesis in shapefiles :grin:

valeriupredoi:
@bouweandela at your leisure here, mate :beer:

bouweandela:
@JaroCamphuijsen promised to help me out with reviewing this one again :tada:

valeriupredoi:
poor guy just got volunteered? :grin: Fire at will @JaroCamphuijsen

JaroCamphuijsen:
Hi V, 
Sorry for the long wait, due to personal circumstances (:baby_bottle:) I'm pretty unreliable at the moment.
However, I started looking into these whole earth shape files and I agree that a solution like the one you propose is necessary due to the non circular nature of these shape files. Still I'd like to investigate whether a less drastic solution (maybe shifting a tiny bit instead of a whole degree?) also works. I think that should work since the bounds of these shape files are only off a tiny bit from -180 degrees, I think shifting 0.0001 degree also works, otherwise you skip quite a broad band with perhaps some islands etc. 

I will also try to replicate the same behavior and try some things. Let me know when you finish your thesis, I'd love to join your defense. I'm sure it will be exhilarating...

esmvalcore/preprocessor/_area.py
```diff
@@ -475,18 +511,30 @@ def extract_shape(cube,
...
 
+        # get parameters specific to the shapefile (NE used case
+        # eg longitudes [-180, 180] or latitude missing
+        # or overflowing edges)
+        cmor_coords = True
+        pad_north_pole = False
+        pad_hawaii = False
+        if geometries.bounds[0] < 0:
+            cmor_coords = False
+        if geometries.bounds[1] > -90. and geometries.bounds[1] < -85.:
```
JaroCamphuijsen, line 525: @valeriupredoi 
I can reproduce the problem you mentioned with bands of NaN's at the edges of the NE shapefiles when I set pad_north_pole and pad_hawaii to be False and cmor_coords also to be True for the run (so your defaults in line 520-522). However the problem is solved already for me if I enable the cmor_coords check from line 523-524. Seems to me that the paddings are not needed.

valeriupredoi:
@JaroCamphuijsen 
> I will also try to replicate the same behavior and try some things. Let me know when you finish your thesis, I'd love to join your defense. I'm sure it will be exhilarating..

Awesome, but you'd have to travel back in time about eight years to 2012 mate :grin: 

> Sorry for the long wait, due to personal circumstances (baby_bottle) I'm pretty unreliable at the moment.

Absolutely no problem and great many congratulations assuming that means a new baby and not that they're feeding you with a :baby_bottle: :laughing: 

Yeh, cheers, let's keep talking about these fixes, they're pretty nasty file-specific :beer:

esmvalcore/preprocessor/_area.py
valeriupredoi: so that means `cmor_coords`, `pad_hawaii` and `pad_north_pole` are all `False`?

JaroCamphuijsen: Affirmative!

JaroCamphuijsen: > so that means cmor_coords, pad_hawaii and pad_north_pole are all False?

@valeriupredoi Did you check this? What do you think?

ledm:
I'm deploying this PR to make IPCC plots. I hope that it can be merged in soon! Cheers!

esmvalcore/preprocessor/_area.py
valeriupredoi: looking at it now :beer:

valeriupredoi: aha, took the fixes out the conditional if the coords are not cmor-like, it should be fixed now - the test is there to confirm :beer:

JaroCamphuijsen: Hi V,
Are you sure we cannot delete these padding fixes all together? My point was that the ne_files did look perfectly fine to me without the pad_hawaii and pad_north_pole fixes altogether. I do see NaNs when the cmor_coords fix is disabled, but I do not see any NaNs anymore when the cmor_coords fix is activated. This seems to me like these padding fixes are unnecessary. Can you confirm this, or otherwise give me the exact shapefile on which it doesn't work for you without the paddings? I would really like to get this merged today, to get it in with the release!

valeriupredoi: @JaroCamphuijsen mate yes they are needed - please see the test for nan's, both the land and ocean NE files we have in the `ne_masks` dir need padding otherwise you end up with NaN strips; these files are ona -180-180 grid, so `cmor_coords` is `False` in this case, whereas this is needed for cmor-like coords too (0-360) I wouldn't know but better not risk it. Note that the check on cmor-like coords just rotates the lon to be on a 0-360 axis, no operation on the missing data strips. Yes, let's get this in today! :beer:

JaroCamphuijsen: @valeriupredoi 
I'm running the tests, but these also pass when I turn off the padding fixes (i.e. comment out line 538-541 in _area.py), so for the unit tests, this is not needed.

valeriupredoi: yes, you are right - because they are masked and not NaN's - I got carried away by your previous comment (about NaN's) and plugged in the test, they are in fact masked strips

valeriupredoi: ok fixed the test and simplified it now - you will see that if you set those pad params to False in the code that bit will be fully masked

JaroCamphuijsen: You are right, it was my sleep deprivation after all. I found the problem and see how your implementation fixes this, however I still think we should talk about this, since correcting for the quirks of the ne_shapefiles in the preprocessors of esmvalcore, doesn't seem right to me.

valeriupredoi: yup! totally agree :beer:

JaroCamphuijsen:
This PR can be merged, however I still think we should talk about this, since correcting for the quirks of the ne_shapefiles in the preprocessors of esmvalcore, doesn't seem right to me.

valeriupredoi:
@bouweandela - @JaroCamphuijsen and meself have proper dissected this and agreed it's time to merge but we need to find a robust method how to fix shapefiles in the future w/o manipulating data. Anyway, time to merge this! Can you OK please mate and squish&merge :beer:

valeriupredoi:
nay bother - did it meself!

PR #323, opened 2019-10-17
Improve extract levels

### LINKED ISSUE #317: extract_levels preprocessor only support 1D coordinates
Currently, `extract_levels` assume that levels are identical at every point and timestep. This is not true always, i.e. when using model levels. The preprocessor function must be enhanced to allow processing this kind of variables

jvegasbsc:
This pull request allow us to extract levels even if they are not regular in all grid points and / or times

This includes two main changes:

- Add an option to specify the coordinate to extract, and do not always use the z-axis dimension coord
- Adapat the code itself to be able to manage levels that are not a 1D array

Tested for atmospheric variables in model levels and air_pressure coordinate (4D)

Closes #317


axel-lauer:
Tested the code for variable *cl* (historical, Amon) and it worked for the following CMIP6 models:
* BCC-CSM2-MR
* CAMS-CSM1-0
* GISS-E2-1-G
* GISS-E2-1-H
* MIROC6
* MPI-ESM1-2-HR
* MRI-ESM2-0

The following models produce various errors:
* BCC-ESM1
* CESM2
* CESM2-WACCM
* CNRM-CM6-1
* CNRM-ESM2-1
* GFDL-CM4
* HadGEM3-GC31-LL
* IPSL-CM6A-LR
* UKESM1-0-LL

The most common error message is for these models is:

`iris.exceptions.CoordinateNotFoundError: 'Expected to find exactly 1 air_pressure coordinate, but found none.'`

[main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/3739694/main_log_debug.txt)
[recipe_esacci-wv3d.yml.txt](https://github.com/ESMValGroup/ESMValCore/files/3739703/recipe_esacci-wv3d.yml.txt)

jvegasbsc:
> UKESM1-0-LL: CoordinateNotFoundError

This model's formula is not providing height in pressure levels but real height. I will test the `extract_levels`with it, as it is a case of a 3D height coordinate (time independent in this case), but I don't know if it will be useful for your analysis

axel-lauer:
I tested the changes and for all models except for BCC-ESM1 (which is working now), I still get the same error:

`iris.exceptions.CoordinateNotFoundError: 'Expected to find exactly 1 air_pressure coordinate, but found none.'`

It seems that those model already provide data on pressure levels, so in those cases only interpolation to the selected pressure levels is needed but no calculation of the pressure?!

jvegasbsc:
The thing is that some models provide the formula to go from model levels to pressure levels. 
Others only have the formula to go from model levels to height in m. I can not compute the pressure levels interpolation on that models because the `extract levels` does not have the required info

By the way, that information is in variable `pfull` (we have its monthly climatology in `Amon`, its instant values in `6hrLev) but this should probably require its own preprocessor function. Also requires combining two variables but not to create a new one (as we do in derive) but to add coordinates. I am not sure how this could be implemented in our actual preprocessor. Maybe @bouweandela or @valeriupredoi can think of some way to do it

jvegasbsc:
I checked all the models, except `GFDL-CM4` which is not on Jasmin

Formula for `height`, not `air_pressure` levels
-`HadGEM3-GC31-LL`
-`UKESM1-0-LL`

Formula missing:
-`CESM2`: Does not have formula terms in lev. Only in levbnds
-`CESM2-WACCM`: not info, but levels are provided in hPa

Iris can not understand formula:
- `CNRM-CM6-1`
- `CNRM-ESM2-1`

And then we have `IPSL-CM6A-LR`: it has levels as pressure levels but also provide the terms for the `air_pressure` 

I think all of them can be fixes using `fix_metadata`

.gitignore
```diff
@@ -82,6 +82,9 @@ test-reports/
 /test_bash.sh
 /python_test_out.txt
 
+# Auto generated doc
+_sidebar.rst.inc
+
```
valeriupredoi: thank you! I have been meaning to add this bloody sidebar to ignore for a while now :grin:

doc/esmvalcore/preprocessor.rst
```diff
@@ -218,6 +218,20 @@ extract the levels and vertically regrid onto the vertical levels of
           # levels: {dataset: ERA-Interim, coordinate: air_pressure}
           scheme: linear_horizontal_extrapolate_vertical
 
+By default, vertical interpolation is performed in the dimension coordinate of
+the z axis. If you want to use another coordinate (for example, ``air_pressure``'
+in variables that are provided in model levels and not pressure levels) you can
```
valeriupredoi: I think `explicitly declaring the z-axis coordinate name` is the key phrase here

esmvalcore/preprocessor/_regrid.py
```diff
@@ -4,10 +4,11 @@
 import re
 from copy import deepcopy
 
-import iris
 import numpy as np
 import stratify
+import iris
 from iris.analysis import AreaWeighted, Linear, Nearest, UnstructuredNearest
+from iris.util import broadcast_to_shape
```
valeriupredoi: you ran `isort` on this?

jvegasbsc: No, never used it

doc/esmvalcore/preprocessor.rst
```diff
@@ -53,7 +53,7 @@ Overview
 
 The ESMValTool preprocessor can be used to perform a broad range of operations
 on the input data before diagnostics or metrics are applied. The preprocessor
-performs these operations in a centralized, documented and efficient way, thus
+performs these operations in a centralied, documented and efficient way, thus
```
mattiarighi:
```suggestion
performs these operations in a centralized, documented and efficient way, thus
```

mattiarighi:
Please fix the tests and the conflicts.

PR #230, opened 2019-09-05
Updated derivation scripts using new fx file definitions

valeriupredoi:
Closes #229, closes #318.

Waiting for #170
UPDATE: #170 has been merged and is fully integrated for this here

PR #331, opened 2019-10-23
Added land/sea fraction weighting preprocessor

### LINKED ISSUE #298: Add land/ocean-fraction weighting preprocessor
Hi - i'm new to this forum, so hopefully this is the right way to request a new functionality.

It would be great to have a pre-processor which applied a weighting by land-fraction. This is different from simply masking data in/out according to a threshold value. What is required is to multiply the given input field by a fraction from 0-1 to account for the fact that not all grid points are completely land/ocean - some are fractional.

The application is important for most carbon cycle variables, and other land-surface outputs, which are reported in units of "carbon per square-metre". But this actually refers to "per square metre of land" and NOT "per square metre of gridbox". So in order to integrate these globally or regionally one has to both area-weight the quantity but also weight by the land fraction.

It would be great to get a pre-processor which could do this to produce a dataset of carbon stores or fluxes which are "per square metre of gridbox".

so, for example, for the carbon flux called "gpp" I would want to:
gpp(t) = gpp(t) * land-frac
where land_frac is the time-static fractional cover of land in each gridbox

A converse weighting for ocean fluxes is also useful of course. (i.e. scaled by 1-land_frac)

I hope this description makes sense. I'd be happy to try to help test or better define.

thanks!
Chris

schlunma:
This PR adds a land/sea fraction weighting preprocessor as suggested by @ChrisJones-MOHC in #298 (I borrowed some text of your issue for the documentation, I hope that's okay :grin:). With this change, many derived variables are not needed anymore, so there are a few changes necessary in ESMValTool after that.

Requires #230.

-   [x] Add unit tests

* * *

Closes #298. Necessary for ESMValGroup/ESMValTool#473.


doc/esmvalcore/preprocessor.rst
```diff
@@ -242,6 +245,50 @@ extract the levels and vertically regrid onto the vertical levels of
...
+Weighting
+=========
+
+.. _land/sea fraction weighting:
+
+Land/sea fraction weighting
+---------------------------
+
+This preprocessor allows weighting of data by land or sea fractions. In other
+words, this function multiplies the given input field by a fraction from 0-1 to
```
mattiarighi:
```suggestion
words, this function multiplies the given input field by a fraction in the range 0-1 to
```

doc/esmvalcore/preprocessor.rst
```diff
@@ -242,6 +245,50 @@ extract the levels and vertically regrid onto the vertical levels of
...
+account for the fact that not all grid points are completely land/sea.
```
mattiarighi:
```suggestion
account for the fact that not all grid points are completely land- or sea-covered.
```

doc/esmvalcore/preprocessor.rst
```diff
@@ -242,6 +245,50 @@ extract the levels and vertically regrid onto the vertical levels of
...
+
+The application of this is very important for most carbon cycle variables (and
```
mattiarighi:
```suggestion
The application of this preprocessor is very important for most carbon cycle variables (and
```

doc/esmvalcore/preprocessor.rst
```diff
@@ -242,6 +245,50 @@ extract the levels and vertically regrid onto the vertical levels of
...
+other land surface outputs), which are e.g. reported in units of
+:math:`kgC~m^{-2}`. This actually refers to 'per square meter of land/sea' and
```
mattiarighi:
```suggestion
:math:`kgC~m^{-2}`. Here, the surface unit actually refers to 'square meter of land/sea' and
```

doc/esmvalcore/preprocessor.rst
```diff
@@ -242,6 +245,50 @@ extract the levels and vertically regrid onto the vertical levels of
...
+NOT 'per square meter of gridbox'. In order to integrate these globally or
```
mattiarighi:
```suggestion
NOT 'square meter of gridbox'. In order to integrate these globally or
```

doc/esmvalcore/preprocessor.rst
```diff
@@ -242,6 +245,50 @@ extract the levels and vertically regrid onto the vertical levels of
...
+regionally one has to both area-weight the quantity but also weight by the
```
mattiarighi:
```suggestion
regionally one has to weight by both the surface quantity and the
```

doc/esmvalcore/preprocessor.rst
```diff
@@ -242,6 +245,50 @@ extract the levels and vertically regrid onto the vertical levels of
...
+land/sea fraction.
+
+For example, to weight an input field with the land fraction, use the following
```
mattiarighi:
```suggestion
For example, to weight an input field with the land fraction, the following
```

doc/esmvalcore/preprocessor.rst
```diff
@@ -242,6 +245,50 @@ extract the levels and vertically regrid onto the vertical levels of
...
+preprocessor:
```
mattiarighi:
```suggestion
preprocessor can be used:
```

doc/esmvalcore/preprocessor.rst
```diff
@@ -242,6 +245,50 @@ extract the levels and vertically regrid onto the vertical levels of
...
+
+.. code-block:: yaml
+
+    preprocessors:
+      preproc_weighting:
+        weighting_landsea_fraction:
+          area_type: land
+          strict: true
+
+Allowed arguments for the keyword ``area_type`` are ``land`` (fraction is 1.0
```
mattiarighi:
```suggestion
Allowed arguments for the keyword ``area_type`` are ``land`` (fraction is 1
```

doc/esmvalcore/preprocessor.rst
```diff
@@ -242,6 +245,50 @@ extract the levels and vertically regrid onto the vertical levels of
...
+for grid cells with only land surface, 0.0 for grid cells with only sea surface
```
mattiarighi:
```suggestion
for grid cells with only land surface, 0 for grid cells with only sea surface
```

doc/esmvalcore/preprocessor.rst
```diff
@@ -242,6 +245,50 @@ extract the levels and vertically regrid onto the vertical levels of
...
+and values in between 0.0 and 1.0 for coastal regions) and ``sea`` (1.0 for
```
mattiarighi:
```suggestion
and values in between 0 and 1 for coastal regions) and ``sea`` (1 for
```

doc/esmvalcore/preprocessor.rst
```diff
@@ -242,6 +245,50 @@ extract the levels and vertically regrid onto the vertical levels of
...
+sea, 0.0 for land, in between for coastal regions). The optional argument
```
mattiarighi:
```suggestion
sea, 0 for land, in between for coastal regions). The optional argument
```

mattiarighi:
You can remove the draft mode if ready.

esmvalcore/preprocessor/_weighting.py
```diff
@@ -0,0 +1,103 @@
...
+    order to integrate these globally or regionally one has to both area-weight
+    the quantity but also weight by the land/sea fraction.
+
+    Parameters
+    ----------
+    cube : iris.cube.Cube
+        Data cube to be weighted.
+    fx_files : dict
+        Dictionary holding ``var_name`` (keys) and full paths (values) to the
+        fx files.
```
valeriupredoi: ...as lists of none or single elements

esmvalcore/preprocessor/_weighting.py
```diff
@@ -0,0 +1,103 @@
...
+    (land_fraction, errors) = _get_land_fraction(cube, fx_files)
+    if land_fraction is None:
+        if strict:
+            raise ValueError(
+                f"Weighting with land/sea fraction is not possible, the "
+                f"following errors occured: {' '.join(errors)}")
+        logger.debug(
+            "Weighting of '%s' with '%s' fraction not possible because of "
+            "the following problems: %s", cube.var_name, area_type,
+            " ".join(errors))
```
valeriupredoi: just raise the exception with full details; replace debug-logging with warning visible with any logging level

esmvalcore/preprocessor/_weighting.py
```diff
@@ -0,0 +1,103 @@
...
+        return cube
+    core_data = cube.core_data()
+    if area_type == 'land':
```
valeriupredoi: I think you need to unindent this block by a tab or add `else`

tests/unit/preprocessor/_weighting/test_weighting_landsea_fraction.py
```diff
@@ -0,0 +1,231 @@
...
+    fx_cubes = []
+    if fx_files.get('sftlf'):
+        fx_cubes.append(CUBE_SFTLF)
+    if fx_files.get('sftof'):
+        fx_cubes.append(CUBE_SFTOF)
+    mock_iris.load_cube.side_effect = fx_cubes
+    weighted_cube = weighting.weighting_landsea_fraction(
+        cube, fx_files, area_type, strict=strict)
+    assert weighted_cube == cube
+    assert weighted_cube is cube
```
valeriupredoi: ooh this is dangerous, this will most probably fail since if you do `c1 = iris.load_cube(file)` and `c2 = iris.load_cube(file)` then `assert c1 == c2` will pass whereas `assert c1 is c2` will fail

esmvalcore/preprocessor/_weighting.py
schlunma: I don't think so, but I replaced the second `if` with an `elif`.

tests/unit/preprocessor/_weighting/test_weighting_landsea_fraction.py
schlunma: No, the weighting happens in-place, so I want that `cube_out is cube_in` (no new object is created). Also, the tests are not failing, right? :wink:

schlunma:
> awesome! cheers @schlunma beer A couple minor stuffs from me -> also, why did all the gpp and lai custom cmor tables get killed?

I implemented your changes. The custom CMOR tables are gone because their derivation functions are gone and they were only needed for that.

tests/unit/preprocessor/_weighting/test_weighting_landsea_fraction.py
valeriupredoi: yes, was a bit confused as to why the tests were passing, have missed the fact that the cube is modified in place :grin:

schlunma:
> Nice work! One question about the `strict` argument though: why is this needed? One of the design choices we have always adhered to so far is that if you specify that you need certain variables from certain datasets in the recipe, they should exist and this seems to go against that.

For some datasets we simply do not  have `sftlf` or `sftof` (especially for `OBS`) and this preprocessor is not necessary. I think it should be possible to use one preprocessor for all datasets and not write a different one. Not applying this preprocessor is also less critical than others, because it does not reshape the data and only affects a really small part of it.

Also, the default behavior is `strict: true`, so the user has to explicitly agree to use that feature.

bouweandela:
To expand a bit on the above: the reason that we require all data to be present, is that we want the behaviour of the tool to be both reliable and easy to understand. This means that errors are not silently ignored. Having an option to maybe run a preprocessor function or maybe not, really goes against that. Is there a reason to want to use a single preprocessor profile for all datasets? E.g. because you want to apply multi model mean or masking functions?

schlunma:
In my opinion explicitly having to set an option to `False` is not silently ignoring it (note that there is also a warning). I think it's comparable to the `--skip-nonexistent` switch, which also allows missing data if **explicitly** desired.

> Is there a reason to want to use a single preprocessor profile for all datasets? E.g. because you want to apply multi model mean or masking functions?

Because it's way more user friendly. Using two different preprocessors requires another block in the `variables` section for the same variable with its own `additional_datasets`, etc.

BTW, all the `_grid` variables (which implicitly used this preprocessor) had the same behavior (and there was no even a warning).

schlunma:
It would be really nice if we could merge this soon, since I need it for ESMValGroup/ESMValTool#473.

Maybe someone else has an opinion on this? @valeriupredoi @mattiarighi

valeriupredoi:
Good by me, man, travelling but will approve it in an hour or so when I get
back to Reading :beer:

Dr Valeriu Predoi.
Computational scientist
NCAS-CMS
University of Reading
Department of Meteorology
Reading RG6 6BB
United Kingdom

mattiarighi:
I'm fine with it, although I had no time to test it.

schlunma:
@bouweandela, do you have further comments?

bouweandela:
I'm still not convinced that this is a good approach, for the reasons mentioned above. If you really insist on using one preprocessor profile to do two different preprocessing chains, maybe an `exclude` argument taking a list with the names of datasets, [like we have for the multi model functions](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/esmvalcore/preprocessor.html#multi-model-statistics), could be an option instead of `strict`? At least that makes it immediately clear from the recipe which datasets will skip this preprocessing step.

schlunma:
All right, that sounds like a reasonable compromise to me. I will implement this as soon as possible.

schlunma:
@bouweandela I made all necessary changes, hopefully we can merge this soon now.

bouweandela:
@mattiarighi Please test before merging, I only looked at the code.

schlunma:
Will do 👍

mattiarighi:
I can only test this together with the corresponding PR in ESMValTool where the recipes are adjusted.

schlunma:
You can test [this](https://github.com/ESMValGroup/ESMValTool/blob/Anav13JClim_v2/esmvaltool/recipes/recipe_anav13jclim.yml) recipe in [this](https://github.com/ESMValGroup/ESMValTool/tree/Anav13JClim_v2) branch.

mattiarighi:
Tested successfully! :tada:

bouweandela:
@schlunma @mattiarighi Is there a corresponding pull request in the ESMValTool repository already? Or is this not needed? Already done?

schlunma:
@bouweandela Only small changes were necessary, they have already been addressed in ESMValGroup/ESMValTool#1466.

Issue #286, opened 2019-10-02
Data finder cannot find data spread over different versions


schlunma:
On DKRZ, it can happen that data for a specific dataset is spread over different versions. For example, the first 150 years of `CESM2`'s `tas` are located in a different subdirectory than the rest of the data:

```
.../CMIP/NCAR/CESM2/abrupt-4xCO2/r1i1p1f1/Amon/tas/gn
├── v20190425
│   └── tas_Amon_CESM2_abrupt-4xCO2_r1i1p1f1_gn_000101-015012.nc
└── v20190829
    ├── tas_Amon_CESM2_abrupt-4xCO2_r1i1p1f1_gn_015101-019912.nc
    ├── tas_Amon_CESM2_abrupt-4xCO2_r1i1p1f1_gn_020001-024912.nc
    ├── tas_Amon_CESM2_abrupt-4xCO2_r1i1p1f1_gn_025001-029912.nc
    ├── tas_Amon_CESM2_abrupt-4xCO2_r1i1p1f1_gn_030001-034912.nc
    ├── tas_Amon_CESM2_abrupt-4xCO2_r1i1p1f1_gn_035001-039912.nc
    ├── tas_Amon_CESM2_abrupt-4xCO2_r1i1p1f1_gn_040001-044912.nc
    ├── tas_Amon_CESM2_abrupt-4xCO2_r1i1p1f1_gn_045001-049912.nc
    ├── tas_Amon_CESM2_abrupt-4xCO2_r1i1p1f1_gn_050001-054912.nc
    ├── tas_Amon_CESM2_abrupt-4xCO2_r1i1p1f1_gn_055001-059912.nc
    ├── tas_Amon_CESM2_abrupt-4xCO2_r1i1p1f1_gn_060001-064912.nc
    ├── tas_Amon_CESM2_abrupt-4xCO2_r1i1p1f1_gn_065001-069912.nc
    ├── tas_Amon_CESM2_abrupt-4xCO2_r1i1p1f1_gn_070001-074912.nc
    ├── tas_Amon_CESM2_abrupt-4xCO2_r1i1p1f1_gn_075001-079912.nc
    ├── tas_Amon_CESM2_abrupt-4xCO2_r1i1p1f1_gn_080001-084912.nc
    ├── tas_Amon_CESM2_abrupt-4xCO2_r1i1p1f1_gn_085001-089912.nc
[...]

PR #337, opened 2019-10-28
Generalize data finder DRAFT

valeriupredoi:
This is a last ditch bid to correct as much as possible what @schlunma points out in #286 - my take on it is rather peasant but might work after all:

- use all available data in the `latesteversion` or `latest`, and `vXXXXX` directories, but:
  - filter it by the newest files if duplicates;
  - do not look into `files` directory;
  - replace `os.walk` with `glob.glob` as @jvegasbsc suggests in one of his issues;
- use the *smart concatenation* in #280 to merge all the various files;

### DIFF
--- a/esmvalcore/_data_finder.py
+++ b/esmvalcore/_data_finder.py
@@ -17,7 +17,7 @@ logger = logging.getLogger(__name__)
 
 def find_files(dirnames, filenames):
     """Find files matching filenames in dirnames."""
-    logger.debug("Looking for files matching %s in %s", filenames, dirnames)
+    logger.info("Looking for files matching %s in %s", filenames, dirnames)
 
     result = []
     for dirname in dirnames:
@@ -230,6 +230,21 @@ def _find_input_files(variable, rootpath, drs):
     return files
 
 
+def _filter_newest_files(input_files):
+    """Filter input files to get newest files if duplicates."""
+    if not len(input_files) > 1:
+        return input_files
+    else:
+        unique_newest_files = []
+        base_files = [os.path.basename(file) for file in input_files]
+        unique_base_files = list(set(base_files))
+        for unique_file in unique_base_files:
+            matches = fnmatch.filter(input_files, '*' + unique_file)
+            newest_match = max(matches, key=os.path.getctime)
+            unique_newest_files.append(newest_match)
+        return unique_newest_files
+
+
 def get_input_filelist(variable, rootpath, drs):
     """Return the full path to input files."""
     # change ensemble to fixed r0i0p0 for fx variables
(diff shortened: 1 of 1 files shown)


esmvalcore/_data_finder.py
```diff
@@ -230,6 +230,21 @@ def _find_input_files(variable, rootpath, drs):
...
 
 
+def _filter_newest_files(input_files):
+    """Filter input files to get newest files if duplicates."""
+    if not len(input_files) > 1:
+        return input_files
+    else:
+        unique_newest_files = []
+        base_files = [os.path.basename(file) for file in input_files]
+        unique_base_files = list(set(base_files))
```
bouweandela: You can simplify this to
```python
unique_base_files = {os.path.basename(file) for file in input_files}
```

esmvalcore/_data_finder.py
```diff
@@ -230,6 +230,21 @@ def _find_input_files(variable, rootpath, drs):
...
+        for unique_file in unique_base_files:
+            matches = fnmatch.filter(input_files, '*' + unique_file)
+            newest_match = max(matches, key=os.path.getctime)
```
```python
            if matches:
                newest_match = sorted(matches)[-1]
                unique_newest_files.append(newest_match)
```
I would do something like this, I don't think the creation time is reliable, but the drs usually contains the version number somewhere. Make sure the ignore the 'latest' directory which is symlinked to the actual latest version on e.g. Jasmin.

esmvalcore/_data_finder.py
```diff
@@ -17,7 +17,7 @@
 
 def find_files(dirnames, filenames):
     """Find files matching filenames in dirnames."""
-    logger.debug("Looking for files matching %s in %s", filenames, dirnames)
+    logger.info("Looking for files matching %s in %s", filenames, dirnames)
```
bouweandela: This should probably stay a debug message, because it will generate a huge amount of output

valeriupredoi:
hey @schlunma you reckon this will be of any use to fix the crazy data cases you been seeing last year? I think we should probably forget about it for now since I really don't feel like rocking the boat too much before the stable v2 release :beer:

schlunma:
> hey @schlunma you reckon this will be of any use to fix the crazy data cases you been seeing last year? I think we should probably forget about it for now since I really don't feel like rocking the boat too much before the stable v2 release beer

I think so too. Anyway, the crazy data has disappeared, at least on DKRZ, so I guess this was just a temporary issue with the synchronization :grin:

valeriupredoi:
OK so we can then put this one on the long-term to-do list then :beer:

valeriupredoi:
gonna close this for now coz I gots no time to deal with it in an exhaustive manner before v2.0 release - maybe we can revisit at a later stage :beer:

Issue #355, opened 2019-11-04
Fixes for MPI-ESM1-2-HR and MPI-ESM1-2-XR


sloosvel:
**Is your feature request related to a problem? Please describe.**
Variables tas, ta, va, zg, ua and sfcWind require metadata fixes related to coordinates plev, height2m and height10m.

**Would you be able to help out?**
I will open a pull request with the fixes.

PR #356, opened 2019-11-04
Fixes for MPI-ESM1-2-HR and MPI-ESM1-2-XR

sloosvel:

**Tasks**

-   [ ] Add unit tests

* * *

Closes {#355}


valeriupredoi:
thanks a bunch @sloosvel :beer: Note that `MPI-ESM1-2-XR` is not currently present in `/cmor/tables/cmip6/Tables/CMIP6_CV.json` so that either needs to be added there (via an update of the CMOR tables, not by hand by us) or we temporarily add it to `config-developer.yml`

sloosvel:
I added the institute and activity manually in the recipes where I use the model, but sure I can add it to the config-developer if you prefer.

valeriupredoi:
well we need something that's foolproof, ie not something the user would add or forget to add to their recipes, ideally the CV should be updated centrally, but in the absence of that config-developer would do :beer:

valeriupredoi:
ooh can you also add `'CESM2-WACCM': ['NCAR']` since you're so nice to add the XR plese :cat2:

sloosvel:
Sure, even though in https://github.com/ESMValGroup/ESMValCore/pull/107#issuecomment-528847487 @jvegasbsc was against doing that. But whatever you prefer.

esmvalcore/config-developer.yml
```diff
@@ -27,6 +27,9 @@ CMIP6:
     ETHZ: '{exp}/{mip}/{short_name}/{dataset}/{ensemble}/{grid}/'
   input_file: '{short_name}_{mip}_{dataset}_{exp}_{ensemble}_{grid}*.nc'
   output_file: '{project}_{dataset}_{mip}_{exp}_{ensemble}_{short_name}_{start_year}-{end_year}'
+  institutes:
+    'CESM2-WACCM': ['NCAR']
```
bouweandela: This should not be needed if we just update the cmor tables to the latest version. I made a pull request here: https://github.com/ESMValGroup/ESMValCore/pull/363

sloosvel: So this list will have to be updated every time the tables get updated?

bouweandela: I think the other institute can also be removed.

esmvalcore/cmor/_fixes/cmip6/mpi_esm1_2_hr.py
```diff
@@ -0,0 +1,152 @@
...
+
+        """
+        for cube in cubes:
+            plev = cube.coord('air_pressure')
+            plev.var_name = 'plev'
+
+        return cubes
+
+
+class Va(Fix):
```
bouweandela: Maybe you could just replace this by 
```
class Va(Ta):
    pass
```

Also, the very lengthy numpy style docstrings are not really needed, since this is not part of the public api and it really reduces readability in this case.

esmvalcore/config-developer.yml
valeriupredoi: careful what you merge first - if you merge this PR before #363 then it won't work proper :beer:

esmvalcore/cmor/_fixes/cmip6/mpi_esm1_2_hr.py
sloosvel: Since they both need the same ones and are the same model but in different resolutions, for the XR, can the fixes from the HR be imported? I mean like it's done in the cesm2 and cesm2-wacc fixes. Or is it better not to mix imports between models?

bouweandela: >  can the fixes from the HR be imported?

Yes, I think that should be fine.

esmvalcore/cmor/_fixes/cmip6/mpi_esm1_2_hr.py
```diff
@@ -0,0 +1,83 @@
+"""Fixes for MPI-ESM1-2-HR model."""
+
+from ..fix import Fix
+from ..shared import add_scalar_height_coord
```
bouweandela: I think it would be a good idea to change these functions for adding auxiliary coordinates so they check if the coordinate is already present and do nothing if this is the case. That way we will avoid the situation that ESMValTool will crash as soon as corrected data is released.

schlunma: #372

bouweandela: Yes, that's a good example, but it would be better to do it in the function, to avoid needing this code that checks if the coordinate exists in every fix.

schlunma: Yeah, I think so too. I only wanted to mention this PR because the code can be used without changes in the shared functions.

PR #358, opened 2019-11-04
Support for ERA5 data in native format

### LINKED ISSUE #311: Make datafinder understand dates in the middle of the file name
Add support for custom position of year information in data filename.

E.g.: 
```
historical_____198401_Amon.nc
```
This will be the function `get_start_end_year`

Note: test case need to added for this.

### LINKED ISSUE #447: Read ERA5 data in native format
The ERA5 dataset is too large to CMORize and then store, e.g. one year of a single hourly variable takes up 34 GB after CMORization. Instead ESMValCore should be able to find ERA5 data already available (either on a cluster or downloaded through era5cli) and CMORize while reading it using fixes.

This can also be used as an example of how to add support for analyzing model data in native format, e.g. EMAC data.

bouweandela:
This pull request makes it possible to use ERA5 data directly in ESMValCore. CMORization is implemented as a fix.

Related to issue https://github.com/ESMValGroup/ESMValTool/issues/1120.

Related ESMValTool PR ESMValGroup/ESMValTool#1432

**Tasks**

-   [x] Add unit tests

* * *

Closes #447
Also closes #311


bouweandela:
TODO:
- [ ] Check that latitude and vertical coordinate are correctly inverted
- [ ] If original  unit is m of water, convert to kg/m^2 (read original units from netcdf file, filename in cube attributes)
- [ ] if original unit is (0 - 1), set to 1 (read original units from netcdf file, filename in cube attributes)
- [ ] Ignore iris warnings about invalid units at load time, these will be found by cmor checker anyway
- [ ] Create separate classes for variable specific fixes
- [ ] Take https://github.com/ESMValGroup/ESMValCore/pull/413 into account
- [ ] Update corresponding recipe in https://github.com/ESMValGroup/ESMValTool/pull/1432 and check that output filenames can be used with project set to OBS6
- [ ] Also fix hourly time coordinate, should be shifted back half an hour for accumulated/mean/min/max variables, see https://confluence.ecmwf.int/display/CKB/ERA5+data+documentation#ERA5datadocumentation-Meanratesandaccumulations
- [ ] Add unit tests
- [ ] Add documentation
- [ ] Optionally, support per dataset DRS. Currently this is defined per project, but this does not make sense for the datasets in the Native6 project

Peter9192:
I completed the tests. Two tests now fail:

- orog: this is an Fx variable. In the previous version, we let it keeps its time dimension of length 1, but the new cmorizer assumes there is no time axis (in `FixEra5._frequency()`)
- iris is unable to parse the unit 'm of water equivalent'. Need to fix that in the cmorizer as well.

Peter9192:
I made a lot of progress with this issue. Now I think it's time that @bouweandela looks at it again (and  hopefully can finish it ;-) ). 

A remaining issue is that era5(cli) variable names and frequency need to be passed on via the recipe, which is not nice. Another thing is that some of the tests not directly related to ERA5 cmorization fail due to the changes in this PR. 

I've compared a single cmorized cube (uas E1hr) between the old and new cmorizers, and there are still some differences: the old cmorizer adds a lot of attributes, like `:title = "ERA5 data reformatted for ESMValTool v2.0.0b1" ;` Also, time bounds exist for the old version, but not for the new one. In terms of output filenames, I'm not sure what you intended. Do you want to add a short diagnostic renaming the preprocessed file? For the current output filename is: `native6_ERA5_reanaly_1_E1hr_uas_1990-1990` and you want it to be `OBS6_ERA5_reanaly_1_E1hr_uas_199001-199012`

esmvalcore/_data_finder.py
```diff
@@ -34,71 +32,42 @@ def find_files(dirnames, filenames):
...
-      or
-    *[-,_]YYYY*.*
-      or
-    YYYY*[-,_]*.*
-      or
-    YYYY*[-,_]YYYY*[-,_]*.*
-      or
-    YYYY*[-,_]*[-,_]YYYY*.* (Does this make sense? Is this worth catching?)
+    We assume that the filename *must* contain at least 1 year/date.
+    We look for four-to-eight-digit numbers.
```
```suggestion
    We look for four-to-twelve-digit numbers.
```

esmvalcore/_data_finder.py
```diff
@@ -34,71 +32,42 @@ def find_files(dirnames, filenames):
...
-        if pos_ydates_r[ind] or pos_ydates_l[ind]
-    ]
-    start_year = None
-    end_year = None
-    if len(dates) == 1:
-        start_year = int(dates[0][:4])
-        end_year = start_year
-    elif len(dates) == 2:
-        start_year, end_year = int(dates[0][:4]), int(dates[1][:4])
+    # Strip only filename from full path (and discard extension?)
```
```suggestion
    # Strip only filename from full path and discard extension
```

esmvalcore/_data_finder.py
```diff
@@ -34,71 +32,42 @@ def find_files(dirnames, filenames):
...
-            break
-
-    if start_year is None or end_year is None:
-        raise ValueError(
-            'File {0} dates do not match a recognized pattern and time can '
-            'not be read from the file'.format(filename)
-        )
-    return start_year, end_year
+        dates = re.findall(r'([0-9]{4,12})', filename)
+        if not dates:
```
bouweandela: If the filename does not contain dates, try reading them using iris, as the original code did. Only raise an exception if even that failed.

esmvalcore/cmor/_fixes/fix.py
```diff
@@ -146,9 +162,9 @@ def get_fixes(project, dataset, variable):
 
             classes = inspect.getmembers(fixes_module, inspect.isclass)
             classes = dict((name.lower(), value) for name, value in classes)
-            for fix_name in ('allvars', variable):
+            for fix_name in (short_name, 'allvars'):
```
bouweandela: I'm not sure if it's a good idea to change the order here, because you will have to check that all fixes that are already implemented still work. @jvegasbsc Do you have an opinion on this?

esmvalcore/cmor/_fixes/native6/era5.py
```diff
@@ -0,0 +1,242 @@
...
+from ..fix import Fix
+from ..shared import add_scalar_height_coord
+
+
+class FixEra5(Fix):
+    """Fixes for ERA5 variables."""
+
+    @staticmethod
+    def _frequency(cube):
+
```
bouweandela: A one line docstring would be nice here

esmvalcore/cmor/_fixes/native6/era5.py
```diff
@@ -0,0 +1,242 @@
...
+        if not cube.coords(axis='T'):
+            return 'fx'
+        time = cube.coord(axis='T')
+        if len(time.points) == 1:
+            return 'fx'
+
+        interval = time.points[1] - time.points[0]
+        unit = 'hours' if 'hour' in time.units.name else 'days'
```
bouweandela: Why the complicated logic trying to distinguish hours since from days since? You could simply convert the coordinate units to `'days since 1850-1-1 00:00:00.0'` if that hasn't been done yet by the time you reach this code

esmvalcore/cmor/_fixes/native6/era5.py
```diff
@@ -0,0 +1,242 @@
...
+class FixEra5(Fix):
+    """Fixes for ERA5 variables."""
+
+    @staticmethod
+    def _frequency(cube):
+
+        if not cube.coords(axis='T'):
+            return 'fx'
+        time = cube.coord(axis='T')
+        if len(time.points) == 1:
```
bouweandela: Not sure if this is safe to assume

esmvalcore/cmor/_fixes/native6/era5.py
```diff
@@ -0,0 +1,242 @@
...
+            self._fix_direction(cube)
+        return cubes
+
+
+class Fx(FixEra5):
+    """Fixes for time invariant variables."""
+
+    @staticmethod
+    def _remove_time_coordinate(cube):
+        cube = iris.util.squeeze(cube)
```
bouweandela:
```suggestion
        cube = cube[0]
```
may be a bit more readable

esmvalcore/cmor/_fixes/native6/era5.py
```diff
@@ -0,0 +1,242 @@
...
+        if (unit == 'days' and interval - 1 / 24 < 1e-4):
+            return 'hourly'
+        return 'monthly'
+
+
+class Accumulated(FixEra5):
+    """Fixes for accumulated variables."""
+
+    def _fix_frequency(self, cube):
+        if cube.var_name in ['mx2t', 'mn2t']:
```
bouweandela: Could you put the cmor variable names in a comment? That will make the code easier to understand.

tests/unit/data_finder/test_get_start_end_year.py
```diff
@@ -1,99 +1,33 @@
...
-        self.assertEqual(1980, end)
-
-    def test_end_and_date_in_name(self):
-        """Test parse one date at the end and one in experiment's name"""
-        start, end = get_start_end_year(
-            'var_control-1950_whatever_19800101.nc')
-        self.assertEqual(1980, start)
-        self.assertEqual(1980, end)
-
-    def test_read_file_if_no_date_present(self):
```
bouweandela: This test should not be deleted

Peter9192:
@bouweandela and I addressed the remaining issues above, and this PR is now ready for review.

Verifying the output against old cmorizer, I found some differences that I think are actually better in the new version than in the old cmorizer:

- The time units in old are not changed (hours since 1990.....). In new (days since 1850...), they agree with the cmip6 coordinate definition from the cmor table. 
- long_names for lon and lat are written without capitals in the old cmorizer. The new cmorizer uses capitals, in accordance with the CMIP table.
- uas van vas are not in the CMIP6_E1hr table. Their variable definition gets looked up in another table, but the time coordinate is inferred from the E1hr table (time1 in this case, instead of 'time'). The result is that the new cmorizers does not add bounds (the old cmorizer just adds bounds to all coordinates). I think the new behaviour makes more sense, because uas and vas are instantaneous variables, but we can change this (e.g. just add bounds to all vars like the old cmorizer) if people disagree.
- The new cmorizer addresses #398 by shifting the time points half an hour back, but the result is that we end up with a time coordinate that has 1 point less, because 1st of january 00:00 is moved to the previous year, and the last point 31st december 00:30 cannot be inferred because it should have come from the next year.

PS: I don't know why the test is failing on circleci, it passes without errors on my local computer.

mattiarighi:
Asking @valeriupredoi to have a look at the changes in `_data_finder.py`. :beer:

mattiarighi:
All variables successfully tested, except `monthly/orog`, see [main_log.txt](https://github.com/ESMValGroup/ESMValCore/files/4169785/main_log.txt).

Peter9192:
> All variables successfully tested, except `monthly/orog`, see [main_log.txt](https://github.com/ESMValGroup/ESMValCore/files/4169785/main_log.txt).

Thanks @mattiarighi ! I removed monthly/orog because all we use it for is to create an fx variable out of it, and since we're doing that for hourly already, we don't need it in daily and monthly as well.

mattiarighi:
@valeriupredoi are you ok with this?

valeriupredoi:
> @valeriupredoi are you ok with this?

looking at it right now, hold yer horses :horse:

esmvalcore/cmor/fix.py
```diff
@@ -152,8 +148,7 @@ def fix_data(cube,
              short_name,
              project,
              dataset,
-             cmor_table=None,
-             mip=None,
+             mip='',
```
valeriupredoi: what's the reason for changing to an empty string?

esmvalcore/preprocessor/_io.py
```diff
@@ -61,6 +61,13 @@ def load(file, callback=None):
             category=UserWarning,
             module='iris',
         )
+        filterwarnings(
+            'ignore',
+            message="Ignoring netCDF variable '.*' invalid units '.*'",
+            category=UserWarning,
+            module='iris',
+        )
+
```
valeriupredoi: hmmm this is probably a bit dangerous since this is something the user needs to see - why turn it off?

esmvalcore/_data_finder.py
```diff
@@ -32,56 +31,32 @@ def find_files(dirnames, filenames):
...
-    start_year = None
-    end_year = None
-    if len(dates) == 1:
-        start_year = int(dates[0][:4])
-        end_year = start_year
-    elif len(dates) == 2:
-        start_year, end_year = int(dates[0][:4]), int(dates[1][:4])
+    """Get the start and end year from a file name."""
+    stripped_name = os.path.splitext(filename)[0]
+    stripped_name = stripped_name.split(os.sep)[-1]
```
valeriupredoi: these two lines can be replaced by `os.path.basename(filename)`

esmvalcore/_data_finder.py
```diff
@@ -32,56 +31,32 @@ def find_files(dirnames, filenames):
...
+    start_year = end_year = None
+
+    # First check for a block of two potential dates separated by _ or -
+    daterange = re.findall(r'([0-9]{4,12}[-_][0-9]{4,12})', stripped_name)
+    if daterange:
+        start_date, end_date = re.findall(r'([0-9]{4,12})', daterange[0])
+        start_year = start_date[:4]
+        end_year = end_date[:4]
```
valeriupredoi: a bit too many assumptions here, but then again, the previous implementation was even more specific, we need a more elegant and general method in the future. Not a change-request

esmvalcore/cmor/_fixes/native6/era5.py
```diff
@@ -0,0 +1,328 @@
...
+        return 'hourly'
+
+    return 'monthly'
+
+
+def fix_hourly_time_coordinate(cube):
+    """Shift aggregated variables 30 minutes back in time."""
+    if get_frequency(cube) == 'hourly':
+        time = cube.coord(axis='T')
+        time.points = time.points - 1 / 48
```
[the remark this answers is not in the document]
Peter9192: For me it doesn't. Isn't that a python 2 thing?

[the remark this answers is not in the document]
valeriupredoi: correct! I forgot about the change in py3 :beer:

esmvalcore/cmor/_fixes/native6/era5.py
```diff
@@ -0,0 +1,328 @@
...
+        time.guess_bounds()
```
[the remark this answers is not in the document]
Peter9192: Yes, this is always the case. However, I think we could remove this line anyway, since they will be added later in the AllVars fix. Agree @bouweandela ?

esmvalcore/cmor/_fixes/native6/era5.py
```diff
@@ -0,0 +1,328 @@
...
+    return cube
+
+
+def fix_accumulated_units(cube):
+    """Convert accumulations to fluxes."""
+    if get_frequency(cube) == 'monthly':
+        cube.units = cube.units * 'd-1'
+    elif get_frequency(cube) == 'hourly':
+        cube.units = cube.units * 'h-1'
```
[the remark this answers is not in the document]
Peter9192: This function is never called on the time coordinate, and none of the era5 variables that are fixed here have units of day. This is a very specific fix for some variables in [table 3 here](https://confluence.ecmwf.int/display/CKB/ERA5%3A+data+documentation#ERA5:datadocumentation-Parameterlistings) which we convert from hourly/daily accumulations to fluxes.

esmvalcore/cmor/_fixes/native6/era5.py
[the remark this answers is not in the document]
bouweandela: Yes, looks fine to me

esmvalcore/preprocessor/_io.py
bouweandela: Because the cmor check will report wrong units anyway and this will emit unnecessary warnings if the units are wrong but corrected in a fix.

Peter9192: Apparently, because [these will be found by the cmor checker anyway](https://github.com/ESMValGroup/ESMValCore/pull/358#issuecomment-568496093)

valeriupredoi: @bouweandela that needs to be reported to the user even more so in the case when we run in dry run mode so that these, in turn, can be reported to the data providers for possible corrections; any of these cmor hiccups should not be kept under the hood

esmvalcore/cmor/_fixes/native6/era5.py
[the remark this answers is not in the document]
valeriupredoi: aye, understood :beer:

esmvalcore/preprocessor/_io.py
valeriupredoi: ow wait this is in `_io.py` -> true, no need for it here then, am blabbering

esmvalcore/_data_finder.py
Peter9192: Even better: `stem = Path(filename).stem`

esmvalcore/cmor/fix.py
bouweandela: Good point, I fixed this.

Peter9192:
I ran again with Bouwe's most recent changes and everything works fine. Think it can be merged now.

bascrezee:
Nice work @bouweandela, @Peter9192 and others. I want to get started with this in combination with the variable soil moisture. Could you very briefly help me get started. E.g.: 

-where to put the data? (we used to put download instructions at top of the cmorization scripts, where to put them for native6? --> I would be a big fan of having this included in ESMValCore)
-how to specify the dataset in the recipe? 

Thanks.

bouweandela:
Hi Bas,

> where to put the data? (we used to put download instructions at top of the cmorization scripts, where to put them for native6? --> I would be a big fan of having this included in ESMValCore)

I guess we (again) failed to provide enough documentation ;-( The `native6` project can be configured in the config-user.yml and config-developer.yml files with a DRS, similar to other projects. Still to do is to make it possible to have a dataset specific DRS, so we can have more than just ERA5 under the native6 project:
https://github.com/ESMValGroup/ESMValTool/blob/master/config-user-example.yml#L15
https://github.com/ESMValGroup/ESMValCore/blob/a9312a7d5be4fa3aac55c0b2ef089c6b4e1a61a9/esmvalcore/config-developer.yml#L191-L201

> how to specify the dataset in the recipe?

For an example recipe using ERA5, see here: https://github.com/ESMValGroup/ESMValTool/blob/master/esmvaltool/recipes/cmorizers/recipe_era5.yml

PR #374, opened 2019-11-13
Implement three levels of CMOR checks strictness

### LINKED ISSUE #338: Allow for CMOR checks but DO NOT raise exceptions (at the user's risk!)
Not good, guys, not good - analyses that **need** to go through past nit-picky CMOR exceptions still stumble at them and return exceptions with the tool stopping. I thought the whole point of `cmor_strict: false` was to bypass all these roadblocks :beer: Running an IPCC stuff for @LisaBock and am getting pestered by `height2m` not present. Who cares about bloody `height2m`? :angry:

valeriupredoi:

**Tasks**

-   [x] Add unit tests

* * *

Closes #338


valeriupredoi:
@sloosvel and @jvegasbsc pls have a look at what I haz so far - I introduced the STRICT CMOR error (`report_strict_error()`) handled in a very particular way by `report_errors()` func - the logic behind this that we want, no matter what, to have the issues reported by these types of errors logged as errors to the very least, if not raised. Whereas regular errors can be not logged, can be logged but not raised or can be raised. Also bear in mind the `--dry-run` option that controls only the `raise_exception` flag (and nothing else). If you agree with this type of functionality, all's left is to replace `report_error` to `report_strict_error` for the cases in Table https://github.com/ESMValGroup/ESMValCore/issues/338#issuecomment-551877315 with the edit from https://github.com/ESMValGroup/ESMValCore/issues/338#issuecomment-552448498 :beer:

sloosvel:
Thanks @valeriupredoi for starting the work. I am going to test it to fully understand what is being done.

sloosvel:
I think that the messages are a bit confusing, the distinction between regular and strict errors is something that you have come up with according to your own judgement, not something that is given by cmor. I would keep the messages as they are now, with the only changes being  the way they will get reported (warning / error) according to the flag. I can try it for instance with the standard_name to see better how this will work.

valeriupredoi:
hi @sloosvel sorry for the late reply, I was out yesterday. Correct, the messages are my interpretation, please feel free to change as you think it's fit :beer: Also, please take it from here with the changes for the different levels in #338 - as it's now all's let to be done is to replace `report_error` with `report_strict_error` for where's needed. I will correct for the unit tests when that's done :beer:

esmvalcore/_main.py
```diff
@@ -98,6 +98,15 @@ def get_args():
         '--diagnostics',
         nargs='*',
         help="Only run the named diagnostics from the recipe.")
+    parser.add_argument(
+        '--cmor-checks',
+        type=str,
+        action='store_true',
```
bouweandela:
```suggestion
        choices=('none', 'relaxed', 'default'),
```

bouweandela: and add `default='default',`

esmvalcore/_main.py
```diff
@@ -143,6 +152,7 @@ def main(args):
         pattern if TASKSEP in pattern else pattern + TASKSEP + '*'
         for pattern in args.diagnostics or ()
     }
+    cfg['cmor_checks_type'] = args.cmor_checks
```
bouweandela:
```suggestion
    cfg['cmor_checks'] = args.cmor_checks
```

esmvalcore/_recipe.py
```diff
@@ -332,11 +332,19 @@ def _get_default_settings(variable, config_user, derive=False):
...
+        if config_user['cmor_checks_type'] in ['none', 'relaxed']:
+            raise_exception = False
+            if config_user['cmor_checks_type'] == 'relaxed':
+                report_only_warning = True
         settings['cmor_check_metadata'] = {
             'cmor_table': variable['cmor_table'],
             'mip': variable['mip'],
             'short_name': variable['short_name'],
             'frequency': variable['frequency'],
+            'raise_exception': raise_exception,
```
bouweandela: Two extra arguments seems overkill, just one argument:
`'raise_on': config_user['cmor_checks'],`
should be sufficient.

bouweandela: Or something similar, maybe the levels could be translated to `'never', 'warning', 'error'` or something similar here..

esmvalcore/cmor/check.py
```diff
@@ -57,11 +57,16 @@ def __init__(self,
                  var_info,
                  frequency=None,
                  fail_on_error=False,
+                 raise_exception=True,
```
bouweandela: One extra argument should be sufficient, e.g.
`raise_on=` with possible values `'never', 'warning', 'error'`

esmvalcore/_recipe.py
valeriupredoi: no, the reason behind this is to recycle the functionality for dry-run where we don't necessarily need to use the `--cmor-checks` flag, I should have explained it better

esmvalcore/cmor/check.py
valeriupredoi: see my comment above about dry-run

valeriupredoi:
regular tests now pass, current fail due to py3.8/numba, PR #376

sloosvel:
I don't think there was a need to create a report_strict_errors function. Just send the messages to error, send them to warning or do not report them for the specific checks on the table. In this way, there are less modifications for the way regular checks are being reported. It seemed to work for the standard_name, but I have not tested properly.

valeriupredoi:
cheers! sounds good @sloosvel :beer:

valeriupredoi:
OK @sloosvel so I actually made it work proper based on your implementation from last week and tested with a file that is like
```
c.coord("latitude").standard_name = 'grid_latitude'
c.coord("longitude").standard_name = 'grid_longitude'
c.long_name = 'Cows per tank'
```
-> it goes through no problem for `none` and `relaxed` and stops in the old way for `default` - can you pls have a look and test as well? If you plug in changes I'd very much appreciate you testing them thoroughly before so that we actually have a working version at all times here :beer:

Pending your suggestions/new commits, I'll fix the automated testing after :beer:

valeriupredoi:
also for a unit discrepancy (needed `K`, having `c.units = 'C'`) it fails at all levels of `--cmor-checks` as it should be, so well done with the implementation @sloosvel :beer: (note: I fixed the tests, we need to implement specific tests for these tiered cmor checks though)

sloosvel:
> also for a unit discrepancy (needed `K`, having `c.units = 'C'`) it fails at all levels of `--cmor-checks` as it should be, so well done with the implementation @sloosvel beer (note: I fixed the tests, we need to implement specific tests for these tiered cmor checks though)

Which units do you mean? Because for var units the check should be relaxed. For coord units it should fail at all levels though.

esmvalcore/cmor/check.py
```diff
@@ -123,7 +127,8 @@ def report_errors(self):
             msg = 'There were errors in variable {}:\n{}\nin cube:\n{}'
             msg = msg.format(self._cube.var_name, '\n '.join(self._errors),
                              self._cube)
-            raise CMORCheckError(msg)
+            if self._raise_exception:
```
sloosvel: I would say that putting this condition here will affect the reporting of "regular" errors.  And for errors that have a flexible reporting this condition has already been checked at the report_strict_error function.  So I removed it to be safe. I am writing some tests now.

valeriupredoi: OK @sloosvel if you are testing thoroughly then all's great since behaviors will emerge from testing. Cheers :beer:

sloosvel:
The tests seem to be working, but I have a question regarding the none flag because it is confusing me. Is it supposed not to report those particular errors or just skip the whole cmor checks entirely? Because if it's the latter I will change things.

valeriupredoi:
Hi Saskia, sorry for the no-show, been clogged in IPCC and other systems stuff. About `none` -> it should perform the checks but not fail on any exception, however crazy the data is. Another thing, as it is now it doesn't fully work, see example below: I have set of HiResMIP data that is OK CMORized but there are still a few hitches - the current implementation fails at all three cmor check levels:
```
variable: tas 
default
=======
esmvalcore.cmor.check.CMORCheckError: There were errors in variable tas:
height2m: does not exist
in cube:
air_temperature / (K)               (time: 12; latitude: 192; longitude: 384)
Raised exception

relaxed
=======
same
Raised exception

none
====
same
Raised exception
```

sloosvel:
Yes I found out when I pushed to the branch that the rank check is not working properly and breaks the tests... But it's strange because my local tests are all running fine. I still need to find out why is this happening.

valeriupredoi:
another thing is for Unit conversion tests for instance when I have a cube with `tas` in `C` and not in `K`: for `none` and `relaxed` it passes the check but fails to run `fix_data`: `2019-11-27 16:12:39,559 UTC [2409] ERROR   Failed to run fix_data(air_temperature / (C)               (time: 312; latitude: 96; longitude: 192)` - with my previous implementation this was taken care of :cry:

sloosvel:
But this is because it's failing at a later step  (which is one of the risks of running with the non-strict flags, things may fail unexpectedly). So far the relax of the checks is only applied at the fix_metadata step, the fix_data step comes in later. And that is why I am not sure the way to go would be to directly relax the report_errors function, because it gets called by all checks and at different steps. Which means that checks that are supposed to be reported as errors (even with the relaxed flag) would not get reported. 

But in any case, if you think that the checks at the fix_data step should also be relaxed, then it needs to be implemented differently for sure.

valeriupredoi:
if you don't check for raising you get cases like this:
```
esmvalcore.cmor.check.CMORCheckError: There were errors in variable tas:
lat: units should be degrees_north, not kg m-2 s-1
```
which is raised even if cmor checks are none

I think you need to put back
```
            if self._raise_exception:
                raise CMORCheckError(msg)
```
:beer:

valeriupredoi:
> But in any case, if you think that the checks at the fix_data step should also be relaxed, then it needs to be implemented differently for sure.

Let's leave the data for now, it's already enough what we're doing here and even if we decide to change it it will go in another PR. Have a look at the Units case above (not the C -> K, that was me dumb trying to convert from K to K via the C name)

valeriupredoi:
Very nice set of unit tests btw! I looked at them and saw the rank tests (2 of them) fail and they fail irrespective of the status they are run under -> that's coz of the raiser raising at all times; if you replace the raiser with the condition `if raise_exception: kill the show` then these will be fixed; that indeed introduces a bit of a flaky behavior wrt warnings (some warnings are not displayed, even if they should be) but I'd go look and see why that is happening rather than reinforce the raiser. Anyway, off home now but will have a closer look myself tomorrow :beer:

esmvalcore/_recipe.py
bouweandela: Please do?

bouweandela:
> I have a question regarding the none flag because it is confusing me. Is it supposed not to report those particular errors or just skip the whole cmor checks entirely? Because if it's the latter I will change things.

My initial thought was to just not run the preprocessor functions `check_metadata` etc. if it would be set to `'none'`, since this is easiest to implement. However, it might be nicer to still run the checks (as far as this is possible, e.g. it's impossible to check the units of a coordinate that is missing), because that will allow using ESMValCore as a tool for checking CMOR compliance. This is also more in line with what would be needed for the dry-run option.

> So far the relax of the checks is only applied at the fix_metadata step, the fix_data step comes in later. And that is why I am not sure the way to go would be to directly relax the report_errors function, because it gets called by all checks and at different steps. Which means that checks that are supposed to be reported as errors (even with the relaxed flag) would not get reported.
>
> But in any case, if you think that the checks at the fix_data step should also be relaxed, then it needs to be implemented differently for sure.

Let's stick with the CMORCheck.check_metadata method for now. Actually the whole concept of having 3 different fix functions and two different checking functions seems overly complicated at this stage:
- [fix_file](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/api/esmvalcore.preprocessor.html#esmvalcore.preprocessor.fix_file) -> never needed
- [fix_metadata](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/api/esmvalcore.preprocessor.html#esmvalcore.preprocessor.fix_metadata) -> all fixes can be done here
- [fix_data](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/api/esmvalcore.preprocessor.html#esmvalcore.preprocessor.fix_data) -> no longer needed: we use dask arrays now, so only those chunks that are actually computed in the end will be fixed. This was an optimization not to apply fixes to data that is e.g. outside of the time range.
- [cmor_check_metadata](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/api/esmvalcore.preprocessor.html#esmvalcore.preprocessor.cmor_check_metadata) -> almost all checks are done here
- [cmor_check_data](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/api/esmvalcore.preprocessor.html#esmvalcore.preprocessor.cmor_check_data) -> the only check here is (part of) the check of the values of the coordinates. I think it would be fine to move this check to the metadata check

valeriupredoi:
Agreed with what Le Bouwe @bouweandela is saying (which is pretty much the plan for this PR as we started it) but let's not overhaul all those funcs and start moving a lot of code; I think this PR should be minimum impact wrt code changes and maximum outcome wrt to the 3 cmor checks levels; we should however bear in mind that the checks will be overhauled in a future release :beer:

valeriupredoi:
Saskia @sloosvel I fixed a wee bug that was making the rank tests fail (altering the actual needed functionality) and now I am super happy with the behavior of the whole show! I tested all sorts of scenarios and all goes to plan. Nice one! :beer:

valeriupredoi:
I'll update the documentation :tractor:

valeriupredoi:
although...I still get this as before:
```
esmvalcore.cmor.check.CMORCheckError: There were errors in variable tas:
height2m: does not exist
in cube:
air_temperature / (K)               (time: 12; latitude: 384; longitude: 768)
```
we should not make the missing coordinate so strict :hammer:

sloosvel:
> Saskia @sloosvel I fixed a wee bug that was making the rank tests fail (altering the actual needed functionality) and now I am super happy with the behavior of the whole show! I tested all sorts of scenarios and all goes to plan. Nice one! beer

Ah ok, I forgot to commit those changes. That's why the tests were running fine in my machine. Thanks for noticing!

>we should not make the missing coordinate so strict hammer

Ok, I will relax that as well. In fact, I have to go over the checks table again to see if I missed anything else.

valeriupredoi:
It's always good to perform a check via either `git status` or `git diff` in your local repo see if there's any uncommitted changes :grin: 

> Ok, I will relax that as well. In fact, I have to go over the checks table again to see if I missed anything else.

Cheers Saskia, let you to it, lemme know when you done so I can test again with that corner case :beer:

bouweandela:
@jvegasbsc It would be really good if you could have a look at this, now that you're back.

jvegasbsc:
> @jvegasbsc It would be really good if you could have a look at this, now that you're back.

I already discussed this with Saskia. I am trying to find a way to generalize it so it is easy to extend in the future

sloosvel:
Relaxed if some coordinate like height2m is missing. Lats, lons and time missing is still reported as an error though.

Other than that, the other open issue is what the none flag is supposed to do. Because right now it's like a relaxed flag but without reporting.

valeriupredoi:
very nice @sloosvel my tests now all pass fine (I yapf-ed your test_check to please the gods of pep8) :beer:

jvegasbsc:
Ok. I have reworked a bit the checker for this:

- Checker messages now are one of the following categories: DEBUG, WARNING, ERROR or CRITICAL

- Users can define the level of message that will be treated as an ERROR: --check-level=WARNING, ERROR or CRITICAL

- I added to special options: NEVER_FAIL: all treated as warnings, IGNORE_ALL: all treated as debug messages

It should be easier to extend if we want more levels of strictness in the future that the previous implementation

@valeriupredoi : one question about the dry run. Why we decided to do not launch the exception in the checker instead of catching it outside?

jvegasbsc:
@bouweandela , @mattiarighi  The missing test is due to the checker trying to add the time auxiliary coordinates to a cube that does not have time when checking with the IGNORE_ALL config

We wanted to remove this behaviour for a while, but maybe is better to just catch this exception now in the checker and open a pull request for it later. What do you think?

jvegasbsc:
By the way: it would be very easy to add similar customization to define the level of error to be plotted as warnings, if you think that will be useful

valeriupredoi:
> one question about the dry run. Why we decided to do not launch the exception in the checker instead of catching it outside?

We wanted to let the code run and just log all exceptions instead of raising and stopping at the first encountered exception

I like the stuff @jvegasbsc - am testing now with my previous tests. But I find it a bit confusing with all these check-level options -> I reckon it's too many dials and buttons for the regular user. I like it when we have plenty of options but I am an engineer, I think a regular scientist user might be a bit confused at first

valeriupredoi:
here, running with `NEVER_FAIL` I get:
```
2019-12-03 11:29:30,112 UTC [116770] ERROR   Program terminated abnormally, see stack trace below for more information
"""
  File "/home/users/valeriu/esmvalcore/esmvalcore/cmor/check.py", line 302, in _check_dim_names
    cube_coord = self._cube.coord(var_name=coordinate.out_name)
  File "/home/users/valeriu/anaconda3R/envs/esmvaltool_users/lib/python3.7/site-packages/iris/cube.py", line 1463, in coord
    raise iris.exceptions.CoordinateNotFoundError(msg)
iris.exceptions.CoordinateNotFoundError: 'Expected to find exactly 1  coordinate, but found none.'

  File "/home/users/valeriu/esmvalcore/esmvalcore/cmor/check.py", line 313, in _check_dim_names
    coord = self._cube.coord(coordinate.standard_name)
  File "/home/users/valeriu/anaconda3R/envs/esmvaltool_users/lib/python3.7/site-packages/iris/cube.py", line 1463, in coord
    raise iris.exceptions.CoordinateNotFoundError(msg)
iris.exceptions.CoordinateNotFoundError: 'Expected to find exactly 1 height coordinate, but found none.'

  File "/home/users/valeriu/anaconda3R/envs/esmvaltool_users/lib/python3.7/multiprocessing/pool.py", line 121, in worker
  File "/home/users/valeriu/esmvalcore/esmvalcore/_task.py", line 686, in _run_task
    output_files = task.run()
  File "/home/users/valeriu/esmvalcore/esmvalcore/_task.py", line 238, in run
  File "/home/users/valeriu/esmvalcore/esmvalcore/preprocessor/__init__.py", line 403, in _run
  File "/home/users/valeriu/esmvalcore/esmvalcore/preprocessor/__init__.py", line 271, in apply
  File "/home/users/valeriu/esmvalcore/esmvalcore/preprocessor/__init__.py", line 213, in preprocess
  File "/home/users/valeriu/esmvalcore/esmvalcore/preprocessor/__init__.py", line 199, in _run_preproc_function
  File "/home/users/valeriu/esmvalcore/esmvalcore/cmor/fix.py", line 122, in fix_metadata
    cube = checker(cube).check_metadata()
  File "/home/users/valeriu/esmvalcore/esmvalcore/cmor/check.py", line 106, in check_metadata
    self._check_dim_names()
  File "/home/users/valeriu/esmvalcore/esmvalcore/cmor/check.py", line 342, in _check_dim_names
    self._does_msg, coordinate.name, 'exist')
  File "/home/users/valeriu/esmvalcore/esmvalcore/cmor/check.py", line 682, in report_error
    self.report(CheckLevels.ERROR, message, *args)
  File "/home/users/valeriu/esmvalcore/esmvalcore/cmor/check.py", line 645, in report
    elif level < self._check_level:
TypeError: '<' not supported between instances of 'CheckLevels' and 'str'
"""

  File "/home/users/valeriu/esmvalcore/esmvalcore/_main.py", line 248, in run
    conf = main(args)
  File "/home/users/valeriu/esmvalcore/esmvalcore/_main.py", line 176, in main
  File "/home/users/valeriu/esmvalcore/esmvalcore/_main.py", line 226, in process_recipe
    recipe.run()
  File "/home/users/valeriu/esmvalcore/esmvalcore/_recipe.py", line 1290, in run
    max_parallel_tasks=self._cfg['max_parallel_tasks'])
  File "/home/users/valeriu/esmvalcore/esmvalcore/_task.py", line 607, in run_tasks
  File "/home/users/valeriu/esmvalcore/esmvalcore/_task.py", line 653, in _run_tasks_parallel
    _copy_results(task, running[task])
  File "/home/users/valeriu/esmvalcore/esmvalcore/_task.py", line 674, in _copy_results
    task.output_files, updated_products = future.get()
  File "/home/users/valeriu/anaconda3R/envs/esmvaltool_users/lib/python3.7/multiprocessing/pool.py", line 657, in get
    raise self._value
TypeError: '<' not supported between instances of 'CheckLevels' and 'str'
```

valeriupredoi:
btw this is `height2m` which is indeed missing but I don't need it and I want the code to go past it (like with Saskia's previous implementation)

jvegasbsc:
A problem with parsing of the args. Will solve it in a minute

valeriupredoi:
done :grin:

valeriupredoi:
AWESOME!! so for `CRITICAL` I get:
```
esmvalcore.cmor.check.CMORCheckError: There were errors in variable tas:
lon: standard_name should be longitude, not grid_longitude
 lat: standard_name should be latitude, not grid_latitude
```
which is the way to be, and for `IGNORE_ALL` I get the same but only as `WARNING` and my show goes one. This is tiptop, man! :beer:

valeriupredoi:
and the safety of `fix_data` not being able to perform correctly is still there `ValueError: Unable to convert from 'Unit('kg m-2 s-1')' to 'Unit('K')'.`

valeriupredoi:
ok I fail to see the difference between `NEVER_FAIL` and `IGNORA_ALL`: I get the same warning (and no fail, which is lovely) for this scenario:
```
Standard name for tas changed from convection_time_fraction to air_temperature
 Long name for tas changed from Cows per tank to Near-Surface Air Temperature
 longitude: does not exist
 lat: standard_name should be latitude, not grid_latitude
 tas: does not match coordinate rank
```
I think we should compress the options - too many for a user to be 100% sure what they must use (or at least, compress them only in the documentation and leave the functionality under the hood)

valeriupredoi:
so here's my suggestion:
```
WARNING: fail at any warning: nobody will use this -> remove from documentation
ERROR(DEFAULT): keep, equivalent to previous functionality
CRITICAL: fail only on critical issues: very useful BUT we need to say what are the actual cases
NEVER_FAIL, IGNORE_ALL: identical: keep just one in the documentation
```

valeriupredoi:
Javi, note that `IGNORE_ALL` still reports warnings, so your tests checking for no warning output will fail

valeriupredoi:
OK I fixed the attribute error and a fix_data test, now the only test that fails is tests/unit/cmor/test_cmor_check.py: l.496 but I am not at all sold on the idea that we should add time-like aux coords when there may not actually be a time coord at all, I would add those only if time is in coords() - the strict check will have alterted the user if the data is missing an actual time coord and if they chose to ignore that then the code should still go on without stoppage (iris exception coord not found resulting from trying to add the aux time coords)

valeriupredoi:
I added some documentation - can you pls check @jvegasbsc and @sloosvel and amend stuffs plss - I am also looking to add some examples where `CRITICAL` should be used :beer:

valeriupredoi:
Hahaha, editing documentation Ernest Hemingway-style: delete everything and start from scratch when sober :grin: Much better, cheers, Javi! OK so this should be ready for merge, no? :beer:

jvegasbsc:
For me yes. It would be nice to have a look from @axel-lauer and @bouweandela, specially on the naming of the levels, command line opetions and this kind of things. We want this to be intuitive and more eyes on this will help

valeriupredoi:
yeps, @zklaus as well

bouweandela:
My suggestion for the command line argument would be `--cmor-checks` with options `ignore`, `relaxed`, `default`, `strict` (instead of NEVER_FAIL, CRITICAL, ERROR, WARNING respectively). I agree with @valeriupredoi that having two different ways to ignore the results of the checks is too much.

It would be nice if in addition to writing the errors to the log, there would also be a separate text file listing the errors per file, e.g. in the preprocessor output directories. This would be useful for checking large numbers of files and reporting issues to the modelling groups.

sloosvel:
> It would be nice if in addition to writing the errors to the log, there would be a separate text file listing the errors per file, e.g. in the preprocessor output directories. 

I have been working on that, but in the case of running with a relaxed flag the preproc dir would get removed if the diagnostic finishes. So I don't know if it would be better to save the log somewhere else (maybe with the option that in the case it's empty, remove the log), or just assume that whoever is interested in keeping the log should set the remove_preproc_dir to false.

bouweandela:
I think such an option would be most interesting with the --dry-run flag from the other pull request anyway, when you're not saving data. That way esmvaltool could be used as a CMOR compliance checking tool. I think this could be a very nice feature, especially if the resulting output file would be written in a format that can be used with the batch CMIP6 data issue reporting tool that is apparently available.

valeriupredoi:
@bouweandela great minds think alike, see https://github.com/ESMValGroup/ESMValCore/issues/387 :grin:

valeriupredoi:
But I think this is of high priority and we shouldn't dilly-doll around it no more, get it in and then tackle the other related suggestions in upcoming PR's :beer:

doc/esmvalcore/fixing_data.rst
```diff
@@ -245,3 +245,42 @@ missing coordinate you can create a fix for this model:
...
+and therefore make the checker fail or warnings or debug messages.
+For this purpose there is an optional command line option `--check-level`
+that can take a number of values, listed below from the lowest level of
+strictness to the highest:
+
+- ``ignore``: all issues, regardless of severity, will be reported as
+    warnings. Checker will never fail. Use this at your own risk.
+
+- ``relaxed``: only CRITICAL issues are treated as errors. We recommend not to
+    relay on this mode, although it can be useful if there are errors preventing
```
bouweandela:
```suggestion
    rely on this mode, although it can be useful if there are errors preventing
```

esmvalcore/cmor/check.py
```diff
@@ -91,23 +98,56 @@ def check_metadata(self, logger=None):
...
+            all checks and then raises.
+
+        """
+        if logger is not None:
+            self._logger = logger
+
+        if self._cmor_var.units:
+            units = self._get_effective_units()
+            if str(self._cube.units) != units:
+                self._cube.convert_units(units)
```
bouweandela: I would expect some sort of warning when the data has the wrong units?

esmvalcore/cmor/check.py
```diff
@@ -90,23 +97,55 @@ def check_metadata(self, logger=None):
...
 
-        self.report_debug_messages(logger)
-        self.report_warnings(logger)
+        self.report_debug_messages()
+        self.report_warnings()
         self.report_errors()
 
         return self._cube
 
+    def check_data(self, logger=None):
```
bouweandela: The `logger` argument is not documented, also for the other functions that have it. This class is part of the public api: https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/api/esmvalcore.cmor.html#module-esmvalcore.cmor.check

esmvalcore/cmor/check.py
```diff
@@ -672,12 +738,15 @@ def _checker(cube):
...
+            check_level=check_level,
             automatic_fixes=automatic_fixes)
 
     return _checker
 
 
-def cmor_check_metadata(cube, cmor_table, mip, short_name, frequency):
+def cmor_check_metadata(cube, cmor_table, mip,
+                        short_name, frequency,
+                        check_level):
```
bouweandela: The new argument should be mentioned in the docstring

esmvalcore/cmor/check.py
```diff
@@ -696,12 +765,14 @@ def cmor_check_metadata(cube, cmor_table, mip, short_name, frequency):
...
-    checker = _get_cmor_checker(cmor_table, mip, short_name, frequency)
+    checker = _get_cmor_checker(cmor_table, mip,
+                                short_name, frequency,
+                                check_level=check_level)
     checker(cube).check_metadata()
     return cube
 
 
-def cmor_check_data(cube, cmor_table, mip, short_name, frequency):
+def cmor_check_data(cube, cmor_table, mip, short_name, frequency, check_level):
```
bouweandela: Same here

esmvalcore/cmor/check.py
```diff
@@ -10,6 +11,9 @@
 
 from .table import CMOR_TABLES
 
+CheckLevels = IntEnum(
```
bouweandela: Should this be added to the API documentation?

esmvalcore/cmor/fix.py
```diff
@@ -56,7 +56,8 @@ def fix_metadata(cubes,
                  dataset,
                  cmor_table=None,
                  mip=None,
-                 frequency=None):
+                 frequency=None,
+                 check_level=CheckLevels.DEFAULT):
```
bouweandela: Documentation is missing

esmvalcore/cmor/fix.py
```diff
@@ -148,13 +125,43 @@ def fix_metadata(cubes,
...
+
 def fix_data(cube,
              short_name,
              project,
              dataset,
              cmor_table=None,
              mip=None,
-             frequency=None):
+             frequency=None,
+             check_level=CheckLevels.DEFAULT):
```
bouweandela: and here too

sloosvel:
I'm trying to solve the conflicts.  From the master branch, where is the argument warning_as_errors in the checker coming from? Isn't that what the strict flag is supposed to do?

sloosvel:
According to @jvegasbsc , the argument can be removed.

esmvalcore/cmor/check.py
bouweandela: This comment has been addressed for coordinate units, but not for the units of the variable. Could you have a look please?

sloosvel: It's being reported while checking, in line 264. But I can add the warning here if you prefer.

tests/integration/test_recipe.py
```diff
@@ -352,13 +352,15 @@ def test_default_preprocessor(tmp_path, patched_datafinder, config_user):
             'output_dir': fix_dir,
         },
         'fix_data': {
+            'check_level': None,
```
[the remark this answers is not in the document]
sloosvel: Sorry for the spam, just saw the Add suggestion to branch

mattiarighi:
What would be a good strategy for testing this?
Running a few representative recipes with different level of cmor strictness?

valeriupredoi:
wow boy am glad to see this ready for testing - awesome work guys! I recommend using a few datasets that we don't have fixes for yet and would normally fail a number of checks and running with different strictness levels. I can think of the HIRES mip's data that is, by all means, okay from a CF point of view but quite down with the sickness from a CMOR standard. I can post a recipe here in a few minutes

valeriupredoi:
eg @mattiarighi do you have access to these models on DKRZ (I ran them on Jasmin but not sure if you have them on DKRZ too):
```
  # LOWRES models in /gws/nopw/j04/primavera5/stream1/CMIP6/HighResMIP
  - {dataset: CMCC-CM2-HR4, exp: hist-1950, grid: gn}  # CMCC
  - {dataset: CNRM-CM6-1, exp: hist-1950, ensemble: r1i1p1f2} # CNRM-CERFACS
  #- {dataset: EC-Earth3P, exp: hist-1950, ensemble: r1i1p2f1} # EC-Earth-Consortium
  - {dataset: ECMWF-IFS-LR, exp: hist-1950, ensemble: r1i1p1f1} # ECMWF
  - {dataset: HadGEM3-GC31-LL, exp: hist-1950, ensemble: r1i1p1f1, grid: gn} # MOHC
  - {dataset: MPI-ESM1-2-HR, exp: hist-1950, ensemble: r1i1p1f1, grid: gn} # MPI-M

  # HIRES models in /gws/nopw/j04/primavera5/stream1/CMIP6/HighResMIP
  - {dataset: CMCC-CM2-VHR4, exp: hist-1950, grid: gn}  # CMCC
  - {dataset: CNRM-CM6-1-HR, exp: hist-1950, ensemble: r1i1p1f2} # CNRM-CERFACS
  #- {dataset: EC-Earth3P-HR, exp: hist-1950, ensemble: r1i1p2f1} # EC-Earth-Consortium issues with lat-lon coords
  - {dataset: ECMWF-IFS-HR, exp: hist-1950, ensemble: r1i1p1f1} # ECMWF
  - {dataset: HadGEM3-GC31-HM, exp: hist-1950, ensemble: r1i1p1f1, grid: gn} # MOHC
  - {dataset: MPI-ESM1-2-XR, exp: hist-1950, ensemble: r1i1p1f1, grid: gn} # MPI-M
```

mattiarighi:
I tested using [this issue](https://github.com/ESMValGroup/ESMValTool/issues/1561) a test case:

- with `strict` it stops with what usually is a warning (`rlut: attribute positive not present`), so it's correct
- with `default` it stops because of the above issue, also correct
- with `relaxed` same as `default`, I guess also correct?
- with `ignore` it still stops.

I'm not sure the above issue is the right test case, though.

jvegasbsc:
> I'm not sure the above issue is the right test case, though.

It is not, because it is an iris issue that is making us fail in an automatic fix

mattiarighi:
Ok, I'll look for a different test case.

valeriupredoi:
lemme test with the HIRES/LOWRES Primavera models

valeriupredoi:
OK for `CNRM-CM6-1` in HiRes mips (LowRes here):

 - strict:
```
esmvalcore.cmor.check.CMORCheckError: There were errors in variable tas:
Added guessed bounds to coordinate lon from var lon
 Added guessed bounds to coordinate lat from var lat
in cube:
air_temperature / (K)               (time: 120; latitude: 128; longitude: 256)
     Dimension coordinates:

```
- default:
PASS but report warning:
```
2020-03-10 16:15:58,179 UTC [121341] WARNING There were warnings in variable tas:
Added guessed bounds to coordinate lon from var lon
 Added guessed bounds to coordinate lat from var lat
```
- relaxed:
PASS but report warning:
```
2020-03-10 16:19:09,868 UTC [124370] WARNING There were warnings in variable tas:
Added guessed bounds to coordinate lon from var lon
 Added guessed bounds to coordinate lat from var lat
```
 - ignore:
PASS but report warning:
```
2020-03-10 16:20:44,096 UTC [125197] WARNING There were warnings in variable tas:
Added guessed bounds to coordinate lon from var lon
 Added guessed bounds to coordinate lat from var lat
```
- no flag:
identical to default

valeriupredoi:
Unfortunately I can not reproduce any other data issues with these types because the HIRES Primavera guys did their job and corrected/updated the data in Dec-Jan, well done to them :beer:

jvegasbsc:
> HIRES Primavera guys did their job and corrected/updated the data in Dec-Jan, well done to them

PRIMAVERA people is very professional and competente. Don't you agree, @sloosvel? 😄

valeriupredoi:
Handling of a regular data issue we see from https://github.com/ESMValGroup/ESMValCore/issues/537

 - strict:
FAIL with
```
esmvalcore.cmor.check.CMORCheckError: There were errors in variable tos:
tos: does not match coordinate rank
in cube:
sea_surface_temperature / (degC)    (time: 1980; latitude: 232; longitude: 360)
```
- default: 
same as strict
- relaxed:
PASS with warning:
```
[21341] WARNING There were warnings in variable tos:
tos: does not match coordinate rank
```
- ignore:
same as relaxed
- no flag:
same as strict

Guys, am happy with the tests - if @mattiarighi is OK I reckon it's safe to merge - and well done, this is a very useful bit of shtuff! :beer:

PR #376, opened 2019-11-15
pinned python lower than 3.8

valeriupredoi:
Bunch of packages from our environment are not yet able to deal with `python=3.8` that gets picked up automatically by an `env update` or `update all` (eg `numba` and possible a bunch more). Can we get this in asap, please :beer:

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -9,5 +9,5 @@ dependencies:
   - graphviz
   - gcc_linux-64
   - libunwind  # Needed for Python3.7+
-  - python>=3.6
+  - python>=3.6,<3.8
   - python-stratify
--- a/meta.yaml
+++ b/meta.yaml
@@ -29,13 +29,13 @@ build:
 requirements:
   build:
     - git
-    - python>=3.6
+    - python>=3.6,<3.8
     # Normally installed via pip:
     - pytest-runner
     - setuptools_scm
   run:
     # esmvaltool
-    - python>=3.6
+    - python>=3.6,<3.8
     - libunwind  # specifically for Python3.7+
     - graphviz
     - iris>=2.2.1


bouweandela:
@jvegasbsc This is one of those cases where it would have been nice not to have numba as a dependency.

valeriupredoi:
fairly sure it's not just Numba that's been caught sleeping, man

valeriupredoi:
> Can you also pin in meta.yaml?

done, bro

bjoernbroetz:
Do we need to address this in the https://github.com/ESMValGroup/ESMValTool repository as well?

zklaus:
> fairly sure it's not just Numba that's been caught sleeping, man

Could you please post the list? This way we will know when we can move back to normal.

valeriupredoi:
> Do we need to address this in the https://github.com/ESMValGroup/ESMValTool repository as well?

no, the environment is restrictive enough not to allow conda to update Python (I just tested)

valeriupredoi:
> > fairly sure it's not just Numba that's been caught sleeping, man
> 
> Could you please post the list? This way we will know when we can move back to normal.

I would have done that if I had a list. Unfortunately it was stopping at Numba compilation. We could turn it off (remove Numba from deps) and try again to see where next it may stop but I think it may be overkill for now

zklaus:
Well, numba is really special in that it is affected by the changes to the internal byte-code format in python 3.8, so if there is no extra reason to belief otherwise it seems reasonable to assume that no other package has problems.

bouweandela:
I think the only other package affected was fiona, you can look this up in
the installation log file, which is recorded as an artifact by circle ci.

bouweandela:
And I believe we're actually using some deprecated functions that have been
removed in Python 3.8 to check that there are no invalid preprocessor
arguments in the recipe, so we might need to update
esmvalcore/preprocessor/__init__.py too before it works.

PR #385, opened 2019-12-03
Update _task.py due to typo

BenMGeo:
fix typo in making files executable
@bascrezee


bouweandela:
This is actually wrong. That error occurs if a file has it's executable bit set (e.g. someone did `chmod +x random_diagnostic.py`, but the file is not executable (e.g. does not contain `#!/usr/bin/env python` as a first iine).

PR #391, opened 2019-12-09
Distinguish between out_name and var_name

### LINKED ISSUE #333: Problem loading zg for 6hrPlevPt
**Describe the bug**

When trying to work with the geopotential height for `mip: 6hrPlevPt`, the variable on BADC (and on ESGF) is called `zg`. But the tool gives the following error:

`esmvalcore._recipe_checks.RecipeError: Unable to load CMOR table 'CMIP6' for variable 'zg' with mip '6hrPlevPt'`

Which is right because there is no such variable called `zg` in this table. Is this an issue with the way the files are named? Or is this something the tool should be able to handle? 

I tried to attach an example recipe but apparently attaching yml files is not supported, here is an example dataset:

 - {project: CMIP6, mip: 6hrPlevPt, dataset: CNRM-CM6-1, exp: historical, ensemble: r1i1p1f2, grid: gr, start_year: 1982, end_year: 1982}

jvegasbsc:
@sloosvel (#333) found that some tables contain variables with the 'out_name' different from the 'var_name'. We are currently assuming that both are the same, this pull request fixes this.

Closes #333


sloosvel:
Thanks @jvegasbsc , it looks like it works well. Although, the variable needs to be called in the recipe by the var_name, not the out_name. Maybe this could become confusing if someone is trying to work with 6hourly files named zg_... .nc , because in the recipe they would need to call zg7h (or zg27) to be able to load the files properly.

valeriupredoi:
yeah am with Saskia on this one

jvegasbsc:
Yes, that is inconvenient. But i have no idea on how I can distinguish between them, as they belong to the same table

jvegasbsc:
> I think the choice var_name is not very clear, because this is also the name of the iris cube attribute that contains the short_name. Would it be possible to use something else? Maybe cmor_name instead of var_name?

Good idea. I will change it

> Note that this pull request breaks existing recipes in which short_name is defined, i.e.:
examples/recipe_variable_groups.yml
examples/recipe_check_obs.yml
recipe_flato13ipcc.yml
recipe_tcr.yml
recipe_ecs.yml
recipe_anav13jclim.yml

Ok. I will create a companion branch in ESMValTool to fix them

esmvalcore/cmor/fix.py
```diff
@@ -110,23 +110,23 @@ def fix_metadata(cubes,
         if len(cube_list) != 1:
             cube = None
             for raw_cube in cube_list:
-                if raw_cube.var_name == short_name:
+                if raw_cube.var_name == cmor_name:
```
bouweandela: I think this would be the `short_name`, or am I missing something?

jvegasbsc: You are right

bouweandela:
Can you also check what name is used in the drs? The short name or the cmor name?

jvegasbsc:
It is the `short_name`. It is not a problem because the variablse that share table and out_name are never present together in the data request for any activity. We need to distinguish them to be able to check against the correct definition, though

bouweandela:
@jvegasbsc Could you have a look at https://github.com/ESMValGroup/ESMValCore/pull/391#discussion_r364770086? Apart from that, I think this pull request looks fine.

bouweandela:
@jvegasbsc Could you a have a look at the merge conflicts and failing tests?

mattiarighi:
I tested successfully a few recipes and cmorizers, other needs still to be adjusted (I reported [here](https://github.com/ESMValGroup/ESMValTool/pull/1493)).

bouweandela:
@jvegasbsc Instead of the massive renaming approach, I would be more in favour of solving this with minimal changes, similar to how it was done for the variables with different names between CMIP5 and CMIP6 in #595. Do you think that would be possible?

jvegasbsc:
> @jvegasbsc Instead of the massive renaming approach, I would be more in favour of solving this with minimal changes, similar to how it was done for the variables with different names between CMIP5 and CMIP6 in #595. Do you think that would be possible?
Not, because it is not a massive rename but a splitting one name in two. Currently we are treating two different names as if they are only one.

I kept short_name for the out_name, as it is the most widely used (specially in diagnostics) and only use the new `cmor_name` when  working with the tables, which is mostly internal stuff.

bouweandela:
It will lead to a massive rename in the ESMValTool repository as it is now (partly implemented in ESMValGroup/ESMValTool#1493), because we will need to adjust the recipes/diagnostics/cmorizers to the fact that one name is split into two.

Would it lead to less renaming if we keep `short_name` for the variable name in the table and call `out_name` `var_name` so it aligns with iris?

jvegasbsc:
> Would it lead to less renaming if we keep short_name for the variable name in the table and call out_name var_name so it aligns with iris?

In the recipes, yes. In ESMValCore code I think we will have more, but involve less people...

bouweandela:
Shall we plan a short meeting at the workshop to discuss how to go forward with this @jvegasbsc @sloosvel?

sloosvel:
Fine for me!

bouweandela:
Summary of the discussion: instead of the approach here, an extra name, e.g. `cmor_name` will become available in the variable definition section of the recipe to specify the name of the variable in the cmor table if it differs from `short_name`. This name will default to `short_name` if it is not provided.

jvegasbsc:
Replaced by #1099

Issue #91, opened 2019-06-14
For each figure, provide citation information in a standard format


bouweandela:
It would be nice if we could provide citation information along with the figures produced by the diagnostics. This should be in a standard format (e.g bibtex?), for easy use.

The following should be included:
- papers describing the diagnostic and recipe
- model data citation information 
- observational data citation information
- esmvaltool and other scientific software citation
- scientific compute cluster citation information, if applicable

It should be fairly easy to extract the required information from the provenance.

Additional information on CMIP6 model data citation provided by Martina:
- Data Citation is provided on two granularities. The data user has to 
decide which one is appropriate for his/her study. The number of data 
citations in a reference list need to be in balance with paper 
references. Both should be included as two data reference list options. 
The granularities in DRS terms are:
 CMIP6.<mip_era>.<institution_id>.<source_id>
 CMIP6.<mip_era>.<institution_id>.<source_id>.<experiment_id>
Example: https://cera-www.dkrz.de/WDCC/ui/cerasearch/cerarest/exportcmip6?input=CMIP6.CMIP.CNRM-CERFACS.CNRM-CM6-1.historical

- Accessing Data Citation information:
I would recommend to use the API for ESGF data node managers at
https://cera-www.dkrz.de/WDCC/ui/cerasearch/cerarest/cmip6Citations?
with a JSON formatted response and the following AND-connected 
attributes: institutionId, sourceId, complete (true|false), drsId

- Link for information on CMIP6 data citation information: http://cmip6cite.wdc-climate.de

bouweandela:
Regarding the implementation: all required information should be extracted from the recorded provenance. If some extra information is needed, it should also be recorded as provenance.

In principle, this functionality could be a command line tool that reads a provenance file and writes a references file in e.g. bibtex format, though in practice it probably makes more sense to integrate this into the normal `esmvaltool` run. A good way to add this would probably be to create a `DiagnosticTask._write_references_file()` method and calling it just after this line: https://github.com/ESMValGroup/ESMValCore/blob/0cfe7fb12bcacd91a5d908f443bb9be0c474d437/esmvalcore/_task.py#L508

SarahAlidoost:
@mattiarighi: the [config-references.yml](https://github.com/ESMValGroup/ESMValTool/blob/master/esmvaltool/config-references.yml) will be used in the implementation of this issue. In this file, there is an instruction to fill out the entries. 
[...]

PR #402, opened 2019-12-17
Add a method to save citation information

SarahAlidoost:

**Tasks**

-   [x] Add unit tests
-   [ ] ~Public functions should have a numpy-style docstring so they appear properly in the [API documentation](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/api/esmvalcore.html). For all other functions a one line docstring is sufficient.~
-   [ ] ~If writing a new/modified preprocessor function, please update the [documentation](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/esmvalcore/preprocessor.html)~
-   [ ] ~Please use `yamllint` to check that your YAML files do not contain mistakes~
-   [x] If you make backward incompatible changes, make a new pull request in the [ESMValTool repository](https://github.com/ESMValGroup/ESMValTool) and add the link below

Link to corresponding ESMValTool pull request https://github.com/ESMValGroup/ESMValTool/pull/1545

* * *

Closes https://github.com/ESMValGroup/ESMValCore/issues/91


SarahAlidoost:
Please note that failing circleCI tests should be fixed after this ESMValGroup/ESMValCore#452

esmvalcore/_task.py
```diff
@@ -577,11 +582,62 @@ def _collect_provenance(self):
...
+        # collect CMIP6 citation, if any
+        if citation['info_url']:
+            citation['entry'], citation['url'] = _collect_cmip_citation(
+                citation['info_url'])
+        if citation['url']:
+            with open(citation['file'][0], 'w') as file:
+                file.write(citation['url'])
+
+        # collect recipe citation, if any
+        if citation['reference']:
```
Peter9192: What if there are multiple references in the list?

esmvalcore/_task.py
```diff
@@ -593,6 +649,82 @@ def __str__(self):
...
+
+def _json_to_bibtex(data):
+    """Make a bibtex entry from CMIP6 Data Citation json data."""
+    url = ''.join(['https://doi.org/', data['identifier']['id']])
+    author_list = []
+    for item in data['creators']:
+        author_list.append(item['creatorName'])
+    bibtex_entry = ('@misc{' + url + ',\n\t'
+                    'url = {' + url + '},\n\t'
+                    'title = {' + data['titles'][0] + '},\n\t'
```
Peter9192: What happens if any of these fields (like `data['titles']) are missing? Or does that never happen?

esmvalcore/_task.py
```diff
@@ -593,6 +649,82 @@ def __str__(self):
...
+                    'publisher = {' + data['publisher'] + '},\n\t'
+                    'year = ' + data['publicationYear'] + ',\n\t'
+                    'author = {' + ' and '.join(author_list) + '},\n\t'
+                    'doi = {' + data['identifier']['id'] + '},\n'
+                    '}')
+    return bibtex_entry
+
+
+def _replace_entry(product_entry):
+    """Find tags of the references in provenance"""
```
Peter9192:
```suggestion
    """Find tags of the references in provenance."""
```
https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

esmvalcore/_task.py
```diff
@@ -593,6 +649,82 @@ def __str__(self):
...
+    tag_list = []
+    for key in entry_tags.keys():
+        for entry in product_entry:
+            if key in entry and entry_tags[key] not in tag_list:
+                tag_list.append(entry_tags[key])
+    return tag_list
+
+
+def _collect_bibtex_citation(tags):
+    """Collect informtion from bibtex files"""
```
Peter9192:
```suggestion
    """Collect information from bibtex files."""
```

esmvalcore/_config.py
```diff
@@ -27,6 +27,18 @@ def find_diagnostics():
...
+def find_references():
+    """Try to find bibtex files in references folder."""
+    try:
+        import esmvaltool
+    except ImportError:
+        return ''
+    return os.path.join(os.path.dirname(esmvaltool.__file__), 'references')
+
+
+REFERENCES_PATH = find_references()
```
Peter9192: Why not just call `find_references()` in `_task.py` instead of importing `REFERENCES_PATH`? Is this used somewhere else?

bouweandela: I think it's fine to keep the configuration information together like this. At some point in the near future we will probably need to improve the way we configure esmvaltool anyway.

esmvalcore/_task.py
SarahAlidoost: @Peter9192 Thanks. I added a function that test for those fields.

esmvalcore/_task.py
```diff
@@ -577,11 +582,61 @@ def _collect_provenance(self):
...
+            Write citation information provided by the recorded provenance.
+            Recipe and cmip6 data references are saved into one bibtex file.
+            cmip6 data references are provided by CMIP6 data citation service.
+            each cmip6 data reference has a json link. In the case of internet
+            connection, cmip6 data references are saved into a bibtex file.
+            Otherwise, cmip6 data reference links are saved into a text file.
+        """
+        # collect info from provenance
+        file_name = os.path.splitext(product.filename)[0]
+        citation = {
```
bouweandela: It would be more readable to use normal local variables instead of storing everything in a dict

esmvalcore/_task.py
```diff
@@ -577,11 +582,61 @@ def _collect_provenance(self):
...
+        if citation['info_urls']:
+            citation['entry'], citation['url'] = _collect_cmip_citation(
+                citation['info_urls'])
+        if citation['url']:
+            with open(citation['file'][0], 'w') as file:
+                file.write(citation['url'])
+
+        # collect recipe citation, if any
+        if citation['references']:
+            citation['tag'] = _replace_entry(citation['references'])
```
bouweandela: Instead of doing this reverse lookup, it would be easier to just replace the `references` tags with the values after using them to find the associated bibtex file. The content of the bibtex file should then be used to construct a meaningful short description.

esmvalcore/_task.py
```diff
@@ -593,6 +648,92 @@ def __str__(self):
...
+def _get_response(url):
+    """Return information from CMIP6 Data Citation service in json format."""
+    json_data = False
+    if url.lower().startswith('https'):
+        try:
+            response = requests.get(url)
+            if response.status_code == 200:
+                json_data = response.json()
+            else:
+                logger.info('Error in the CMIP json link')
```
bouweandela:
```suggestion
                logger.info('Error in the CMIP json link: %s', url)
```

esmvalcore/_task.py
```diff
@@ -593,6 +648,92 @@ def __str__(self):
...
+        except IOError:
+            logger.info('Error in receiving the CMIP json file')
+    return json_data
+
+
+def _valid_json_data(data):
```
bouweandela: Instead of validating, it might be a bit more robust to just use the `dict.get` method with a default option to retrieve the values from the json. That way you still get something meaningful even if a single field is missing (does this ever happen?)

esmvalcore/_task.py
```diff
@@ -593,6 +648,92 @@ def __str__(self):
...
+        else:
+            logger.info('The reference file %s does not exist.',
+                        bibtex_file)
+    return entry
+
+
+def _collect_cmip_citation(info_url):
+    """Collect information from CMIP6 Data Citation Service."""
+    split_str = 'cmip6?input=CMIP6.CMIP.'
+    url_stem = ''.join([CMIP6_CITATION_URL.split(split_str)[0],
```
bouweandela: It would be more clear to just add another global constant containing url_stem

esmvalcore/_task.py
```diff
@@ -24,6 +25,9 @@
     'mip',
 }
 
+CMIP6_CITATION_URL = 'https://cera-www.dkrz.de/WDCC/ui/cerasearch/' \
+                     'cmip6?input=CMIP6.CMIP.'
```
bouweandela:
```suggestion
                     'cmip6?input=CMIP6.'
```
`CMIP` is the `activity_id`, so should not be hardcoded, or this will not work for other activities than CMIP

esmvalcore/_task.py
```diff
@@ -593,6 +648,92 @@ def __str__(self):
...
+    return valid_data
+
+
+def _json_to_bibtex(data):
+    """Make a bibtex entry from CMIP6 Data Citation json data."""
+    url = ''.join(['https://doi.org/', data['identifier']['id']])
+    author_list = []
+    for item in data['creators']:
+        author_list.append(item['creatorName'])
+    bibtex_entry = ('@misc{' + url + ',\n\t'
```
bouweandela: Use an [f-string](https://docs.python.org/3/reference/lexical_analysis.html#formatted-string-literals) instead of all the plusses

esmvalcore/_task.py
```diff
@@ -593,6 +648,92 @@ def __str__(self):
...
+def _collect_bibtex_citation(tags):
+    """Collect information from bibtex files."""
+    entry = ''
+    for tag in tags:
+        bibtex_file = os.path.join(REFERENCES_PATH, tag + '.bibtex')
+        if os.path.isfile(bibtex_file):
+            with open(bibtex_file, 'r') as file:
+                entry += '{}\n'.format(file.read())
+        else:
+            logger.info('The reference file %s does not exist.',
```
bouweandela: raise an exception instead, this is an error

esmvalcore/_task.py
```diff
@@ -577,11 +582,61 @@ def _collect_provenance(self):
...
+            'file': [
+                file_name + '_data_citation_url.txt',
+                file_name + '_citation.bibtex',
+            ]}
+        for item in product.provenance.records:
+            for key, value in item.attributes:
+                if key.namespace.prefix == 'attribute':
+                    if key.localpart in {'reference', 'references'}:
+                        citation['references'].append(value)
+                    elif key.localpart == 'further_info_url':
```
bouweandela: Instead of the info url, I think it would be better to use the `mip_era`, `institution_id`, `source_id`, and `experiment_id` attributes to construct the url, only if the `project_id` is `'CMIP6'` of course.

esmvalcore/_task.py
SarahAlidoost: Thanks @bouweandela please see changes in new file _citation.py.

SarahAlidoost:
> Nice job! Here are some points for improvement:
> 
> * [x]  Move the code that is added to esmvalcore._task.py to a new file, e.g. esmvalcore/_citation.py
> * [x]  The references are now stored in two places, in config-references.yml and in the bibtex files. This is inconvenient and will lead to errors, as people may forget to add the entry in one of two places. I would remove the entries from config-references.yml and use the bibtex files as the single source.
> * [x]   It is important that the `esmvalcore` package can be used stand-alone as well, i.e. without installing the `esmvaltool` package, so you need to test that it stays working with this pull request.
> * [x]  Some unit testing would be nice, though it does not have to be too much, because I do not foresee that this is a part of the code that is changed often.

@bouweandela please also see the PR https://github.com/ESMValGroup/ESMValTool/pull/1545 for removing the reference section from config-references.yml.

esmvalcore/_citation.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+    # collect info from provenance
+    product_name = os.path.splitext(filename)[0]
+    info_urls = []
+    json_urls = []
+    product_tags = []
+    for item in provenance.records:
+        for key, value in item.attributes:
+            if key.namespace.prefix == 'attribute':
+                if key.localpart in {'reference', 'references'}:
```
bouweandela:
```suggestion
                if str(key.localpart == 'references':
```
and also you need to check that `item` is a diagnostic output file or a recipe, because netcdf input data can also have a 'references' global attribute that may contain some text string which is not a tag/bibtex file. Try running examples/recipe_python.yml to see an example of this.

esmvalcore/_citation.py
```diff
@@ -0,0 +1,208 @@
...
+
+    # save CMIP6 url_info, if any
+    if info_urls:
+        for info_url in info_urls:
+            citation_urls += '{}\n'.format(info_url)
+        with open(f'{product_name}_data_citation_url.txt', 'w') as file:
+            file.write(citation_urls)
+
+    # convert json_urls to bibtex entries
+    if json_urls:
```
bouweandela: there is no need for this `if` statement, the loop below will do nothing if json_urls is an empty list.

esmvalcore/_citation.py
```diff
@@ -0,0 +1,208 @@
...
+    REFERENCES_PATH = ''
+
+logger = logging.getLogger(__name__)
+
+CMIP6_URL_STEM = 'https://cera-www.dkrz.de/WDCC/ui/cerasearch'
+
+# it is the technical overview and should always be cited
+ESMVALTOOL_PAPER_TAG = 'righi19gmdd'
+ESMVALTOOL_PAPER = (
+    '@article{righi19gmdd,\n\t'
```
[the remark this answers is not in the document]
SarahAlidoost: @bouweandela thanks, a tab is needed for contents in the bibtex format.

esmvalcore/_citation.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+def _save_citation_info(product_name, product_tags, json_urls, info_urls):
+    citation_entries = [ESMVALTOOL_PAPER]
+    citation_urls = ''
+
+    # save CMIP6 url_info, if any
+    if info_urls:
+        for info_url in info_urls:
+            citation_urls += '{}\n'.format(info_url)
```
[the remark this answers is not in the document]
SarahAlidoost: @bouweandela thanks, I applied your suggestion.

esmvalcore/_citation.py
SarahAlidoost: @bouweandela now it checks whether an item is related to a diagnostics, recipe or input data with attributes "references".

esmvalcore/_citation.py
SarahAlidoost: @bouweandela , thanks, I applied your suggestion.

SarahAlidoost:
Some commits are related to the PR here https://github.com/ESMValGroup/ESMValTool/pull/1545
one work is still in the progress: the test_recipe.py needs some changes accounting for moving reference lists to bibtex files.

SarahAlidoost:
@bouweandela This PR is now ready to review. I don't think that the failed test because of flake8 is related to the files changed in this PR. Thanks in advance.

esmvalcore/_citation.py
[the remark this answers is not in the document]
bouweandela: Ok, I understand. Could you move the tabs to the beginning of the line that they are on? I think this will be easier to read if you always put newline characters at the end of a line.

esmvalcore/_citation.py
```diff
@@ -0,0 +1,229 @@
...
+    each cmip6 data reference has a json link. In the case of internet
+    connection, cmip6 data references are saved into a bibtex file.
+    Otherwise, cmip6 data reference links are saved into a text file.
+    """
+    product_name = os.path.splitext(filename)[0]
+    info_urls = []
+    json_urls = []
+    product_tags = []
+    for item in provenance.records:
+        reference_attr = item.get_attribute('attribute:' + 'references')
```
bouweandela:
```suggestion
        reference_attr = item.get_attribute('attribute:references')
```

esmvalcore/_citation.py
```diff
@@ -0,0 +1,229 @@
...
+        # get cmip6 citation info
+        value = item.get_attribute('attribute:' + 'mip_era')
```
bouweandela:
```suggestion
        value = item.get_attribute('attribute:mip_era')
```

esmvalcore/_citation.py
```diff
@@ -0,0 +1,229 @@
...
+        if 'CMIP6' in list(value):
```
bouweandela:
```suggestion
        if 'CMIP6' in value:
```
there is no need to convert a `set` to a `list` before you can test if some item is in it

esmvalcore/_citation.py
```diff
@@ -0,0 +1,229 @@
...
+                product_tags += list(reference_attr)
+
+    # get other references information recorded by provenance
+    tags = set(_clean_tags(product_tags + [ESMVALTOOL_PAPER_TAG]))
+    for item in provenance.records:
+        reference_attr = item.get_attribute('attribute:' + 'references')
+        if reference_attr:
+            value = set(_clean_tags(reference_attr))
+            if not value.issubset(tags):
+                info_urls += list(reference_attr)
```
bouweandela:
```suggestion
                product_tags.extend(reference_attr)
            # get diagnostics citation tags
            elif item.get_attribute('attribute:' + 'script_file'):
                product_tags.extend(reference_attr)
            else:
                # get other references information recorded by provenance
                info_urls += list(reference_attr)            
```
I think you could simplify this a bit like this

esmvalcore/_citation.py
```diff
@@ -0,0 +1,229 @@
...
+    # convert json_urls to bibtex entries
+    for json_url in json_urls:
+        cmip_citation = _collect_cmip_citation(json_url)
+        if cmip_citation:
+            citation_entries.append(cmip_citation)
+
+    # convert tags to bibtex entries
+    if REFERENCES_PATH and product_tags:
+        # make tags clean and unique
+        tags = list(set(_clean_tags(product_tags)))
```
bouweandela:
```suggestion
        tags = _clean_tags(product_tags)
```
I would move getting rid of duplicates to the clean tags function

esmvalcore/_citation.py
```diff
@@ -0,0 +1,229 @@
...
+            citation_entries.append(_collect_bibtex_citation(tag))
+
+    with open(f'{product_name}_citation.bibtex', 'w') as file:
+        file.write('\n'.join(citation_entries))
+
+
+def _clean_tags(tags):
+    """Clean the tags that are recorded as str by provenance."""
+    pattern = re.compile(r'\w+')
+    return pattern.findall(str(tags))
```
bouweandela:
```suggestion
    return {pattern.findall(t) for t in tags}
```
I would move making sure there are only unique tags here. For me it's also slightly easier to understand what is going on if you get the tags from the strings individually instead of by converting a list of strings to a string and then looking for tags in that.

esmvalcore/_citation.py
```diff
@@ -0,0 +1,229 @@
...
+def _get_response(url):
+    """Return information from CMIP6 Data Citation service in json format."""
+    json_data = False
+    if url.lower().startswith('https'):
+        try:
+            response = requests.get(url)
+            if response.status_code == 200:
+                json_data = response.json()
+            else:
+                logger.info('Error in the CMIP json link: %s', url)
```
bouweandela:
```suggestion
                logger.warning('Error in the CMIP6 citation link: %s', url)
```
This should not happen unless the data is badly formatted, so I think this should be a warning, right?

esmvalcore/_citation.py
```diff
@@ -0,0 +1,229 @@
...
+        except IOError:
+            logger.info('Error in receiving the CMIP json file')
```
bouweandela:
```suggestion
            logger.info('No network connection, unable to retrieve CMIP6 citation information')
```
would be a bit more clear

esmvalcore/_citation.py
```diff
@@ -0,0 +1,229 @@
...
+        author_list = [
+            item.get('creatorName', '') for item in data['creators']
+        ]
+    if author_list:
+        if author_list[0] == author_list[-1]:
+            authors = author_list[0]
+            if not authors:
+                authors = 'creatorName not found'
+        else:
+            authors = ' and '.join(author_list)
```
bouweandela:
```suggestion
    authors = ' and '.join(author_list)
    if not authors:
         authors = 'Unknown'
```
would be a bit simpler, you can do this because `'xyz'.join([])` returns `''`.

esmvalcore/_citation.py
```diff
@@ -0,0 +1,229 @@
...
+    bibtex_entry = (
+        f'{"@misc{"}{url},\n\t'
+        f'url = {{{url}}},\n\t'
+        f'title = {{{title}}},\n\t'
+        f'publisher = {{{publisher}}},\n\t'
+        f'year = {year},\n\t'
+        f'author = {{{authors}}},\n\t'
+        f'doi = {{{doi}}},\n'
+        f'{"}"}\n'
+    )
```
bouweandela:
```suggestion
    bibtex_entry = textwrap.dedent(
        f"""
        @misc{{{url}
        \turl = {{{url}}},
        \ttitle = {{{title}}},
        \tpublisher = {{{publisher}}},
        \tyear = {year},
        \tauthor = {{{authors}}},
        \tdoi = {{{doi}}},
        }}
        """".lstrip()
    )
```
would be easier to read

esmvalcore/_citation.py
```diff
@@ -0,0 +1,229 @@
...
+    return bibtex_entry
+
+
+def _collect_bibtex_citation(tag):
+    """Collect information from bibtex files."""
+    bibtex_file = REFERENCES_PATH / f'{tag}.bibtex'
+    if bibtex_file.is_file():
+        entry = bibtex_file.read_text()
+    else:
+        logger.info(
```
bouweandela:
```suggestion
        logger.warning(
```
I think this should be a warning, because the bibtex files should be available, right?

esmvalcore/_citation.py
```diff
@@ -0,0 +1,229 @@
...
+    return entry
+
+
+def _collect_cmip_citation(json_url):
+    """Collect information from CMIP6 Data Citation Service."""
+    json_data = _get_response(json_url)
+    if json_data:
+        bibtex_entry = _json_to_bibtex(json_data)
+    else:
+        logger.info('Invalid json link %s', json_url)
```
bouweandela:
```suggestion
```
It looks like an appropriate log message is already generated by the `_get_response` function if something goes wrong, or did I miss something?

esmvalcore/_citation.py
```diff
@@ -0,0 +1,229 @@
...
+        bibtex_entry = False
```
bouweandela:
```suggestion
        bibtex_entry = ''
```
The convention is to always return a value with the same type from a function. In some cases you may want to use the return value to signal that something went wrong, but then the usual choice is to return `None` instead of `False`. So here it would be best to return either an empty string `''` or `None`.

esmvalcore/_citation.py
```diff
@@ -0,0 +1,229 @@
...
+    return json_url
+
+
+def _make_info_url(url_prefix):
+    """Make info url based on CMIP6 Data Citation Service."""
+    info_url = f'{CMIP6_URL_STEM}/cmip6?input=CMIP6.{url_prefix}'
+    return info_url
+
+
+def cite_tag_value(tags):
```
bouweandela: This function should be moved to ` esmvaltool/cmorizers/obs/utilities.py` in the ESMValTool repository, because that is the only place where it is used.

Provided that the information is available, the return value should be a string looking like
```
"Author X., et al., Journal Y, doi:xyz, year."
```
instead of the complete contents of the bibtex file. I think we discussed this, my apologies if that was not clear.

esmvalcore/_provenance.py
```diff
@@ -31,9 +32,12 @@ def create_namespace(provenance, namespace):
...
     provenance = ProvDocument()
-    namespace = 'software'
-    create_namespace(provenance, namespace)
-    attributes = {}  # TODO: add dependencies with versions here
+    for namespace in ('software', 'attribute'):
+        create_namespace(provenance, namespace)
+
+    # TODO: add dependencies with versions here
+    attributes_value = ESMVALTOOL_PAPER_TAG
+    attributes = {'attribute:references': attributes_value}
```
bouweandela:
```suggestion
    attributes = {'attribute:references': ESMVALTOOL_PAPER_TAG}
```
would be a bit shorter

tests/integration/test_recipe.py
```diff
@@ -1289,6 +1289,18 @@ def test_diagnostic_task_provenance(
     assert os.path.exists(prefix + '.svg')
 
 
+def _test_bibtex_files(product_tags):
```
bouweandela: This test should be implemented in the ESMValTool repository instead. It should read all the yaml files in the `esmvaltool/recipes` directory and check that there is a bibtex file available for each reference in the 'references' list in the `documentation` section of the recipes. I think we discussed this, my apologies if that was not clear.

bouweandela:
> I don't think that the failed test because of flake8 is related to the files changed in this PR. 

The failing tests should be solved if you merge the latest version of the master branch into this branch.

tests/integration/test_recipe.py
bouweandela: Could be in `unit/test_recipes.py`, function called `test_bibtex_available` or something similar

esmvalcore/_provenance.py
bouweandela: There is no need for these changes, because you are not using the tag from the activity to collect the citation information.

esmvalcore/_citation.py
bouweandela:
```suggestion
    # tags is a list of strings, the content of each string is a list of tags
    return pattern.findall(str(tags))
```

esmvalcore/_citation.py
bouweandela: Discussed this, will use only DOI for now

tests/integration/test_recipe.py
SarahAlidoost: The test is found here: 
https://github.com/ESMValGroup/ESMValTool/pull/1545/files?file-filters%5B%5D=.py#diff-2266a659a150d71d8451a1568a6948a9

esmvalcore/_citation.py
SarahAlidoost: To stick to bibtex format, I removed the .lstrip() .

bouweandela: Then the bibtex file will start with a newline, which is probably not what you want?

esmvalcore/_citation.py
```diff
@@ -0,0 +1,218 @@
...
+            citation_entries.append(_collect_bibtex_citation(tag))
+
+    with open(f'{product_name}_citation.bibtex', 'w') as file:
+        file.write('\n'.join(citation_entries))
+
+
+def _extract_tags(tags):
+    """
+    Extract tags that are recorded by provenance as
+    for example, "['acknow_project', 'acknow_author']".
```
bouweandela: This docstring doesn't follow the recommended format for docstrings, see [pep257](https://www.python.org/dev/peps/pep-0257/). Could you have a look? Prospector should also warn you about this. I'm a bit surprised this prospector warning doesn't show up on Codacy..

esmvalcore/_citation.py
```diff
@@ -0,0 +1,218 @@
...
+    """
+    pattern = re.compile(r'\w+')
+    return list(set(pattern.findall(str(tags))))
+
+
+def _get_response(url):
+    """Return information from CMIP6 Data Citation service in json format."""
+    json_data = False
```
bouweandela:
```suggestion
    json_data = None
```
or `{}` would be better, because then the function has a consistent return type

esmvalcore/_citation.py
```diff
@@ -0,0 +1,218 @@
...
+def _json_to_bibtex(data):
+    """Make a bibtex entry from CMIP6 Data Citation json data."""
+    url = 'url not found'
+    title = data.get('titles', ['title not found'])[0]
+    publisher = data.get('publisher', 'publisher not found')
+    year = data.get('publicationYear', 'publicationYear not found')
+    authors = 'creators not found'
+    doi = 'doi not found'
+
+    if data.get('creators', ''):
```
bouweandela:
```suggestion
    if data.get('creators'):
```
because the default return value of `dict.get` is `None`, which also evaluates to `False` in the `if` statement. Maybe writing `if 'creators' in data` would be even more clear, since the case where `creators` is an empty list is already covered on line 147.

esmvalcore/_citation.py
```diff
@@ -0,0 +1,218 @@
...
+        author_list = [
+            item.get('creatorName', '') for item in data['creators']
+        ]
+        authors = ' and '.join(author_list)
+        if not authors:
+            authors = 'creators not found'
+
+    if data.get('identifier', ''):
+        doi = data.get('identifier').get('id', 'doi not found')
```
bouweandela:
```suggestion
    if 'identifier' in data:
        doi = data['identifier'].get('id', 'doi not found')
```
would be a bit more clear I think, or do you think it's less robust?

SarahAlidoost: @bouweandela thanks. I applied your suggestion. I think the usage of get in if statement wasn't correct.

esmvalcore/_citation.py
SarahAlidoost: I run prospector locally and got the errors. thanks.

esmvalcore/_citation.py
```diff
@@ -0,0 +1,217 @@
...
+            # get diagnostics citation tags
+            elif item.get_attribute('attribute:script_file'):
+                product_tags.extend(reference_attr)
+            else:
+                info_urls += list(reference_attr)
+
+    _save_citation_info(product_name, product_tags, json_urls, info_urls)
+
+
+def _save_citation_info(product_name, product_tags, json_urls, info_urls):
```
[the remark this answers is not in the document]
SarahAlidoost: @bouweandela thanks. There are now two functions.

bouweandela:
We should probably also make a new release for this.

esmvalcore/_citation.py
```diff
@@ -0,0 +1,244 @@
...
+from ._config import DIAGNOSTICS_PATH
+
+logger = logging.getLogger(__name__)
+
+REFERENCES_PATH = DIAGNOSTICS_PATH / 'references'
+
+CMIP6_URL_STEM = 'https://cera-www.dkrz.de/WDCC/ui/cerasearch'
+
+# The technical overview paper should always be cited
+ESMVALTOOL_PAPER = (
```
mattiarighi: This is now published in final form, please update the reference to `righi20gmd`, see https://www.geosci-model-dev.net/13/1179/2020/.

mattiarighi:
`recipe_permetrics_CMIP5.yml` runtimes:
- in `master`: [25:36.240417, 27:53.705252, 27:59.009339, 25:46.094682, 25:23.925156]; **median = 25:46.094682**.
- in `save_citation`: [25:40.530607, 25:36.464112, 26:06.032161, 25:30.391732, 25:20.122022]; **median = 25:36.464112**.

No difference.

esmvalcore/_citation.py
bouweandela: Done

PR #416, opened 2020-01-03
Update documentation so it uses the master branch by default

### LINKED ISSUE #115: Simplify branches
Currently we have two branches

`development` -> this is where we work
`master` -> we never use this, it is identical to the latest tagged version

I would propose to start working in the `master` branch instead and remove the `development` branch, because we never use the current `master` branch. This is what most open source projects do.

If you want to see the code of the released version, you can just get the tag, i.e. `git checkout v2.0.0a1`, or download the zipped source code from the [releases page](https://github.com/ESMValGroup/ESMValCore/releases).

bouweandela:
Closes #115

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -201,7 +201,6 @@ workflows:
           filters:
             branches:
               only:
-                - development
                 - master
     jobs:
       - test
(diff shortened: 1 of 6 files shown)


mattiarighi:
Just noticed that this should have been merged into master...

bouweandela:
Haha, you're right

Issue #212, opened 2019-08-28
Preprocessor functions that made checker fail


jvegasbsc:
In pull request #87 I added the option to compute variances for several time periods. That functionality made the checker crash because units were squared.

The question is, we want to allow this kind of functions on the preprocessor? If that's true, when we want to apply the data checker?

Other (potential) preprocessor functions that can be affected:

- Corner cases in regrid than can make values go outside `valid_min` and `valid_max` range
- Anomalies: can make values go outside `valid_min` and `valid_max` range
- Standarized anomalies ( anomalies / std ): they don't have units

bouweandela:
We could also consider merging the function `cmor_check_data` with `cmor_check_metadata`, since we do not actually check the data anyway. That would also solve this problem, since `cmor_check_metadata` is run right after loading the data. @jvegasbsc @mattiarighi Do you think this could be an acceptable solution?

PR #418, opened 2020-01-03
Remove derived variable gtfgco2

bouweandela:
Because it can be computed from fgco2 with the area_statistics preprocessor, see https://github.com/ESMValGroup/ESMValCore/pull/83#issuecomment-551676316

**Tasks**

-   If writing a new/modified preprocessor function, please update the [documentation](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/esmvalcore/preprocessor.html) -> I don't think this is documented anywhere
-   If you make backward incompatible changes to the recipe format, make a new pull request in the [ESMValTool repository](https://github.com/ESMValGroup/ESMValTool) and add the link below -> The only recipe where this derived variable is mentioned is [recipe_ocean_bgc.yml](https://github.com/ESMValGroup/ESMValTool/blob/19262286ef14fa84acf3a0dc09ac0bc739d9c464/esmvaltool/recipes/recipe_ocean_bgc.yml#L225), but it is commented out.

* * *

Partially addresses https://github.com/ESMValGroup/ESMValCore/issues/42


ledm:
I'm just in the process of preparing a PR for the GMD paper (https://github.com/ESMValGroup/ESMValTool/pull/1621) and I need to re-open this issue (sorry). I  realise that I'm a bit late to the party here, but it's still not strictly possible to fully replace this derived variable with a preprocessor at the moment.  (Similarly for the global total integrated primary production derived variable, which is not mentioned here but also needed for this publication.)

The `area_statistics` preprocessor sum operator multiplies and sums the data and the area, and this process change the units. In this case, the units change from from `kg m-2 s-1` in fgco2 to `kg s-1` for the global total dervied variable gtfgco2, which is then converted into `Pg/year`.

Even now, it's not clear that there's a solution where preprocessors can change the dimensions of the output units. @valeriupredoi like to bring up Iris's `cube.convert_units(new_units)` function, but this function will be able to change the dimensionality of the units, and it is not sensible for it to do so either.  Also the preprocessed version may only use the units listed in the CMOR tables, but the  derived output netcdf can have any units.

Can we annul this PR and re-instate the global total fgco2?

ledm:
Instead of annulling this PR, is there a solution where area_statistics sum can change the cube's units?

valeriupredoi:
no, I don't think it's a good idea to revert this PR. Instead, maybe you can create a preprocessor that calculates a flux like in the derivation script that got deleted but rather using dask arrays for laziness (that may be useful for other variables and being a preprocessing function preserves a good degree of generality)? :beer:

ledm:
@valeriupredoi, I think that you're missing the point here. The `area_statistics` sum operation calculates the area-weighted sum. This means that it multiplies the value, either a concentration [per cubic meter] or a flux [per square meter] by the area [square meters]. This changes the units, which isn't reflected by the preprocessor at the moment.  

This means that every preprocessor that uses the `area_statistics` sum operation with fx files is scientifically incorrect. However, if you attempt to change the units in a preprocessor, the netCDF is no longer CMOR compliant. This is why we needed a global total air sea flux of CO2 derived variable.

bouweandela:
@ledm Changing the units shouldn't be that big of a problem, so long as you do it after the cmor checks are performed, so I guess you could work around issues #212 and #505 by using a custom order for now. Would that be an option? See also discussion in #212.

PR #351, opened 2019-10-31
Fix regrid time function to add year yr frequency and remove resetting calendars

valeriupredoi:
Before you start, please read [CONTRIBUTING.md](https://github.com/ESMValGroup/ESMValTool/blob/version2_development/CONTRIBUTING.md): yadayada, I did, I bloody co-wrote that!

**Tasks**

-   [x] [Create an issue] Created #339 by @ledm  
-   [x] This pull request has a descriptive title that can be used in a changelog
-   [x] Add unit tests
-   [x] Public functions should have a numpy-style docstring so they appear properly in the [API documentation](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/api/esmvalcore.html). For all other functions a one line docstring is sufficient.
-   [x] If writing a new/modified preprocessor function, please update the [documentation](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/esmvalcore/preprocessor.html)
-   [x] Circle/CI tests pass. Status can be seen below your pull request. If the tests are failing, click the link to find out why.
-   [x] Codacy code quality checks pass. Status can be seen below your pull request. If there is an error, click the link to find out why. If you suspect Codacy may be wrong, please ask by commenting.
-   [x] Please use `yamllint` to check that your YAML files do not contain mistakes
-   [ ] If you make backward incompatible changes to the recipe format, make a new pull request in the [ESMValTool repository](https://github.com/ESMValGroup/ESMValTool) and add the link below

If you need help with any of the tasks above, please do not hesitate to ask by commenting in the issue or pull request.

* * *

Closes #339 

@ledm here you go, bud! Cheers for raising that, I did remove completely the calendar resetting in `regrid_tim` since that was total bollocks and was actuallt resetting calendars and changing the time axis in a nasty way; no need to worry about cubes with different calendars since after the CMOR checks all cubes will have the same calendar :beer:

PR #420, opened 2020-01-08
Remove automatic addition of time auxiliary coordinates

jvegasbsc:
Remove auxiliary coordinates

**Tasks**

-   [x] [Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do, if you haven't done so already (and add the link at the bottom)
-   [x] This pull request has a descriptive title that can be used in a changelog
-   [x] Add unit tests
-   [x] Public functions should have a numpy-style docstring so they appear properly in the [API documentation](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/api/esmvalcore.html). For all other functions a one line docstring is sufficient.
-   [x] If writing a new/modified preprocessor function, please update the [documentation](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/esmvalcore/preprocessor.html)
-   [x] Circle/CI tests pass. Status can be seen below your pull request. If the tests are failing, click the link to find out why.
-   [x] Codacy code quality checks pass. Status can be seen below your pull request. If there is an error, click the link to find out why. If you suspect Codacy may be wrong, please ask by commenting.
-   [x] Please use `yamllint` to check that your YAML files do not contain mistakes
-   [x] If you make backward incompatible changes to the recipe format, make a new pull request in the [ESMValTool repository](https://github.com/ESMValGroup/ESMValTool) and add the link below

If you need help with any of the tasks above, please do not hesitate to ask by commenting in the issue or pull request.

* * *

Closes #195

PR #424, opened 2020-01-13
Fix missing refs after 351 merge

valeriupredoi:
merging #351 missed a whole lot of refs  that I re-added here
A good example not to trust 100% gitHub when it tells you able to merge and do a `git merge master` in the branch before; that was an old PR that was initially pointing to `development` and the able to merge message has not been redirected towards `master` @mattiarighi can you pls approve so we can fix the `master` (currently the tests are failing there)

### DIFF
--- a/esmvalcore/preprocessor/_time.py
+++ b/esmvalcore/preprocessor/_time.py
@@ -199,14 +199,13 @@ def daily_statistics(cube, operator='mean'):
     iris.cube.Cube
         Daily statistics cube
     """
-    iris.coord_categorisation.add_day_of_year(cube, 'time')
-    iris.coord_categorisation.add_year(cube, 'time')
+    if not cube.coords('day_of_year'):
+        iris.coord_categorisation.add_day_of_year(cube, 'time')
+    if not cube.coords('year'):
+        iris.coord_categorisation.add_year(cube, 'time')
 
     operator = get_iris_analysis_operation(operator)
     cube = cube.aggregated_by(['day_of_year', 'year'], operator)
-
-    cube.remove_coord('day_of_year')
-    cube.remove_coord('year')
     return cube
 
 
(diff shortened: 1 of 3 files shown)


bouweandela:
@valeriupredoi Can you please give this pull request a meaningful title that I can use in the changelog? It looks like you're fixing some incompatibility between functionality you added in #351 and #420 that got merged before  #351.

valeriupredoi:
indeed it was a compatibility issue that should have been highlighted as conflict; Ill change it now, but also note that there is actually missing code that was not pushed to master in the test module

valeriupredoi:
ow hang on, not the case; from #420 the add aux coord func has been removed from the test module and is actually needed by the tests I introduced

esmvalcore/preprocessor/_time.py
```diff
@@ -199,8 +199,10 @@ def daily_statistics(cube, operator='mean'):
     iris.cube.Cube
         Daily statistics cube
     """
-    iris.coord_categorisation.add_day_of_year(cube, 'time')
-    iris.coord_categorisation.add_year(cube, 'time')
+    if not cube.coords('day_of_year'):
+        iris.coord_categorisation.add_day_of_year(cube, 'time')
+    if not cube.coords('year'):
+        iris.coord_categorisation.add_year(cube, 'time')
```
valeriupredoi: this is from #420 but I don't agree with the change from there!

tests/unit/preprocessor/_time/test_time.py
```diff
@@ -762,6 +773,7 @@ def setUp(self):
             ),
             0,
         )
+        add_auxiliary_coordinate([self.cube_1, self.cube_2])
```
valeriupredoi: the tests need aux coords for testing so I suggest reintroducing them

PR #427, opened 2020-01-14
Log max memory used by the recipe in the main log

jvegasbsc:
In the last few days I had to look several time in the resource logger for the maximum consumed memory. It is tedious.

I modified a bit the resource logger so it prints at the end that value in the main log, to make getting this value easier for all. 

I am not very happy with the implementation, so any suggestion on how to improve it is even more welcome than usual

### DIFF
--- a/esmvalcore/_task.py
+++ b/esmvalcore/_task.py
@@ -120,8 +120,8 @@ def resource_usage_logger(pid, filename, interval=1, children=True):
                     logger.info(
                         'Maximum memory used (estimate): %.1f GB', max_mem)
                     logger.info(
-                        'Sampled every second. May be inaccurate if short but '
-                        'high spikes in memory consumption are present.')
+                        'Sampled every second. It may be inaccurate if short but '
+                        'high spikes in memory consumption occur.')
                     return
 
     thread = threading.Thread(target=_log_resource_usage)
--- a/esmvalcore/_task.py
+++ b/esmvalcore/_task.py
@@ -48,12 +48,13 @@ def _get_resource_usage(process, start_time, children=True):
         'Disk write (GB)',
     ]
     fmt = '{}\t' * len(entries[:-1]) + '{}\n'
-    yield fmt.format(*entries)
+    yield (fmt.format(*entries), 0.)
 
     # Compute resource usage
     gigabyte = float(2**30)
     precision = [1, 1, None, 1, None, 3, 3]
     cache = {}
+    max_memory = 0.
     try:
         process.io_counters()
     except AttributeError:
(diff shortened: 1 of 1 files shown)


valeriupredoi:
can use memory profiler https://pypi.org/project/memory-profiler/ for nicer and more info :beer:

valeriupredoi:
but am quite OK with how it's done now, why you unhappy @jvegasbsc

jvegasbsc:
>but am quite OK with how it's done now, why you unhappy @jvegasbsc
I am not quite pleased with my hack to print it, the rest of the code is ok

valeriupredoi:
our sampling frequency is high enough to catch any memory spike no? If the RES memory is going through the roof for a very short period of time that may be to do with the machine not the program running on it

jvegasbsc:
> This way of retrieving the maximum memory used can be very inaccurate because of the sampling frequency.

I am happy with a coarse option. I was just looking for a way to quickly tweak the MAX_MEMORY for the job scheduler or to give a rough estimate of how much resources you need to run a recipe

bouweandela:
> I am happy with a coarse option. I was just looking for a way to quickly tweak the MAX_MEMORY for the job scheduler or to give a rough estimate of how much resources you need to run a recipe

@jvegasbsc In that case, can you please add a message that this may not be accurate since it's sampled at a 1 second rate?

> our sampling frequency is high enough to catch any memory spike no? If the RES memory is going through the roof for a very short period of time that may be to do with the machine not the program running on it

@valeriupredoi No, this typically happens just before the program crashes because of lack of memory

bouweandela:
Looks good to me now. @mattiarighi can you please test?

esmvalcore/_task.py
```diff
@@ -110,10 +112,16 @@ def _log_resource_usage():
...
-            for msg in _get_resource_usage(process, start_time, children):
+            for msg, max_mem in _get_resource_usage(process, start_time,
+                                                    children):
                 file.write(msg)
                 time.sleep(interval)
                 if halt.is_set():
+                    logger.info(
+                        'Maximum memory used (estimate): %.1f GB', max_mem)
+                    logger.info(
+                        'Sampled every second. May be inaccurate if short but '
```
mattiarighi:
```suggestion
                        'Sampled every second. It may be inaccurate if short but '
```

esmvalcore/_task.py
```diff
@@ -110,10 +112,16 @@ def _log_resource_usage():
...
+                        'high spikes in memory consumption are present.')
```
mattiarighi:
```suggestion
                        'high spikes in memory consumption occur.')
```

PR #433, opened 2020-01-16
[URGENT] Rename zonal_means to zonal_statistics

### LINKED ISSUE #24: Zonal and meridional statistics preprocessor

mattiarighi:
In Table 1 of the paper, we mention `zonal_statistics` and `meridional_statistics`, but in the preprocessor this is called `zonal_means` and covers both the zonal (longitude stats) and the meridional (latitude stats) case.

This PR attempts to quickly solve the problem.

`recipe_flato13ipcc.yml` and `recipe_collins13ipcc.yml` should be adjusted accordingly.

Closes #24.

### DIFF
--- a/doc/esmvalcore/preprocessor.rst
+++ b/doc/esmvalcore/preprocessor.rst
@@ -987,7 +987,7 @@ See also :func:`esmvalcore.preprocessor.extract_shape`.
 
 
 ``zonal_statistics``
----------------
+--------------------
 
 The function calculates the zonal statistics by applying an operator along the longitude coordinate. This function takes one argument:
 
@@ -997,7 +997,7 @@ See also :func:`esmvalcore.preprocessor.zonal_means`.
 
 
 ``meridional_statistics``
----------------
+-------------------------
 
 The function calculates the meridional statistics by applying an operator along the latitude coordinate. This function takes one argument:
 
--- a/doc/esmvalcore/preprocessor.rst
+++ b/doc/esmvalcore/preprocessor.rst
@@ -916,8 +916,9 @@ The area manipulation module contains the following preprocessor functions:
 * extract_named_regions_: Extract a specific region from in the region
   cooordinate.
 * extract_shape_: Extract a region defined by a shapefile.
-* zonal_means_: Calculates the zonal or meridional means.
-* area_statistics_: Calculates the average value over a region.
+* zonal_statistics_: Calculates zonal statistics.
+* meridional_statistics_: Calculates meridional statistics.
+* area_statistics_: Calculates area statistics.
 
 
 ``extract_region``
@@ -985,18 +986,24 @@ Examples:
 See also :func:`esmvalcore.preprocessor.extract_shape`.
 
 
-``zonal_means``
+``zonal_statistics``
 ---------------
 
-The function calculates the zonal or meridional means. While this function is
-named ``zonal_mean``, it can be used to apply several different operations in
-an zonal or meridional direction. This function takes two arguments:
+The function calculates the zonal statistics by applying an operator along the longitude coordinate. This function takes one argument:
 
-* ``coordinate``: Which direction to apply the operation: latitude or longitude
-* ``mean_type``: Which operation to apply: mean, std_dev, variance, median,
-  min, max or sum
+* ``operator``: Which operation to apply: mean, std_dev, median, min, max or sum
 
 See also :func:`esmvalcore.preprocessor.zonal_means`.
+
+
+``meridional_statistics``
+---------------
+
+The function calculates the meridional statistics by applying an operator along the latitude coordinate. This function takes one argument:
+
+* ``operator``: Which operation to apply: mean, std_dev, median, min, max or sum
+
+See also :func:`esmvalcore.preprocessor.meridional_means`.
 
 
 ``area_statistics``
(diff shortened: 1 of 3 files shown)


ledm:
If we really need to get this pushed through urgently, I'd recommend removing the merdional statistics preprocessor altogether (from here and from the paper), and then we need to put an raise fatal error in the `zonal_statistics` when the latitude data is two dimensional.

mattiarighi:
There are probably other preprocessor features that might have problems with irregular grids (like NE masking), so I do not think this is a reason against the renaming.

mattiarighi:
> If we really need to get this pushed through urgently, I'd recommend removing the merdional statistics preprocessor altogether (from here and from the paper), and then we need to put an raise fatal error in the zonal_statistics when the latitude data is two dimensional.

Fine with me.

Note however that the two functions are basically identical, they only differ by the target coordinate.

ledm:
> Note however that the two functions are basically identical, they only differ by the target coordinate.


Unfortunately, the cell descriptions are fundamentally different in the  latitude and longitude directions and a function that works in one direction won't work in the other.

mattiarighi:
> Unfortunately, the cell descriptions are fundamentally different in the latitude and longitude directions and a function that works in one direction won't work in the other.

Yes, but `zonal_means` was designed to work with both, by taking `coordinate` as an argument.
What I'm doing here is just to rename `_means` to `_statistics` and to split the function into `zonal` and `meridional`, since by definition zonal refers to the longitude coordinate only.

mattiarighi:
- Option 1: leave everything as is and remove `zonal_statistics` and `meridional_statistics` from Table 1.
- Option 2: just rename `zonal_means` to `zonal_statistics` and remove `meridional_statistics` from Table 1.
- Option 3: this PR, split `zonal_means` into `zonal_statistics` and `meridional_statistics` to be consistent with Table 1.

I'm not happy with 1, since the functionality is there and is used by some diagnostics, so it should be documented. 
Option 2 is not fully consistent, since zonal statistics means "along longitude".

ledm:
From your list, my preference is option 2. 

I guess what I'm trying to say that that both these functions are wrong and neither will work as intended in the real world. The calculations will output invalid numbers, or break, except for in the case of a zonal mean of an dataset with a regular grid.

Most CMIP6 ocean models have iregular grids (ie 2d arrays for latitude and longitude coordinates.) Iris isn't clever enough to do these calculations along the latitude or longitude directions. It is only able to do the calculation in the direction of the cube dimensions (x, y, or z). These directions do not coincide with the latitude or longitude direction. It's a non-trivial calculation to produce the mean of an irregular grid in the latitude or longitude direction. (My hack was to re-grid the data to a regular grid before running this calculation.)

In theory, both meridional and zonal means should require the fx data for volume. However, in an regular grid, we're lucky that the cell volume is the same for a given cell depth for a given latitude band. This means that we can calculate the zonal mean without the cell volume (or more correctly: the zonal cross section). This isn't the case for the meridional direction, as the cell volume changes in the North-South direction, so the meridional mean will always need the cell volume as a weight. 

Hope this makes sense - I spend quite a bit of time working on this issue and wasn't able to come to a satisfying solution.

mattiarighi:
Thanks for clarifying!

> I guess what I'm trying to say that that both these functions are wrong and neither will work as intended in the real world. The calculations will output invalid numbers, or break, except for in the case of a zonal mean of an dataset with a regular grid.

In the two recipes where this function is used, the data are first regridded to a `1x1` grid.

What about option 3 with a check on grid type and error if irregular?

mattiarighi:
Something similar to what is done in `_mask.py`:
https://github.com/ESMValGroup/ESMValCore/blob/a0552bd1a33108156d9453ddf884e21087932150/esmvalcore/preprocessor/_mask.py#L156-L166

valeriupredoi:
yes, placing that check would probably the best bet (for longer term too)

mattiarighi:
Done.

axel-lauer:
I am in favor of Mattia's option 3 with an additional error message in case of irregular grids. No need to trash potentially useful preprocessor functions just because they do not work on every single grid yet.

mattiarighi:
> looks good tome, who's gonna change the preprocessors in recipes?

Guess who... :unamused:

mattiarighi:
Thanks everybody for the quick response! :+1:

valeriupredoi:
> > looks good tome, who's gonna change the preprocessors in recipes?
> 
> Guess who... unamused

I can help if needed (please say no) :grin:

mattiarighi:
No worries :smile:, already done https://github.com/ESMValGroup/ESMValTool/pull/1491

PR #439, opened 2020-01-21
Apply select preprocessor to area, volume and zonal statistics fx variables and add use of fx_files as list of dictionaries

valeriupredoi:
Before you start, please read [CONTRIBUTING.md](https://github.com/ESMValGroup/ESMValCore/blob/master/CONTRIBUTING.md).

**Tasks**

-   [x] [Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do, if you haven't done so already (and add the link at the bottom)
-   [x] This pull request has a descriptive title that can be used in a changelog
-   [x] Add unit tests
-   [x] Public functions should have a numpy-style docstring so they appear properly in the [API documentation](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/api/esmvalcore.html). For all other functions a one line docstring is sufficient.
-   [x] If writing a new/modified preprocessor function, please update the [documentation](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/esmvalcore/preprocessor.html)
-   [x] Circle/CI tests pass. Status can be seen below your pull request. If the tests are failing, click the link to find out why.
-   [x] Codacy code quality checks pass. Status can be seen below your pull request. If there is an error, click the link to find out why. If you suspect Codacy may be wrong, please ask by commenting.
-   [x] Please use `yamllint` to check that your YAML files do not contain mistakes
-   [ ] If you make backward incompatible changes to the recipe format, make a new pull request in the [ESMValTool repository](https://github.com/ESMValGroup/ESMValTool) and add the link below

If you need help with any of the tasks above, please do not hesitate to ask by commenting in the issue or pull request.

* * *

Closes #436 #440

PR #446, opened 2020-01-23
Fix area tiling issue with fx files in area_statistics, added calculate_grid option, added several unit tests to area statistics, clarified several exceptions in area_statistics

### LINKED ISSUE #445: Tiling problem with fx files in `area_statistics`
There are a few issues with `area_statistics`. 
- [ ] The tiling of the areacello/a data was done by numpy, which is too memory heavy. It needs to be replaced by a dask version.
- [ ] There is no option to try to calculate the cell areas. (This is neccesairy for observational datasets, sometimes as FX files are rarely available for these datasets.)
- [ ] There is no way to calculate a weightless mean or a weightless sum. Issue described in more detail in #442

ledm:

**Tasks**

-   [x] Add unit tests

* * *

Closes #445

### DIFF
--- a/esmvalcore/preprocessor/_area.py
+++ b/esmvalcore/preprocessor/_area.py
@@ -152,7 +152,7 @@ def meridional_statistics(cube, operator):
         raise ValueError(msg)
 
 
-def tile_grid_areas(cube, fx_files):
+def load_fx(cube, fx_files):
     """
     Tile the grid area data to match the dataset cube.
 
@@ -177,22 +177,42 @@ def tile_grid_areas(cube, fx_files):
             fx_cube = iris.load_cube(fx_file)
 
             grid_areas = fx_cube.core_data()
+    return grid_areas
 
-            if cube.ndim == 4 and grid_areas.ndim == 2:
-                tiled = da.array([grid_areas for i in range(cube.shape[1])])
-                grid_areas = da.stack(tiled, axis=0)
-                tiled = da.array([grid_areas for i in range(cube.shape[0])])
-                grid_areas = da.stack(tiled, axis=0)
-            elif cube.ndim == 4 and grid_areas.ndim == 3:
-                tiled = da.array([grid_areas for i in range(cube.shape[0])])
-                grid_areas = da.stack(tiled, axis=0).compute()
-            elif cube.ndim == 3 and grid_areas.ndim == 2:
-                tiled = da.array([grid_areas for i in range(cube.shape[0])])
-                grid_areas = da.stack(tiled, axis=0).compute()
-            else:
-                raise ValueError('Grid and dataset number of dimensions not '
-                                 'recognised: {} and {}.'
-                                 ''.format(cube.ndim, grid_areas.ndim))
+
+def tile_grid_areas(cube, grid_areas):
+    """
+    Tile the grid area data to match the dataset cube.
+
+    Parameters
+    ----------
+    cube: iris.cube.Cube
+        input cube.
+    fx_files: dict
+        dictionary of field:filename for the fx_files
+
+    Returns
+    -------
+    iris.cube.Cube
+        Freshly tiled grid areas cube.
+    """
+    if cube.ndim == grid_areas.ndim:
+        return grid_areas
+    if cube.ndim == 4 and grid_areas.ndim == 2:
+        tiled = da.array([grid_areas for i in range(cube.shape[1])])
+        grid_areas = da.stack(tiled, axis=0)
+        tiled = da.array([grid_areas for i in range(cube.shape[0])])
+        grid_areas = da.stack(tiled, axis=0)
+    elif cube.ndim == 4 and grid_areas.ndim == 3:
+        tiled = da.array([grid_areas for i in range(cube.shape[0])])
+        grid_areas = da.stack(tiled, axis=0).compute()
+    elif cube.ndim == 3 and grid_areas.ndim == 2:
+        tiled = da.array([grid_areas for i in range(cube.shape[0])])
+        grid_areas = da.stack(tiled, axis=0).compute()
+    else:
+        raise ValueError('Grid and dataset number of dimensions not '
+                         'recognised: {} and {}.'
+                         ''.format(cube.ndim, grid_areas.ndim))
     return grid_areas
 
 
@@ -267,17 +287,38 @@ def area_statistics(cube, operator, fx_files=None, calculate_grid=False):
     if not operator_accept_weights(operator):
         return cube.collapsed(['longitude', 'latitude'], operation)
 
+    # The following are for operators with means
     if not fx_files and cube.coord('latitude').points.ndim == 2:
         logger.error(
-            'fx_file needed to calculate grid cell area for irregular grids.')
+            'fx_file needed for weighted calcuions on irregular grids.')
         raise iris.exceptions.CoordinateMultiDimError(cube.coord('latitude'))
 
-    if not calculate_grid:
-        grid_areas = tile_grid_areas(cube, fx_files)
-    else:
+    if fx_files and calculate_grid:
+        logger.error(
+            'Cannot calculate_grid when FX file is provided.')
+        raise ValueError('Incompatible recipe arguments (area_statistics): '
+                         'fx_files and calculate_grid do not work together.')
+
+    # Load grid area
+    grid_areas_loaded = False
+    if fx_files:
+        grid_areas = load_fx(cube, fx_files)
+        grid_areas_loaded = True
+
+    if calculate_grid:
         cube = guess_bounds(cube, ['longitude', 'latitude'])
         grid_areas = iris.analysis.cartography.area_weights(cube)
         logger.info('Calculated grid area shape: %s', grid_areas.shape)
+        grid_areas_loaded = True
+
+    if not grid_areas_loaded:
+        logger.error(
+            'Cannot calculate_grid without Grid area data. '
+            'Please provide fx_files or set calculate_grid: True.')
+        raise ValueError('Cannot calculate_grid without Grid area data.')
+
+    # Tile grid area to match cube
+    grid_areas = tile_grid_areas(cube, grid_areas)
 
     if cube.shape != grid_areas.shape:
         raise ValueError('Cube shape ({}) doesn`t match grid area shape '
--- a/esmvalcore/preprocessor/_area.py
+++ b/esmvalcore/preprocessor/_area.py
@@ -225,6 +225,10 @@ def area_statistics(cube, operator, fx_files=None, calculate_grid = False):
     +------------+--------------------------------------------------+
     | `max`      | Maximum value                                    |
     +------------+--------------------------------------------------+
+    | `weightless_mean` | Mean without weighting by area            |
+    +------------+--------------------------------------------------+
+    | `weightless_sum` | Calculate the Sum without weighing by area |
+    +------------+--------------------------------------------------+
 
     Parameters
     ----------
--- a/esmvalcore/preprocessor/_area.py
+++ b/esmvalcore/preprocessor/_area.py
@@ -287,7 +287,7 @@ def area_statistics(cube, operator, fx_files=None, calculate_grid=False):
     # The following are for operators with means
     if not fx_files and cube.coord('latitude').points.ndim == 2:
         logger.error(
-            'fx_file needed for weighted calcuions on irregular grids.')
+            'fx_file needed for weighted calculations on irregular grids.')
         raise iris.exceptions.CoordinateMultiDimError(cube.coord('latitude'))
 
     if fx_files and calculate_grid:
--- a/esmvalcore/preprocessor/_area.py
+++ b/esmvalcore/preprocessor/_area.py
@@ -323,7 +323,7 @@ def area_statistics(cube, operator, fx_files=None, calculate_grid=False):
 
     if isinstance(grid_areas, da.Array):
         grid_areas = grid_areas.compute()
-        
+
     return cube.collapsed(['longitude', 'latitude'], operation,
                           weights=grid_areas)
 
--- a/esmvalcore/preprocessor/_area.py
+++ b/esmvalcore/preprocessor/_area.py
@@ -202,7 +202,7 @@ def tile_grid_areas(cube, grid_areas):
         tiled = da.array([grid_areas for i in range(cube.shape[1])])
         grid_areas = da.stack(tiled, axis=0)
         tiled = da.array([grid_areas for i in range(cube.shape[0])])
-        grid_areas = da.stack(tiled, axis=0)
+        grid_areas = da.stack(tiled, axis=0).compute()
     elif (cube.ndim, grid_areas.ndim) in [(4, 3), (3, 2)]:
         tiled = da.array([grid_areas for i in range(cube.shape[0])])
         grid_areas = da.stack(tiled, axis=0).compute()
@@ -321,6 +321,7 @@ def area_statistics(cube, operator, fx_files=None, calculate_grid=False):
         raise ValueError('Cube shape ({}) doesn`t match grid area shape '
                          '({})'.format(cube.shape, grid_areas.shape))
 
+    if isinstance(grid_areas, da.Array):
     return cube.collapsed(['longitude', 'latitude'], operation,
                           weights=grid_areas)
 
--- a/esmvalcore/preprocessor/_area.py
+++ b/esmvalcore/preprocessor/_area.py
@@ -175,7 +175,7 @@ def tile_grid_areas(cube, fx_files):
                 continue
             logger.info('Attempting to load %s from file: %s', key, fx_file)
             fx_cube = iris.load_cube(fx_file)
-             
+
             grid_areas = fx_cube.core_data()
 
             if cube.ndim == 4 and grid_areas.ndim == 2:
@@ -197,7 +197,7 @@ def tile_grid_areas(cube, fx_files):
 
 
 # get the area average
-def area_statistics(cube, operator, fx_files=None, calculate_grid = False):
+def area_statistics(cube, operator, fx_files=None, calculate_grid=False):
     """
     Apply a statistical operator in the horizontal direction.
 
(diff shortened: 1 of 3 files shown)


ledm:
Question to the reviewers. At the moment, the tests are failing because of the absence of FX files. How do I write a test that provides an `fx_files`?

Should we add an small sample areacello fx file somewhere to the repository?

valeriupredoi:
no need to, you can create a wee cube and save it with a patched func like in `tests/integration/test_recipe.py` - I can do that for you in this test on Monday :beer:

ledm:
Making some progress on this, and I've added several sensible tests (with assert raises) for the area_statistics calculation.

ledm:
I've changed the implementation of `area_statistics` so that if you want to use iris to calculate the grid cell area, you need to explicitly ask for this with the option `calculate_grid`. 

To make sure the functionaility of this preprocessor is explicitly clear, I've added a few failing clauses: 
- If `calculate_grid` is set to `True, an exception will be raised if your grid is irregular (latitide has more than 1 dimensions). This is because iris can't calculate the cell area for complex/irregular grids.

- If `calculate_grid` is set to `True, an exception will be raised if you also provide FX files in your recipe. You can't provide both an fx file and attempt to calculate the grid. 

- These tests only occur if the operator requires the grid definitions (many do not).

esmvalcore/preprocessor/_area.py
```diff
@@ -177,13 +177,18 @@ def tile_grid_areas(cube, fx_files):
...
+                tiled = da.array([grid_areas for i in range(cube.shape[0])])
+                grid_areas = da.stack(tiled, axis=0)
             elif cube.ndim == 4 and grid_areas.ndim == 3:
-                grid_areas = da.tile(grid_areas, [cube.shape[0], 1, 1, 1])
+                tiled = da.array([grid_areas for i in range(cube.shape[0])])
+                grid_areas = da.stack(tiled, axis=0).compute()
             elif cube.ndim == 3 and grid_areas.ndim == 2:
-                grid_areas = da.tile(grid_areas, [cube.shape[0], 1, 1])
+                tiled = da.array([grid_areas for i in range(cube.shape[0])])
+                grid_areas = da.stack(tiled, axis=0).compute()
```
valeriupredoi: replace the two `elif`'s with:
```
elif (cube.ndim, grid_areas.ndim) in [(4, 3), (3, 2)]:
```
and write the defs for `tiled` and `grid_areas` only once

esmvalcore/preprocessor/_area.py
```diff
@@ -243,33 +255,36 @@ def area_statistics(cube, operator, fx_files=None):
...
     if not fx_files and cube.coord('latitude').points.ndim == 2:
         logger.error(
             'fx_file needed to calculate grid cell area for irregular grids.')
         raise iris.exceptions.CoordinateMultiDimError(cube.coord('latitude'))
 
-    coord_names = ['longitude', 'latitude']
-    if grid_areas is None or not grid_areas.any():
-        cube = guess_bounds(cube, coord_names)
+    if not calculate_grid:
+        grid_areas = tile_grid_areas(cube, fx_files)
```
valeriupredoi: is `fx_files` an optional argument initialized with `None` for `tile_grid_areas()`?

esmvalcore/preprocessor/_area.py
```diff
@@ -243,33 +255,36 @@ def area_statistics(cube, operator, fx_files=None):
...
+    else:
+        cube = guess_bounds(cube, ['longitude', 'latitude'])
```
valeriupredoi: check if previous bounds exist

esmvalcore/preprocessor/_area.py
ledm: This code is outdated now...

esmvalcore/preprocessor/_area.py
ledm: `guess_bounds` does this check already.

valeriupredoi: no it don't, it fails if there are previous bounds

ledm:
```
# guess bounds tool
def guess_bounds(cube, coords):
    """Guess bounds of a cube, or not."""
    # check for bounds just in case
    for coord in coords:
        if not cube.coord(coord).has_bounds():
            cube.coord(coord).guess_bounds()
    return cube
```

esmvalcore/preprocessor/_area.py
```diff
@@ -177,22 +177,47 @@ def tile_grid_areas(cube, fx_files):
...
+        tiled = da.array([grid_areas for i in range(cube.shape[1])])
+        grid_areas = da.stack(tiled, axis=0)
+        tiled = da.array([grid_areas for i in range(cube.shape[0])])
+        grid_areas = da.stack(tiled, axis=0)
+    elif cube.ndim == 4 and grid_areas.ndim == 3:
+        tiled = da.array([grid_areas for i in range(cube.shape[0])])
+        grid_areas = da.stack(tiled, axis=0).compute()
+    elif cube.ndim == 3 and grid_areas.ndim == 2:
+        tiled = da.array([grid_areas for i in range(cube.shape[0])])
+        grid_areas = da.stack(tiled, axis=0).compute()
```
valeriupredoi: new code but still the same repetition of lines, man, pls use the suggestion from my previous comment

esmvalcore/preprocessor/_area.py
valeriupredoi: ah yes, that's the wrapper of guess bounds from iris, my bad (I put the wrapper in and I forget about it)

tests/unit/preprocessor/_area/test_area.py
```diff
@@ -15,6 +15,83 @@
...
+                                     long_name='longitude',
+                                     var_name='longitude',
+                                     units=Unit('degrees'),
+                                     bounds=lon2d_bounds)
+
+    gridir.add_aux_coord(lat_coord, data_dims=(0, 1))
+    gridir.add_aux_coord(lon_coord, data_dims=(0, 1))
+    fn = 'fx_test_areacello.nc'
+    iris.save(gridir, fn)
+    return fn
```
ledm: @valeriupredoi I wrote some tests (`test_area_statistics_mean_ir` and test_area_statistics_sum_ir`) and they are failing because IRIS doesn't like the fx cube I'm making here:

Any thoughts?

valeriupredoi: line 3277 from `cube.py` is trying to view (memory map) your **numpy** array but your weights is a dask array that doesn't have `view` as attribute:
```
In [42]: b                                                                                                                                        
Out[42]: dask.array<array, shape=(2,), dtype=int64, chunksize=(2,), chunktype=numpy.ndarray>

In [43]: b.view()                                                                                                                                 
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-43-4a91f95886bf> in <module>
----> 1 b.view()

TypeError: view() missing 1 required positional argument: 'dtype'
```
you need to realize your weights before calling the collapse (I don't think they've implemented lazy weights yet in iris)

valeriupredoi: just replace:
```
@@ -322,7 +322,7 @@ def area_statistics(cube, operator, fx_files=None, calculate_grid=False):
                          '({})'.format(cube.shape, grid_areas.shape))
 
     return cube.collapsed(['longitude', 'latitude'], operation,
-                          weights=grid_areas)
+                          weights=np.array(grid_areas))
```
note that the two tests will fail again but this time the result is not the same as expected (first is 1.5 when you check on 1.0 and the second is 7.381511e+12 when you check on 1.0) - I suspect you are not adding the right results spat by area statistics

ledm: I did something like this, but used:
```
    if isinstance(grid_areas, da.Array):
        grid_areas = grid_areas.compute()
```

valeriupredoi: ya that'll do! bit confused as to why your area statistic is such a monster number (7e12)

valeriupredoi:
can you pls give it a meaningful title?

ledm:
> can you pls give it a meaningful title?
Done

valeriupredoi:
> > can you pls give it a meaningful title?
> > Done

Almost :grin: All good, man! Can you pls tell me why you are expecting such a monster number from your test result (order 7e12)?

ledm:
yes, it's a weighted sum and the weights are real cell areas for 5x5 degree grid cells. It's a big number.

valeriupredoi:
> yes, it's a weighted sum and the weights are real cell areas for 5x5 degree grid cells. It's a big number.

ok so if you are expecting it to be this of a biggie, all good from me buds

ledm:
@schlunma @zklaus any chance that you could take a look at this please!

ledm:
It's worth mentionning that this PR changes the interface to the `area_statistics` preprocessor (in a sensible direction). However, some recipes may need to be changed as the `area_statistics` operators `mean` and `sum` will now not work without either a `fx_files` or a `calculate_grid: True` option. 

I guess that this means that we need to make a new ESMValTool branch and change a couple recipes.

valeriupredoi:
Can you pls do that on ESMValTool so  @mattiarighi can test the coupled core/tool modules? Code-wise this is all ok by me :beer:

schlunma:
What about cases where some data sets provide fx files and others don't (e.g. `OBS` ones)? With these changes it's not possible to calculate the areas in case there are no fx files, right? Maybe we change the name from `calculate_grid` to `allow_grid_calculation` and allow the use of `allow_grid_calculation` together with `fx_files`?

Otherwise I'm okay with the changes :+1:

ledm:
> What about cases where some data sets provide fx files and others don't (e.g. `OBS` ones)? With these changes it's not possible to calculate the areas in case there are no fx files, right? Maybe we change the name from `calculate_grid` to `allow_grid_calculation` and allow the use of `allow_grid_calculation` together with `fx_files`?
> 
> Otherwise I'm okay with the changes 👍

In that case, I've been assuming that we just apply a different preprocessor to the obs datasets. Is that sensible or are there cases where we need obs and models to use the same preprocessor chain?

schlunma:
That's of course always possible, but I think this is really not user-friendly, since it implies adding a new entry to the `variable` section of a diagnostic and adding `additional_datasets` at the correct places. Not hard if you are used to that, but might be problematic for new users.

ledm:
I don't like the idea of having that level of (hidden) flexibility. If people want to calculate their own grid areas, they should explicitly ask for that. 

Just so that everyone is aware that that is whats happening and that that is what they want.

schlunma:
But if the user has to explicitly specify `allow_grid_calculation` the feature is not hidden, isn't it? If that is not desired, only `fx_files` should be used, which gives an error if no files are available. If only `allow_grid_calculation` is given, no `fx_files` are used. Fairly transparent in my opinion.

ledm:
So, what are we waiting for on this PR?

mattiarighi:
To be reviewed by @bouweandela and tested, as usual.

valeriupredoi:
> But if the user has to explicitly specify `allow_grid_calculation` the feature is not hidden, isn't it? If that is not desired, only `fx_files` should be used, which gives an error if no files are available. If only `allow_grid_calculation` is given, no `fx_files` are used. Fairly transparent in my opinion.

@ledm have you addressed in any way @schlunma comment above? If no code change is needed then pls give a motivation that might be good to put in the documentation as well :beer: Anyways, I *think* this PR will only work after #439 has been merged no? But by all means let's hurry up and get both this and 439 reviewed pls

esmvalcore/preprocessor/_area.py
```diff
@@ -243,33 +272,60 @@ def area_statistics(cube, operator, fx_files=None):
...
 
+    # Many IRIS analysis functions do not accept weights arguments.
+    if not operator_accept_weights(operator):
+        return cube.collapsed(['longitude', 'latitude'], operation)
+
+    # The following are for operators with means
     if not fx_files and cube.coord('latitude').points.ndim == 2:
         logger.error(
-            'fx_file needed to calculate grid cell area for irregular grids.')
+            'fx_file needed for weighted calcuions on irregular grids.')
```
bouweandela:
```suggestion
            'fx_file needed for weighted calculations on irregular grids.')
```

esmvalcore/preprocessor/_area.py
```diff
@@ -243,33 +272,60 @@ def area_statistics(cube, operator, fx_files=None):
...
     if cube.shape != grid_areas.shape:
         raise ValueError('Cube shape ({}) doesn`t match grid area shape '
                          '({})'.format(cube.shape, grid_areas.shape))
 
-    operation = get_iris_analysis_operation(operator)
-
-    # TODO: implement weighted stdev, median, s var when available in iris.
-    # See iris issue: https://github.com/SciTools/iris/issues/3208
+    if isinstance(grid_areas, da.Array):
+        grid_areas = grid_areas.compute()
```
bouweandela: Why are you realizing the data? Is it not possible to use a dask array as a weight?

ledm: Nope, iris only takes numpy arrays as weights.

esmvalcore/preprocessor/_area.py
```diff
@@ -177,22 +177,44 @@ def tile_grid_areas(cube, fx_files):
...
+    else:
+        raise ValueError('Grid and dataset number of dimensions not '
+                         'recognised: {} and {}.'
+                         ''.format(cube.ndim, grid_areas.ndim))
     return grid_areas
 
 
 # get the area average
-def area_statistics(cube, operator, fx_files=None):
+def area_statistics(cube, operator, fx_files=None, calculate_grid=False):
```
bouweandela: I think it would be nicer to define this function as
```python
def area_statistics(cube, operator, weights=None):
```
and `weights` can be either the path to an fx file, the string `'calculate'` or `None` for unweighted statistics.

esmvalcore/preprocessor/_area.py
```diff
@@ -177,22 +177,44 @@ def tile_grid_areas(cube, fx_files):
...
+        input cube.
+    fx_files: dict
+        dictionary of field:filename for the fx_files
+
+    Returns
+    -------
+    iris.cube.Cube
+        Freshly tiled grid areas cube.
+    """
+    if cube.ndim == grid_areas.ndim:
```
bouweandela: I think this entire function could be replaced by
```python
grid_areas = np.broadcast_to(grid_areas, cube.shape[:-grid_areas.ndim] + grid_areas.shape)
```

bouweandela:
> That's of course always possible, but I think this is really not user-friendly, since it implies adding a new entry to the `variable` section of a diagnostic and adding `additional_datasets` at the correct places. Not hard if you are used to that, but might be problematic for new users.

If this is problematic for new users, I think we need to improve our documentation. I agree with @ledm that it should be no problem to use different preprocessors for different datasets.

esmvalcore/preprocessor/_area.py
bouweandela: or even:
```python
grid_areas = np.broadcast_to(grid_areas, cube.shape)
```
or am I missing something?

bouweandela:
> using only fx files, using only calculated grid areas or using both with a preference on using fx files

I think that would also be fine, so long as it doesn't silently fall back to calculated areas when you specify you want to use an fx file.

The reason why land/sea masks are different, at least intuitively, is that you can really easily see the difference in masking when you plot the data, while with statistics you'll only get some numbers out, so it's less easy to see if things go wrong. Maybe it would be good to make the land/sea mask preprocessor function more explicit in whether or not you want natural earth as a fallback.

sloosvel:
Sorry but I've been waiting quite some time for the area statistics to be fixed. I do not mind doing workarounds but now we have users wanting to use the preprocessor on irregular grids, and for some reason the code in master fails at `da.tile`. If this is not ready, could at least the code on master go back to using `np.tile` ?

zklaus:
@ledm I'm afraid this is on your desk. I particularly agree with @bouweandela that `tile` is probably uncalled for. Let me know if you would like to discuss the broadcast approach.

ledm:
What are the differences between `np.tile`, `da.tile` and `np.broadcast_to`. As far as I can tell, from the users perspective, they are all just different ways to produce the same output. I'm not wise enough to know which one is optimised for our use case. 

Note that this entire problem was caused by `np.tile` and `da.tile` having a different interface, so they are not perfectly compatible.

Is there a dask version of `broadcast_to` to consider as well?

Having said that, functionality should really be included in iris, when cell methods are built into the data cube.

bouweandela:
> Note that this entire problem was caused by np.tile and da.tile having a different interface, so they are not perfectly compatible.

I think this may have been solved since dask version 2.8.0 (see [here](https://docs.dask.org/en/latest/changelog.html#id128)).

zklaus:
@ledm, tiling creates a new array with copies of data, broadcasting creates a new [view](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.view.html) using the same data memory with only a bit of logic to make the indices point to the right place.

As a consequence, broadcasting works in constant time and memory (O(1)), tiling needs memory and time that depends on the shape of the target, hence is much more expensive. Always prefer broadcasting, if possible! Only when you actually need the new array to have and manipulate further consider using tiling. (Cf the numpy documentation on tiling:
> Note: Although tile may be used for broadcasting, it is strongly recommended to use numpy’s broadcasting operations and functions.

)


Ok, that's the two concepts, now how can we use them? Tiling using `np.tile`/`da.tile` seems clear enough. Broadcasting is a bit more comprehensive, but in many cases simpler. Broadcasting always happens when combining two arrays of different shape with operations that would work on same-shape arrays. Examples:
```python
a = np.random.randn(3, 3)  # create random 3x3 matrix
3. * a  # multiply scalar (shape: (,)) with matrix (shape: (3, 3)); scalar will be broadcast to shape (3, 3)
b = np.array([1, 2, 3])  # create 3 vector
a * b  # multiply `a` (shape: (3, 3)) with b (shape: (3,)); b will be broadcast to (3, 3)
a * b[np.newaxis, :]  # the same as the previous line
a * b[:, np.newaxis]  # almost the same, but interpret b as column vector instead of a row vector
```
So, you likely have been broadcasting all the time, for a long time, without knowing it :smile:.

Sometimes, numpy/dask needs a bit of help to do the broadcasting correctly. The simplest way is when you need to specify where to insert axes to get one array to the shape of the other; this you can see in the examples above with `np.newaxis`. Sometimes even that is not enough, for example when you already know the shape of the target, but don't have the target array yet. In that case, numpy/dask has no chance to figure out the shape on its own and you can use `broadcast_to` to give that information explicitly.

More information on broadcasting can be found in [the numpy docu (also note the links at the top)](https://numpy.org/doc/stable/user/basics.broadcasting.html).

Does that make sense?

valeriupredoi:
@ledm dis still needed?

valeriupredoi:
cheers @ledm - I be nuking this then :bomb:

Issue #453, opened 2020-02-03
Regrid with esmpy tests failing on CI due to mock=4


valeriupredoi:
```
    def __new__(cls, *args, **kw):
        # every instance has its own class
        # so we can create magic methods on the
        # class without stomping on other mocks
        bases = (cls,)
        if not issubclass(cls, AsyncMockMixin):
            # Check if spec is an async object or function
            bound_args = _MOCK_SIG.bind_partial(cls, *args, **kw).arguments
            spec_arg = bound_args.get('spec_set', bound_args.get('spec'))
>           if spec_arg and _is_async_obj(spec_arg):
E           ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

.eggs/mock-4.0.0b1-py3.7.egg/mock/mock.py:411: ValueError
```

Honestly I gots no clues where the CI machine picks up `mock=4.0.0b1` and not `3.0.5` that's in my env and in all the conda-forge and pypi repos. Will investigate more - any clues @bouweandela ?

zklaus:
@valeriupredoi can I have the full log?

valeriupredoi:
@zklaus e.g. here mate https://circleci.com/gh/ESMValGroup/ESMValCore/6951?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

bouweandela:
We no longer need the mock library, we needed it for Python 2 and we no longer support that. In Python 3 it's available in the standard library as `unittest.mock`. I created #452 to fix this issue.

valeriupredoi:
and I closed #454 cheers Bouwe!

valeriupredoi:
Solved by #452

PR #452, opened 2020-01-30
Clean up unit tests and fix warnings

bouweandela:
This pull request
- Removes no longer required test dependencies, that currently break the tests on CirclCI in the `master` branch
- Fixes a lot of warnings

PR #454, opened 2020-02-03
Restrict mock to <4

valeriupredoi:
#453 `mock=4` beta seems dodgy at the moment and the CI is picking it up from God knows where (nowhere on the conda forge or pypi official repos)

### DIFF
--- a/meta.yaml
+++ b/meta.yaml
@@ -49,6 +49,7 @@ requirements:
     - netCDF4
     - numba
     - numpy
+    - mock>=3.0.5,<4.0
     - prov
     - psutil
     - pydot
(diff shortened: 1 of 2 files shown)


valeriupredoi:
OK so it looks like we have to put up with the three test fails once we unpin `mock` -> they have actually released 4.0 https://github.com/testing-cabal/mock/pull/476 but the updated packages have yet to arrive on pypi and conda-forge -> @zklaus can you pls have a look at #453 and advise as to how to proceed? This PR will do for in the meantime so we dont have our tests failed

bouweandela:
This problem is already solved in #452.

valeriupredoi:
as Bouwe says -> goto #452

PR #432, opened 2020-01-15
Allowed arbitrary CMIP6 fx variables in special preprocessors and added a catch on project not found in config-developer (invalid project)

schlunma:
**Tasks**

-   [x] [Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do, if you haven't done so already (and add the link at the bottom)
-   [x] This pull request has a descriptive title that can be used in a changelog
-   [x] Add unit tests
-   [x] Public functions should have a numpy-style docstring so they appear properly in the [API documentation](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/api/esmvalcore.html). For all other functions a one line docstring is sufficient.
-   [x] If writing a new/modified preprocessor function, please update the [documentation](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/esmvalcore/preprocessor.html)
-   [x] Circle/CI tests pass. Status can be seen below your pull request. If the tests are failing, click the link to find out why.
-   [x] Codacy code quality checks pass. Status can be seen below your pull request. If there is an error, click the link to find out why. If you suspect Codacy may be wrong, please ask by commenting.
-   [x] Please use `yamllint` to check that your YAML files do not contain mistakes
-   [x] If you make backward incompatible changes to the recipe format, make a new pull request in the [ESMValTool repository](https://github.com/ESMValGroup/ESMValTool) and add the link below

If you need help with any of the tasks above, please do not hesitate to ask by commenting in the issue or pull request.

* * *

Closes #431.

PR #455, opened 2020-02-03
Added test for check of data availability and fixed some checks for _recipe.py

schlunma:
This adds missing tests for #98 (where @valeriupredoi was too quick in merging :wink:).

I also fixed some syntactic errors in the tests for `_recipe.py` which prevented them to work correctly (e.g. this explains why a test didn't fail after #432 even though it should have failed). This revealed some errors in the old tests. I couldn't fix two of them since I'm not quite sure how. Can you have a look at them, @bouweandela please? I think you wrote them (`test_extract_shape_raises` in `tests.integration.test_recipe`).

This PR only affects tests, so I think we can quickly merge it.


valeriupredoi:
In principle approved, but then I saw the failed tests (not the ones fixed by #452 ) - can you check them out pls :beer:

schlunma:
> In principle approved, but then I saw the failed tests (not the ones fixed by #452 ) - can you check them out pls beer

These are the ones where @bouweandela needs to step in since I'm not sure about them (I don't know if the test itself is wrong or the function).

valeriupredoi:
Right on, Ill have a closer look too, after coffee :coffee:

valeriupredoi:
that test is badly written in `master` and you have actually uncovered the badness - the recipe fails with `RecipeError(In preprocessor function `extract_shape`: Unable to find 'shapefile: ')` at all times irrespective of the functional argument passed - so it should fail exactly 3 times as you see it fail

valeriupredoi:
ie if the `extract_shape` is run without `shapefile: test.shp` it'll always fail first coz the shapefile is not found due to `extract_shape(settings)` from `_recipe_checks.py` - @bouweandela or @JaroCamphuijsen please fix this catch22 :beer:

valeriupredoi:
> ie if the `extract_shape` is run without `shapefile: test.shp` it'll always fail first coz the shapefile is not found due to `extract_shape(settings)` from `_recipe_checks.py` - @bouweandela or @JaroCamphuijsen please fix this catch22 beer

any chance @bouweandela or @JaroCamphuijsen had a look at this pls :beer:

bouweandela:
The goal of the check function is to check that the arguments are within valid range and the goal of the test is to check that an RecipeError is raised if an invalid argument is provided. We should probably use a schema for this kind of checking at some point though.

valeriupredoi:
Ah biutiful, cheers monsieur @bouweandela :beer:

PR #457, opened 2020-02-06
Fix areacello in CMIP5 EC-Earth

jvegasbsc:

**Tasks**

-   [x] Add unit tests

### DIFF
--- a/esmvalcore/cmor/_fixes/cmip5/ec_earth.py
+++ b/esmvalcore/cmor/_fixes/cmip5/ec_earth.py
@@ -1,8 +1,9 @@
 """Fixes for EC-Earth model."""
 from dask import array as da
+import iris
 
 from ..fix import Fix
-from ..shared import add_scalar_height_coord
+from ..shared import add_scalar_height_coord, cube_to_aux_coord
 
 
 class Sic(Fix):
@@ -101,3 +102,30 @@ class Tas(Fix):
                 cube.coord('time').long_name = 'time'
 
         return cubes
+
+
+class Areacello(Fix):
+    """Fixes for areacello."""
+
+    def fix_metadata(self, cubes):
+        """
+        Fix potentially missing scalar dimension.
+
+        Parameters
+        ----------
+        cubes: iris CubeList
+            List of cubes to fix
+
+        Returns
+        -------
+        iris.cube.CubeList
+
+        """
+        areacello = cubes.extract('Areas of grid cell')[0]
+        lat = cubes.extract('latitude')[0]
+        lon = cubes.extract('longitude')[0]
+
+        areacello.add_aux_coord(cube_to_aux_coord(lat), (0, 1))
+        areacello.add_aux_coord(cube_to_aux_coord(lon), (0, 1))
+
+        return iris.cube.CubeList([areacello, ])
--- a/esmvalcore/cmor/_fixes/shared.py
+++ b/esmvalcore/cmor/_fixes/shared.py
@@ -69,6 +69,7 @@ def add_scalar_typesea_coord(cube, value='default'):
         cube.add_aux_coord(typesea_coord, ())
     return cube
 
+
 def cube_to_aux_coord(cube):
     """Convert cube to iris AuxCoord"""
     return iris.coords.AuxCoord(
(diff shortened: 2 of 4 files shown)


jvegasbsc:
Required for ESMValGroup/ESMValTool#1299

zklaus:
Is there an issue for this PR?

jvegasbsc:
> Is there an issue for this PR?

No, just found that Iris can not read latitudes and longitudes for areacello in ESMValGroup/ESMValTool#1299

valeriupredoi:
@jvegasbsc Javi man, it'd be good to stick to what @bouweandela has been preaching for a while now: descriptive title for the PR, tick the boxes in the PR template, link and give a nice description in the original https://github.com/ESMValGroup/ESMValTool/pull/1299 (open an issue here before a PR too) - you know, the boring stuff, but albeit very useful for later :beer:

zklaus:
The actual issue with the original file is that the `areacello` variable in it is missing the `coordinates` attribute. Hence, imho, the correct fix is to use `fix_file` to add `areacello:coordinates = "lat lon"`.

jvegasbsc:
>  Hence, imho, the correct fix is to use fix_file to add areacello:coordinates = "lat lon".

That way is much slower, as we don't modify the files but create a new copy. Fix file must be used only if Iris can not read the cubes.

zklaus:
I thought about that, but since this is an fx file the difference should be very small.

valeriupredoi:
In #439 we use the output file as input for fx preproc steps so the file needs to be fixed proper via `fix_file`

bouweandela:
I think that we might want to consider removing fix_file altogether at some point, because so far we have not encountered a single dataset on ESGF that needs it.

schlunma:
> I think that we might want to consider removing fix_file altogether at some point, because so far we have not encountered a single dataset on ESGF that needs it.

I found a lot of them, but the fixes are not merged (regarding `cl` of almost all CMIP5 and CMIP6 datesets):

https://github.com/ESMValGroup/ESMValCore/blob/97a242cd0ccc544d16ca9c083bafdf1164f78a7d/esmvalcore/cmor/_fixes/cmip5/canesm2.py#L12-L58

Note that setting the attribute `formula_terms` is not allowed in `iris`, thus we definitely need to use another package for that. Without that attribute, `iris` is not able to read the derived coordinate correctly.

jvegasbsc:
> Note that setting the attribute formula_terms is not allowed in iris, thus we definitely need to use another package for that. Without that attribute, iris is not able to read the derived coordinate correctly.

Can't you overcome that using any of this, at least for a few datasets? 
https://scitools.org.uk/iris/docs/latest/iris/iris/aux_factory.html

zklaus:
One alternative might be to rethink the `fix_file` in terms of iris' callback mechanism.

jvegasbsc:
> One alternative might be to rethink the fix_file in terms of iris' callback mechanism.

I had open a new issue to discuss the future of the fix_file

zklaus:
And the issue is #463. Thanks, good idea to separate that discussion.

jvegasbsc:
> And the issue is #463. Thanks, good idea to separate that discussion.

Ups, I forgot the link. Thanks!

bouweandela:
@mattiarighi I think this can be merged, considering that the discussion is on a different topic. Right?

zklaus:
I still think this is the wrong fix, but I guess we'll have to revisit the whole `cell_measures` complex after some iris improvements anyway.

zklaus:
Actually, philosophical differences aside, V [seems to hint](https://github.com/ESMValGroup/ESMValCore/pull/457#issuecomment-582975979) above that the file needs to be fixed. @valeriupredoi, can you please elaborate and tell us whether this fix works or not?

valeriupredoi:
> Actually, philosophical differences aside, V [seems to hint](https://github.com/ESMValGroup/ESMValCore/pull/457#issuecomment-582975979) above that the file needs to be fixed. @valeriupredoi, can you please elaborate and tell us whether this fix works or not?

dont mind me, I was walking in the desert, the fx file will have gone through the full fix metadata/fx data/all needed preproc steps before being used - that's the whole point of #439

zklaus:
Alright, hope you found some water :). Then this is closed for good.

PR #462, opened 2020-02-07
Fix for bad time bounds in CMIP5 GFDL-CM2p1 sit

### LINKED ISSUE #461: Dataset problem:  GFDL-CM2p1 sit has invalid time bounds
I have found a sit file for CMIP5 GFDL-CM2p1 that has time bounds of about 1e9  with times of abot 40000. Iris breaks up at the checker, when changing the time units:

```yaml
- {dataset: GFDL-CM2p1, project: CMIP5, exp: historical, ensemble: r1i1p1, start_year: 2001, end_year: 2004}
```

Pull request with the fix is coming soon

jvegasbsc:
Add fix for time bounds of sit  in CMIP5 GFDL-CM2p1. Not all files are affected, so fix is applied only after checking that the error is really ther

**Tasks**

-   [x] Add unit tests

* * *

Closes #461 
Required by ESMValGroup/ESMValTool#1299


esmvalcore/cmor/_fixes/cmip5/gfdl_cm2p1.py
```diff
@@ -56,6 +56,32 @@ def fix_data(self, cube):
...
 
 
+class Sit(Fix):
+    """Fixes for sit"""
+
+    def fix_metadata(self, cubes):
+        """
+        Fix metadata.
+
+        Fixes wrong units.
```
zklaus: Probably this is a copy-and-paste leftover? Improve docstring.

esmvalcore/cmor/_fixes/cmip5/gfdl_cm2p1.py
```diff
@@ -56,6 +56,32 @@ def fix_data(self, cube):
...
+        -------
+        iris.cube.Cube
+
+        """
+        cube = self.get_cube_from_list(cubes)
+        time = cube.coord('time')
+        if time.bounds.max() > 1e8:
+            time.bounds = None
+            time.guess_bounds()
+        return cubes
```
zklaus: Did you check that this procedure leads to the same time bounds as in unaffected variables?

jvegasbsc: No, I do not check. And you are right I should, time bounds are usually not as easy to guess as lat - lon ones in regular grids

jvegasbsc: Not, and in fact they are wrong. Fixing them manually now and restricting the fix to monthly data

esmvalcore/cmor/_fixes/cmip5/gfdl_cm2p1.py
```diff
@@ -56,6 +58,41 @@ def fix_data(self, cube):
...
+        cube: iris.cube.Cube
+
+        Returns
+        -------
+        iris.cube.Cube
+
+        """
+        cube = self.get_cube_from_list(cubes)
+        time = cube.coord('time')
+        if time.bounds.max() > 1e8 and self.vardef.frequency == 'mon':
```
zklaus: Looking at one example file I found that this test for necessity is not reliable. For early times (~1850 in the historical) even the wrong bounds are below 1e8. Maybe a better test is to check whether the point lies inside its corresponding bounds?

jvegasbsc: Can you point me to that file? I put this because in mine the values where of the order of 1e9 and I assume that all problems of this kind with the bounds will be in the same range.

zklaus: Sure: `sit_OImon_GFDL-CM2p1_historical_r1i1p1_186101-186512.nc`

jvegasbsc: Thanks!

esmvalcore/cmor/_fixes/cmip5/gfdl_cm2p1.py
```diff
@@ -56,6 +58,50 @@ def fix_data(self, cube):
...
+        Returns
+        -------
+        iris.cube.Cube
+
+        """
+        cube = self.get_cube_from_list(cubes)
+        time = cube.coord('time')
+        if self._fix_required(time):
+            new_bounds = np.empty(time.bounds.shape, time.bounds.dtype)
+            for x, point in enumerate(time):
```
bouweandela:
```suggestion
            for i, point in enumerate(time):
```
would be more in line with the conventions

esmvalcore/cmor/_fixes/cmip5/gfdl_cm2p1.py
jvegasbsc: Done!

esmvalcore/cmor/_fixes/cmip5/gfdl_cm2p1.py
jvegasbsc: It looks that they are using different units for the bounds (hours since 1850, maybe?), or more likely  that they converted the time units and forgot converting the time bounds

esmvalcore/cmor/_fixes/cmip5/gfdl_cm2p1.py
```diff
@@ -56,6 +58,50 @@ def fix_data(self, cube):
...
+            for i, point in enumerate(time):
+                date = point.units.num2date(point.points[0])
+                start = cftime.DatetimeJulian(date.year, date.month, 1)
+                if date.month < 12:
+                    end = cftime.DatetimeJulian(date.year, date.month + 1, 1)
+                else:
+                    end = cftime.DatetimeJulian(date.year + 1, 1, 1)
+                new_bounds[i, 0] = point.units.date2num(start)
+                new_bounds[i, 1] = point.units.date2num(end)
+            time.bounds = new_bounds
```
zklaus: I personally don't care much for `for` loops, so would suggest something a bit more pythonic, maybe

```suggestion
times = time.units.num2date(time.points)
starts = map(lambda c: cftime.DatetimeJulian(c.year, c.month, 1), times)
ends = map(lambda c: (cftime.DatetimeJulian(c.year, c.month + 1, 1) 
                                      if c.month < 12
                                      else cftime.DatetimeJulian(c.year + 1, 1, 1)), times)
time.bounds = time.units.date2num(np.stack([starts, ends], -1))
```

esmvalcore/cmor/_fixes/cmip5/gfdl_cm2p1.py
```diff
@@ -56,6 +58,50 @@ def fix_data(self, cube):
...
+        return cubes
+
+    def _fix_required(self, time):
+        if self.vardef.frequency != 'mon':
+            return False
+        if time.bounds[-1, 0] > time.points[-1]:
+            return True
+        if time.bounds[-1, 1] < time.points[-1]:
+            return True
+        return False
```
zklaus:
```suggestion
        return (self.vardef.frequency == 'mon'
                and not (time.bounds[-1, 0] < time.points[-1] < time.bounds[-1, 1]))
```

zklaus: I also wonder where the frequency part of the condition is coming from. Does `sit` occur in another table? Is it ok in that table if the bounds are not surrounding the points?

jvegasbsc: > Does sit occur in another table

It is also in day

> Is it ok in that table if the bounds are not surrounding the points?

Of course not, but I have not tested it and prefer to leave the fix out in case it is not needed

jvegasbsc:
Suggestions added

bouweandela:
@mattiarighi Could you please test and merge?

Issue #93, opened 2019-06-17
Too many unnessary settings in config-user.yml


bouweandela:
At the moment there are many settings in config-user.yml that users will very rarely change. Therefore I propose to reduce the number of options exposed to users in the following way:

**Proposed for removal (can be moved to the recipe for those diagnostics that support these options)**
- `write_plots: true`
- `write_netcdf: true`
- `output_file_type: ps  # [ps]/pdf/png/eps/epsi`

These are options that should be implemented by diagnostic developers, so this puts a burden on diagnostic developers to implement it and in many cases it does not work, because they are either not aware of these options, or for the last option, the library they use for plotting does not support a particular plot type. The amount of time that is saved by not creating plots or netcdf files or all possible plot types is probably only a small fraction of the total runtime, so the advantage of having these settings is rather unclear, but having them creates a bad user experience because they do not work in many cases and a bad diagnostic developer experience because they have to implement and test additional requirements.

**Proposed for moving to an advanced users file, e.g config-developer.yml**
- `profile_diagnostic: false` -> only interesting for Python diagnostic developers
- `save_intermediary_cubes: false` -> for debugging purposes, probably only interesting for diagnostic developers
- `exit_on_warning: false` -> specific to NCL diagnostics, probably only interesting for diagnostic developers

ledm:
While we're working on config-user.yml, can we change the way that ESMValTool locates the data (rootpath?

I don't understand why user's need to edit their config-user.yml file on jasmin by adding a rootpath here?
Ie:
```
rootpath:
  # Rootpath to CMIP5 data
  CMIP5: /badc/cmip5/data/cmip5/output1
```

This path is never going to change. It will always be the same for all JASMIN users. Why do we need to specify it in our config-user.yml files? Can't it be added to the config-developer.yml by default?

bouweandela:
I think that is a good suggestion, but please open a new issue for it.

bouweandela:
Setting `max_data_filesize: 100` has been removed since this issue was opened.

jvegasbsc:
With the new CLI, all `config-user.yml` options can be set as flags in the run command, maybe we should think in some command line options to be moved to it

bouweandela:
[...]

PR #465, opened 2020-02-07
Better default configuration

bouweandela:
This pull request aims to provide better default settings in config-user.yml. Main changes:
- run in parallel by default, because users seem to be afraid to use this option and then complain that the tool is slow.
- unused option max_data_filesize removed

**Tasks**

* * *

Related to #93


esmvalcore/_config.py
```diff
@@ -38,13 +38,12 @@ def read_config_user_file(config_file, recipe_name):
...
-        'output_file_type': 'ps',
-        'output_dir': './output_dir',
-        'auxiliary_data_dir': './auxiliary_data',
+        'output_file_type': 'png',
+        'output_dir': 'esmvaltool_output',
+        'auxiliary_data_dir': 'auxiliary_data',
         'save_intermediary_cubes': False,
-        'remove_preproc_dir': False,
-        'max_parallel_tasks': 1,
+        'remove_preproc_dir': True,
```
mattiarighi: I would keep `False` as a default.
For a first-time user it may be more convenient to get the full output.

bouweandela: I think that's only convenient if you're going to develop a diagnostic, not if you're a first time user.

bouweandela: Anyway, the default has been true in all the example configuration files we provide for some time now, so this pull request just aligns the defaults in the code with those in the example files.

mattiarighi: Yes, but the [documentation](https://esmvaltool.readthedocs.io/en/latest/getting_started/outputdata.html) says that the output directory will contain 4 subdirs, including `preproc/`. It could be confusing if `preproc/` is deleted by default.

bouweandela: If you scroll down a bit in that documentation, it says that the directory is deleted by default. I've updated the section to make it more clear.

PR #468, opened 2020-02-07
Try to add is-enes3 community and is-enes3 grant to zenodo and update authors

bouweandela:
Let's see if it works.

I found the instructions here: https://developers.zenodo.org/#representation

This pull request also puts the core development team at the top of the list.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -92,5 +92,18 @@
     "license": {
         "id": "Apache-2.0"
     },
-    "title": "ESMValCore"
+    "title": "ESMValCore",
+    "communities": [
+        {
+            "identifier": "is-enes3"
+        },
+        {
+            "identifier": "ecfunded"
+        }
+    ],
+    "grants": [
+        {
+            "id": "10.13039/501100000780::824084"
+        }
+    ]
 }
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -7,20 +7,12 @@
         },
         {
             "affiliation": "DLR, Germany",
-            "name": "Bock, Lisa",
-            "orcid": "0000-0001-7058-5938"
-        },
-        {
-            "affiliation": "DLR, Germany",
             "name": "Broetz, Bjoern"
         },
         {
-            "affiliation": "NLeSC, Netherlands",
-            "name": "Diblen, Faruk"
-        },
-        {
-            "affiliation": "MetOffice, UK",
-            "name": "Dreyer, Laura"
+            "affiliation": "PML, UK",
+            "name": "de Mora, Lee",
+            "orcid": "0000-0002-5080-3149"
         },
         {
             "affiliation": "NLeSC, Netherlands",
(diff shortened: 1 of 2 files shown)


mattiarighi:
Note that the list of authors has changed order (Core Team alphabetically + Others alphabetically).

bouweandela:
Did you make a pull request for that already? See also: https://github.com/ESMValGroup/ESMValCore/blob/master/CONTRIBUTING.md#list-of-authors

mattiarighi:
No, actually I just changed it by editing the item on zenodo.

bouweandela:
I'm not sure how that will interact with having a .zenodo.json file in our repository, but I suspect the changes might be overwritten with the next release.

bouweandela:
Also, I think it would be good to keep the same the list of authors everywhere.

valeriupredoi:
yes, I agree with @bouweandela for a common list. Surely you have to create PR after you've added your name as big honcho on the creator's list no? Also how do we accept/reject such cases?

mattiarighi:
> Also, I think it would be good to keep the same the list of authors everywhere.

But for it would still be different between ESMValCore and ESMValTool, right?

mattiarighi:
Should I update the list in the .json file in this PR to be consistent with the one currently on Zenodo?

.zenodo.json
```diff
@@ -92,5 +92,18 @@
...
     },
-    "title": "ESMValCore"
+    "title": "ESMValCore",
+    "communities": [
+        {
+            "identifier": "is-enes3"
+        },
+        {
+            "identifier": "ecfunded"
+        }
```
mattiarighi:
```suggestion
        {
        {
            "identifier": "dlr_de"
        }
```
This is something I added recently.

bouweandela:
>> I think it would be good to keep the same the list of authors everywhere.

> But for it would still be different between ESMValCore and ESMValTool, right?

Yes. What I meant should be the same is the files in our repositories, i.e. the list of authors should be the same in the citation file (CITATION.cff) and the Zenodo configuration file (.zenodo.json). The latter is generated from the former, using the procedure described here: https://github.com/ESMValGroup/ESMValCore/blob/master/CONTRIBUTING.md#list-of-authors

> Should I update the list in the .json file in this PR to be consistent with the one currently on Zenodo?

Yes, but update the CITATION.cff file and then generate the .zenodo.json file

.zenodo.json
bouweandela: Added

mattiarighi:
Question to @ledm: should you be listed under "D" (de Mora, Lee) or under "M" (More, Lee de)?

mattiarighi:
> Yes, but update the CITATION.cff file and then generate the .zenodo.json file

Done.

valeriupredoi:
> Question to @ledm: should you be listed under "D" (de Mora, Lee) or under "M" (More, Lee de)?

@ledm 's nickname is Alpaca so under A (that'd make him first author too :grin: )

mattiarighi:
My one is AbbottAbbot, from the author of [Flatland](https://en.wikipedia.org/wiki/Flatland) stuck_out_tongue:

valeriupredoi:
d'you know we used to have an Abbott in LIGO and a lot of the papers were Abbott et al, then he left and a certain Aasi came along :grin:

mattiarighi:
I opened a complementary PR in ESMValTool https://github.com/ESMValGroup/ESMValTool/pull/1512

mattiarighi:
Btw, @ledm is listed under "D" on his institute's [https://www.pml.ac.uk/People], so it should be fine the way I did it.

ledm:
Yes, it's listed under D. (and you correctly used the lowercase "d". Thanks

PR #483, opened 2020-02-12
Increase version to v2.0.0b6

bouweandela:
New release needed for finishing ESMValGroup/ESMValTool#1432


valeriupredoi:
we gonna run out of beta numbers :grin:

Issue #487, opened 2020-02-13
New preprocessor mask_multimodel


bascrezee:
**Is your feature request related to a problem? Please describe.**
I want to propagate the mask over several datasets such that when an element is masked in one dataset, it is also masked in all other datasets.

**Would you be able to help out?**
Yes, working on it. Initially I thought `mask_fillvalues` would do the job, but it doesn't. Also Github kindly points me to other issues, and just now I realize this has been discussed before here https://github.com/ESMValGroup/ESMValCore/issues/116

mattiarighi:
Can you be more specific on why `mask_fillvalues` doesn't do what you need.
From the feature request it looks like this is exactly what `mask_fillvalues` was designed for.

@valeriupredoi

bascrezee:
Yes, let me try to clarify.

```
datasets:
  - {dataset: cds-era5-monthly, type: reanaly, project: OBS, mip: Lmon,
     version: 1, tier: 3, start_year: 2008, end_year: 2010}
  - {dataset: CDS-SATELLITE-SOIL-MOISTURE, type: sat, project: OBS, mip: Lmon,
     version: COMBINED-TCDR-v201812.0.0, tier: 3, start_year: 2008, end_year: 2010}

preprocessors:
  test_b:
    regrid:
      target_grid: 0.25x0.25
      scheme: nearest
    mask_fillvalues:
      threshold_fraction: 0.00    # keep all missing values
      min_value: -1e+20           # small enough not to alter the data
```

In my diagnostic script:
```
def _get_cubelist(cfg):
    input_data = cfg['input_data'].values()
    grouped_input_data = group_metadata(
        input_data, 'dataset', sort='dataset')
    cubes = []
    for datasetname in grouped_input_data.keys():
        cubes.append(iris.load_cube(grouped_input_data[datasetname][0]['filename']))
    return iris.cube.CubeList(cubes)


def main(cfg):
    """Calculate linear regression between albedo and xxfrac.

    Arguments:
    ---------
        cfg - nested dictionary of metadata
    """
    cubelist = _get_cubelist(cfg)
    assert(np.array_equal(cubelist[0].data.mask, cubelist[1].data.mask))
```

Executing raises `AssertionError`

mattiarighi:
Maybe we should try to extend or adapt `mask_fillvalues` rather than create a new preprocessor.
@valeriupredoi any suggestion?

bascrezee:
[...]

PR #492, opened 2020-02-13
added test for simple mask combination for a small threshold for mask_fillvalues

valeriupredoi:
Solves the no-need for #487

### DIFF
--- a/doc/esmvalcore/preprocessor.rst
+++ b/doc/esmvalcore/preprocessor.rst
@@ -451,22 +451,23 @@ is set to 19.0 (in units of the variable units).
 
 See also :func:`esmvalcore.preprocessor.mask_fillvalues`.
 
-.. note::
+Common mask for multiple models
+-------------------------------
 
-   It is possible to use ``mask_fillvalues`` to create a combined multimodel
-   mask (all the masks from all the analyzed models combined into a single
-   mask); for that purpose setting the ``threshold_fraction`` to 0 will not
-   discard any time windows, essentially keeping the original model masks and
-   combining them into a single mask; here is an example:
+It is possible to use ``mask_fillvalues`` to create a combined multimodel
+mask (all the masks from all the analyzed models combined into a single
+mask); for that purpose setting the ``threshold_fraction`` to 0 will not
+discard any time windows, essentially keeping the original model masks and
+combining them into a single mask; here is an example:
 
-   .. code-block:: yaml
+.. code-block:: yaml
 
-       preprocessors:
-         missing_values_preprocessor:
-           mask_fillvalues:
-             threshold_fraction: 0.0     # keep all missing values
-             min_value: -1e20            # small enough not to alter the data
-             #  time_window: 10.0        # this will not matter anymore
+    preprocessors:
+      missing_values_preprocessor:
+        mask_fillvalues:
+          threshold_fraction: 0.0     # keep all missing values
+          min_value: -1e20            # small enough not to alter the data
+          #  time_window: 10.0        # this will not matter anymore
 
 Minimum, maximum and interval masking
 -------------------------------------
--- a/esmvalcore/preprocessor/_mask.py
+++ b/esmvalcore/preprocessor/_mask.py
@@ -399,7 +399,10 @@ def count_spells(data, threshold, axis, spell_length):
         # just cope with negative axis numbers
         axis += data.ndim
     # Threshold the data to find the 'significant' points.
-    data_hits = data > threshold
+    if not threshold:
+        data_hits = data
+    else:
+        data_hits = data > threshold
     # Make an array with data values "windowed" along the time axis.
     ###############################################################
     # WARNING: default step is = window size i.e. no overlapping
@@ -520,7 +523,7 @@ def mask_outside_range(cube, minimum, maximum):
 
 def mask_fillvalues(products,
                     threshold_fraction,
-                    min_value=-1.e10,
+                    min_value=None,
                     time_window=1):
     """
     Compute and apply a multi-dataset fillvalues mask.
@@ -542,7 +545,8 @@ def mask_fillvalues(products,
         Must be between 0 and 1.
 
     min_value: float
-        minumum value threshold; default set to -1e10.
+        minumum value threshold; default None
+        If default, no thresholding applied so the full mask will be selected.
 
     time_window: float
         time window to compute missing data counts; default set to 1.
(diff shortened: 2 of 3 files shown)


valeriupredoi:
Good call, will replace it in this PR. About the dox - am not sure how to make it more spoonfeeding - maybe Bas can give some advice as you mention :beer:

bascrezee:
Sure, I can give some feedback. For me at first sight it is not clear that this preprocessor is a multimodel function. The nice box in `ESMValCore Preprocessor` (docs) helps a lot, maybe it would be good to also add some info to the doc string, such that it becomes clear when browsing the `ESMValCore API Reference` as well. 

Also I find it confusing that it is called `mask_fillvalues`, when reading in data, the `fillvalues` are already masked thanks to Iris correctly interpreting the CF conventions. This function actually goes two steps further, it (1) masks a certain time window if number of values are below a certain threshold and (2) it applies the masks from each single model into one mask applied to each model. I expected a separate preprocessor for step (2) that would be named something like `mask_multimodel`.

Not sure what to recommend concrete. Renaming this preprocessor is not really an option I guess. Two possible suggestions: 

a) make as default the following values: 
```
    mask_fillvalues:
      threshold_fraction: 0.0     # keep all missing values
      min_value: None            # small enough not to alter the data 
```
such that a user can simply call this in one line (advantage: user who just wants to apply the same mask to all datasets does not need to think about setting the right thresholds)
```
my_masking:
  mask_fillvalues  # apply same mask to all datasets
```

b) create a new preprocessor named `mask_multiproduct`. (additional advantage: the preprocessor has a more intuitive name):
```
def mask_multiproduct():
    mask_fillvalues(threshold_fraction=0.0, min_value: -1e20)
```

@BenMGeo might have input on this as well.

valeriupredoi:
Cheers @bascrezee - much appreciated :beer: Since you are such a nice dude wrt this issue, can I be cheeky and ask you to create a PR with the changes in the documentation that you regard as useful and making the use of the function more straightforward? You and @BenMGeo , if you guys do that, will receive a German :beer: each from me at the workshop :grin:

valeriupredoi:
I will plug the changed defaults now...

valeriupredoi:
also I defer the consideration to rename the function to @mattiarighi since I coded it up in his service back during the hedy times of ESMValTool v2.0SuperInitialWhenWeWereYoungIn2017 :grin:

mattiarighi:
`mask_multimodel` could be misleading, e.g. that this preprocessor is only applied to the dataset created by `multimodel_stats`.

`mask_multiproduct` is also confusing, since we are not using the term `product` anywhere (and it is usually applied to satellite data, rather than models).

I think `fillvalues` should remain in the preprocessor name, maybe we could change `mask` to something like `propagate`, `uniform` or `homogenize`.

BenMGeo:
what about `common_mask_all_fillvalues`. It's a bit lengthy, but may be describing the functionality at best.

mattiarighi:
or just `common_fillvalues` to make it shorter.

bascrezee:
> or just `common_fillvalues` to make it shorter.

Nice suggestions @BenMGeo and @mattiarighi, and how about `mask_common_fillvalues` ? 

I like it when the first part of the preprocessor describes the "action". Like `extract_levels`, `clip`, `fix_data`, etc.

mattiarighi:
Nice!

BenMGeo:
> 
> 
> > or just `common_fillvalues` to make it shorter.
> 
> Nice suggestions @BenMGeo and @mattiarighi, and how about `mask_common_fillvalues` ?
> 
> I like it when the first part of the preprocessor describes the "action". Like `extract_levels`, `clip`, `fix_data`, etc.

This can be misinterpreted. Why masking common fillvalues when they are already common?

bascrezee:
> > > or just `common_fillvalues` to make it shorter.
> > 
> > 
> > Nice suggestions @BenMGeo and @mattiarighi, and how about `mask_common_fillvalues` ?
> > I like it when the first part of the preprocessor describes the "action". Like `extract_levels`, `clip`, `fix_data`, etc.
> 
> This can be misinterpreted. Why masking common fillvalues when they are already common?

That is a good point. Hmmmm... the search continues. Some more thoughts: 

a) we can use a more general verb e.g. `apply` or `set` 

My suggestion than would be `set_common_mask`. This new name suggestion would be solely for a new preprocessor which basically wraps the existing (not renamed) preprocessor `mask_fillvalues`, like this:

```
def set_common_mask(products):
    mask_fillvalues(products, threshold_fraction=0.0, min_value: -1e20)
```

mattiarighi:
I think we should avoid wrappers: it would make recipes and documentation more confusing.

bascrezee:
> I think we should avoid wrappers: it would make recipes and documentation more confusing.

Does that mean that you are in favour of renaming the function `mask_fillvalues` ?

mattiarighi:
Yes, as I said [above](https://github.com/ESMValGroup/ESMValCore/pull/492#issuecomment-586213116).

BenMGeo:
> I think we should avoid wrappers: it would make recipes and documentation more confusing.

The question is, if we want to provide functionality to different levels of user experience. The wrapped one is easier to use and understand for beginners. 

If no wrappers are wanted, just merge it as is. We try to extend the documentation examples then.

bascrezee:
> If no wrappers are wanted, just merge it as is. 

Yes, I agree. I do not see a way of renaming `mask_fillvalues` in a way that reflects everything that can be done with this preprocessor. 

> We try to extend the documentation examples then.
Agree.

valeriupredoi:
you guys had a name-the-baby party and I wasn't here :grin: Can we then accept this PR (mind you, just a wee test) and sail into the sunset :ocean:

mattiarighi:
Waiting for @bouweandela's approval.

valeriupredoi:
@bouweandela says:
> The test looks fine.
We can forget about the However... bit :grin:

bouweandela:
>> We try to extend the documentation examples then.

>Agreed.

Who will do this? It would be good to get this done in this pull request, because the entire discussion lives here.

valeriupredoi:
:eyes: :eyes:

BenMGeo:
I just looked at the docs and I think it is already there. Maybe we should upgrade the information from a note to a section?!

valeriupredoi:
Cheers @BenMGeo - proves out I can't remember what I wrote, a case of old age surely :grin: @bouweandela all good wrt your comment? :beer:

PR #495, opened 2020-02-14
Remove attributes DimCoord and/or AuxCoord attribute from time coordinate

### LINKED ISSUE #490: Cube concatenation failure with IPSL cmip6 Atmos
**Describe the bug**
Running the `hyint` recipe on the IPSL-CM6A-LR cmip6 model fails when attempting to concatenate historical with the ssp585 pr day projection. The files look fine and can be concatenated with e.g. cdo. Both historical and ssp585 data run smoothly through the preprocessing independently.

Error message:
```
2020-02-13 13:16:56,433 UTC [56691] ERROR   From Cell(point=real_datetime(1850, 1, 1, 12, 0), bound=(real_datetime(1850, 1, 1, 0, 0), real_datetime(1850, 1, 2, 0, 0))) to Cell(point=real_datetime(2014, 12, 31, 12, 0), bound=(real_datetime(2014, 12, 31, 0, 0), real_datetime(2015, 1, 1, 0, 0)))

2020-02-13 13:16:56,437 UTC [56691] ERROR   From Cell(point=real_datetime(2015, 1, 1, 12, 0), bound=(real_datetime(2015, 1, 1, 0, 0), real_datetime(2015, 1, 2, 0, 0))) to Cell(point=real_datetime(2100, 12, 31, 12, 0), bound=(real_datetime(2100, 12, 31, 0, 0), real_datetime(2101, 1, 1, 0, 0)))

2020-02-13 13:16:56,437 UTC [56691] ERROR   Failed to run concatenate([<iris 'Cube' of precipitation_flux / (kg m-2 s-1) (time: 60265; latitude: 143; longitude: 144)>, <iris 'Cube' of precipitation_flux / (kg m-2 s-1) (time: 31411; latitude: 143; longitude: 144)>], {})

```

Here is the [main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/4199252/main_log_debug.txt).

Here is a test recipe:
[recipe_hyint_test_IPSL_cmip6_yml.txt].(https://github.com/ESMValGroup/ESMValCore/files/4199150/recipe_hyint_test_IPSL_cmip6_yml.txt)
It crosses 2014/2015 where the concatenation occurs. It takes a few minutes with these settings if everything is ok. If you change the dates make sure `norm_years` fall within and is shorter than the given range. 

@valeriupredoi  @jvegasbsc - same problem as #282?

valeriupredoi:
Closes #490 and is motivated by expert comment https://github.com/ESMValGroup/ESMValCore/issues/490#issuecomment-586322028

### DIFF
--- a/esmvalcore/cmor/check.py
+++ b/esmvalcore/cmor/check.py
@@ -580,7 +580,7 @@ class CMORCheck():
                     break
 
         if hasattr(coord, "attributes"):
-            coord.attributes = None
+            coord.attributes.pop('time_origin', None)
 
     @staticmethod
     def _simplify_calendar(calendar):
--- a/esmvalcore/cmor/check.py
+++ b/esmvalcore/cmor/check.py
@@ -579,8 +579,8 @@ class CMORCheck():
                     self.report_error(msg, var_name, freq)
                     break
 
-        if hasattr(coord, "attributes"):
-            coord.attributes.pop('time_origin', None)
+        # remove time_origin from attributes
+        coord.attributes.pop('time_origin', None)
 
     @staticmethod
     def _simplify_calendar(calendar):
--- a/esmvalcore/cmor/check.py
+++ b/esmvalcore/cmor/check.py
@@ -579,6 +579,9 @@ class CMORCheck():
                     self.report_error(msg, var_name, freq)
                     break
 
+        if hasattr(coord, "attributes"):
+            coord.attributes = None
+
     @staticmethod
     def _simplify_calendar(calendar):
         calendar_aliases = {
--- a/tests/unit/cmor/test_cmor_check.py
+++ b/tests/unit/cmor/test_cmor_check.py
@@ -169,10 +169,10 @@ class TestCMORCheck(unittest.TestCase):
 
     def test_check_with_time_attributes(self):
         """Test checks succeeds for a good cube with year."""
-        self.cube.coord('time').attributes["cow"] = "moo"
-        assert self.cube.coord('time').attributes["cow"] == "moo"
+        self.cube.coord('time').attributes['time_origin'] = "cow"
+        assert self.cube.coord('time').attributes['time_origin'] == "cow"
         self._check_cube()
-        assert not self.cube.coord('time').attributes
+        assert 'time_origin' not in self.cube.coord('time').attributes
 
     def test_check_bad_standard_name_auto_fix(self):
         """Test check pass for a bad standard_name with automatic fixes."""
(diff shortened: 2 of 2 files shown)


valeriupredoi:
arite, fellers, a quickie one just as pub o'clock is fast approaching :beer: -> @earnone 's recipe from #490 is running fine now

earnone:
@valeriupredoi sorry, don't know how to run ESMValCore once I checkout this branch.

valeriupredoi:
@earnone no need to run ESMValCore - if you checked out this branch and have installed using `pip` as in the instructions (`pip install -e .[develop]` in the esmvalcore repo) you are ready to run your recipe (the `esmvalcore` package is linked as a dependency of esmvaltool)

esmvalcore/cmor/check.py
```diff
@@ -579,6 +579,9 @@ def _check_time_coord(self):
                     self.report_error(msg, var_name, freq)
                     break
 
+        if hasattr(coord, "attributes"):
+            coord.attributes = None
```
zklaus: Hm. I think even an empty cube will have a `iris._cube_coord_common.LimitedAttributeDict` here. Maybe something like `coord.attributes.clear()` would be more appropriate? If one really wants to be so callous as to always destroy all the metadata.

valeriupredoi: but `coord.attributes = None` still preserves the `attributes` and it makes it an empty dict `{}` so I don't see why use `clear()` - would that be safer?

jvegasbsc: Do we really want to remove everything from all posible datasets?. I prefer to do 
```
coord.attributes.pop('time_origin', None)
```
This attribute looks redundant with the units, so it's unlikely to lose relevant information, but this may not be the case with other ones in other datasets

zklaus: @valeriupredoi, you are right. I personally find `.clear()` clearer.

Regardless, the `if` line can go since all `coord`s have the `attributes` attribute and I am with @jvegasbsc in that I'd prefer only to remove the offender.

valeriupredoi: Yeah I just remembered about attributes: positive: up/down for the Z coord that's really needed but note this is done only for the T coord - anything we want to keep from the time coords attrs? I think we should just nuke them all just so we don't run into this again - up to yous, it's all the same for me

zklaus: There could be. For example for climatological time axis. Let's do what we must and keep what we can.

valeriupredoi: good stuff, gents, cheers for the comments - Ive changed it so we remove just `time_origin` but kept the check on actual existence of attrs member, just in case - you know, stranger things may happen :beer:

zklaus: Like what? :thinking: But seriously, that's part of the api and we should avoid unnecessary bloat.

valeriupredoi: ok, boss, removed the if statement :beer:

PR #496, opened 2020-02-17
Test module for _main.py with different run scenarios

valeriupredoi:
No issue, just improving the testing for `_main.py`


valeriupredoi:
Yep, but the coverage is happening ok only if run in installation mode, that's why I wanted to add more testing even if run in just test mode. I mean, the ci machine runs the installation, but users and most developers won't. So unit tests will defo come in handy for this reason too - good call! what do you think - a set of unit tests for main.py that will perform imports of each of esmvalcore's components?

bouweandela:
Yes, create tests for the functions in _main.py only, so mock the imports using e.g. [unittest.mock.create_autospec](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.create_autospec) or [unittest.mock.Mock](https://docs.python.org/3/library/unittest.mock.html#the-mock-class) and/or [monkeypatch](https://docs.pytest.org/en/latest/monkeypatch.html).

valeriupredoi:
Monkey Patch :monkey: Will do on Monday, man :beer:

valeriupredoi:
I'll close this - it is incredibly old and not adding too much to what we need, frankly, I think the the tests for main have improved massively in the past...four years :rofl:

PR #507, opened 2020-02-21
Check horizontal grid before regridding

### LINKED ISSUE #485: regrid preprocessor regrids dataset though it has the same grid specification as target grid
The regrid preprocessor currently reacts "dumb" and regrids all datasets, even if a dataset is set as the targetgrid. This takes a serious amount of time, depending on your grid resolution, and has no visible benefits (at least for me).

Same if the horizontal grid is defined the same as the original dataset's.

BenMGeo:
Check if horizontal grid of cube is same as of target grid before regridding. Exits on first difference.
Note: No warning or feedback, if regridding is skipped as there is no logging functionality implemented in file yet.

**Tasks**

-   [x] Add unit tests

* * *

Closes #485


BenMGeo:
I'm not sure why the tests fail. Is it, because the tests assume different coordinate names?

valeriupredoi:
@BenMGeo the test is failing beause the first conditional in your function is failing ie `cube.coord(coord)` is a `MagicMock` object eg `<Mock name='mock.coord().shape' id='140630128663888'>` that differs from cube1 to cube2. In any case, please don't compare bounds at regrid level, especially horizontal regrid level, and here's why:

- cubes that get passed to the regrid module already have been checked on bounds existence by the CMOR checks and fixes module
- your conditional is way too strict ie 

```
                if (c1bndval != c2bndval
                        and not np.allclose(c1bndval, c2bndval)):
```

will fail many times since bounds are not always equal and the differences can be of order 1 that will immediately fail `allclose` test since `allclose` is sensitive to <1e-5 see https://docs.scipy.org/doc/numpy/reference/generated/numpy.allclose.html

- regridding should happen even if `latitude` or `longitude` are not x-y coord names -> your first conditional will return coord not found error
- as a good practice, if we really want to allow for bounds checks we should introduce it as a separate regrid preprocessor option that can be turned on or off by the user

:beer:

valeriupredoi (dismissed):
please see my comment :beer:

valeriupredoi:
right - so I totally misinterpreted the purpose of this PR - just saw the issue related to it :grin: For that case there is no need to go this deep in checks (note that the comments above still stand but they are irrelevant wrt to the purpose of the PR) - you can just check if `cube == target_grid` provided an extra check that `target_grid` is a cube :beer:

doc/esmvalcore/preprocessor.rst
```diff
@@ -44,7 +44,7 @@ Overview
    * ease of maintenance (including unit and integration testing) of existing
      routines;
    * a straightforward manner of importing and using the preprocessing routines as
-     part  of the overall usage of the software and, as a special case, the use
+     part  of the overall usage of the software and, as a special case, the useFH
```
bouweandela:
```suggestion
     part  of the overall usage of the software and, as a special case, the use
```

esmvalcore/preprocessor/_regrid.py
```diff
@@ -252,15 +252,64 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
...
+        Exits on first difference.
+    """
+    # Go through the 2 expected horizontal coordinates longitude and latitude.
+    for coord in ['latitude', 'longitude']:
+        # Compare shapes.
+        if cube1.coord(coord).shape != cube2.coord(coord).shape:
+            # Different shapes detected.
+            return False
+        else:
+            # Iterate through the combined bounds.
```
bouweandela: what about the points values, shouldn't those be equal too?

esmvalcore/preprocessor/_regrid.py
```diff
@@ -252,15 +252,64 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
...
+            for c1bndval, c2bndval in np.nditer(
```
bouweandela: this function looks a bit more complicated than needed, I think it would be fine to just have
```python
coord1 = cube1.coord(coord)
coord2 = cube2.coord(coord)
return (coord1.shape == coord2.shape
        and np.allclose(coord1.points, coord2.points)
        and np.allclose(coord1.bounds, coord2.bounds))
```
because the coordinate arrays are typically much smaller than the data chunks anyway and we can probably rely on numpy to implement optimizations

valeriupredoi: yeah, I was saying cube equality, that'll do straight out the box, minimal hassle

esmvalcore/preprocessor/_regrid.py
```diff
@@ -252,15 +252,64 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
                 [coord] = coords
                 cube.remove_coord(coord)
 
-    # Perform the horizontal regridding.
-    if _attempt_irregular_regridding(cube, scheme):
-        cube = esmpy_regrid(cube, target_grid, scheme)
-    else:
-        cube = cube.regrid(target_grid, HORIZONTAL_SCHEMES[scheme])
+    # Return unaltered cube if hotizontal grid is the same.
```
bouweandela: Note that the cube is not necessarily unaltered at this point, because the code above removes certain coordinates. Maybe this check should go before that?

bouweandela:
> I'm not sure why the tests fail. Is it, because the tests assume different coordinate names?

The faling test is trying to check that the following code is executed and the `coord` method is called with the correct arguments:
https://github.com/ESMValGroup/ESMValCore/blob/637bd47e6766d66dd15f453851c8d3891a785368/esmvalcore/preprocessor/_regrid.py#L237-L238
with the correct arguments as on the line above. The reason that the test fails, is that you have added an extra call to the `coord` method with a different argument `'longitude'` here:
https://github.com/ESMValGroup/ESMValCore/blob/77a255600752c45ecff26e9a9c43b78c5b93bfad/esmvalcore/preprocessor/_regrid.py#L293-L295

esmvalcore/preprocessor/_regrid.py
BenMGeo: Don't need to be. Assume the points are upper left coordinates, so bounds are more precise here.

esmvalcore/preprocessor/_regrid.py
BenMGeo: As said before, I would not check points. I targeted the use case that the artificial target grid is the same as the dataset, but the coordinates are differently set. within the bounds.

esmvalcore/preprocessor/_regrid.py
BenMGeo: maybe call it "unregridded". It is not going back to before what was done in this method, you are right.

BenMGeo:
> > I'm not sure why the tests fail. Is it, because the tests assume different coordinate names?
> 
> The faling test is trying to check that the following code is executed and the `coord` method is called with the correct arguments:
> 
> https://github.com/ESMValGroup/ESMValCore/blob/637bd47e6766d66dd15f453851c8d3891a785368/esmvalcore/preprocessor/_regrid.py#L237-L238
> 
> with the correct arguments as on the line above. The reason that the test fails, is that you have added an extra call to the `coord` method with a different argument `'latitude'` here:
> 
> https://github.com/ESMValGroup/ESMValCore/blob/77a255600752c45ecff26e9a9c43b78c5b93bfad/esmvalcore/preprocessor/_regrid.py#L293-L295

So, I assumed that usually the preprocessor already checked for these specific coord names before. I could simply add x and y, as alternative. Reading the coordinates is an issue as we would also read time and z-coord, if available.

BenMGeo:
> right - so I totally misinterpreted the purpose of this PR - just saw the issue related to it grin For that case there is no need to go this deep in checks (note that the comments above still stand but they are irrelevant wrt to the purpose of the PR) - you can just check if `cube == target_grid` provided an extra check that `target_grid` is a cube beer

This does not catch the case that a self defined target_grid is fitting the cube in the function "accidentally". This is why I tried to include the checks on a different level.

BenMGeo:
> It would be nice if you could also add a unit test for the new functionality

Will do. But I wanted to understand the failing of the current tests before.

esmvalcore/preprocessor/_regrid.py
BenMGeo: > this function looks a bit more complicated than needed, I think it would be fine to just have
> 
> ```python
> coord1 = cube1.coord(coord)
> coord2 = cube2.coord(coord)
> return (coord1.shape == coord2.shape
>         and np.allclose(coord1.points, coord2.points)
>         and np.allclose(coord1.bounds, coord2.bounds))
> ```
> 
> because the coordinate arrays are typically much smaller than the data chunks anyway and we can probably rely on numpy to implement optimizations

We have to only return True, if it is true for both horizontal dimensions. We still need an 'if', otherwise it would return true whenever the first one is ok.

bouweandela:
> So, I assumed that usually the preprocessor already checked for these specific coord names before.

Yes, the preprocessor has checked that the coordinates have the correct names, so I think it's probably fine to use the names to find the horizontal coordinates.

> I could simply add x and y, as alternative.

It would be good if you could have a careful look at the failing test. While simply using the same arguments in the second call to the `coord` method might make the test pass, I'm not sure it will still test what the author of the test intended, i.e. that this code is correctly executed:
https://github.com/ESMValGroup/ESMValCore/blob/637bd47e6766d66dd15f453851c8d3891a785368/esmvalcore/preprocessor/_regrid.py#L237-L238

BenMGeo:
> 
> 
> > So, I assumed that usually the preprocessor already checked for these specific coord names before.
> 
> Yes, the preprocessor has checked that the coordinates have the correct names, so I think it's probably fine to use the names to find the horizontal coordinates.
> 
> > I could simply add x and y, as alternative.
> 
> It would be good if you could have a careful look at the failing test. While simply using the same arguments in the second call to the `coord` method might make the test pass, I'm not sure it will still test what the author of the test intended, i.e. that this code is correctly executed:
> 
> https://github.com/ESMValGroup/ESMValCore/blob/637bd47e6766d66dd15f453851c8d3891a785368/esmvalcore/preprocessor/_regrid.py#L237-L238

So, the reason for the failed tests is, that grids are tested that are supposed to be regridded. Fails with the latitude call and, thus, I needed to add this call to the unit tests check.

I now focus on the unit tests for the new functionality. I assume they should be added in the same object.

BenMGeo:
> 
> 
> please see my comment 🍺

@valeriupredoi Not sure what I should review here:

Point 1: bounds do exist, so I could check if they are similar. I assume this is good?
Point 2: If differences are to big, the dataset should regrid. Do you suggest another treshold?
Point 3: I thought the CMOR checks mentioned in your Point 1 also check for the coordinate names, so I could assume lat/lon names?
Point 4: So should we call these checks only if we have something like a 'avoid_unnecessary_regriding: True' and 'accepted_closeness: 1e-5' options? I think any comparison fails above a specific grid deviation and we should simply have the same restrictions here. I assumed the default numpy closeness is the one used in iris.

BenMGeo:
I added a test. This is as close as I can get, as I do not fully understand the previous tests. I currently do not have experience with mock calls and have no idea how to set it up in the existing object.

tests/unit/preprocessor/_regrid/test_regrid.py
```diff
@@ -108,5 +117,46 @@ def test_regrid__cell_specification(self):
...
 
+class TestCloseness(tests.Test):
+    def test_regrid__closeness_cell_specification(self):
+        latitude = iris.coords.DimCoord(np.linspace(-85, 85, 18),
+                                        standard_name='latitude',
+                                        units='degrees')
+        longitude = iris.coords.DimCoord(np.linspace(5, 355, 36),
+                                         standard_name='longitude',
+                                         units='degrees')
+        latitude2 = iris.coords.DimCoord(np.linspace(-85, 85, 17),
```
bouweandela: Maybe you could give this coordinate the same shape as `latitude`, but different bounds? Because now only the first check (if the shape of the coordinates is the same) gets executed in the test, the second check (if the bounds are the same) is never executed in a test run.

BenMGeo: Yes you are Right. I only tested for shape. So I'd add 2 more cubes for longitude and latitude shifted Bounds, just to be complete?! Push a new test as soon as possible.

BenMGeo:
> It would be really great if you could also add a test that shows that cubes with a different grid actually get regridded to the target grid, while cubes with an identical grid are not regridded at all. There should be no need for mocking to do that ;-)

I thought that this is already included in the tests that are there at the moment (excluding shifted bounds).

bouweandela:
> I thought that this is already included in the tests that are there at the moment (excluding shifted bounds).

I checked and indeed there are tests that regridding works in `tests/integration/preprocessor/_regrid/test_regrid.py`. However, for the new functionality you added, i.e. regridding is skipped if the grids are the same, there is of course no test yet. Maybe you could add it there?

tests/unit/preprocessor/_regrid/test_regrid.py
bouweandela: Yes, that should be fine. Or you could just make one of the coordinates of equal shape but with different bounds, then you also have this case covered.

BenMGeo: I thought I would have to test that for both axes independently.

bouweandela: The minimum amount of testing I would recommend is that at least all code is executed once by a test, that way you're sure all code actually works. You're using the same code for both axes, because it's in a loop, so I think that testing for only one of the axes in one of the test cases should be OK.

Of course more thorough testing is better, but it's good to think about the balance between the amount of effort you put in creating the tests and subsequently maintaining those tests and how sure you want to be that the code does what it is supposed to do.

bouweandela:
Friendly ping @BenMGeo Any plans to complete this pull request?

BenMGeo:
> 
> 
> Friendly ping @BenMGeo Any plans to complete this pull request?

Planned for the next 2 weeks while processing huge data. hopefully can implement reasonable tests in the end.

stefsmeets:
Hi @BenMGeo , this PR has not seen much activity lately. No worries! Since it is so close to being finished, I hope you do not mind that I help out to get it over the finish line! 👍

stefsmeets:
Hi @bouweandela and @valeriupredoi , from my understanding there are still two open issues with this PR:

> It would be really great if you could also add a test that shows that cubes with a different grid actually get regridded to the target grid, while cubes with an identical grid are not regridded at all. There should be no need for mocking to do that ;-)

I have added a test for this case (`test_regrid_is_skipped_if_grids_are_the_same`). 

> Yes, that should be fine. Or you could just make one of the coordinates of equal shape but with different bounds, then you also have this case covered.

I added a test case with the same shape, but different bounds both lat and lon.

I also refactored the new tests a bit to use `pytest` instead of the old `unittest` library. Could you have a look whether your concerns have been addressed?

bouweandela (approved):
Thanks for the contribution @BenMGeo and thanks for wrapping this up @stefsmeets! :partying_face:

BenMGeo:
Thanks Peeps for finishing this. 
I'm sorry that I answer so late. Hard time to contribute further, these days :(

PR #514, opened 2020-02-24
Improving test coverage for _task.py

valeriupredoi:

**Tasks**

-   [x] Add unit tests

* * *

No issue open, just upping the test coverage for `_task.py` from 45% to 87% like a `Boss` :grin:


valeriupredoi:
hey @bouweandela I reckon am done here, pretty decent test coverage now, would you mind having a looksee maybe you can add another one or two tests? :beer:

tests/integration/test_task.py
```diff
@@ -80,3 +82,219 @@ def _run(self, input_files):
...
+
+    def _run(self, input_filesi=[]):
+        print(f'running task {self.name}')
+
+    monkeypatch.setattr(BaseTask, '_run', _run)
+
+    _run_tasks_parallel(tasks, 2)
+
+
+@pytest.mark.install
```
bouweandela: You could probably just monkeypatch `which` so it returns a fake path to the NCL executable instead of using the install marker, since you're not actually going to run the NCL executable

tests/integration/test_task.py
```diff
@@ -80,3 +82,219 @@ def _run(self, input_files):
...
+    """Run DiagnosticTask in parallel with OK NCL script."""
+    diagnostic_text = _py2ncl({'cow': 22}, 'tas')
+    tasks = _get_diagnostic_tasks(tmp_path, diagnostic_text, 'ncl')
+
+    def _run(self, input_filesi=[]):
+        print(f'running task {self.name}')
+
+    monkeypatch.setattr(BaseTask, '_run', _run)
+
+    _run_tasks_parallel(tasks, 2)
```
bouweandela: `run_tasks`, `_run_tasks_sequential`, and `_run_tasks_parallel` are already reasonably well tested with the current tests, so I do not think there is a need to test them more. Maybe you could just use the `run` method of the Task object to test running?

valeriupredoi:
yeah `master` is at 45% when running the regular tests, whereas this gets up to 70% :grin: and yes, 87% when running with `--addopts --installation` which the CI runs part of the `install` packet

I added a few tests which look more like unit tests to me but I didnt want to create a special test module in `/tests/unit` since I thought hey, since we're testing all here

tests/integration/test_task.py
valeriupredoi: no that actually runs the ncl executable via the diagnostic task, hence the two outcome cases - fail and warning

tests/integration/test_task.py
valeriupredoi: well they're wrappers of the `task.run()` object anyway no? I can just remove one of the tests so there's no double-testing, what you think?

tests/integration/test_task.py
bouweandela: I see. But why are you monkey patching `BaseTask._run`? If you would patch `DiagnosticTask._run` instead, there would be no need to run the NCL interpreter.

tests/integration/test_task.py
```diff
@@ -80,3 +82,219 @@ def _run(self, input_files):
...
+        _py2ncl([1, "cow"], 'tas')
+    assert 'NCL array cannot be mixed type:' in str(ex_err.value)
+    with pytest.raises(ValueError) as ex_err:
+        _py2ncl({"cow": 22}, None)
+    assert 'NCL does not support nested dicts:' in str(ex_err.value)
+
+
+def test_base_task():
+    """Test BaseTask basic attributes."""
+    tasks = set()
```
bouweandela: Remove the `tasks` set, it does not seem to add anything to this test, but it is confusing to the reader

tests/integration/test_task.py
```diff
@@ -80,3 +82,219 @@ def _run(self, input_files):
...
+    task_names = [task.name for task in tasks]
+    assert task_names == ['task0']
+    ancestor_names = [anc.name for anc in list(tasks)[0].ancestors]
+    assert ancestor_names == ['task0-ancestor0', 'task0-ancestor1']
+    task = list(tasks)[0]
+    assert task.script == diag_script
+    assert task.settings == diag_settings
+    assert task.output_dir == diag_output_dir
+
+    # diag doesn't exist
```
bouweandela: Code below here should probably go into a separate test

tests/integration/test_task.py
bouweandela: It would be better test smaller amounts of code per test. I know I have not been setting a very good example here, this is because of lack of time. It would probably be better to use the tests of @jvegasbsc as an example of how to write good tests (but do use pytest rather than the unittest.UnitTest class please) than anything I implemented.

tests/integration/test_task.py
valeriupredoi: I actually want to see the running of the task ie running the executables (python, ncl, R) with a small basic diag as part of the test - is that beyond the scope of this test purpose you reckon?

tests/integration/test_task.py
valeriupredoi: Good call! am only using pytest, will break them up in smaller bits

tests/integration/test_task.py
valeriupredoi: Was about to but then I got distracted with smth else :grin:

tests/integration/test_task.py
bouweandela: I think that's fine, but it would be really good to also have a version that doesn't require all these extra's installed.

valeriupredoi: Plan! I'll `mock` the calls and build those extra tests too, on Monday though :beer:

valeriupredoi:
hey @bouweandela it looks like running an `ncl` task is not working anymore in parallel:
```
def test_ncl_diagnostic_run_parallel_task_fails(monkeypatch, tmp_path):
    """Run DiagnosticTask in parallel with OK NCL script."""
    diagnostic_text = "cows on the [river]"
    tasks = _get_diagnostic_tasks(tmp_path, diagnostic_text, 'ncl')

    def _run(self, input_filesi=[]):
        print(f'running task {self.name}')

    if shutil.which('ncl') is not None:
        monkeypatch.setattr(BaseTask, '_run', _run)
        with pytest.raises(DiagnosticError) as err_mssg:
            _run_tasks_parallel(tasks, 2)
```
is now hanging

valeriupredoi:
ctrl-c-ing renders:
```
Process ForkPoolWorker-1:
  File "/home/valeriu/anaconda3R/envs/esmvaltool_py/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/home/valeriu/anaconda3R/envs/esmvaltool_py/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/valeriu/anaconda3R/envs/esmvaltool_py/lib/python3.8/multiprocessing/pool.py", line 114, in worker
    task = get()
  File "/home/valeriu/anaconda3R/envs/esmvaltool_py/lib/python3.8/multiprocessing/queues.py", line 356, in get
    res = self._reader.recv_bytes()
  File "/home/valeriu/anaconda3R/envs/esmvaltool_py/lib/python3.8/multiprocessing/connection.py", line 216, in recv_bytes
    buf = self._recv_bytes(maxlength)
  File "/home/valeriu/anaconda3R/envs/esmvaltool_py/lib/python3.8/multiprocessing/connection.py", line 414, in _recv_bytes
    buf = self._recv(4)
  File "/home/valeriu/anaconda3R/envs/esmvaltool_py/lib/python3.8/multiprocessing/connection.py", line 379, in _recv
    chunk = read(handle, remaining)
KeyboardInterrupt
```

valeriupredoi:
`_run_tasks_sequential(tasks)` works fine though

valeriupredoi:
solved by #534 - need to merge master here now

valeriupredoi:
let's get this one in @bouweandela - it already proved useful - making you create PR's at 10pm on a Monday evening :grin:

valeriupredoi:
waddup, @bouweandela ? :beer:

valeriupredoi:
le @bouweandela and segnor @jvegasbsc - let's get this in guys! :beer:

valeriupredoi:
Earth calling Major @bouweandela :guitar:

tests/integration/test_task.py
```diff
@@ -80,3 +83,245 @@ def _run(self, input_files):
...
+    diagnostic_text = "import os\n\nprint(cow)"
+    tasks = _get_diagnostic_tasks(tmp_path, diagnostic_text, 'py')
+
+    def _run(self, input_filesi=[]):
+        print(f'running task {self.name}')
+
+    monkeypatch.setattr(BaseTask, '_run', _run)
+
+    with pytest.raises(DiagnosticError) as err_mssg:
+        _run_tasks_sequential(tasks)
```
bouweandela, line 230: It would be better if you just used the `task.run()` method instead of the `_run_tasks_sequential` and `_run_tasks_parallel` methods, because these are already tested by `test_run_tasks`. It's better to test smaller amounts of code, because that 1) makes it easier to find where the bug is if a test fails and 2) requires fewer changes to the tests if you change the code.

tests/integration/test_task.py
```diff
@@ -80,3 +83,245 @@ def _run(self, input_files):
...
+        exp_mssg_2 = "diag_cow.ncl"
+        assert exp_mssg_1 in str(err_mssg.value)
+        assert exp_mssg_2 in str(err_mssg.value)
+    else:
+        with pytest.raises(DiagnosticError) as err_mssg:
+            tasks = _get_diagnostic_tasks(tmp_path, diagnostic_text, 'ncl')
+        exp_mssg_1 = "Cannot execute script "
+        exp_mssg_2 = "program 'ncl' not installed."
+        assert exp_mssg_1 in str(err_mssg.value)
+        assert exp_mssg_2 in str(err_mssg.value)
```
bouweandela, line 287: This test would test a smaller amount of code and would be more logically placed if you added it below this test:
https://github.com/ESMValGroup/ESMValCore/blob/54180f7aee4fbad0f6b44cf74963a9d956860c0c/tests/unit/test_task.py#L59-L61

tests/integration/test_task.py
```diff
@@ -80,3 +83,245 @@ def _run(self, input_files):
...
+            fil.write("import os\n\nprint(os.getcwd())")
+
+    task = DiagnosticTask(
+        name='task0',
+        ancestors=[BaseTask(name=f'task0-ancestor{j}') for j in range(2)],
+        script=diag_script,
+        settings=diag_settings,
+        output_dir=diag_output_dir,
+    )
+    tasks = {task}
```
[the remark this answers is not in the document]
valeriupredoi: because I am being silly - I have changed the code to account for me being serious now :grin:

tests/integration/test_task.py
```diff
@@ -80,3 +83,245 @@ def _run(self, input_files):
...
+def test_diagnostic_diag_script_none(tmp_path):
+    """Test case when diagnostic script doesnt exist."""
+    diag_script = tmp_path / 'diag_cow.py'
+    with pytest.raises(DiagnosticError) as err_msg:
+        _get_single_diagnostic_task(tmp_path, diag_script, write_diag=False)
+    diagnostics_root = os.path.join(DIAGNOSTICS_PATH, 'diag_scripts')
+    script_file = os.path.abspath(os.path.join(diagnostics_root, diag_script))
+    suf = "Cannot execute script "
+    ept = suf + "'{}' ({}): file does not exist.".format(script_file,
+                                                         script_file)
```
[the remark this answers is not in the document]
valeriupredoi: bleh, can't be arsed - it's that I can't fit the whole message in one line and I don't want to use `\` :grin:

tests/integration/test_task.py
valeriupredoi: done! good call, B-man

tests/integration/test_task.py
valeriupredoi: I dont get it - we are testing for the existence of ncl and R via `shutil`

tests/integration/test_task.py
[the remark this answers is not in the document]
bouweandela: You can use brackets to concatenate strings:
```python
>>> ('a' "b") == "ab"
True
```
instead of backslashes:
```python
    ept = (
        "Cannot execute script "
        "'{}' ({}): file does not exist.".format(script_file, script_file)
    )
```
but I think you could also just mention the `script_file` parameter only once?

tests/integration/test_task.py
bouweandela: Thanks, did you `git push` the changes? I cannot see them yet.

tests/integration/test_task.py
bouweandela: Sorry for not being clear :blush: What I mean is that the test I referred to above tests just the function `DiagnosticTask._initialize_cmd`, which raises the exception. However, there is no test yet for the cases where an exception is raised, only for the normal case. So from a code organization point of view it would be better to add this test there, because then it's easy to see that both normal run and exception are tested.

tests/integration/test_task.py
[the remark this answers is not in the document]
valeriupredoi: done with `()` nice one"!

tests/integration/test_task.py
valeriupredoi: ya, take off your sunglasses :sunglasses:

tests/integration/test_task.py
valeriupredoi: I still don't get it - you can't really use `parametrize` since the tests differ slightly in terms of the actual diagnostic to be executed - you want to check first for the existence of the executable **and** the interpreter correctly installed? Meh, think this is fine, man

bouweandela: This test should be taken apart into two separate tests, the bit if the executable is present can stay here, the bit if the executable is not present should be moved to tests/unit/test_task.py

valeriupredoi: yes makes sense, I thought I can get away but no chance :racehorse:

valeriupredoi: dang! Just realized you were all this time pointing to an existing code in the unit tests! sorry for making you chase around dude, yes, now I completely understand the call

tests/integration/test_task.py
bouweandela: Now I see it, you did it everywhere except above this comment ;-)

tests/integration/test_task.py
```diff
@@ -80,3 +83,229 @@ def _run(self, input_files):
...
+
+    task = DiagnosticTask(
+        name='task0',
+        ancestors=None,
+        script=diag_script.as_posix(),
+        settings=diag_settings,
+        output_dir=diag_output_dir.as_posix(),
+    )
+
+    return {task}
```
bouweandela, line 216: Why are you making a set with this task in it? In the tests below you only use the task and need extra code again to take it out of the set

tests/integration/test_task.py
```diff
@@ -80,3 +83,245 @@ def _run(self, input_files):
...
+
+def _get_single_base_task():
+    """Test BaseTask basic attributes."""
+    task = BaseTask(
+        name='task0',
+        ancestors=[BaseTask(name=f'task0-ancestor{j}') for j in range(2)],
+    )
+    for task0 in task.flatten():
+        task0.priority = 0
+    tasks = {task}
```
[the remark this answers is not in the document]
bouweandela: Could you have a look at this comment?

tests/integration/test_task.py
```diff
@@ -80,3 +83,229 @@ def _run(self, input_files):
     print(order)
     assert len(order) == 12
     assert order == sorted(order)
+
+
+@pytest.mark.parametrize('executables', ['ls', 'mv'])
+def test_which(executables):
```
bouweandela, line 89: there is no need to test this function anymore, because now it's just imported from the shutil library instead of having our own implementation

tests/integration/test_task.py
```diff
@@ -80,3 +83,229 @@ def _run(self, input_files):
...
+    diagnostic_text = 'var0 <- "zg"'
+
+    def _run(self, input_filesi=[]):
+        print(f'running task {self.name}')
+
+    if shutil.which('Rscript') is not None:
+        tasks = _get_diagnostic_tasks(tmp_path, diagnostic_text, 'R')
+        monkeypatch.setattr(BaseTask, '_run', _run)
+        for task in tasks:
+            task.run()
```
bouweandela, line 311: It would be much shorter to use `pytest.mark.parametrize` to combine these two tests (and `test_py_diagnostic_run_parallel_task` above) into a single parametrized test.

tests/integration/test_task.py
```diff
@@ -80,3 +83,229 @@ def _run(self, input_files):
...
+
+
+def test_ncl_diagnostic_run_parallel_task_fails(monkeypatch, tmp_path):
+    """Run DiagnosticTask in parallel with OK NCL script."""
+    diagnostic_text = "cows on the [river]"
+
+    def _run(self, input_filesi=[]):
+        print(f'running task {self.name}')
+
+    if shutil.which('ncl') is not None:
```
bouweandela, line 273: It would be better to skip the test if ncl is not installed, rather than silently passing it. You can use [`pytest.mark.skipif`](https://docs.pytest.org/en/latest/skipping.html#id1) to do that.

tests/unit/task/test_diagnostic_task.py
```diff
@@ -90,6 +91,18 @@ def test_initialize_cmd(ext_profile, cmd, tmp_path, monkeypatch):
 
     assert task.cmd == cmd
 
+    # test for no executable
+    monkeypatch.setattr(esmvalcore._task, 'which', lambda x: None)
+    if ext_profile[0] != '':
+        with pytest.raises(DiagnosticError) as err_mssg:
+            task = esmvalcore._task.DiagnosticTask(script,
```
bouweandela, line 98:
```suggestion
            esmvalcore._task.DiagnosticTask(script,
```
since you're not using the variable anywhere

tests/integration/test_task.py
valeriupredoi: :facepalm:

tests/integration/test_task.py
valeriupredoi: :facepalm:

tests/integration/test_task.py
valeriupredoi: yarp, good call, removed it - I was looking at it in a dubious manner yesterday but then I forgot I actually put it there myself :beer:

valeriupredoi:
ok @bouweandela I have now Bouwe-ified the test module (cheers for the pointers, I've learned a few new things!) :beer:

valeriupredoi:
@bouweandela we can include this in the release, you know... :smiling_face_with_three_hearts:

stefsmeets:
Hi @valeriupredoi @bouweandela , I hope you don't mind that I had a look at this PR. It seems really close to being ready, so I merged master and resolved the merge conflicts. Then I went through all the open conversations, and fixed one remaining review comment from @bouweandela (https://github.com/ESMValGroup/ESMValCore/pull/514#pullrequestreview-411802363). Would that be enough to get this merged?

tests/integration/test_task.py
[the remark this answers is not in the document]
stefsmeets: I simplified these two tests by getting rid of the set and just returning the single task 👍

valeriupredoi:
great many thanks @stefsmeets - what a hero, I completely forgot/buried this one under snow. Yes, I reckon this should be merged now, @bouweandela would you be OK to have a last look and push this baby :baby:

zklaus:
I was just chatting with Bouwe on a different issue. We'll push this one to 2.4.0 due to acute overload.

valeriupredoi:
sounds good, it's not something urgent, it may qualify for the title of the oldest PR to be merged :grin:

bouweandela (approved):

PR #523, opened 2020-02-27
Minor improvements to installation

bouweandela:
- Add codespell to CONTRIBUTING.md and as a developer dependency
- Remove unused stickler configuration file
- ~No longer install python-stratify from conda, it can be installed from PyPI~ -> this does not work when trying to run tests without first installing
- ~No longer install iris from conda, it can be installed from PyPI~ -> stopped trying because there were so many dependencies to install
- Unpin Python


setup.py
```diff
@@ -58,6 +58,7 @@
     # Use pip install -e .[develop] to install in development mode
     'develop': [
         'isort',
+        'codespell',
```
mattiarighi: Alphabetically listed?

PR #530, opened 2020-03-02
Unpin on Python<3.8 for conda package

valeriupredoi:
@bouweandela any reason you kept the pin in `meta.yml`? Anyways, I tested the unpinning and all works fine :beer:

### DIFF
--- a/meta.yaml
+++ b/meta.yaml
@@ -29,7 +29,7 @@ build:
 requirements:
   build:
     - git
-    - python>=3.6,<3.8
+    - python>=3.6
     # Normally installed via pip:
     - pytest-runner
     - setuptools_scm


valeriupredoi:
pretty dam weird why the `test_read_time_from_cube` fails in `installation` mode :confused:

bouweandela:
Thanks for the pull request! Indeed I overlooked the conda package.

> pretty dam weird why the test_read_time_from_cube fails in installation mode confused

That test failed for me too recently. I suspect it fails occasionally because it uses a randomly generated filename, so sometimes that will be a name containing something that looks sufficiently like a date for the function that is tested, so it will read the date from the filename instead of from the cube and fail because it's not the same.

valeriupredoi:
cool! I am plugging in a PR to correct for that transitory test fluke right now :beer:

PR #532, opened 2020-03-02
Test get start end year replace with tmp_dir/test.nc so no fluke dates can be found in temporary file name

valeriupredoi:
Motivated by transient test fails due to random combinations that look like dates as @bouweandela explains in https://github.com/ESMValGroup/ESMValCore/pull/530#issuecomment-593466236

### DIFF
--- a/tests/unit/data_finder/test_get_start_end_year.py
+++ b/tests/unit/data_finder/test_get_start_end_year.py
@@ -33,10 +33,9 @@ def test_get_start_end_year(case):
     assert case_end == end
 
 
-def test_read_time_from_cube():
+def test_read_time_from_cube(tmp_path):
     """Try to get time from cube if no date in filename"""
-    temp_file = tempfile.NamedTemporaryFile(suffix='.nc')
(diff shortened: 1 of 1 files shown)


tests/unit/data_finder/test_get_start_end_year.py
```diff
@@ -34,6 +36,7 @@ def test_get_start_end_year(case):
 def test_read_time_from_cube():
     """Try to get time from cube if no date in filename"""
     temp_file = tempfile.NamedTemporaryFile(suffix='.nc')
+    temp_file.name = re.sub(r'\d', 'v', temp_file.name)
```
bouweandela:
```suggestion
def test_read_time_from_cube(tmp_path):
    """Try to get time from cube if no date in filename"""
    temp_file = str(tmp_path / 'test.nc')
```
I would recommend using the [`tmp_path`](https://docs.pytest.org/en/latest/tmpdir.html) fixture instead, it's much easier to understand

valeriupredoi: absolutely! good call

PR #546, opened 2020-03-03
Implemented concatenation of cubes with derived coordinates

### LINKED ISSUE #544: Cubes lose derived coordinates during concatenation
**Describe the bug**
Due to a bug in `iris` (SciTools/iris#2478) , cubes with a derived coordinate (`AuxFactory`) lose this during the concatenation step of our preprocessor.

I will provide a PR for this.

schlunma:
Implemented for two special cases (hybrid height coordinate and hybrid pressure coordinate), can easily be expanded to more cases.

**Tasks**

-   [x] Add unit tests

* * *

Closes #544.

### DIFF
--- a/esmvalcore/preprocessor/_io.py
+++ b/esmvalcore/preprocessor/_io.py
@@ -13,6 +13,7 @@ import numpy as np
 import yaml
 
 from .._task import write_ncl_settings
+from ..cmor._fixes.shared import AtmosphereSigmaFactory
 from ._time import extract_time
 
 logger = logging.getLogger(__name__)
@@ -58,6 +59,19 @@ def _fix_aux_factories(cube):
         )
         for aux_factory in cube.aux_factories:
             if isinstance(aux_factory, iris.aux_factory.HybridHeightFactory):
+                break
+        else:
+            cube.add_aux_factory(new_aux_factory)
+
+    # Atmosphere sigma coordinate
+    if 'atmosphere_sigma_coordinate' in coord_names:
+        new_aux_factory = AtmosphereSigmaFactory(
+            pressure_at_top=cube.coord(var_name='ptop'),
+            sigma=cube.coord(var_name='lev'),
+            surface_air_pressure=cube.coord(var_name='ps'),
+        )
+        for aux_factory in cube.aux_factories:
+            if isinstance(aux_factory, AtmosphereSigmaFactory):
                 break
         else:
             cube.add_aux_factory(new_aux_factory)
--- a/esmvalcore/preprocessor/_io.py
+++ b/esmvalcore/preprocessor/_io.py
@@ -28,6 +28,41 @@ VARIABLE_KEYS = {
 }
 
 
+def _fix_aux_factories(cube):
+    """Fix :class:`iris.aux_factory.AuxCoordFactory` after concatenation.
+
+    Necessary because of bug in :mod:`iris` (see issue #2478).
+
+    """
+    coord_names = [coord.name() for coord in cube.coords()]
+
+    # Hybrid sigma pressure coordinate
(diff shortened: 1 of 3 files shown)


schlunma:
> nice addition, cheers @schlunma - also cool tests! One question/suggestion: maybe fixing coords should not happen on the fly in `_io.py` but rather in `fix.py`

I think in this specific case `_io.py` is more suited: it's closely related to the `concatenation` operation and can also be removed once this bug is fixed in `iris`. So it's not really a fix for the datasets, but rather a fix for `iris` :smile:

valeriupredoi:
right, good clarification! In that case I'd add a tag `iris_temporary_fix` in the function name, we could actually create a dedicated module in the `preprocessor/` for such things so it's very clear as to what needs to be removed (and also a separate test module that can be easily removed partly or entirely) - code clog is something that we're being hit by at the moment

valeriupredoi:
ping @bouweandela on this matter too :arrow_up:

schlunma:
I could also add a test that fails once the issue in `iris` is fixed, so we know when we can remove the relevant code.

valeriupredoi:
yeah that's an option too - make the fail message very angry so people don't ignore it :grinning:

bouweandela:
Can you please change the 'test if the iris issue has been fixed' so it emits a warning instead of fails? I'd prefer it if not everyone's CircleCI tests start failing when the next version of iris is released.

valeriupredoi:
> Can you please change the 'test if the iris issue has been fixed' so it emits a warning instead of fails? I'd prefer it if not everyone's CircleCI tests start failing when the next version of iris is released.

Hahahahah :rofl:

schlunma:
@bouweandela any further comments on this? Otherwise I think we can merge this

bouweandela:
It would be great if we could get a review from @jvegasbsc or @zklaus on this, as they have a much better understanding of the CF-conventions than me.

esmvalcore/preprocessor/_io.py
```diff
@@ -28,6 +28,41 @@
...
+def _fix_aux_factories(cube):
+    """Fix :class:`iris.aux_factory.AuxCoordFactory` after concatenation.
+
+    Necessary because of bug in :mod:`iris` (see issue #2478).
+
+    """
+    coord_names = [coord.name() for coord in cube.coords()]
+
+    # Hybrid sigma pressure coordinate
+    if 'atmosphere_hybrid_sigma_pressure_coordinate' in coord_names:
```
bouweandela: This looks like it could only work for atmospheric models? @ledm Any opinion?

schlunma: This can be easily extended to arbitrary [aux factories](https://scitools.org.uk/iris/docs/latest/iris/iris/aux_factory.html), but I don't have any data to test that/don't know if there are any issues which I'm not yet aware of. So I would suggest to stick with those variables in this PR.

esmvalcore/preprocessor/_io.py
```diff
@@ -28,6 +28,41 @@
...
 
+def _fix_aux_factories(cube):
+    """Fix :class:`iris.aux_factory.AuxCoordFactory` after concatenation.
+
+    Necessary because of bug in :mod:`iris` (see issue #2478).
+
+    """
+    coord_names = [coord.name() for coord in cube.coords()]
+
+    # Hybrid sigma pressure coordinate
```
valeriupredoi:
```suggestion
    # Hybrid sigma pressure coordinate
    # TODO possibly add support for other hybrid coordinates
```

schlunma:
> It would be great if we could get a review from @jvegasbsc or @zklaus on this, as they have a much better understanding of the CF-conventions than me.

@jvegasbsc @zklaus any comments? Sorry for the hurry, but I really need this for a diagnostic used in Axel's paper :grimacing:

PR #548, opened 2020-03-04
Implemented regridding of auxiliary coordinates

### LINKED ISSUE #545: Cubes lose auxiliary coordinates during regridding
**Describe the bug**
During our preprocessing step `regrid`, any cube with an `AuxCoord` that includes a horizontal dimension (like e.g. `surface_air_pressure`) loses this. This is caused by [`iris.cube.Cube.regrid`](https://scitools.org.uk/iris/docs/latest/iris/iris/cube.html#iris.cube.Cube.regrid) and probably an intended behavior (@bjlittle is this correct? If not then I will open an issue on the iris GitHub).

I think that at least for coordinates without bounds, the regridding should be possible. I will provide a PR for this. I encountered this problem while dealing with datasets that have a hybrid pressure coordinate and thus a `surface_air_pressure` coordinate.

schlunma:
**Tasks**

-   [x] Add unit tests

* * *

Closes #545.

### DIFF
--- a/esmvalcore/preprocessor/_regrid.py
+++ b/esmvalcore/preprocessor/_regrid.py
@@ -1,12 +1,13 @@
 """Horizontal and vertical regridding module."""
 
+import logging
 import os
 import re
 from copy import deepcopy
 
+import iris
 import numpy as np
 import stratify
-import iris
 from iris.analysis import AreaWeighted, Linear, Nearest, UnstructuredNearest
 from iris.util import broadcast_to_shape
 
@@ -15,6 +16,8 @@ from ..cmor.table import CMOR_TABLES
 from ._io import concatenate_callback, load
 from ._regrid_esmpy import ESMF_REGRID_METHODS
 from ._regrid_esmpy import regrid as esmpy_regrid
+
+logger = logging.getLogger(__name__)
 
 # Regular expression to parse a "MxN" cell-specification.
 _CELL_SPEC = re.compile(
@@ -61,12 +64,13 @@ VERTICAL_SCHEMES = ('linear', 'nearest',
 
 
 def parse_cell_spec(spec):
-    """
-    Parse an MxN cell specification string.
+    """Parse an MxN cell specification string.
 
     Parameters
     ----------
-    spec: str
+    spec : str
+        Cell specs given in the form ``'MxN'`` (``M``: Number of latitudes,
+        ``N``: number of longitudes).
 
     Returns
     -------
@@ -185,8 +189,45 @@ def _attempt_irregular_regridding(cube, scheme):
     return False
 
 
+def _add_regridded_aux_coords(cube, regridded_cube, target_grid):
+    """Get regridded :class:`iris.coords.AuxCoord` and corresponding dims."""
+    if regridded_cube.ndim != cube.ndim:
+        return
+    for coord in cube.coords(dim_coords=False):
+        if regridded_cube.coords(coord.name()):
+            continue
+        if coord.ndim < 2:
+            continue
+        if coord.bounds is not None:
+            logger.warning(
+                "Auxiliary coordinate '%s' cannot be regridded since it "
+                "contains bounds", coord.name())
+            continue
+        logger.debug(
+            "Regridding (bound-less) auxiliary coordinate '%s' which could "
+            "not be regridded by iris", coord.name())
+
+        # Broadcast to shape of whole cube
+        coord_dims = cube.coord_dims(coord)
+        points = iris.util.broadcast_to_shape(coord.core_points(),
+                                              cube.shape,
+                                              coord_dims)
+        aux_coord_cube = cube.copy(points)
+        aux_coord_cube = aux_coord_cube.regrid(
+            target_grid, Linear(extrapolation_mode='extrapolate'))
+
+        # Broadcast to original shape
+        slices = [0] * cube.ndim
+        for coord_dim in coord_dims:
+            slices[coord_dim] = slice(None)
+        aux_coord = coord.copy(aux_coord_cube[tuple(slices)].core_data())
+
+        # Add coordinate to regridded cube
+        regridded_cube.add_aux_coord(aux_coord, coord_dims)
+
+
 def extract_point(cube, latitude, longitude, scheme):
-    """Extract a point, with interpolation
+    """Extract a point, with interpolation.
 
     Extracts a single latitude/longitude point from a cube, according
     to the interpolation scheme `scheme`.
@@ -246,7 +287,6 @@ def extract_point(cube, latitude, longitude, scheme):
     array([ 1,  5, 17, 21, 33, 37, 49, 53])
 
     """
-
     msg = f"Unknown interpolation scheme, got {scheme!r}."
     scheme = POINT_INTERPOLATION_SCHEMES.get(scheme.lower())
     if not scheme:
@@ -332,11 +372,12 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
 
     # Perform the horizontal regridding.
     if _attempt_irregular_regridding(cube, scheme):
-        cube = esmpy_regrid(cube, target_grid, scheme)
+        regridded_cube = esmpy_regrid(cube, target_grid, scheme)
     else:
-        cube = cube.regrid(target_grid, HORIZONTAL_SCHEMES[scheme])
+        regridded_cube = cube.regrid(target_grid, HORIZONTAL_SCHEMES[scheme])
+        _add_regridded_aux_coords(cube, regridded_cube, target_grid)
 
-    return cube
+    return regridded_cube
 
 
 def _create_cube(src_cube, data, src_levels, levels, ):
@@ -356,6 +397,8 @@ def _create_cube(src_cube, data, src_levels, levels, ):
     data : array
         The payload resulting from interpolating the source cube
         over the specified levels.
+    src_levels : iris.coords.DimCoord or iris.coords.AuxCoord
+        Source levels.
     levels : array
         The vertical levels of interpolation.
 
--- a/esmvalcore/preprocessor/_regrid.py
+++ b/esmvalcore/preprocessor/_regrid.py
@@ -196,8 +196,6 @@ def _add_regridded_aux_coords(cube, regridded_cube, target_grid):
     for coord in cube.coords(dim_coords=False):
         if regridded_cube.coords(coord.name()):
             continue
-        if coord.ndim < 2:
-            continue
         if coord.bounds is not None:
             logger.warning(
                 "Auxiliary coordinate '%s' cannot be regridded since it "
--- a/tests/integration/preprocessor/_io/test_concatenate.py
+++ b/tests/integration/preprocessor/_io/test_concatenate.py
@@ -1,179 +1,18 @@
 """Integration tests for :func:`esmvalcore.preprocessor._io.concatenate`."""
 
 import unittest
-from unittest.mock import call
 
 import numpy as np
-import pytest
 from cf_units import Unit
-from iris.aux_factory import HybridHeightFactory, HybridPressureFactory
-from iris.coords import AuxCoord, DimCoord
-from iris.cube import Cube, CubeList
+from iris.coords import DimCoord
+from iris.cube import Cube
 from iris.exceptions import ConcatenateError
 
 from esmvalcore.preprocessor import _io
 
 
-@pytest.fixture
-def mock_empty_cube():
-    """Return mocked cube with irrelevant coordinates."""
-    cube = unittest.mock.create_autospec(Cube, spec_set=True, instance=True)
-    a_coord = AuxCoord(0.0, var_name='a')
-    b_coord = AuxCoord(0.0, var_name='b')
-    cube.coords.return_value = [a_coord, b_coord]
-    return cube
-
-
-@pytest.fixture
-def mock_hybrid_height_cube():
-    """Return mocked cube with hybrid height coordinate."""
-    cube = unittest.mock.create_autospec(Cube, spec_set=True, instance=True)
-    lev_coord = AuxCoord([1.0], bounds=[[0.0, 2.0]], var_name='lev', units='m')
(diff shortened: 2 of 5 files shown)


valeriupredoi:
may be out of context but here's something that may be of use - was just sifting through iris issues waiting for a `du -h BIG_DIR` to finish https://github.com/SciTools/iris/issues/3653

esmvalcore/preprocessor/_regrid.py
```diff
@@ -185,8 +189,45 @@ def _attempt_irregular_regridding(cube, scheme):
...
 
+def _add_regridded_aux_coords(cube, regridded_cube, target_grid):
+    """Get regridded :class:`iris.coords.AuxCoord` and corresponding dims."""
+    if regridded_cube.ndim != cube.ndim:
+        return
+    for coord in cube.coords(dim_coords=False):
+        if regridded_cube.coords(coord.name()):
+            continue
+        if coord.ndim < 2:
+            continue
```
valeriupredoi: can you add a `logger.debug()` message here?

esmvalcore/preprocessor/_regrid.py
```diff
@@ -185,8 +189,45 @@ def _attempt_irregular_regridding(cube, scheme):
...
+            "not be regridded by iris", coord.name())
+
+        # Broadcast to shape of whole cube
+        coord_dims = cube.coord_dims(coord)
+        points = iris.util.broadcast_to_shape(coord.core_points(),
+                                              cube.shape,
+                                              coord_dims)
+        aux_coord_cube = cube.copy(points)
+        aux_coord_cube = aux_coord_cube.regrid(
+            target_grid, Linear(extrapolation_mode='extrapolate'))
```
valeriupredoi: any reason why we fix the extrapolation method?

esmvalcore/preprocessor/_regrid.py
```diff
@@ -185,8 +189,45 @@ def _attempt_irregular_regridding(cube, scheme):
...
+
+        # Broadcast to original shape
+        slices = [0] * cube.ndim
+        for coord_dim in coord_dims:
+            slices[coord_dim] = slice(None)
+        aux_coord = coord.copy(aux_coord_cube[tuple(slices)].core_data())
+
+        # Add coordinate to regridded cube
+        regridded_cube.add_aux_coord(aux_coord, coord_dims)
```
valeriupredoi: maybe make sure the `regridded_cube` doesn't already have the aux coord? It shouldn't, right?

esmvalcore/preprocessor/_regrid.py
```diff
@@ -332,11 +372,12 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
 
     # Perform the horizontal regridding.
     if _attempt_irregular_regridding(cube, scheme):
-        cube = esmpy_regrid(cube, target_grid, scheme)
+        regridded_cube = esmpy_regrid(cube, target_grid, scheme)
     else:
-        cube = cube.regrid(target_grid, HORIZONTAL_SCHEMES[scheme])
+        regridded_cube = cube.regrid(target_grid, HORIZONTAL_SCHEMES[scheme])
+        _add_regridded_aux_coords(cube, regridded_cube, target_grid)
```
valeriupredoi: ah okay so a check on the existence of the aux coord is rather needed then

esmvalcore/preprocessor/_regrid.py
schlunma: The check in the first line is the check that the coordinate is not present in the regridded cube (as you requested in your comment below), the first check can probably be removed (originally I only had this and not the first check, but I see no reason why this should not work for 1D coordinates.

valeriupredoi:
also add `ndim` as Mock Cube attribute in tests/unit/preprocessor/_regrid/test_regrid.py

esmvalcore/preprocessor/_regrid.py
schlunma: Do other methods make sense for coordinates? Nearest neighbor? Area weighting?

esmvalcore/preprocessor/_regrid.py
schlunma: see comment above, l. 197

esmvalcore/preprocessor/_regrid.py
schlunma: See comment above, l. 197

esmvalcore/preprocessor/_regrid.py
valeriupredoi: aye :beer:

schlunma:
> also add `ndim` as Mock Cube attribute in tests/unit/preprocessor/_regrid/test_regrid.py

It is there, see e.g. l.35

schlunma:
> is there no way to add bounds to aux coords in iris so that we can jump over the constraint that is there be no bounds there be no regridding? In a peasant way you can promote that coord to DimCoord and use guess_bounds()

My peasant way of doing it would be the same approach as done for the points now: Create `n` new cubes for each edge of the bounds (in most cases `n=2`) and regrid them separately. I think `guess_bounds` does not work on multidimensional coordinates.

esmvalcore/preprocessor/_regrid.py
valeriupredoi: I dunno, preserving the original functional argument `HORIZONTAL_SCHEMES[scheme]` seems a bit more general, if stuff hits the fan we can always go back and restrict

valeriupredoi:
> > also add `ndim` as Mock Cube attribute in tests/unit/preprocessor/_regrid/test_regrid.py
> 
> It is there, see e.g. l.35

sorry - bad advice from my side! both `src_cube` and `tgt_grid` as Mock members from the UnitTest are built correctly but `regridded_cube` is a floating Sentinel object that gets passed around to `regrid()` func and has no attributes - I'd recommend either keeping the in place transformation of the `cube` member in:
```
    # Perform the horizontal regridding.
    if _attempt_irregular_regridding(cube, scheme):
        regridded_cube = esmpy_regrid(cube, target_grid, scheme)
    else:
        regridded_cube = cube.regrid(target_grid, HORIZONTAL_SCHEMES[scheme])
        _add_regridded_aux_coords(cube, regridded_cube, target_grid)

    return regridded_cube
```
or if you need this current implementation, then create a `regridded_cube` Mock object in the test and for Pete's sake give him `ndim` :grin:

valeriupredoi:
> > is there no way to add bounds to aux coords in iris so that we can jump over the constraint that is there be no bounds there be no regridding? In a peasant way you can promote that coord to DimCoord and use guess_bounds()
> 
> My peasant way of doing it would be the same approach as done for the points now: Create `n` new cubes for each edge of the bounds (in most cases `n=2`) and regrid them separately. I think `guess_bounds` does not work on multidimensional coordinates.

Correct! According to iris bible:
```
An error is raised if the coordinate already has bounds, is not one-dimensional, or is not monotonic.
```
Let's keep it this way then (the way you have it)

esmvalcore/preprocessor/_regrid.py
schlunma: I think that scientifically only `linear` makes sense, but I'm not 100% sure. Consider this example:

Original grid (`lat x lon`): `[1.0, 2.0] x [1.0, 2.0]`
Target grid: `[1.25, 1.75] x [1.25, 1.75]`

If we have an 2D auxiliary coordinate that is basically identical to `lat` coord, i.e.

```
[[1.0, 1.0], [2.0, 2.0]]
```

I except that this is regridded to

```
[[1.25, 1.25], [1.75, 1.75]]
```

That is only true for the `linear` regridding.

schlunma:
> > > also add `ndim` as Mock Cube attribute in tests/unit/preprocessor/_regrid/test_regrid.py
> > 
> > 
> > It is there, see e.g. l.35
> 
> sorry - bad advice from my side! both `src_cube` and `tgt_grid` as Mock members from the UnitTest are built correctly but `regridded_cube` is a floating Sentinel object that gets passed around to `regrid()` func and has no attributes - I'd recommend either keeping the in place transformation of the `cube` member in:
> 
> ```
>     # Perform the horizontal regridding.
>     if _attempt_irregular_regridding(cube, scheme):
>         regridded_cube = esmpy_regrid(cube, target_grid, scheme)
>     else:
>         regridded_cube = cube.regrid(target_grid, HORIZONTAL_SCHEMES[scheme])
>         _add_regridded_aux_coords(cube, regridded_cube, target_grid)
> 
>     return regridded_cube
> ```
> 
> or if you need this current implementation, then create a `regridded_cube` Mock object in the test and for Pete's sake give him `ndim` grin

No, in the unit test `cube` and `regridded_cube` are `Mock` objects, `target_grid` is simply set to the string `'target_grid'` because its only used in this call

https://github.com/ESMValGroup/ESMValCore/blob/d8d607385a9362dc2567ca4c99d47e85a5fe512c/esmvalcore/preprocessor/_regrid.py#L216-L217

and not for anything else (no calls, no attributes, no methods), so it doesn't need to be mocked.

schlunma:
Ahh, now I get what you mean @valeriupredoi, we were talking about two different files. Sorry, will fix the test now

valeriupredoi:
> Ahh, now I get what you mean @valeriupredoi, we were talking about two different files. Sorry, will fix the test now

yeah sorry mate, had a flash meeting that's why I stopped replying :grin:

valeriupredoi:
if you write something similar to `return_mock_stock_cube()` to patch `add_regridded_aux_coords()` should do I reckons

esmvalcore/preprocessor/_regrid.py
valeriupredoi: what are the other methods giving?

valeriupredoi: my worry is not necessarily a technical one but rather - if one method is applied to DimCoords and another method to AuxCoords that is mixing statistics in one operation, bit of a black box for the user and that might bite at some point and nobody will know why (all silent). If so then I'd add a logger message saying `Hell man, you might in trouble, am changing your regridding method...`

schlunma: `nearest` gives

```
[[1.0, 1.0], [2.0, 2.0]]
```

`area_weighted` gives

```
[[--, --], [--, 2.0]]
```

valeriupredoi: Gah, pretty funky! OK keep the fixed one then, clearly, but I'd the message still pls, we should be as transparent as possible :tumbler_glass:

schlunma: > my worry is not necessarily a technical one but rather - if one method is applied to DimCoords and another method to AuxCoords that is mixing statistics in one operation, bit of a black box for the user and that might bite at some point and nobody will know why (all silent). If so then I'd add a logger message saying `Hell man, you might get in trouble, am changing your regridding method...`

But the `DimCoords` are technically not regridded; they rather define what the target grid looks like, so they are fixed. Like I said, I'm not 100% sure if the regridding of other aux coords makes scientifically sense in the first place (also since `iris` does obviously not support it), but my intuition tells me to use a linear scheme. Maybe @zklaus can comment on this since he's done the irregular regridding?

valeriupredoi:
*writes awesome mock and patch tests*/*has codacy complain too many blank lines* :rofl:

esmvalcore/preprocessor/_regrid.py
zklaus: I think doing it linearly is indeed the only thing that might make some sense, so I think this conversation can be resolved. I'll but general comments below.

valeriupredoi: Ok let's but this converstion :grin:

valeriupredoi:
waddup, guys? :beer:

schlunma:
> waddup, guys? beer

I think @zklaus had some comments on this, right?

zklaus:
I am not sure regridding is generally applicable to aux coords. I'd like to see the actual use case that triggered that to get a better idea. @schlunma, could you mention the issue of the use case here (again?)?

schlunma:
Sure thing. Some FGOALS model data has a derived coordinate `atmosphere_sigma_coordinate` which cannot be read by iris automatically (see #564). Since we save nc files at then end of the preprocessing and load them again in the diagnostic, adding an `aux_factory` to those files during the preprocessing step is useless, since at loading step `iris` does not automatically know about this `aux_factory` and we have to add it manually again in the diagnostic.

For this, it is necessary that the file contains the `surface_pressure` coordinate (which is an aux coord). Without regridding, this works fine, but as mentioned, regridding drops this coordinate.

The funny thing about that is that `iris` does **not** drop these coordinates if they belong to one of the [aux factories](https://scitools.org.uk/iris/docs/latest/iris/iris/aux_factory.html) that `iris` supports automatically. Otherwise, every aux coord is lost after `iris.cube.Cube.regrid`.

esmvalcore/preprocessor/_regrid.py
```diff
@@ -185,8 +189,43 @@ def _attempt_irregular_regridding(cube, scheme):
...
+            continue
+        logger.debug(
+            "Regridding (bound-less) auxiliary coordinate '%s' which could "
+            "not be regridded by iris", coord.name())
+
+        # Broadcast to shape of whole cube
+        coord_dims = cube.coord_dims(coord)
+        points = iris.util.broadcast_to_shape(coord.core_points(),
+                                              cube.shape,
+                                              coord_dims)
```
zklaus: Do I understand correctly that this has the potential to incur a hefty additional cost? A 1d aux coord, even a scalar coord might be blown into a long timeseries of 3d data, right? If I understand [iris](https://github.com/SciTools/iris/blob/06e9e2b5eae6e903e486ce12162c242d25ee4e5d/lib/iris/analysis/_regrid.py#L1003) correctly, this is not necessary. Source and target only need to share x and y dim. What happens without the broadcast?

schlunma:
Okay, I justed tested again with the FGOALS model and I realized that adding `atmosphere_sigma_coordinate` to #546 solves the problem without this fix. I completely forgot about this, sorry!

I'm really sorry for wasting your time with this PR @valeriupredoi @zklaus!

zklaus:
Don't worry. Might still be an interesting concept in the future.

PR #529, opened 2020-03-02
Added various fixes for hybrid pressure coordinates

schlunma:
Added various fixes for CMIP5 and CMIP6 models concerning hybrid pressure coordinates (variables `cl`, `cli` and `clw`).

**Tasks**

-   [x] [Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do, if you haven't done so already (and add the link at the bottom)
-   [x] This pull request has a descriptive title that can be used in a changelog
-   [x] Add unit tests
-   [x] Public functions should have a numpy-style docstring so they appear properly in the [API documentation](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/api/esmvalcore.html). For all other functions a one line docstring is sufficient.
-   [x] If writing a new/modified preprocessor function, please update the [documentation](https://esmvaltool.readthedocs.io/projects/esmvalcore/en/latest/esmvalcore/preprocessor.html)
-   [x] Circle/CI tests pass. Status can be seen below your pull request. If the tests are failing, click the link to find out why.
-   [ ] Codacy code quality checks pass. Status can be seen below your pull request. If there is an error, click the link to find out why. If you suspect Codacy may be wrong, please ask by commenting.
-   [x] Please use `yamllint` to check that your YAML files do not contain mistakes
-   [x] If you make backward incompatible changes to the recipe format, make a new pull request in the [ESMValTool repository](https://github.com/ESMValGroup/ESMValTool) and add the link below

If you need help with any of the tasks above, please do not hesitate to ask by commenting in the issue or pull request.

* * *

Related to #536, #540, #541, #542. (Note: Automatic closing of these issues is not desired; they should be closed when the issue is reported to the modelers).

PR #551, opened 2020-03-05
Added clw, cli, clcalipso to CMIP6 fixes

axel-lauer:
Added variables clw, cli, clcalipso to CMIP6 fix files.


schlunma:
Included in #529, #559, #562 and #565.

Issue #537, opened 2020-03-03
Dataset problem: Issues with latitude and longitude for tos of BCCs CMIP6 models


schlunma:
**Describe the dataset issue**
The names for the 1D `grid_latitude` and `grid_longitude` on irregular grids are set to `latitude` and `longitude`, causing `iris` to find duplicate coordinates. Multidimensional `latitude` and `longitude` coordinate do not have bounds.

- ~~Data file has been changed by you in any way~~
- Project (CMIP5/CMIP6/OBS/obs4mips/etc): CMIP6
- Full dataset description (fill out as many as you know, please):
  - dataset: BCC-CSM2-MR, BCC-ESM1
  - experiment: historical
  - mip: Omon
  - grid: gn
  - variable used: tos
- Pointer to existing copy of data on ESGF node (it would be very useful if you could provide a physical fullpath to the file(s) that are causing the problem, e.g. on CEDA Jasmin or DKRZ):
`/mnt/lustre02/work/ik1017/CMIP6/data/CMIP6/CMIP/BCC/BCC-CSM2-MR/historical/r1i1p1f1/Omon/tos/gn/v20181126/tos_Omon_BCC-CSM2-MR_historical_r1i1p1f1_gn_185001-201412.nc`
`/mnt/lustre02/work/ik1017/CMIP6/data/CMIP6/CMIP/BCC/BCC-ESM1/historical/r1i1p1f1/Omon/tos/gn/v20181129/tos_Omon_BCC-ESM1_historical_r1i1p1f1_gn_185001-201412.nc`

ncdump (without global attributes):
```
netcdf tos_Omon_BCC-CSM2-MR_historical_r1i1p1f1_gn_185001-201412 {
dimensions:
        time = UNLIMITED ; // (1980 currently)
        lat = 232 ;
        lon = 360 ;
        bnds = 2 ;
variables:
        double time(time) ;
                time:bounds = "time_bnds" ;
                time:units = "days since 1850-1-1" ;
                time:calendar = "365_day" ;
                time:axis = "T" ;
                time:long_name = "time" ;
                time:standard_name = "time" ;
        double time_bnds(time, bnds) ;
        double lat(lat) ;
                lat:bounds = "lat_bnds" ;
                lat:units = "degrees_north" ;
                lat:axis = "Y" ;
                lat:long_name = "latitude" ;
                lat:standard_name = "latitude" ;
        double lat_bnds(lat, bnds) ;
        double lon(lon) ;
[...]

PR #562, opened 2020-03-09
Added various fixes for hybrid height coordinates

schlunma:
This PR adds fixes for `cl`, `cli` and `clw` datasets with a hybrid height coordinate.

**Tasks**

-   [x] Add unit tests

* * *

Related to #537. (Note: Automatic closing of these issues is not desired; they should be closed when the issue is reported to the modelers).


schlunma:
Please do not merge yet, I will refactor this similar to #529.

> Holy cheese, there are a **lot** of fixes here - well done, man! One comment: it would be very useful if we documented proper all funcs in ` esmvalcore/cmor/_fixes/shared.py` so that they can be used via the API - not necessarily here, but we should have that in mind (open an issue) for the future beer

Yes, definitely! I'll add proper docstrings for all public method

schlunma:
I refactored all the code, now it's ready for another review.

Please note that I had to adapt some fixes introduce in #529 since they were not working properly (I didn't test this thoroughly, sorry!): By doing a simple `Cli = ClFixForHybridPressure`, the function [`Fix.get_cube_from_list`](https://github.com/ESMValGroup/ESMValCore/blob/e5827ad0f9121c5851c7686ff57bcd08f9e55664/esmvalcore/cmor/_fixes/fix.py#L70-L96) did not work anymore, since it relies on the name of the class to find the correct `cube` out of `cubes`. Since `Cli = ClFixForHybridPressure` simply adds a new reference for the class `ClFixForHybridPressure`, `Cli.__name__` still gives `'ClFixForHybridPressure'`, leading to an exception for the variables `cli` and `clw`.

I fixed this by introducing new classes

```python
Cli = CliFixHybridHeightCoord
```

with

```python
class CliFixHybridHeightCoord(ClFixHybridHeightCoord):
    """Fixes for ``cli`` regarding hybrid sigma height coordinates."""

    SHORT_NAME = 'cli'
```

with the static attribute `SHORT_NAME`, which is used as parameter for [`Fix.get_cube_from_list`](https://github.com/ESMValGroup/ESMValCore/blob/e5827ad0f9121c5851c7686ff57bcd08f9e55664/esmvalcore/cmor/_fixes/fix.py#L70-L96). I think without using metaclasses this is the most elegant solution.

I tested all affected models and variables (`cl`, `cli` and `clw`), everything works as expected now.

schlunma:
Any comments on this @zklaus @valeriupredoi?

valeriupredoi:
> I refactored all the code, now it's ready for another review.
Great! I am going to have a look in a jiffy, trying to sort out smth else right at the moment. You don't happen to know how to assign `dtype` to an iris cube so you change data type from float64 to 32 but keeping the data lazy? :beer:

schlunma:
> > I refactored all the code, now it's ready for another review.
> 
> Great! I am going to have a look in a jiffy, trying to sort out smth else right at the moment. You don't happen to know how to assign `dtype` to an iris cube so you change data type from float64 to 32 but keeping the data lazy? beer

Great, thanks!!

I think something like this should work, just replace `coord.core_points` by `cube.core_data`.

https://github.com/ESMValGroup/ESMValTool/blob/322cc0b98e257113d6a6d6abb1e9a4ff1716dac9/esmvaltool/cmorizers/obs/utilities.py#L294-L312

bouweandela:
The changes mentioned in this https://github.com/ESMValGroup/ESMValCore/pull/562#issuecomment-598723119 should not be needed.

There is a much simpler solution, which is to use the name from the CMOR table definition. I already added that here https://github.com/ESMValGroup/ESMValCore/pull/391/files#diff-dcd9100a443a7b362eceac557168cc7b, but unfortunately @jvegasbsc does not seem to have time fix the corresponding ESMValTool PR for that one, so it has become stale and it will probably take some time before it gets merged. Maybe you can simply copy the changes from that other pull request?

Alternatively, I guess you could simply inherit instead of assign, i.e. `class Cli(ClFixForHybridPressure): pass` instead of `Cli = ClFixForHybridPressure`?

schlunma:
> Maybe you can simply copy the changes from that other pull request?

Sounds great!

schlunma:
@bouweandela I implemented your suggestion and refactored the code again (hopefully for the last time now). Can you take another look at it? Thanks!

schlunma:
> @bouweandela I implemented your suggestion and refactored the code again (hopefully for the last time now). Can you take another look at it? Thanks!

@bouweandela? Sorry for the hurry, but I really need this for a diagnostic used in Axel's paper :grimacing:

PR #587, opened 2020-03-25
Added derivation script for co2s

### LINKED ISSUE #586: Add derivation script for co2 mole fraction at surface
The variable `co2` (`mole_fraction_of_carbon_dioxide_in_air`) is 3D. I need a 2D version of this variable which gives the CO2 mole fraction at surface level.

For unmasked datasets this is trivial (just use the lowest pressure level), but for masked this is a little bit more tricky (extract the lowest value which is not masked for every grid cell).

I will open a PR for this.

schlunma:
This PR adds the derivation script for `co2s` (CO2 mole fraction at surface level). It can be tested with this sample recipe:

```yaml
# recipe_test.yml
---
documentation:

  description: |
    Test.

  authors:
    - schlund_manuel

  maintainer:
    - schlund_manuel

  projects:
    - crescendo


DATASET_ANCHOR: &datasets
  - {dataset: BNU-ESM}
  - {dataset: CanESM2}
  - {dataset: CESM1-BGC}
  - {dataset: GFDL-ESM2G}
  - {dataset: GFDL-ESM2M}
  - {dataset: MIROC-ESM}
  - {dataset: NorESM1-ME}


diagnostics:

  test:
    variables:
      co2s:
        project: CMIP5
        exp: esmHistorical
        ensemble: r1i1p1
        mip: Amon
        additional_datasets: *datasets
        start_year: 1950
        end_year: 1955
        derive: true
    scripts:
      null
```

Please note that using `core_data()` instead of `data` [here](https://github.com/ESMValGroup/ESMValCore/blob/dede748b065741e76590982ca0e1c5e5b94a7a40/esmvalcore/preprocessor/_derive/co2s.py#L35) results in an `NotImplementedError: Don't yet support nd fancy indexing` of `dask`.

**Tasks**

-   [x] Add unit tests

* * *

Closes #586.

### DIFF
--- a/esmvalcore/cmor/tables/custom/CMOR_co2s.dat
+++ b/esmvalcore/cmor/tables/custom/CMOR_co2s.dat
@@ -1,0 +1,20 @@
+SOURCE: CMIP5
+!============
+variable_entry:    co2s
+!============
+modeling_realm:    atmos
+!----------------------------------
+! Variable attributes:
+!----------------------------------
+standard_name:     mole_fraction_of_carbon_dioxide_in_air
+units:             mol mol-1
+cell_methods:      time: mean
+cell_measures:     area: areacella
+long_name:         Mole Fraction of CO2 at surface level
+!----------------------------------
+! Additional variable information:
+!----------------------------------
+dimensions:        longitude latitude time
+type:              real
+!----------------------------------
+!
--- a/esmvalcore/cmor/_fixes/cmip5/cesm1_bgc.py
+++ b/esmvalcore/cmor/_fixes/cmip5/cesm1_bgc.py
@@ -5,29 +5,6 @@ from dask import array as da
 from ..fix import Fix
 
 
-class Co2(Fix):
-    """Fixes for co2 variable."""
-
-    def fix_data(self, cube):
-        """Fix data.
-
-        Fixes discrepancy between declared units and real units
-
-        Parameters
-        ----------
-        cube: iris.cube.Cube
-
-        Returns
-        -------
-        iris.cube.Cube
-
-        """
-        metadata = cube.metadata
-        cube *= 28.966 / 44.0
-        cube.metadata = metadata
-        return cube
-
-
 class Gpp(Fix):
     """Fixes for gpp variable."""
 
--- a/esmvalcore/preprocessor/_derive/co2s.py
+++ b/esmvalcore/preprocessor/_derive/co2s.py
@@ -1,0 +1,39 @@
+"""Derivation of variable ``co2s``."""
+import dask.array as da
+import iris
+
+from ._baseclass import DerivedVariableBase
+
+
+class DerivedVariable(DerivedVariableBase):
+    """Derivation of variable ``co2s``."""
+
+    @staticmethod
+    def required(project):
+        """Declare the variables needed for derivation."""
+        required = [{'short_name': 'co2'}]
+        return required
+
+    @staticmethod
+    def calculate(cubes):
+        """Compute mole fraction of CO2 at surface."""
+        cube = cubes.extract_strict(
+            iris.Constraint(name='mole_fraction_of_carbon_dioxide_in_air'))
+        mask = da.ma.getmaskarray(cube.core_data())
+        if not mask.any():
+            cube = cube[:, 0, :, :]
+        else:
+            numerical_mask = da.where(mask, -1.0, 1.0)
+            indices_first_positive = da.argmax(numerical_mask, axis=1)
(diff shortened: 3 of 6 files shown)


schlunma:
I think the fix for `co2` of CESM1-BGC is wrong:

https://github.com/ESMValGroup/ESMValCore/blob/735f4160da4d2ae113dd87f93f25cb38d943c656/esmvalcore/cmor/_fixes/cmip5/cesm1_bgc.py#L25-L28

For the year 1950, the resulting CO2 fraction is ~200ppm, while all other models have around 300ppm. When I compare the original file to that of BNU-ESM (which has the same fix), BNU-ESM shows CO2 fractions of over 500ppm in 2005, while `CESM1-BGC` has about 400ppm (`historical` and `esmHistorical` exp), which is the correct value. Thus, I think that BNU-ESM needs that fix, but CESM1-BGC does not.

Can someone with access to mistral confirm that? These are the files

`/pf/b/b309141/work/CMIP5_DKRZ/BNU/BNU-ESM/esmHistorical/mon/atmos/Amon/r1i1p1/v20120510/co2/co2_Amon_BNU-ESM_esmHistorical_r1i1p1_185001-200512.nc`

`/pf/b/b309141/work/CMIP5_DKRZ/NSF-DOE-NCAR/CESM1-BGC/esmHistorical/mon/atmos/Amon/r1i1p1/v20130216/co2/co2_Amon_CESM1-BGC_esmHistorical_r1i1p1_195001-200512.nc`

mattiarighi:
I also think the fix is wrong.

The multiplication factor `molarmass(C) / molarmass(CO2)` is usually applied to mass units, but here it is applied to a volume mixing ratio.

mattiarighi:
It is actually molarmass(dry_air) / molarmass(co2).

esmvalcore/cmor/tables/custom/CMOR_co2s.dat
```diff
@@ -0,0 +1,20 @@
+SOURCE: CMIP5
+!============
+variable_entry:    co2s
+!============
+modeling_realm:    atmos
+!----------------------------------
+! Variable attributes:
+!----------------------------------
+standard_name:     mole_fraction_of_carbon_dioxide_in_air
+units:             mol mol-1
```
mattiarighi: Is there a reason for not using the same units as the parent variable (`1.e-6`)?

schlunma: I used the one from the CMIP6 table:

https://github.com/ESMValGroup/ESMValCore/blob/735f4160da4d2ae113dd87f93f25cb38d943c656/esmvalcore/cmor/tables/cmip6/Tables/CMIP6_Amon.json#L305-L321

mattiarighi: Ok. I was looking into CMIP5.

schlunma:
@mattiarighi I just realized that there is already a variable `co2s` in the CMIP6 tables :grimacing: :

https://github.com/ESMValGroup/ESMValCore/blob/cd5cf1f622fd277323008c054a1f19f54f73de64/esmvalcore/cmor/tables/cmip6/Tables/CMIP6_Emon.json#L827-L844

Unfortunately, this can't be used for all projects which use `CMIP5`-like tables, but I need to use that variable with CMIP5. What should we do here? I see two options:

1. Rename the custom `co2s`

2. Change the custom `co2s` so that it is similar to the CMIP6 entry.

What do you think?

mattiarighi:
I would go for the second option, using two names for the same variables sounds confusing.
But we should discuss this in the @ESMValGroup/esmvaltool-coreteam, since I think we never had this case before, did we?

valeriupredoi:
You can map CMIP6<->CMIP5 variables as @jvegasbsc implemented in the cmor module, then you can add this variable as a CMIP5 variable only and map it to the actual cmor tables CMIP6 variable. Hows that sound? :beer:

schlunma:
@valeriupredoi can you show me where this CMIP5<->CMIP6 translation is in the code?

valeriupredoi:
`l.90` in `esmvalcore/cmor/table.py`: `_CMIP_5to6_varname` dict :beer:

schlunma:
But this is only necessary if the the variables have different names in CMIP5 and CMIP6, right? If we use the second option, the variable is called `co2s` in both cases.

jvegasbsc:
Yes, you need to do this only for those that differ in both

valeriupredoi:
yeh I was mistaken in thinking it needs a different name for CMIP5 :beer:

PR #610, opened 2020-04-06
make ncl or R tests not fail if package not installed

valeriupredoi:

**Tasks**

-   [ ] Add unit tests

* * *

### DIFF
--- a/tests/integration/test_diagnostic_run.py
+++ b/tests/integration/test_diagnostic_run.py
@@ -125,15 +125,15 @@ def interpreter_not_installed(script):
 
 @pytest.mark.parametrize('script_file, script', [
     pytest.param(
+        script_file,
         script,
-        content,
         marks=[
             pytest.mark.installation,
-            pytest.mark.xfail(interpreter_not_installed(script),
+            pytest.mark.xfail(interpreter_not_installed(script_file),
                               run=False,
                               reason="Interpreter not available"),
         ],
-    ) for script, content in SCRIPTS.items()
+    ) for script_file, script in SCRIPTS.items()
 ])
 def test_diagnostic_run(tmp_path, script_file, script):
 
--- a/tests/integration/test_diagnostic_run.py
+++ b/tests/integration/test_diagnostic_run.py
@@ -2,6 +2,7 @@
 import contextlib
 import shutil
 import sys
+from pathlib import Path
 from textwrap import dedent
 
 import pytest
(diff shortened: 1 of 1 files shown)


valeriupredoi:
@bouweandela skipping is something we want since the regular Joe will run `python setup.py test` rather than look up what extra options and how to run in advanced ways the tests. They'll immediately ask 'oh tests failing, guess smth is wrong' - I got asked this a lot of times.

bouweandela:
Maybe you're right, but note that if they do have an interpreter installed, but not the required libraries, the tests will still fail.

bouweandela:
It would be nicer to use the pytest xfail marker to skip the tests, e.g.
```python
def interpreter_not_installed(script):
    """Check if an interpreter is installed for script."""
    interpreters = {
        '.jl': 'julia',
        '.ncl': 'ncl',
        '.py': 'python',
        '.R': 'Rscript',
    }
    ext = Path(script).suffix
    interpreter = interpreters[ext]
    return which(interpreter) is None


@pytest.mark.parametrize('script_file, script', [
    pytest.param(
        script,
        content,
        marks=[
            pytest.mark.installation,
            pytest.mark.xfail(interpreter_not_installed(script),
                              run=False,
                              reason="Interpreter not available"),
        ],
    ) for script, content in SCRIPTS.items()
])
def test_diagnostic_run(tmp_path, script_file, script):
...
```

valeriupredoi:
> Maybe you're right, but note that if they do have an interpreter installed, but not the required libraries, the tests will still fail.

which is probably the desired behaviour? In any case, I don't think that's true, since if they have an interpreter in the path it means they are using a mixed environment (ie conda path for the esmvaltool environment + other libraries paths) which is a major headache and that'll hit the fan when they run the first R or ncl diagnostic. In fact, we should probably produce a test that fails if `ncl` and `R` are not the ones installed in the esmvaltool environment

bouweandela:
> which is probably the desired behaviour? 

Not really, because you can use ESMValCore as a standalone package for running diagnostics that are not part of the ESMValTool package. If you want to use it only with Python diagnostics, there is no need to install R or R dependencies. Yet the test will fail if you happen to have R installed as a system package but not the R yaml library. In that case you will need to use the `installation` pytest marker to skip the test, because the workaround in this pull request will not skip the test and it will fail.

> if they have an interpreter in the path it means they are using a mixed environment 

Conda environments are not standalone, they are always mixed with whatever else is installed on your machine

>  In fact, we should probably produce a test that fails if ncl and R are not the ones installed in the esmvaltool environment

I think this will cause more confusion than be helpful and I think this is a feature we might want to support. If someone already has ncl or R interpreters installed on their machine with whatever libraries they need, I see no point in forcing them into using the packages from conda. For Julia this is not even possible, as it cannot be installed from conda. Once we have ESMValTool on PyPI as well, it might even be possible to use it without conda (though the iris dependencies may not be that easy to install without it).

valeriupredoi:
cheers Bouwe man, I took your `xfail` implementation, it works nicely :beer:

PR #618, opened 2020-04-20
Change volume stats to handle and output masked array result

### LINKED ISSUE #611: Empty cubes after volume_statistics for certain datasets
**Describe the bug**
CMIP6 EC-Earth ocean 3D data has the whole bottom depth layer set to masked values. Both the variables and the volcello. The volume_statistics preproc computes the weighted mean per layer just fine, and for the bottom layer the results are masked as well. But when it comes to computing the weighted mean over the water volume, the presence of masked values sets the results to NaNs and the data comes out of the preprocessing empty:

`result.append(np.average(column, weights=depth_volume))`

It looks like the function `np.average` is not handling well the presence of layers with only masked values. Would it be correct check and remove this sort of layers before calling `np.average` ?

**Please attach**
I'm working with EC-Earth local data and can't seem to find Ecearth data with volcello on /badc, sorry.

valeriupredoi:

**Tasks**

-   [x] Add unit tests

no, it's just being petty, it can bugger off

* * *

Closes #611

### DIFF
--- a/tests/unit/preprocessor/_volume/test_volume.py
+++ b/tests/unit/preprocessor/_volume/test_volume.py
@@ -21,11 +21,6 @@ class Test(tests.Test):
         coord_sys = iris.coord_systems.GeogCS(iris.fileformats.pp.EARTH_RADIUS)
         data1 = np.ones((3, 2, 2))
         data2 = np.ma.ones((2, 3, 2, 2))
-        data2_masked = np.ma.ones((2, 3, 2, 2))
-        data2_masked[0] = 200.
-        data2_masked.mask = (False, True, True, False)
-        data2_fully_masked = np.ma.ones((2, 3, 2, 2))
-        data2_fully_masked.mask = True
         data3 = np.ma.ones((4, 3, 2, 2))
         mask3 = np.full((4, 3, 2, 2), False)
         mask3[0, 0, 0, 0] = True
@@ -71,11 +66,7 @@ class Test(tests.Test):
 
         coords_spec4 = [(time, 0), (zcoord, 1), (lats2, 2), (lons2, 3)]
         self.grid_4d = iris.cube.Cube(data2, dim_coords_and_dims=coords_spec4)
-        self.grid_4d_masked = iris.cube.Cube(data2_masked,
-                                             dim_coords_and_dims=coords_spec4)
-        self.grid_4d_fully_masked = iris.cube.Cube(
-            data2_fully_masked,
-            dim_coords_and_dims=coords_spec4)
+
         coords_spec5 = [(time2, 0), (zcoord, 1), (lats2, 2), (lons2, 3)]
         self.grid_4d_2 = iris.cube.Cube(
             data3, dim_coords_and_dims=coords_spec5)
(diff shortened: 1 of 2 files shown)


sloosvel:
Thanks @valeriupredoi . I'm still getting NaNs though. This is what the last elements of `column` and `depth_volume` look like respectively when debugging:

`column`: 
72:array(0.13211952, dtype=float32)
73:array(0.13278724, dtype=float32)
74:masked_array(data=--, mask=True, fill_value=1e+20, dtype=float64)

`depth_volume`:
72:4726634600000000.0
73:3180468400000000.0
74:masked

 I think it's because of the masked `depth_volume`, because I changed the last value to 0 just to test with a non-masked weight and the results I got are not nans anymore.  So it looks like `np.ma.average` is not able to handle masked weights after all.

valeriupredoi:
@sloosvel indeed, `np.ma.average()` is being silly:

```python
In [10]: np.ma.average(np.ma.array([1, 2, 3, 5]), weights=np.ma.array([2,2,2,2], mask=[True, True, True, False]))
Out[10]: 5.0

In [11]: np.ma.average(np.ma.array([1, 2, 3, 5]), weights=np.ma.array([2,2,2,2], mask=[True, True, True, True]))
Out[11]: nan
```

and it looks like the actual logic is correct from the [documentation](https://numpy.org/devdocs/reference/generated/numpy.ma.average.html) since the statistic is computed using sum(weights) as denominator (obv, weighted mean heh) so in case of a fully masked weights array we are a bit stuck - should probably perform a simple mean in that case since weights are completely irrelevant but that could assign a bit more weight to that particular column- I am not sure how to approach from the statistical point of vew - @ledm and advice?

valeriupredoi:
or we can just completely mask that column?

valeriupredoi:
OK I masked it completely to get rid of `nan`'s

sloosvel:
I think the routine is computing the 2D average first and then the average over all levels. So the output of averaging `column` should be the total average. What these last commits seem to be doing is just set the total result from nan to masked, which would mean that the output would still be empty. 

Anyway, I think it's a bit strange to have a completely masked level in volcello.

valeriupredoi:
so that's the average volume per time point with computed average over a set of columns at different depths, weighted by each column's volume; the code changes mask those values that have weights masked equivalent to saying if the column volume is not viable as a weight there (coz it's masked) it means I may not have enough information about that particular column, and extrapolating this - if all my columns have masked values I don't have enough information to compute an average volume at that time point so I am masking it - I think it's fine

sloosvel:
I don't know that's seems overly complicated to me. I took a closer look and it turns out that depth_volume is a list. I changed its type to np.ma.array and I don't get nans anymore.

valeriupredoi:
@sloosvel this is the generalized approac, note that even if `depth_volume` is a masked array and its values are all masked you will still get `nan`'s ie `np.ma.average(np.ma.array([1, 2, 3, 5]), weights=np.ma.array([2,2,2,2], mask=[True, True, True, True]))` returns `nan`

valeriupredoi:
any news on this @sloosvel :beer:

sloosvel:
I am not sure if a case in which all `depth_volume` values are masked should be taken into account. That would mean that the volcello files consist of only masked values. Which would make the data not really useful.

There are several models that have the last layer of the ocean data fully masked. This routine is not handling well this case because the `depth_volume` used in `np.average` (or `np.ma.average`) has a masked value in the last level and is declared as a list instead of being a masked array. If the type is changed to masked array before the final averaging, I get proper results instead of a result with just NaN values.

valeriupredoi:
cool! then apply whatever changes are needed here, job done by the sounds of it :beer:

jvegasbsc:
@ledm @bouweandela Can you please review this? It is blocking us

sloosvel:
I added a test for nans and looks like it's passing

valeriupredoi:
Indeed it do :grin:

sloosvel:
I copy-pasted code and probably messed the format while doing so, will correct now.

bouweandela:
> @ledm @bouweandela Can you please review this? It is blocking us

Please feel free to help out with reviewing, it is impossible for me to review every single pull request in a timely manner.

tests/unit/preprocessor/_volume/test_volume.py
```diff
@@ -100,6 +100,16 @@ def test_volume_statistics_long(self):
...
         self.assert_array_equal(result.data, expected)
 
+    def test_volume_statistics_masked_level(self):
+        """
+        Test to take the volume weighted average of a (2,3,2,2) cube
+        where the last depth level is fully masked.
+        """
+        self.grid_4d.data[:, -1, :, :] = np.ma.masked_all((2, 2, 2))
+        result = volume_statistics(self.grid_4d, 'mean')
+        expected = np.array([1., 1.])
```
bouweandela, line 110: I would expect `volume_statistics` to return a cube with masked data when the input also has masked data.

sloosvel: This is one spatial average value per timestep. Ocean data always contains masked values anyway and the output for this kind of operations is a non-masked value. It should not be different in this case.

bouweandela: What if all values for a particular time step are masked? Are you sure that never happens?

valeriupredoi: ya that was my previous implementation that looked a bit overkill (inc a test for it) but @sloosvel assures me that can't happen

sloosvel:
>What if all values for a particular time step are masked? Are you sure that never happens?

@bouweandela Well if you want to consider that case I think the easiest is to save the result in line 275 in np.ma.array(result) instead of np.array(result), let me know what you think. I don't know whether or not this happens very often or rarely. It's not the case for which I opened the issue. The main problem that I was having is that the average was being computed with lists that where not handling well the presence of a masked value in the volume. So I changed the implementation to use masked arrays instead.

>ya that was my previous implementation that looked a bit overkill (inc a test for it) but @sloosvel assures me that can't happen

@valeriupredoi Your implementation was not working for the datasets that I was testing. Because the average was still being computed with lists instead of masked arrays, so the result was a NaN that was getting masked at the very end. I cannot have average_volume = NaN (which is what happens when using master) or average_volume = masked (which is what happens when using your implementation) for all ocean variables in all of our EC-Earth experiments, it does not make sense and it's not the result that we get using other old packages in the department.

bouweandela:
> Well if you want to consider that case I think the easiest is to save the result in line 275 in np.ma.array(result) instead of np.array(result), let me know what you think.

Yes, that would be a good solution.

bouweandela:
@mattiarighi Could you please test and merge when you're happy with the result?

mattiarighi:
@sloosvel can you provide a test case (specific model and variable(s) to test)?

sloosvel:
@mattiarighi Sorry for the delay, our data it's on its way to be published but it's not available yet. The HadGEM3-GC31-LL historical data could be another example of a volcello with a fully masked level, but I'm having concatenation errors with this dataset.

mattiarighi:
The changes in code look fine to me.
If this solves your issue please approve the PR and I will merge.

PR #626, opened 2020-04-30
Padding while cropping needs to stay within sane bounds for shapefiles that span the whole Earth

### LINKED ISSUE #625: Padding lon/lat bounds in _area/_crop_cube returns invalid bounds in extract_region if the geometry is all Earth
[Here](https://github.com/ESMValGroup/ESMValCore/blob/d817e69a312c0cabe172f8cb427288c05b703acc/esmvalcore/preprocessor/_area.py#L324) - if geometry is as is from `esmvalcore/preprocessor/ne_masks/ne_10m_ocean.shp` and its bounds are `[-179.99999999999991 -85.22193775799991 180.0 90.00000000000006]` by padding that will introduce a `ValueError: Invalid end_latitude: 90.50000000000006` from `extract_region`:
```bash
  File "test_ne.py", line 49, in <module>
    cube = _crop_cube(cube, *geometries.bounds)
  File "/home/valeriu/ESMValCore/esmvalcore/preprocessor/_area.py", line 340, in _crop_cube
    start_latitude, end_latitude)
  File "/home/valeriu/ESMValCore/esmvalcore/preprocessor/_area.py", line 54, in extract_region
    raise ValueError(f"Invalid end_latitude: {end_latitude}")
ValueError: Invalid end_latitude: 90.50000000000006
```

valeriupredoi:

**Tasks**

-   [x] Add unit tests

* * *

Closes #625


esmvalcore/preprocessor/_area.py
```diff
@@ -331,14 +331,23 @@ def _crop_cube(cube, start_longitude, start_latitude, end_longitude,
         lon_bound = lon_coord.core_bounds()[0]
         lon_step = lon_bound[1] - lon_bound[0]
         start_longitude -= lon_step
+        if start_longitude < -180:
+            start_longitude = -180.
         end_longitude += lon_step
+        if end_longitude > 360:
+            end_longitude = 360.
+        if end_longitude > 180 and start_longitude == -180.:
+            end_longitude = 180.
```
bouweandela, line 340: This looks a bit worrying, is this needed?

tests/unit/preprocessor/_area/test_area.py
```diff
@@ -377,6 +378,40 @@ def square_composite_shape(request, tmp_path):
     return np.ma.masked_array(vals, mask)
 
 
+def _create_sample_full_cube():
+    cube = Cube(np.zeros((24, 180, 360)), var_name='co2', units='J')
```
bouweandela, line 382:
```suggestion
    cube = Cube(np.zeros((2, 180, 360)), var_name='co2', units='J')
```
It would be nicer to use a smaller cube to avoid slowing down the tests and making them require a lot of memory

tests/unit/preprocessor/_area/test_area.py
```diff
@@ -385,6 +420,18 @@ def test_crop_cube(make_testcube, square_shape, tmp_path):
         np.testing.assert_array_equal(result.data, expected)
 
 
+def test_crop_cube_with_ne_file():
+    """Test for cropping a cube by shape bounds."""
+    shp_file = "esmvalcore/preprocessor/ne_masks/ne_10m_ocean.shp"
```
bouweandela, line 425: This shapefile appears to be used nowhere except in this test, why is it in the repository? And why do we use higher resolution shapefiles for masking land than sea?

valeriupredoi: the `10m` is perfectly usable for oceans but it's overkill and very heavy on the speed, and the `50m` is good enough and much faster, we can probably remove the `10m` one and totally forget about it

esmvalcore/preprocessor/_area.py
valeriupredoi: yes, since the padding for files that are full maps goes over 180, specifically needed for this case

tests/unit/preprocessor/_area/test_area.py
valeriupredoi: good point!

tests/unit/preprocessor/_area/test_area.py
bouweandela: Sounds good, let's remove it. We can add it again if we ever need it.

esmvalcore/preprocessor/_area.py
bouweandela: Maybe you could take this out of this pull request and add it in the other pull request where you add support for shapefiles with longitude coordinates between -180 and 180? At the moment these aren't supported anyway.

bouweandela: On line 334 and 335 there is also code related to supporting those files, maybe it would be best to also move that to the other pull request. I think it does not work in all cases when you make it like this, so it would be better to just get this pull request merged with things we know do work and try to add a new feature in the next pull request.

valeriupredoi: > Maybe you could take this out of this pull request and add it in the other pull request where you add support for shapefiles with longitude coordinates between -180 and 180? At the moment these aren't supported anyway.

where's the restriction clause? I can't see it and the thing fails if any of these files is used - the user doesn't know about this restriction and will attempt to use whatevs shapefile that may get into this trouble

valeriupredoi: it's perfectly fine to have this check here since it doesn't affect any other case imho

bouweandela: The problem is that it doesn't work if you do it like this. To make it work, you first need to distinguish between the two cases 1) longitude in domain 0 to 360 or 2) longitude in domain -180 to 180 and then limit the `start_longitude` and `end_longitude` arguments you provide to `extract_shape` to the chosen domain.

valeriupredoi: `start_longitude == -180.` implies we are on a [-180, 180] grid, mate - we can not be on a [0, 360] and if lons go from [-180, 360] then it means we're wrapping coords like Xmas presents and I am a fish :fish: - both unlikely :grin:

bouweandela: Yes, but a start longitude of say -170 would probably also mean a  [-180, 180] grid, but your code does not cover that case.

valeriupredoi: ok @bouweandela you win :checkered_flag: - I have removed that special case and put it in https://github.com/ESMValGroup/ESMValCore/pull/319 where there is an actual checking mechanism for these sort of files :beer:

esmvalcore/preprocessor/_area.py
```diff
@@ -331,14 +331,21 @@ def _crop_cube(cube, start_longitude, start_latitude, end_longitude,
         lon_bound = lon_coord.core_bounds()[0]
         lon_step = lon_bound[1] - lon_bound[0]
         start_longitude -= lon_step
+        if start_longitude < -180:
+            start_longitude = -180.
```
bouweandela, line 335:
```suggestion
        if start_longitude < 0:
            start_longitude = 0.
```

esmvalcore/preprocessor/_area.py
bouweandela: :beers:

esmvalcore/preprocessor/_area.py
bouweandela: Could you also change it here?

valeriupredoi: done and adjusted tests accordingly as well :beer:

valeriupredoi:
cheers @bouweandela for the qwik approval - now I can plop this in #319 shipshape the tests for that one :beer:

tests/unit/preprocessor/_area/test_area.py
```diff
@@ -385,6 +420,22 @@ def test_crop_cube(make_testcube, square_shape, tmp_path):
         np.testing.assert_array_equal(result.data, expected)
 
 
+def test_crop_cube_with_ne_file():
+    """Test for cropping a cube by shape bounds."""
+    shp_file = "esmvalcore/preprocessor/ne_masks/ne_50m_ocean.shp"
+    with fiona.open(shp_file) as geometries:
```
bouweandela, line 426: Actually this test could be done completely without the shapefile, but since you've already merged the pull request I guess that can wait for another time ;-)

valeriupredoi: yes I know - the thing is I want to use it in #319 that's why I didn't change it here

PR #645, opened 2020-05-19
More flexible native6 default DRS

### LINKED ISSUE #641: Simplify default DRS for native6 project
Because the data used with the `native6` project can have any filename, it might be nice to make the default DRS
```
  input_dir:
    default: '{dataset}/{latestversion}/{frequency}/{short_name}'
  input_file:
    default: '*.nc'
```
because this would allow using the filenames that the data comes with without renaming. You could just download the data, put it in the right directory and be done. It would also avoid the issue reported here: https://github.com/ESMValGroup/ESMValTool/pull/1644#issue-413480838.

bouweandela:

**Tasks**

* * *

Closes #641

### DIFF
--- a/esmvalcore/config-developer.yml
+++ b/esmvalcore/config-developer.yml
@@ -192,10 +192,8 @@ native6:
   cmor_strict: false
   input_dir:
     default: 'Tier{tier}/{dataset}/{latestversion}/{frequency}/{short_name}'
-    era5cli: 'Tier{tier}/{dataset}'
   input_file:
     default: '*.nc'
-    era5cli: 'era5_{era5_name}*{era5_freq}.nc'
   output_file: '{project}_{dataset}_{type}_{version}_{mip}_{short_name}'
   cmor_type: 'CMIP6'
   cmor_default_table_prefix: 'CMIP6_'
--- a/esmvalcore/config-developer.yml
+++ b/esmvalcore/config-developer.yml
@@ -191,10 +191,10 @@ OBS6:
 native6:
   cmor_strict: false
   input_dir:
-    default: 'Tier{tier}/{dataset}'
+    default: '{dataset}/{latestversion}/{frequency}/{short_name}'
     era5cli: 'Tier{tier}/{dataset}'
   input_file:
-    default: '{project}_{dataset}_{type}_{version}_{mip}_{short_name}[_.]*nc'
+    default: '*.nc'
     era5cli: 'era5_{era5_name}*{era5_freq}.nc'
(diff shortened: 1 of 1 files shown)


esmvalcore/config-developer.yml
```diff
@@ -191,10 +191,10 @@ OBS6:
 native6:
   cmor_strict: false
   input_dir:
-    default: 'Tier{tier}/{dataset}'
+    default: '{dataset}/{latestversion}/{frequency}/{short_name}'
```
mattiarighi, line 194: Why do we need `{latestversion}` here?

mattiarighi: Also, wouldn't be more straightforward to use `era5_name` instead of `short_name`?

bouweandela: > Why do we need {latestversion} here?

Because we want to support multiple versions of datasets

> Also, wouldn't be more straightforward to use `era5_name` instead of `short_name`?

Using the short name has the advantage that you can also use this DRS for other observational datasets than ERA5, making it easier to add support for those.

mattiarighi: OK. I think I understand why it didn't work before.
Let me try again.

bouweandela: I provided an example of how to use it in the issue (#641), maybe that helps?

mattiarighi: Can we add `Tier{tier}/`, I think it would be good to keep this information in the directory tree.

bouweandela: Do we really need those Tier1, 2, 3 directories? Is it possible for one dataset to be in two different Tiers?

mattiarighi: > Do we really need those Tier1, 2, 3 directories? 
Yes, we should keep this classification, it's also described in the paper.

> Is it possible for one dataset to be in two different Tiers?
No, I don't think so.

I can also add another entry for dkrz with `Tier{tier}` and leave `default` unchanged.

Do we still need the `era5cli` entry? I think that would never work given the ambiguity in variable name.

bouweandela: > I can also add another entry for dkrz with Tier{tier} and leave default unchanged.

Then we still need it in the recipes, so it would not help. Let's keep it then.

> Do we still need the era5cli entry? I think that would never work given the ambiguity in variable name.

I think it can be deleted, I guess I would need to add some documentation somewhere then on what directory structure to use.

bouweandela:
I'll add some documentation after ESMValGroup/ESMValTool#1652 is merged.

PR #672, opened 2020-06-09
Avoid crashing when there is directory 'esmvaltool' in the current working directory

valeriupredoi:

**Tasks**

-   [ ] Add unit tests

* * *

### DIFF
--- a/esmvalcore/_config.py
+++ b/esmvalcore/_config.py
@@ -22,7 +22,8 @@ def find_diagnostics():
         import esmvaltool
     except ImportError:
         return Path.cwd()
-    # catch a lingering failed install
+    # avoid a crash when there is a directory called
+    # 'esmvaltool' that is not a Python package
     if esmvaltool.__file__ is None:
         return Path.cwd()
     return Path(esmvaltool.__file__).absolute().parent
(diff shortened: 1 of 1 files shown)


valeriupredoi:
quick one for you @bouweandela - `import esmvaltool` does not throw an `ImportError` but the installation is non-existent; this may happen when eg you conda uninstall esmvaltool :beer:

esmvalcore/_config.py
```diff
@@ -22,6 +22,9 @@ def find_diagnostics():
         import esmvaltool
     except ImportError:
         return Path.cwd()
+    # catch a lingering failed install
```
bouweandela, line 25:
```suggestion
    # avoid a crash when there is a directory called 'esmvaltool' that is not a Python package
```
Good find, though I think that the actual explanation is that there is an empty directory called `esmvaltool` either in your current working directory or somewhere where Python is looking for modules. That would get interpreted as a [namespace package](https://packaging.python.org/guides/packaging-namespace-packages/#native-namespace-packages) and confuse the tool.

valeriupredoi: yup good call Bouwe man, just plugged in the change :+1:

Issue #639, opened 2020-05-18
Shrink _recipe.py


valeriupredoi:
I think we have reached the size and complexity limit for `_recipe.py` (we may have reached that a while ago) so I reckon we should start shrinking it and optimizing its use; I reckon the first step is to move all the preprocessor module specific functions (like `_update_{STEP}_settings`) to the preprocessor itself. What say you @bouweandela - if you agree I can get cracking with it starting now :beer:

bouweandela:
Hi V,

I agree with you that it would be good to clean this file up. In the next few months we will be working on #498. I expect that once we have a more complete Python API in place, _recipe.py will also be a lot simpler, so I think it would be best to wait a bit with this.

Another issue is that at the moment we have about 90 open pull requests (45 in ESMValCore and 46 in ESMValTool), I would really like to focus on getting those merged before starting new work.

valeriupredoi:
makes sense, man! let's wait a bit more, I'll dedicate my effeorts towards the noble cause of solving as many of those PR's and issues :carousel_horse:

Issue #489, opened 2020-02-13
`mask_fillvalues` fails in some cases when input is parsed as string and not as float


bascrezee:
**Describe the bug**
The example given in the documentation for `mask_fillvalues` fails, because the input is parsed as string and not as float.

```
datasets:
  - {dataset: cds-era5-monthly, type: reanaly, project: OBS, mip: Lmon,
     version: 1, tier: 3, start_year: 2008, end_year: 2010}
  - {dataset: CDS-SATELLITE-SOIL-MOISTURE, type: sat, project: OBS, mip: Lmon,
     version: COMBINED-TCDR-v201812.0.0, tier: 3, start_year: 2008, end_year: 2010}


preprocessors:
  test_b:
    regrid:
      target_grid: 0.25x0.25
      scheme: nearest
    mask_fillvalues:
      threshold_fraction: 0.0     # keep all missing values
      min_value: -1e20            # small enough not to alter the data
```

```
  File "/ESMValCore/esmvalcore/preprocessor/_mask.py", line 402, in count_spells
    data_hits = data > threshold
TypeError: '>' not supported between instances of 'MaskedArray' and 'str'
```

PR #673, opened 2020-06-10
Add ensemble statistics preprocessor and 'groupby' option for multimodel

### LINKED ISSUE #52: Comparing several single model ensembles
Hi all,

two very common things that users will want to do in ESMValTool are:

1. Compare single model ensemble means.
  For instance, I have a HadGEM2-ES 4 member ensemble and UKESM 12 member ensemble. I want to make some time series plots showing the time development of the two single-model ensemble means. 
  How do I do that?  Is there a way to use the multi_model_statistics preprocessor so that it only compares single models? Or perhaps we need a single_model_ensemble_statistics preprocessor? 

2. Calculate the differences between two specific time periods in the preprocessor stage.
  For instance, I want to look at  average surface temperature in the years 1975-2000  minus the average surface temperature between 1875-1900. 
  I don't know of a way to do this in the preprocessor, it seems to be such a common job that it would be great to get it as a preprocessor instead of in the diagnostic stage.

Any ideas?

Lee

sloosvel:
- Add a new preprocessor function to calculate ensemble statistics
- Add support for grouping in multi-model statistics preprocessor
- Separate functions that operate on cubes from functions that operate on products (esmvalcore.PreprocessorFiles). 
- More efficient implementation of multi-cube statistics that can eventually be used for multi-model statistics as well

Link to documentation: https://esmvaltool--673.org.readthedocs.build/projects/ESMValCore/en/673/recipe/preprocessor.html#ensemble-statistics

**Tasks**

-   [x] Add unit tests
-   [x] ~~Codacy code quality checks pass. Status can be seen below your pull request. If there is an error, click the link to find out why. If you suspect Codacy may be wrong, please ask by commenting.~~
-   [x] ~Please use `yamllint` to check that your YAML files do not contain mistakes~
-   [x] ~If you make backward incompatible changes to the recipe format, make a new pull request in the [ESMValTool repository](https://github.com/ESMValGroup/ESMValTool) and add the link below~

* * *

Closes #52
Merge first: #685


stefsmeets:
Minimal working example:

```
# ESMValTool
# recipe_climwip.yml
---
documentation:
  description: Test ensemble statistics

  authors:
    - kalverla_peter

  maintainer:
    - kalverla_peter

  references:
    - acknow_proj

datasets: &model_data
  - {dataset: CanESM2, project: CMIP5, exp: [historical, rcp45], ensemble: "r(1:4)i1p1"}
  - {dataset: CCSM4, project: CMIP5, exp: [historical, rcp45], ensemble: "r(1:4)i1p1"}

preprocessors:
  preproc:
    custom_order: true

    area_statistics:
      operator: mean

    ensemble_statistics:
      operator: mean

diagnostics:

  diagnostic:

    themes:
      - phys
    realms:
      - atmos
    variables:
      pr:
        mip: Amon
        start_year: 2000
        end_year: 2002
        preprocessor: preproc
    scripts: null
```

Peter9192:
Just to continue the discussion in #52 , I think this is a greatstart. I merged #685 into this branch, which is just waiting to get merged in master, and my new colleague Stef added some suggestions as well. 

Additionally, I think it would be nice to use [iris' cube aggregation](https://scitools.org.uk/iris/docs/v1.9.0/html/iris/iris/cube.html#iris.cube.Cube.collapsed) instead of our own `multimodel_statistics`. This works for lazy data as well, and has a much larger range of operators.  Eventually I think we want to use this for multimodel stats as well, but there it's more difficult due to the cubes not being very homogeneous. So this would be a nice first step in that direction.

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -451,3 +383,33 @@ def multi_model_statistics(products, span, statistics, output_products=None):
...
+        product_dict[dataset].add(product)
+
+    for dataset, ensemble_products in product_dict.items():
+        statistic_products = multi_model_statistics(
+            ensemble_products,
+            span,
+            statistics,
+            output_products[dataset],
+        )
+        products |= statistic_products  # set union
```
Peter9192, line 413: Maybe we should not add the ensemble statistics to the set of products, but replace the existing products altogether. Otherwise, if you apply multi-model statistics later, you will apply it to the combination of individual members and the ensemble mean.

stefsmeets:
We got the code to a working state on the minimal example.

Peter9192:
@stefsmeets now consider the following use case that doesn't work yet: 

```YAML
# recipe_ensemble.yml
---
documentation:
  description: Test ensemble
  authors:
    - kalverla_peter
  maintainer:
    - kalverla_peter
  references:
    - acknow_proj

preprocessors:
  preproc:
    custom_order: true
    area_statistics:
      operator: mean
    ensemble_statistics:
      statistics: ['mean']
    multi_model_statistics:
      span: overlap
      statistics: ['max']

diagnostics:
  diagnostic:
    variables:
      pr:
        project: CMIP5
        mip: Amon
        start_year: 2000
        end_year: 2002
        preprocessor: preproc
        additional_datasets:
          - {dataset: CanESM2, exp: historical, ensemble: "r(1:2)i1p1"}
          - {dataset: CCSM4, exp: historical, ensemble: "r(1:2)i1p1"}
    scripts: null
```

The input for multimodel statistics needs to be updated, as it is currently just the four input datasets, whereas we want it to be the 2 ensemble stats files.

stefsmeets:
What doesn't work yet is if you define multiple statistics for ensemble and multi-model statistics:
```yaml
preprocessors:
  preproc:
    custom_order: true
    area_statistics:
      operator: mean
    ensemble_statistics:
      statistics: ['mean', 'max']
    multi_model_statistics:
      span: overlap
      statistics: ['mean', 'max']
```

stefsmeets:
Should: `data_finder.get_output_file` and  `data_finder.get_multiproduct_filename` be a method on `PreprocessorFile`? They are only each used once in the code.

stefsmeets:
When exp is a list, i.e. `[historical, rcp45]`

```
  File "/home/stef/r/esmvalcore/esmvalcore/_main.py", line 430, in run
    fire.Fire(ESMValTool())
  File "/home/stef/miniconda3/envs/esmvaltool/lib/python3.8/site-packages/fire/core.py", line 138, in Fire
  File "/home/stef/miniconda3/envs/esmvaltool/lib/python3.8/site-packages/fire/core.py", line 463, in _Fire
  File "/home/stef/miniconda3/envs/esmvaltool/lib/python3.8/site-packages/fire/core.py", line 672, in _CallAndUpdateTrace
  File "/home/stef/r/esmvalcore/esmvalcore/_main.py", line 407, in run
  File "/home/stef/r/esmvalcore/esmvalcore/_main.py", line 94, in process_recipe
  File "/home/stef/r/esmvalcore/esmvalcore/_recipe.py", line 72, in read_recipe_file
    return Recipe(raw_recipe,
  File "/home/stef/r/esmvalcore/esmvalcore/_recipe.py", line 1018, in __init__
  File "/home/stef/r/esmvalcore/esmvalcore/_recipe.py", line 1366, in initialize_tasks
    task = _get_preprocessor_task(
  File "/home/stef/r/esmvalcore/esmvalcore/_recipe.py", line 984, in _get_preprocessor_task
    task = _get_single_preprocessor_task(
  File "/home/stef/r/esmvalcore/esmvalcore/_recipe.py", line 843, in _get_single_preprocessor_task
    products = _get_preprocessor_products(
  File "/home/stef/r/esmvalcore/esmvalcore/_recipe.py", line 811, in _get_preprocessor_products
    ensemble_products, ensemble_settings = _update_ensemble(products, order, preproc_dir)
  File "/home/stef/r/esmvalcore/esmvalcore/_recipe.py", line 684, in _update_ensemble
    return _update_multi_product_settings(products, order, preproc_dir, step, grouping=groupby)
  File "/home/stef/r/esmvalcore/esmvalcore/_recipe.py", line 654, in _update_multi_product_settings
    grouped_products = groupby(products, keyfunc=lambda p: p.group(grouping))
  File "/home/stef/r/esmvalcore/esmvalcore/_recipe.py", line 624, in groupby
    key = keyfunc(item)
  File "/home/stef/r/esmvalcore/esmvalcore/_recipe.py", line 654, in <lambda>
    grouped_products = groupby(products, keyfunc=lambda p: p.group(grouping))
  File "/home/stef/r/esmvalcore/esmvalcore/preprocessor/__init__.py", line 380, in group
    return '_'.join(identifier)
TypeError: sequence item 2: expected str instance, list found
```

stefsmeets:
CircleCI tests fail on `tests/integration/cmor/_fixes/cmip6/test_gfdl_cm4.py` with a Python segmentation fault, unrelated to this PR.

stefsmeets:
`conda_build` bugs out with 

```Failed to rename work directory despite sleeping and retrying. This is some Windows file locking mis-bahaviour.
ERROR:conda_build.utils:Failed to rename work directory despite sleeping and retrying. This is some Windows file locking mis-bahaviour.
ERROR conda_build.utils:shutil_move_more_retrying(2005): Failed to rename work directory despite sleeping and retrying. This is some Windows file locking mis-bahaviour.
```

@bouweandela I see you made some changes to the conda build recently, is this a new issue? I thought we weren't building for Windows.

Peter9192:
Okay @sloosvel sorry for hijacking your PR... I hope you don't mind. Building upon your first steps we did quite a bit of work, especially:

- refactoring parts of the code so that they can be reused for both multi-model and ensemble statistics
- adding a groupby option to multimodel statistics, such that ensemble statistics is just a special case
- adding a new multi-cube function that uses iris build in operations, so that ensemble statistics is more efficient
- more extensive testing

There's a few remaining codacy issues, that we think should be addressed in different PRs, particularly (#639, #489, https://github.com/google/yapf/issues/744). 

The new functionality is illustrated by this example recipe: 

```yaml
preprocessors:
  preproc:
    custom_order: true
    area_statistics:
      operator: mean
    ensemble_statistics:
      statistics: [mean, median]
    multi_model_statistics:
      span: overlap
      groupby: [ensemble_statistics, tag]
      statistics: [max, min]

diagnostics:
  diagnostic:
    variables:
      pr:
        project: CMIP5
        mip: Amon
        start_year: 2000
        end_year: 2002
        preprocessor: preproc
        additional_datasets:
          - {dataset: CanESM2, exp: historical, ensemble: "r(1:2)i1p1", tag: 'group1'}
          - {dataset: CCSM4, exp: historical, ensemble: "r(1:2)i1p1", tag: 'group2'}
          - {dataset: ACCESS1-3, exp: historical, ensemble: r1i1p1, tag: 'group1'}

    scripts: null
```

Would you be willing to review and test this? Thanks!

sloosvel:
It works great for what I have tried! The tag for grouping is really useful and it's great it can be customized. Even at diagnostic level, it's really easy to load the data by grouping the generated files either by tag or statistic.  The documentation also looks clear and it was easy to follow. So I don't think I have much to add, it works and reads really well.

Peter9192:
@sloosvel thanks for spotting. Also related to #744, which proposes to move the time handling out of the mm-stats altogether. I suppose we would want the time handling preprocessor to return cubes with bounds on their time coords, so then we should probably do that here as well (and in #685). At least we should be consistent between `overlap` and `full`, but adding them everywhere means that the new behaviour for `full` will then be inconsistent with the old behaviour.

sloosvel:
I checked and I can't seem to find any major reason as for why the output cubes should not have time bounds when using `span: full`. Maybe it's something that has gone unnoticed. But if preprocessor steps that can come in later expect the cubes to have bounds, they should be added back at some point.

bouweandela:
> conda_build bugs out with
> ```
> ERROR:conda_build.utils:Failed to rename work directory despite sleeping and retrying. This is some Windows file locking mis-bahaviour.
> ERROR conda_build.utils:shutil_move_more_retrying(2005): Failed to rename work directory despite sleeping and retrying. This is some Windows file locking mis-bahaviour.
>```
> @bouweandela I see you made some changes to the conda build recently, is this a new issue? I thought we weren't building for Windows.

I'm not sure where this comes from. Seeing errors mentioning Windows file locking misbehaviour on a Linux machine is pretty funky!

Peter9192:
@sloosvel I added bounds for both methods now. Can you check again?

sloosvel:
They are there now for both spans. The current version must add them differently than this one because the values are slightly different. 

```
 master time_bnds =
  54786, 54817,
  54817, 54846,
  54846, 54877,
  54877, 54907,
  54907, 54938,
  54938, 54968,
  54968, 54999,
  54999, 55030,
  55030, 55060,
  55060, 55091,
  55091, 55121,
  55121, 55152,
  55152, 55183,
  55183, 55211,
  55211, 55242,
  55242, 55272,
  55272, 55303,
  55303, 55333,
  55333, 55364,
  55364, 55395,
  55395, 55425,
  55425, 55456,
  55456, 55486,
  55486, 55517,
  55517, 55548,
  55548, 55576,
  55576, 55607,
  55607, 55637,
  55637, 55668,
  55668, 55698,
  55698, 55729,
  55729, 55760,
  55760, 55790,
  55790, 55821,
  55821, 55851,
  55851, 55882 ;
}

```

```
dev_multimodel_ensemble  time_bnds =
  54784.5, 54815.5,
  54815.5, 54845.5,
  54845.5, 54875.5,
  54875.5, 54906,
  54906, 54936.5,
  54936.5, 54967,
  54967, 54997.5,
  54997.5, 55028.5,
  55028.5, 55059,
  55059, 55089.5,
  55089.5, 55120,
  55120, 55150.5,
  55150.5, 55181.5,
  55181.5, 55211,
  55211, 55240.5,
  55240.5, 55271,
  55271, 55301.5,
  55301.5, 55332,
  55332, 55362.5,
  55362.5, 55393.5,
  55393.5, 55424,
  55424, 55454.5,
  55454.5, 55485,
  55485, 55515.5,
  55515.5, 55546.5,
  55546.5, 55576,
  55576, 55605.5,
  55605.5, 55636,
  55636, 55666.5,
  55666.5, 55697,
  55697, 55727.5,
  55727.5, 55758.5,
  55758.5, 55789,
  55789, 55819.5,
  55819.5, 55850,
  55850, 55880 ;
}

```

So if the bounds are used later to compute weights for time statistics the results will be slightly different. But maybe it's not that big of a deal, the differences are not huge anyway.

Peter9192:
Okay, so
- The current (master branch) implementation differs for `full` and `overlap`. For `full`, a new time coordinate is constructed from scratch, and no bounds are guessed. For `overlap`, the time coordinate from 1 of the input cubes is re-used, which presumably already has bounds, so they will be retained.
- The new implementation always constructs a new time coordinate, and `guess_bounds` is used to get new bounds.
- For monthly data, the new implementation sets the center points to the 15th of the month, and then guesses the bounds. This is consistent with what happens in `regrid_time`. However, if the time bounds are guessed, it will just place them in the center between two 15th days, which is sometimes the 30th at 12:00 UTC instead of the 1st at 00:00 UTC. 

So, the new functionality makes the behaviour consistent between overlap and full, and consistent with regrid_time (anticipating #744). However, since guess_bounds doesn't understand the concept of start/end of months, we get slightly shifted bounds. I'm not sure if this is a problem or not, but if it should be addressed, I think that should be in a different PR.

sloosvel:
That's fine for me, I think this pull request is working as intended.

bouweandela:
> For monthly data, the new implementation sets the center points to the 15th of the month, and then guesses the bounds. This is consistent with what happens in regrid_time. However, if the time bounds are guessed, it will just place them in the center between two 15th days, which is sometimes the 30th at 12:00 UTC instead of the 1st at 00:00 UTC.

Not necessarily for this pull request, but it would be nicer to set the lower bound to the start of the month, the upper bound to the start of the next month and then set the points by computing 0.5 * (lower_bound + upper_bound).

Peter9192:
> it would be nicer to set the lower bound to the start of the month, the upper bound to the start of the next month and then set the points by computing 0.5 * (lower_bound + upper_bound).

made an issue for that: https://github.com/ESMValGroup/ESMValCore/issues/803

Peter9192:
#685 is finally merged, so hopefully we can now also proceed with this PR. To do:

- [ ] Use the new tests from in #856 to see if the output of the multimodel statistics function has changed due to this PR. 
- [ ] See where the new test failures come from
- [ ] Have a look at the codacy issues

stefsmeets:
We would also want to add some new tests using sample data for the ensemble statistics. Seems like a good fit.
Other than that, would it be an idea to open a new PR to get rid of all the baggage this PR has accumulated?

Peter9192:
> We would also want to add some new tests using sample data for the ensemble statistics. Seems like a good fit.
Yes, good idea. I've added that to the list

> Other than that, would it be an idea to open a new PR to get rid of all the baggage this PR has accumulated?
It would be good to clean up, but how do you propose we go about that?

stefsmeets:
> It would be good to clean up, but how do you propose we go about that?

Maybe start a new branch and just pick all the changes (maybe squash them). We would lose some of the references and linked issues, so maybe not that good of an idea...

Peter9192:
@stefsmeets and I just had a good discussion on this one. A considerable chunck of this PR is dedicated to refactoring the multi-model statistics function, in order to support the new ensemble statistics. We think it would be good to split this off in a separate PR. We can then easily test whether that causes any changes into the results of the MM-stats functionality. This PR will then only focus on the changes needed in order to use the new ensemble statistics preprocessor.

schlunma:
Hi @sloosvel and @Peter9192, do you have any plans to further work on this PR?

I really like the changes here and they would be really helpful in a recipe that I'm developing. I think with the updates in the multi-model statistics preprocessor from recent releases this PR might also be much simpler now, since the refactoring of the multi-model statistics preprocessor is not necessary anymore.

sloosvel:
> Hi @sloosvel and @Peter9192, do you have any plans to further work on this PR?

Back when I tested it, it seemed to work fine. As for further work maybe it would be nice to set an attribute in the output cube that gives information of how many datasets were used to obtain the result.

schlunma:
Cool! I think it would be nice to have this in v2.5.

Do you think you might have the time to look into the merge conflicts? I could also try to solve them, but you clearly have a better understanding of the code.

sloosvel:
I can try but I think @Peter9192 has the last word in this case

zklaus:
> > Hi @sloosvel and @Peter9192, do you have any plans to further work on this PR?
> 
> Back when I tested it, it seemed to work fine. As for further work maybe it would be nice to set an attribute in the output cube that gives information of how many datasets were used to obtain the result.

Without diving into the too much right now, this kind of information should probably be recorded as a cell method as described in [CF Conventions 1.9, Section 7.3](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#cell-methods). Iris has a dedicated interface for that. A good starting point to look is https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/cube.html#iris.cube.Cube.add_cell_method.

Peter9192:
Good point that this is no longer blocked by the refactoring of multimodel statistics! I don't have much time to look into this in the near future, but feel free to take over! I might be able to help out with questions you have

sloosvel:
> A considerable chunck of this PR is dedicated to refactoring the multi-model statistics function, in order to support the new ensemble statistics. We think it would be good to split this off in a separate PR. We can then easily test whether that causes any changes into the results of the MM-stats functionality. This PR will then only focus on the changes needed in order to use the new ensemble statistics preprocessor.

I think despite this comment, this branch was not splitted and it's a bit confusing to know if all the changes in the multi-model should be included. Even after the refactoring, the multi-model module from `main` differs quite a lot from what's in here. Would it work for you if the changes in this branch only include the grouping and not all the work with the `engine` and such?

Peter9192:
I think the "engine" stuff is a bit outdated, so most of the code in main (notably: the multicube and multiproduct functions) can be kept as they are in main.

sloosvel:
Ok thanks! I updated the branch and added the cell_methods information as well as some changes to include the sub_experiment for the ensemble grouping. 

Also, this branch breaks when the `exclude` setting is included in both the multimodel and the ensemble stats steps. I have pending to fix that.

Peter9192:
Just wondering why you moved ensemble stats to end of chain? The original use case I had in mind for this was one where you first average all the ensemble members, then do a multi-model statistics. That would require that you run ensemble stats first.

sloosvel:
It's still before the multi_model though. Just after time and space operations.

sloosvel:
All green now, except for the Codacy but I don't think the issues are that big of a deal.

valeriupredoi:
whoa @sloosvel this is a bit of a monster PR, well done and poor me as a reviewer :laughing: One quick comment before I start diving into it - Codacy can be ignored apart from the func argument `by` which I would definitely replace with eg `groupby` or `group_criterion` etc

valeriupredoi:
cheers @sloosvel for changing the arg! I've not forgotten about this but I had to deal with a lot of maintenance-related guff, I'll review in the morrow, I promise :+1:

doc/recipe/preprocessor.rst
```diff
@@ -864,6 +865,37 @@ See also :func:`esmvalcore.preprocessor.regrid`
    for resolutions of ``< 0.5`` degrees the regridding becomes very slow and
    will use a lot of memory.
 
+.. _ensemble statistics:
+
+Ensemble statistics
+===================
+For certain use cases it may be desirable to compute ensemble statistics. For
+example to prevent models with many ensemble member getting excessive weight in
```
valeriupredoi, line 873:
```suggestion
example to prevent models with many ensemble members getting excessive weight in
```

doc/recipe/preprocessor.rst
```diff
@@ -907,18 +939,33 @@ days in a year may vary between calendars, (sub-)daily data are not supported.
           statistics: [mean, median]
           exclude: [NCEP]
 
-see also :func:`esmvalcore.preprocessor.multi_model_statistics`.
+Multi-model statistics now also supports a ``groupby`` argument. You can group by
+any dataset key (``project``, ``experiment``, etc.) or a combination of keys. You can
```
valeriupredoi, line 943:
```suggestion
any dataset key (``project``, ``experiment``, etc.) or a combination of keys in a list. You can
```

esmvalcore/_data_finder.py
```diff
@@ -484,17 +483,33 @@ def get_output_file(variable, preproc_dir):
...
-    updated_var = copy.deepcopy(variable)
-    updated_var['timerange'] = updated_var['timerange'].replace('/', '-')
-    template = os.path.join(
+def get_multiproduct_filename(attributes, preproc_dir):
+    """Get ensemble/multi-model filename depending on settings."""
+    relevant_keys = [
+        'project', 'dataset', 'exp', 'ensemble_statistics',
+        'multi_model_statistics', 'mip', 'short_name'
+    ]
+
```
valeriupredoi, line 492: do we really want to allow for inter-project grouping? This is tricky: CMIP5+CMIP6 maybe makes some sense, but CMIP+OBS?

esmvalcore/_recipe.py
```diff
@@ -703,31 +705,80 @@ def _update_multi_dataset_settings(variable, settings):
...
+    relevant_settings = {
+        'output_products': defaultdict(dict)
+    }  # pass to ancestors
+
+    output_products = set()
+    for identifier, products in _group_products(products, by_key=grouping):
+        common_attributes = _get_common_attributes(products)
+
+        for statistic in settings.get('statistics'):
+
```
valeriupredoi, line 757:
```suggestion
```

esmvalcore/_recipe.py
```diff
@@ -703,31 +705,80 @@ def _update_multi_dataset_settings(variable, settings):
...
+            common_attributes[step] = _get_tag(step, identifier, statistic)
+
```
valeriupredoi, line 759:
```suggestion
```

esmvalcore/_recipe.py
```diff
@@ -703,31 +705,80 @@ def _update_multi_dataset_settings(variable, settings):
...
+            filename = get_multiproduct_filename(common_attributes,
+                                                 preproc_dir)
+            common_attributes['filename'] = filename
+
```
valeriupredoi, line 763:
```suggestion
```

esmvalcore/_recipe.py
```diff
@@ -703,31 +705,80 @@ def _update_multi_dataset_settings(variable, settings):
...
+            statistic_product = PreprocessorFile(common_attributes,
+                                                 downstream_settings)
+
```
valeriupredoi, line 766:
```suggestion
```

esmvalcore/_recipe.py
```diff
@@ -703,31 +705,80 @@ def _update_multi_dataset_settings(variable, settings):
...
+            output_products.add(statistic_product)
+
```
valeriupredoi, line 768:
```suggestion
```

esmvalcore/_recipe.py
```diff
@@ -891,10 +951,48 @@ def _get_preprocessor_products(variables, profile, order, ancestor_products,
...
             f'{separator.join(sorted(missing_vars))}')
 
-    _update_statistic_settings(products, order, config_user['preproc_dir'])
     check.reference_for_bias_preproc(products)
 
-    for product in products:
+    ensemble_step = 'ensemble_statistics'
+    multi_model_step = 'multi_model_statistics'
+    if ensemble_step in profile:
+
```
valeriupredoi, line 959:
```suggestion
```

esmvalcore/_recipe.py
```diff
@@ -891,10 +951,48 @@ def _get_preprocessor_products(variables, profile, order, ancestor_products,
...
+        update_ancestors(
+            ancestors=products,
+            step=ensemble_step,
+            downstream_settings=ensemble_settings,
+        )
+    else:
+        ensemble_products = products
+
+    if multi_model_step in profile:
+
```
valeriupredoi, line 973:
```suggestion
```

esmvalcore/_recipe_checks.py
```diff
@@ -220,16 +219,88 @@ def extract_shape(settings):
...
+def _verify_statistics(statistics, step):
+    """Raise error if multi-model statistics cannot be verified."""
+    valid_names = ["mean", "median", "std", "min", "max"]
     valid_patterns = [r"^(p\d{1,2})(\.\d*)?$"]
-    if not (statistic in valid_names
-            or re.match(r'|'.join(valid_patterns), statistic)):
+
+    for statistic in statistics:
+        if not (statistic in valid_names
+                or re.match(r'|'.join(valid_patterns), statistic)):
```
valeriupredoi, line 229: I don't understand why the regexp is needed once we have a set in stone list of valid statistics?

esmvalcore/_recipe_checks.py
```diff
@@ -220,16 +219,88 @@ def extract_shape(settings):
...
+            f"Got {span}.")
+
+
+def _verify_groupby(groupby):
+    """Raise error if groupby arguments cannot be verified."""
+    if not isinstance(groupby, list):
+        raise RecipeError(
+            "Invalid value encountered for `groupby` in preprocessor "
+            "`multi_model_statistics`.`groupby` must be defined as a "
+            f"list. Got {groupby}.")
```
valeriupredoi, line 252: wait - you're only checking if it's a list, do the list members get checked anywhere? can I groupby `geese` :duck:

esmvalcore/_recipe_checks.py
```diff
@@ -220,16 +219,88 @@ def extract_shape(settings):
...
+def _verify_arguments(given, expected):
+    """Raise error if arguments cannot be verified."""
+    for key in given:
+        if key not in expected:
+            raise RecipeError(
+                f"Unexpected keyword argument encountered: {key}. Valid "
+                f"keywords are: {expected}.")
+
+
+def multimodel_statistics(settings):
```
valeriupredoi, line 272: I would use a different function name, this is the same as the actual preprocessor, also maybe make it private too since it may get picked up in the API documentation

esmvalcore/_recipe_checks.py
```diff
@@ -220,16 +219,88 @@ def extract_shape(settings):
...
+
+    statistics = settings.get('statistics', None)  # required
+    if statistics:
+        _verify_statistics(statistics, 'multi_model_statistics')
+
+    keep_input_datasets = settings.get('keep_input_datasets', True)
+    _verify_keep_input_datasets(keep_input_datasets)
+
+
+def ensemble_statistics(settings):
```
valeriupredoi, line 293: same here, different name and maybe make it private

valeriupredoi (changes requested):
very nice work! Some comments after I've gone through half the PR :+1:

tests/integration/test_recipe.py
```diff
@@ -1841,6 +1841,184 @@ def test_extract_shape_raises(tmp_path, patched_datafinder, config_user,
...
+    task._initialize_product_provenance()
+    assert next(iter(products)).provenance is not None
+
+
+def test_multi_model_statistics(tmp_path, patched_datafinder, config_user):
+    statistics = ['mean', 'max']
+    diagnostic = 'diagnostic_name'
+    variable = 'pr'
+    preprocessor = 'multi_model_statistics'
+
```
valeriupredoi, line 1915: you can use the list of all sorts of stuff decorator `@pytest.mark.parametrize()` to throw in all sorts of parameters for the test, no need to list them internally and go one by one

tests/unit/preprocessor/_multimodel/test_multimodel.py
```diff
@@ -514,13 +514,37 @@ def test_unify_time_coordinates():
...
+    def __init__(self, cube=None, attributes=None):
         if cube:
             self.cubes = [cube]
+        if attributes:
+            self.attributes = attributes
 
     def wasderivedfrom(self, product):
         pass
 
+    def group(self, keys: list) -> str:
```
valeriupredoi, line 526: isn't this the same function as in the source code? You can just import that one

tests/unit/preprocessor/_other/test_other.py
```diff
@@ -43,5 +44,26 @@ def test_clip(self):
...
+            attributes={
+                'project': 'A',
+                'dataset': 'B',
+                'filename': 'A_B.nc'},
+            settings={}),
+        PreprocessorFile(
+            attributes={
+                'project': 'A',
+                'dataset': 'C',
+                'filename': 'A_C.nc'},
```
valeriupredoi, line 59: they are essentially the same - do you not want to vary a third parameter e.g. `A_B_x_1990-1991.nc` and `A_C_y_1950-1991.nc`?

valeriupredoi:
@schlunma I've done the full code review, but am exhausted (heh :grin: ) - would you mind running a couple tests with this new functionality, since you are the actual testing emeritus, please mate :beer:

schlunma:
Sure, will do after all your comments have been addressed :+1:

I will do the tests with an actual recipe, so I can extend `ESMValTool/esmvaltool/recipes/examples/recipe_preprocessor_test.yml`, will open a PR for that!

valeriupredoi:
@schlunma that's brilliant, cheers muchly!

esmvalcore/preprocessor/__init__.py
```diff
@@ -372,8 +375,7 @@ class PreprocessorFile(TrackedFile):
     """Preprocessor output file."""
 
     def __init__(self, attributes, settings, ancestors=None):
-        super(PreprocessorFile, self).__init__(attributes['filename'],
-                                               attributes, ancestors)
+        super().__init__(attributes['filename'], attributes, ancestors)
```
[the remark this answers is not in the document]
sloosvel: @Peter9192 's work not mine!

esmvalcore/_data_finder.py
sloosvel: You can exclude the obs in the `exclude` parameter

esmvalcore/_recipe_checks.py
sloosvel: For the percentiles I guess?

esmvalcore/_recipe_checks.py
sloosvel: You could but it would not do anything because it would not  find any geese attribute. I don't think you can really predict all the possible grouping options so I am not sure how to check it's values.

tests/unit/preprocessor/_other/test_other.py
sloosvel: This test is just to check that _group_products can work with both strings or lists, as it was something thas was missing in the codecov reports:

```
    grouped_by_string = _group_products(products, 'project')
    grouped_by_list = _group_products(products, ['project'])
```
The value of the filenames is not that relevant in this test, it's something that get's checked in other tests in `test_recipe.py`

esmvalcore/_recipe_checks.py
sloosvel: I'll change the name but plenty of other checks like this one are not private (`valid_time_selection`, `reference_for_bias_preproc`) so I am not sure why this and the ensemble one should be any different.

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -412,14 +432,45 @@ def multi_model_statistics(products,
...
+                        output_products, keep_input_datasets=True):
+    """Entry point for ensemble statistics.
+
+    The products are grouped and the cubes are extracted from
+    the products. Resulting cubes are assigned to `output_products`.
+
+    See Also
+    --------
+    multicube_statistics_iris : core statistics function.
+    """
```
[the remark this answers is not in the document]
sloosvel: I think this comes from a merge conflict and it can be removed.

esmvalcore/_data_finder.py
valeriupredoi: cheers, do you have a clear example for the use of `exclude`?

esmvalcore/_recipe_checks.py
valeriupredoi: aah right, good call!

esmvalcore/_recipe_checks.py
valeriupredoi: hahah, geese was a red herring of course :fish: Let me ask you this way then - if a user wants to groubpy geese, will there be an exception raised saying geese is not a valid grouping parameter, groupby should be either of [project, dataset, mip ...]? Because you can use that as a restrictor in the check here

esmvalcore/_recipe_checks.py
valeriupredoi: ah right, I'd have made them all private, prob legacy of ye olde times, yes please change the name, we need to be careful not to reuse the same function names in different modules

tests/unit/preprocessor/_other/test_other.py
valeriupredoi: excellent then, cheers for the explanation :+1:

esmvalcore/_recipe_checks.py
schlunma: I don't think we should restrict this to the usual facets (like `project`, `dataset`, etc.) but rather allow arbitrary grouping by e.g. using `groupby: animal` with something like

```yml
- {project: CMIP5, dataset: CanESM2, animal: goose}
- {project: CMIP6, dataset: CanESM5, animal: herring}
- {project: OBS, dataset: ERA5, animal: goose}
- {project: CMIP6 dataset: UKESM}
```

BTW, what happens when at least one dataset does not include the `groupby` attribute like the last example listed here?

valeriupredoi: OK I like that! sounds good to me, better coverage for the mess that'll be CMIP7 :grin: Good point about the bits that don't have the groupby parameter :+1:

tests/integration/test_recipe.py
sloosvel: I am not sure this is written in this way, I did not write this test sorry.

esmvalcore/_recipe_checks.py
sloosvel: >BTW, what happens when at least one dataset does not include the groupby attribute like the last example listed here

It would go to another group that has an empty identifier. This case would eventually crash when trying to compute multimodel stats on a single cube.

esmvalcore/_data_finder.py
sloosvel: Just added a test

esmvalcore/preprocessor/__init__.py
```diff
@@ -461,13 +463,35 @@ def is_closed(self):
...
         """Initialize the entity representing the file."""
-        super(PreprocessorFile, self)._initialize_entity()
+        super()._initialize_entity()
         settings = {
             'preprocessor:' + k: str(v)
             for k, v in self.settings.items()
         }
         self.entity.add_attributes(settings)
 
+    def group(self, keys: list) -> str:
```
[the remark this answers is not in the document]
sloosvel: Why do think it should be private?

tests/integration/test_recipe.py
Peter9192: I don't think it's going one by one. I guess it was written like this because those variables are reused multiple times (in the recipe definition and in the check). Indeed, since it's written like this, you can now easily extend the tests with multiple combinations of variables etc., but since it's only one test now, adding `parametrize` is unnecessary.

esmvalcore/_data_finder.py
valeriupredoi: excellent, it'd be good to have one in the preprocessor example recipe - @schlunma is putting one together

esmvalcore/_recipe_checks.py
valeriupredoi: is this explained/pointed to in the docstring/documentation?

esmvalcore/preprocessor/__init__.py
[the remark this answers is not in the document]
valeriupredoi: it don't have to be, but anything that's not private kind of needs proper numpy-style documentation

esmvalcore/preprocessor/_multimodel.py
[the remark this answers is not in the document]
valeriupredoi: OK thanks, this needs addressing before merge sadly :+1:

tests/unit/preprocessor/_multimodel/test_multimodel.py
sloosvel: I am having some issues trying to mock this. And I don't have much time to focus on that. Does it bother you too much if it stays like this?

valeriupredoi: no not at all, leave it like this :+1:

esmvalcore/preprocessor/_multimodel.py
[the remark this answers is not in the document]
sloosvel: Added docstrings here as well.

esmvalcore/preprocessor/__init__.py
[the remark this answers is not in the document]
sloosvel: All methods in this class are documented in the same way, why is this an issue for `group` but not for `check`,  `apply` and other?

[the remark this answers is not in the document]
valeriupredoi: well, if some old methods are poorly documented it don't mean new ones should be too :grin: It's fine, don't worry, I think in the interest of time we should leave it like this - I really like this feature and it would be awesome if we pushed it to v2.5

esmvalcore/preprocessor/_multimodel.py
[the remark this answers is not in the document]
valeriupredoi: awesome, cheers!

esmvalcore/_recipe_checks.py
sloosvel: Added

sloosvel: I think this is the only issue left to be addressed. Does this explanation of this behaviour work for you?

valeriupredoi (changes requested):

esmvalcore/_recipe_checks.py
valeriupredoi: cheers Saskia, please see my wee bitty code suggestion below. Also, one more (and last, I promise) thing: cheers for tidying up `ensemble_statistics(products, statistics,
                        output_products, keep_input_datasets=True)` docstring - could you please add a quick sentence there saying it's essentially multimodel_statistic but for ensembles - it's not really clear what the func does, juging just from the doctring. Cheers :beer:

sloosvel: Yes of course, adding it now.

valeriupredoi: you're a star :star:

sloosvel: Is it clearer now?

valeriupredoi: indeed it is, cheers Saskia, this will get a big green check mark from me :smile:

valeriupredoi (approved):
great many thanks @sloosvel and @Peter9192 - this looks very good and mergeable to me! @schlunma was kind enough to be volunteered to test (cheers, bud!) so she's all yours :beer:

schlunma (commented):
Really great work Saskia and Peter! I just tested the `ensemble_statistics` preprocessor alone and it works just like expected :tada: 

However, when running the combination of `ensemble_statistics` and `multi_model_statistics` in combination as advertised in the documentation I get the following error:

```
  File "/mnt/lustre02/work/bd0854/b309141/mambaforge/envs/esm/lib/python3.9/multiprocessing/pool.py", line 125, in worker
  File "/mnt/lustre02/work/bd0854/b309141/repos/ESMValCore/esmvalcore/_task.py", line 780, in _run_task
    output_files = task.run()
  File "/mnt/lustre02/work/bd0854/b309141/repos/ESMValCore/esmvalcore/_task.py", line 253, in run
  File "/mnt/lustre02/work/bd0854/b309141/repos/ESMValCore/esmvalcore/preprocessor/__init__.py", line 592, in _run
    self.products = _apply_multimodel(self.products, step,
  File "/mnt/lustre02/work/bd0854/b309141/repos/ESMValCore/esmvalcore/preprocessor/__init__.py", line 506, in _apply_multimodel
    result = preprocess(products - exclude, step, **settings)
  File "/mnt/lustre02/work/bd0854/b309141/repos/ESMValCore/esmvalcore/preprocessor/__init__.py", line 342, in preprocess
    result.append(_run_preproc_function(function, items, settings,
  File "/mnt/lustre02/work/bd0854/b309141/repos/ESMValCore/esmvalcore/preprocessor/__init__.py", line 302, in _run_preproc_function
  File "/mnt/lustre02/work/bd0854/b309141/repos/ESMValCore/esmvalcore/preprocessor/_multimodel.py", line 441, in multi_model_statistics
    group_statistics = _multiproduct_statistics(
  File "/mnt/lustre02/work/bd0854/b309141/repos/ESMValCore/esmvalcore/preprocessor/_multimodel.py", line 348, in _multiproduct_statistics
    statistics_product = output_products[statistic]
KeyError: 'max'
```
[recipe_preprocessor_test.yml.txt](https://github.com/ESMValGroup/ESMValCore/files/7996766/recipe_preprocessor_test.yml.txt)
[main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/7996768/main_log_debug.txt)

In addition, I have a couple of smaller comments to the documentation.

Thanks a lot for all your work on this!

sloosvel:
>However, when running the combination of ensemble_statistics and multi_model_statistics in combination as advertised in the documentation I get the following error:

Before the recent changes, `ensemble_statistics` was using an older routine that did not consider the option `keep_input_datasets`. When the branch was updated, I introduced `keep_input_datasets` as valid setting in `ensemble_statistics` to match with `multi_model_statistics` and set it's default behaviour to `True`. It looks like this is breaking the chain. I ran the recipe with `keep_input_datasets=false` in `ensemble_statistics` and it finished successfully.

Maybe it would be better to remove the option to set the parameter in `ensemble_statistics` (not consider it as a valid key in the checks) and set the value to False. Unless there is a good reason to keep the input enters `ensemble_statistics`.

schlunma:
I don't think dropping `keep_input_datasets` completely is a good idea. I can think of many use cases where you want to analyze your ensemble means but also the individual ensemble members.

I don't really understand what's going on here. `ensemble_statistics` runs completely fine with `keep_input_datasets=true`, and `multi_model_statistics` also works fine when a dataset does not have the  `groupby` attribute that we ask for. I would assume that if I use `groupy=[ensemble_statistics]` the ensemble members will simply get an empty entry for `ensemble_statistics`.

I got the same error for `groupby: [tag]`, which also has datasets that do not have a tag, but `groupby: [dataset]` (every dataset has this key) worked just fine. (Had to account for a bug in `iris` regarding scalar coordinates here, see SciTools/iris#3584).

Peter9192:
> Maybe it would be better to remove the option to set the parameter in ensemble_statistics (not consider it as a valid key in the checks) and set the value to False. Unless there is a good reason to keep the input enters ensemble_statistics.

I remember we [discussed this](https://github.com/ESMValGroup/ESMValCore/issues/975) also in the refactor of multi-model. Back then I thought it was strange behaviour, and if you actually want to do this it can easily be done in the recipe. However, because people got used to it and there were several recipes using this "functionality", we decided to keep it (or actually, submitted this option as a fix to add it back). 

Since this is a new feature with no legacy recipes, I think it makes sense to *not* add it from the start.

schlunma:
@Peter9192 Agreed, makes sense. In this case I would suggest to add a note to the documentation about this:

1. For `multi_model_statistics`, add 1 sentence about `keep_input_datasets` (this is not described yet)
2. Add the following nice example from @Peter9192 to `ensemble_statistics` with something along the lines "If a diagnostic requires ensemble means (or other statistics) AND the individual ensemble members, you can use the following preprocessor combination in the recipe:"

```yml
preprocessors:
  everything_else: &everything_else
    area_statistics: ... 
    regrid_time: ...  
  multimodel:
    <<: *everything_else
    ensemble_statistics:
      ...

variables:
  tas_all_cubes:
    short_name: tas
    preprocessor: everything_else
    ...
  tas_mmstats:
    short_name: tas
    preprocessor: multimodel
    ...
```

Maybe also add a short `.. note::` that explicitly states that `ensemble_statistics` does not support `keep_input_datasets`?

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -412,14 +433,69 @@ def multi_model_statistics(products,
...
+
+    See Also
+    --------
+    :func:`esmvalcore.preprocessor.multi_model_statistics` for
+    the full description of the core statistics function.
+    """
+    ensemble_grouping = ('project', 'dataset', 'exp', 'sub_experiment')
+    return multi_model_statistics(
+        products=products,
+        span='overlap',
```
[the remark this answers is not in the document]
sloosvel: All ensemble members for a single dataset I have seen span for the same timerange, so I guess that is why it's fine to assume that they always overlap. But I added the option to keep it flexible, it can always be changed back anyway.

doc/recipe/preprocessor.rst
```diff
@@ -864,6 +865,37 @@ See also :func:`esmvalcore.preprocessor.regrid`
...
+.. _ensemble statistics:
+
+Ensemble statistics
+===================
+For certain use cases it may be desirable to compute ensemble statistics. For
+example to prevent models with many ensemble members getting excessive weight in
+the multi-model statistics functions.
+
+Theoretically, ensemble statistics are a special case (grouped) multi-model
+statistics. However, they should typically be computed earlier in the workflow.
```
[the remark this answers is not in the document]
sloosvel: done

doc/recipe/preprocessor.rst
```diff
@@ -907,18 +939,56 @@ days in a year may vary between calendars, (sub-)daily data are not supported.
...
+
+    preprocessors:
+      example_preprocessor:
+        ensemble_statistics:
+          statistics: [median, mean]
+        multi_model_statistics:
+          span: overlap
+          statistics: [min, max]
+          groupby: [ensemble_statistics, tag]
+          exclude: [NCEP]
```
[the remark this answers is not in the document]
sloosvel: I removed the `tag` from this example, below I think there is another example that already shows how the grouping works.

doc/recipe/preprocessor.rst
```diff
@@ -864,6 +865,42 @@ See also :func:`esmvalcore.preprocessor.regrid`
...
+
+Ensemble statistics
+===================
+For certain use cases it may be desirable to compute ensemble statistics. For
+example to prevent models with many ensemble members getting excessive weight in
+the multi-model statistics functions.
+
+Theoretically, ensemble statistics are a special case (grouped) multi-model
+statistics. This grouping is performed taking into account the dataset tags
+`project`, `dataset`, `experiment`, and (if present) `sub_experiment`.
```
schlunma, line 878:
```suggestion
``project``, ``dataset``, ``experiment``, and (if present) ``sub_experiment``.
```

doc/recipe/preprocessor.rst
```diff
@@ -864,6 +865,42 @@ See also :func:`esmvalcore.preprocessor.regrid`
...
+      example_preprocessor:
+        ensemble_statistics:
+          statistics: [mean, median]
+
+This preprocessor function exposes the iris analysis package, and works with all
+(capitalized) statistics from the :mod:`iris.analysis` package
+that can be executed without additional arguments (e.g. percentiles are not
+supported because it requires additional keywords: percentile.).
+
+Note that `ensemble_statistics` will not return the single model and ensemble files,
```
schlunma, line 899:
```suggestion
Note that ``ensemble_statistics`` will not return the single model and ensemble files,
```

doc/recipe/preprocessor.rst
```diff
@@ -864,6 +865,42 @@ See also :func:`esmvalcore.preprocessor.regrid`
...
+only the requested ensemble statistics results.
```
schlunma, line 900: Can you add the comment about the special preprocessor that can be used to get the enemble statistics and the raw datasets here (l.981f). I think this should go to the `ensemble_statistics` section.

doc/recipe/preprocessor.rst
```diff
@@ -893,6 +930,9 @@ across overlapping times only (``span: overlap``) or across the full time span
 of the combined models (``span: full``). The preprocessor sets a common time
 coordinate on all datasets. As the number of days in a year may vary between
 calendars, (sub-)daily data with different calendars are not supported.
+The preprocessor saves both the input single model files as well as the multi-model
+results. In case of not wanting to keep the single model files, set
```
schlunma, line 934:
```suggestion
results. In case you do not want to keep the single model files, set the
```

doc/recipe/preprocessor.rst
```diff
@@ -893,6 +930,9 @@ across overlapping times only (``span: overlap``) or across the full time span
...
+parameter `keep_input_datasets` to `false` (default value is `true`).
```
schlunma, line 935:
```suggestion
parameter ``keep_input_datasets`` to ``false`` (default value is ``true``).
```

doc/recipe/preprocessor.rst
```diff
@@ -901,24 +941,89 @@ days in a year may vary between calendars, (sub-)daily data are not supported.
 .. code-block:: yaml
 
     preprocessors:
-      multi_model_preprocessor:
+      multi_model_save_input:
```
schlunma, line 944: Can you move this section up so it directly follows "...set the parameter `keep_input_datasets` to `false` (default value is `true`).?

doc/recipe/preprocessor.rst
```diff
@@ -901,24 +941,89 @@ days in a year may vary between calendars, (sub-)daily data are not supported.
...
         multi_model_statistics:
           span: overlap
           statistics: [mean, median]
           exclude: [NCEP]
+            multi_model_save_input:
```
schlunma, line 949:
```suggestion
```

doc/recipe/preprocessor.rst
```diff
@@ -901,24 +941,89 @@ days in a year may vary between calendars, (sub-)daily data are not supported.
...
+          statistics: [median, mean]
+        multi_model_statistics:
+          span: overlap
+          statistics: [min, max]
+          groupby: [ensemble_statistics]
+          exclude: [NCEP]
+
+This will first compute ensemble mean and median, and then compute the multi-model
+min and max separately for the ensemble means and medians. Note that this combination
+will not save the individual ensemble members, only the ensemble and multimodel statistics results.
```
schlunma, line 981: Move this paragraph to `ensemble_statistics`

doc/recipe/preprocessor.rst
```diff
@@ -901,24 +941,89 @@ days in a year may vary between calendars, (sub-)daily data are not supported.
...
+      preprocessor: everything_else
+      ...
+    tas_multimodel:
+      short_name: tas
+      preprocessor: multimodel
+      ...
+
+When grouping by a tag not defined in all datasets, the datasets missing the tag will
+be grouped together. In the example below, datasets `UKESM` and `ERA5` would belong to the same
+group, while the other datasets would belong to either `group1` or `group2`
```
schlunma, line 1006:
```suggestion
group, while the other datasets would belong to either ``group1`` or ``group2``
```

doc/recipe/preprocessor.rst
```diff
@@ -901,24 +941,89 @@ days in a year may vary between calendars, (sub-)daily data are not supported.
...
+.. code-block:: yaml
+  preprocessors:
+    everything_else: &everything_else
+      area_statistics: ...
+      regrid_time: ...
+    multimodel:
+      <<: *everything_else
+      ensemble_statistics:
+
+variables:
```
schlunma, line 994: The indentation is wrong here

schlunma (commented):
Great! I tested this preprocessor extensively with [this](https://github.com/ESMValGroup/ESMValTool/blob/7fc747e1fc44e23aee8a8452b2a31186eb4a0f3a/esmvaltool/recipes/examples/recipe_preprocessor_test.yml) recipe and everything works perfectly! Many thanks!!

I just have some additional comments regarding the doc, afterwards this can be merged :tada:

Peter9192:
Just wanted to step in and say awesome work here @sloosvel! Thanks for picking up this PR. Also cheers to @schlunma and @valeriupredoi for kicking and reviewing it. Have a nice weekend!

valeriupredoi:
yeah man, I second that! Good weekend too, 30min til pub for me :grin: :beer:

doc/recipe/preprocessor.rst
sloosvel: When `ensemble_statistics` is described, `multi_model_statistics` has not been described yet so I thought it would fit better in here. But whatever it can be moved

schlunma:
The documentation build fails with a

```
failed to reach any of the inventories with the following issues:
intersphinx inventory 'https://docs.scipy.org/doc/scipy/reference/objects.inv' not fetchable due to <class 'requests.exceptions.HTTPError'>: 404 Client Error: Not Found for url: https://docs.scipy.org/doc/scipy/reference/objects.inv
```

now. This is most certainly not related to this PR.

Any tips on this @ESMValGroup/esmvaltool-coreteam? I don't think `scipy` migrated their doc, so I guess this is just a temporary problem. Will try again in a couple of hours, but I think we should merge this even when this test is failing. The commit before that passed all tests.

zklaus:
I came to the same conclusion as @schlunma regarding the broken intersphinx link. It is possible that we need to change something (see scipy/scipy#14267), but let's wait for a bit until things get clearer.

schlunma (approved):
Merging this now since the failing tests are not related to this PR.

Thanks for all your work on this @sloosvel and @Peter9192, this is a great addition to the tool! :tada:

Peter9192:
Awesome! Also thanks to @stefsmeets for your contributions 👍

PR #677, opened 2020-06-16
Make multimodel work correctly with yearly data

### LINKED ISSUE #665: Multi-model statistics shift time coordinate
I tried to calculate the multi-model mean of several models with different time axes. So I chose `span: full` . After that the time axis of the MulitModelMean preproc file has different time points. And it does not fit to the overall time period of the models anymore. 
Did I something wrong?
Could anybody help me?
Thanks!

recipe to test:
```
# ESMValTool
---
documentation:

  description: |
    xxx

preprocessors:

  clim_ref:
    regrid:
      target_grid: reference_dataset
      scheme: linear
    multi_model_statistics:
      span: full
      statistics: [mean]
      exclude: [reference_dataset]



datasets:
  - {dataset: bccr_bcm2_0,  institute: BCCR}
  - {dataset: cccma_cgcm3_1, institute: CCCMA}
  - {dataset: cccma_cgcm3_1_t63, institute: CCCMA}
  - {dataset: csiro_mk3_0, institute: CSIRO, start_year: 1871}
  - {dataset: gfdl_cm2_0, institute: GFDL, start_year: 1861}
  - {dataset: gfdl_cm2_1, institute: GFDL, start_year: 1861} 
  - {dataset: giss_aom, institute: NASA}
  - {dataset: giss_model_e_h, institute: NASA, start_year: 1880}
  - {dataset: giss_model_e_r, institute: NASA, start_year: 1880}
  - {dataset: iap_fgoals1_0_g, institute: LASG}
  - {dataset: ingv_echam4, institute: INGV, start_year: 1870}
  - {dataset: inmcm3_0, institute: INM, start_year: 1871}
  - {dataset: ipsl_cm4, institute: IPSL, start_year: 1860}
  #- {dataset: miroc3_2_hires, institute: NIES, start_year: 1900}
  - {dataset: miroc3_2_medres, institute: NIES}
  #- {dataset: miub_echo_g, institute: MIUB-KMA} #something wrong with numeration of years
  - {dataset: mpi_echam5, institute: MPIM, start_year: 1860}
  - {dataset: mri_cgcm2_3_2a, institute: MRI, start_year: 1851}
  - {dataset: ncar_ccsm3_0, institute: NCAR, start_year: 1870}
  - {dataset: ncar_pcm1, institute: NCAR, start_year: 1890}
  - {dataset: ukmo_hadcm3, institute: UKMO, start_year: 1860}
  - {dataset: ukmo_hadgem1, institute: UKMO, start_year: 1860}
  - {dataset: HadCRUT4, project: OBS, type: ground, version: 1, tier: 2,
     mip: Amon, end_year: 2017}


diagnostics:

  fig_1_cmip3: 
    description: CMIP3 timeseries of near-surface temperature anomalies
    variables:
      tas: 
        preprocessor: clim_ref   
        reference_dataset: HadCRUT4   
        project: CMIP3
        mip: A1
        modeling_realm: atm
        exp: 20c3m
        frequency: mo
        ensemble: run1
        start_year: 1850
        end_year: 1999
```

valeriupredoi:

**Tasks**

-   [x] Add unit tests

* * *

Closes #665

### DIFF
--- a/esmvalcore/preprocessor/_multimodel.py
+++ b/esmvalcore/preprocessor/_multimodel.py
@@ -194,8 +194,7 @@ def _datetime_to_int_days(cube, overlap=False):
 
 def _align_yearly_axes(cube):
     """Perform a time-regridding operation to align time axes for yr data."""
-    time_cells = [cell.point for cell in cube.coord('time').cells()]
-    years = [date_obj.year for date_obj in time_cells]
+    years = [cell.point.year for cell in cube.coord('time').cells()]
     # be extra sure that the first point is not in the previous year
     if len(years) > 2:
         delta = int(years[2]) - int(years[1])
--- a/esmvalcore/preprocessor/_multimodel.py
+++ b/esmvalcore/preprocessor/_multimodel.py
@@ -111,10 +111,12 @@ def _put_in_cube(template_cube, cube_data, statistic, t_axis):
     if t_axis is None:
         times = template_cube.coord('time')
     else:
+        unit_name = template_cube.coord('time').units.name
+        tunits = cf_units.Unit(unit_name, calendar="standard")
         times = iris.coords.DimCoord(
             t_axis,
             standard_name='time',
-            units=template_cube.coord('time').units)
+            units=tunits)
 
     coord_names = [c.long_name for c in template_cube.coords()]
     coord_names.extend([c.standard_name for c in template_cube.coords()])
@@ -164,7 +166,7 @@ def _put_in_cube(template_cube, cube_data, statistic, t_axis):
     return stats_cube
 
 
-def _datetime_to_int_days(cube, overlap=False):
+def _datetime_to_int_days(cube):
     """Return list of int(days) converted from cube datetime cells."""
     cube = _align_yearly_axes(cube)
     time_cells = [cell.point for cell in cube.coord('time').cells()]
@@ -175,13 +177,7 @@ def _datetime_to_int_days(cube, overlap=False):
         # real_date resets the actual data point day
         # to the 1st of the month so that there are no
         # wrong overlap indices
-        # NOTE: this workaround is good only
-        # for monthly data
-        if overlap:
-            real_date = datetime(date_obj.year, date_obj.month, 1, 0, 0, 0)
-        else:
-            real_date = datetime(date_obj.year,
-                                 date_obj.month, 15, 0, 0, 0)
+        real_date = datetime(date_obj.year, date_obj.month, 1, 0, 0, 0)
         real_dates.append(real_date)
 
     # get the number of days starting from the reference unit
@@ -215,7 +211,7 @@ def _get_overlap(cubes):
     """
     all_times = []
     for cube in cubes:
-        span = _datetime_to_int_days(cube, overlap=True)
+        span = _datetime_to_int_days(cube)
         start, stop = span[0], span[-1]
         all_times.append([start, stop])
     bounds = [range(b[0], b[-1] + 1) for b in all_times]
@@ -233,7 +229,7 @@ def _slice_cube(cube, t_1, t_2):
     of common time-data elements.
     """
     time_pts = [t for t in cube.coord('time').points]
-    converted_t = _datetime_to_int_days(cube, overlap=True)
+    converted_t = _datetime_to_int_days(cube)
     idxs = sorted([
         time_pts.index(ii) for ii, jj in zip(time_pts, converted_t)
         if t_1 <= jj <= t_2
--- a/esmvalcore/preprocessor/_multimodel.py
+++ b/esmvalcore/preprocessor/_multimodel.py
@@ -192,7 +192,7 @@ def _align_yearly_axes(cube):
     """Perform a time-regridding operation to align time axes for yr data."""
     years = [cell.point.year for cell in cube.coord('time').cells()]
     # be extra sure that the first point is not in the previous year
-    if not 0 in np.diff(years):
+    if 0 not in np.diff(years):
         return regrid_time(cube, 'yr')
     else:
         return cube
--- a/esmvalcore/preprocessor/_multimodel.py
+++ b/esmvalcore/preprocessor/_multimodel.py
@@ -194,8 +194,7 @@ def _align_yearly_axes(cube):
     # be extra sure that the first point is not in the previous year
     if 0 not in np.diff(years):
         return regrid_time(cube, 'yr')
-    else:
-        return cube
+    return cube
 
 
 def _get_overlap(cubes):
--- a/esmvalcore/preprocessor/_multimodel.py
+++ b/esmvalcore/preprocessor/_multimodel.py
@@ -164,12 +164,10 @@ def _put_in_cube(template_cube, cube_data, statistic, t_axis):
     return stats_cube
 
 
-def _datetime_to_int_days(cube):
+def _datetime_to_int_days(cube, overlap=False):
     """Return list of int(days) converted from cube datetime cells."""
     cube = _align_yearly_axes(cube)
     time_cells = [cell.point for cell in cube.coord('time').cells()]
-    time_unit = cube.coord('time').units.name
-    time_offset = _get_time_offset(time_unit)
 
     # extract date info
     real_dates = []
@@ -179,10 +177,18 @@ def _datetime_to_int_days(cube):
         # wrong overlap indices
         # NOTE: this workaround is good only
         # for monthly data
-        real_date = datetime(date_obj.year, date_obj.month, 1, 0, 0, 0)
+        if overlap:
+            real_date = datetime(date_obj.year, date_obj.month, 1, 0, 0, 0)
+        else:
+            real_date = datetime(date_obj.year,
+                                 date_obj.month, 15, 0, 0, 0)
         real_dates.append(real_date)
 
+    # get the number of days starting from the reference unit
+    time_unit = cube.coord('time').units.name
+    time_offset = _get_time_offset(time_unit)
     days = [(date_obj - time_offset).days for date_obj in real_dates]
+
     return days
 
 
@@ -214,7 +220,7 @@ def _get_overlap(cubes):
     """
     all_times = []
     for cube in cubes:
-        span = _datetime_to_int_days(cube)
+        span = _datetime_to_int_days(cube, overlap=True)
         start, stop = span[0], span[-1]
         all_times.append([start, stop])
     bounds = [range(b[0], b[-1] + 1) for b in all_times]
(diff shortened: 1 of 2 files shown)


valeriupredoi:
yeah cheers mate, better implementation :beer: - go for it, I was a bit in a hurry when plugging these in, I think too there's room for improvement defo

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -176,13 +177,31 @@ def _datetime_to_int_days(cube):
...
+        if overlap:
+            real_date = datetime(date_obj.year, date_obj.month, 1, 0, 0, 0)
+        else:
+            real_date = datetime(date_obj.year,
+                                 date_obj.month, 15, 0, 0, 0)
         real_dates.append(real_date)
 
+    # get the number of days starting from the reference unit
+    time_unit = cube.coord('time').units.name
+    time_offset = _get_time_offset(time_unit)
```
Peter9192, line 189: Hey @valeriupredoi , this is one of the points that really confuses me. What happens if different cubes have different time offsets? Don't you want the offset with respect to a common reference time? I'm thinking something like 

```Python
def _datetime_to_int_days(cube, overlap=False):
    """Return list of int(days) with respect to a common reference.

    Cubes may have different calendars. This function extracts the date
    information from the cube and re-constructs a default calendar using
    the datetime library.

    This doesn't work for daily or subdaily data, because different
    calendars may have different number of days in the year.

    The standard calendar used by datetime provides the to_ordinal()
    function which returns "int number of days since the 0001/01/01".
    """
    # Convert to a common time format
    years = [cell.point.year for cell in cube.coord('time').cells()]
    months = [cell.point.month for cell in cube.coord('time').cells()]
    if not 0 in np.diff(years):
        # yearly data
        standard_dates = [datetime(year, 7, 1) for year in years]
    elif not 0 in np.diff(months):
        # monthly data
        standard_dates = [datetime(year, month, 15) for year, month in zip(years, months)]
    else:
        raise ValueError("Multimodel only supports yearly or monthly data")

    return [dt.to_ordinal() for dt in standard_dates]
```

I can work this out further if you like?

valeriupredoi: yes that's exactly that, I use the number of days since the start of the data taken from units forcing a `STANDARD_CALENDAR` eg `days since 1950-01-01, calendar=STANDARD_CALENDAR` - this works fine if there is no variation in calendars since the units are all the same for all cubes after `cmor.fix` unified the units, but I suspect it's the cause of the shifts @LisaBock is seeing when calendars differ

valeriupredoi: btw isn't it `toordinal()`?

Peter9192: > btw isn't it `toordinal()`?

Right you are :beer:

Peter9192: > this works fine if there is no variation in calendars

So I'd rather just fix it, either to the ordinal or to 1950/01/01. Taking it from the cube seems unnecessary and error-prone.

valeriupredoi: I fail to see the error-prone nature if all the cubes have the same units as they should have. Also taking it from the cube assumes nothing, it's safer in my books; I believe the stuff that Lisa is seeing has to do with cmip3 data not being correctly fixed

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -196,7 +215,7 @@ def _get_overlap(cubes):
     """
     all_times = []
     for cube in cubes:
-        span = _datetime_to_int_days(cube)
+        span = _datetime_to_int_days(cube, overlap=True)
         start, stop = span[0], span[-1]
         all_times.append([start, stop])
     bounds = [range(b[0], b[-1] + 1) for b in all_times]
```
Peter9192, line 221: I think this can also be simplified:
```Python
    bounds = []
    for cube in cubes:
        span = _datetime_to_int_days(cube)
        start, stop = span[0], span[-1]
        bounds.append(range(start, stop+1))
```
But why even use the range function? Don't we just want the overlapping time points from span?
```Python
    spans = []
    for cube in cubes: 
        spans.append(_datetime_to_int_days(cube))    
    time_pts = reduce(np.intersect1d, spans)
```

esmvalcore/preprocessor/_multimodel.py
Peter9192: > I fail to see the error-prone nature if all the cubes ...

It's in the "if", right?

> taking it from the cube assumes nothing

Yes, it assumes the calendars offsets have been properly fixed.

> it's safer in my books

I don't agree. You don't need the new time array to have anything to do with the old cubes, except for it to represent the same dates (which you get using the `cell.point.year` and `cell.point.month` methods . So why not just convert all of them to the same, easy to work with calendar?

esmvalcore/preprocessor/_multimodel.py
valeriupredoi: mate can we please not start trying to fix something that's not broken? We need to understand the very specific problem that @LisaBock is having and to allow for the correct handling of yearly data - the second one works fine so let's just focus on the scope of this PR. I really don't want to start changing this MM right before the release :beer:

esmvalcore/preprocessor/_multimodel.py
valeriupredoi: ah right, I only saw now that you are constructing the full datetime object - yes that should work, cheers! Can you pls commit and run the tests? It should pass them

esmvalcore/preprocessor/_multimodel.py
Peter9192: Sure, that's fair. It's just that I thought it might help to debug, for I'm still struggling to understand these parts of the code.

esmvalcore/preprocessor/_multimodel.py
Peter9192: Don't cheer too early, this doesn't pass the tests yet. It removes the overlap argument again, and also the `_get_time_offset` function. Then, the tests assume a different ordinal for the `expected` cubes, and also I think it returns a list of integers, where the old version returned `datetime.timedelta` objects. But I guess I can fix all this and make a commit, if you like.

valeriupredoi: the thing is - we shouldn't need this (apart from the nicer check on determining if it's yearly data, yes) - it all works fine as it is, unless there are issues with the actual ingested data in the MM as I'm suspecting it's Lisa's case. She is running CMIP3, we never had any MM issues with the MM for CMIP5/6

Peter9192: Maybe we don't... but you have to agree that this thing has been a pain to debug, right? So the cleaner we make it, the easier it should get. Let me invest a bit more time, and see if it solves Lisa's problem. I don't have other V2 deadlines ;-)

valeriupredoi: yeah not the best code to understand indeed! but it does do a lot of tricksy stuff nonetheless. Can I suggest we finish up with this PR in terms of functionality for the yearly (normal) data and merge then we look into Lisa's issue, the more I look at it the more it looks to be way separate :beer:

Peter9192: I agree we should respect the scope of the PR, but let's not merge to hastily. I'm not so keen on the overlap argument for the yearly data. But removing it in favor of the above suggestion breaks several tests. Can I look into fixing that first?

valeriupredoi: that would ve very much appreciated mate and will buy a :beer:

Peter9192:
Hey @valeriupredoi, I think I finally got to the bottom of it! 

Internally, we're always working with those integer days according to the standard calendar. But finally, we put them back into the `template_cube` like so (in `_put_in_cube`): 
```Python
    if t_axis is None:
        times = template_cube.coord('time')
    else:
        times = iris.coords.DimCoord(
            t_axis,
            standard_name='time',
            units=template_cube.coord('time').units)
```
Now, if `templace_cube` happens to have a different than standard calendar, this is where things get messy. For example, the CCSM4 calendar is "365 days". So in comparison to the standard calendar, every ~4 years, 1 day is miscounted. That leads to the shifted dates!

So I bet that changing the time units here to `Unit("days since 1950-01-01", calendar='standard') will fix the problem for good!

esmvalcore/preprocessor/_multimodel.py
Peter9192: So I fixed most of it, but some tests still fail because test cube 2 has a *daily* time coordinate, and the code above raises an exception in that case. What to do with that?

valeriupredoi:
excellent call @Peter9192 - I had completely forgotten about the cube assembly right at the very end. Note that forcing a fixed unit will shift the days if they are not from 1950, but this does the job perfectly:
```python
def _put_in_cube(template_cube, cube_data, statistic, t_axis):
    """Quick cube building and saving."""
    if t_axis is None:
        times = template_cube.coord('time')
    else:
        unit_name = template_cube.coord('time').units.name
        tunits = cf_units.Unit(unit_name, calendar="standard")
        times = iris.coords.DimCoord(
            t_axis,
            standard_name='time',
            units=tunits)
```
Cheers for spotting this! :beer:

esmvalcore/preprocessor/_multimodel.py
valeriupredoi: let's not make the changes - your suggestion below saved the day, so let's keep the changes to a minimum and only commit what's necessary - let me take it from here :beer:

Peter9192:
> Note that forcing a fixed unit will shift the days if they are not from 1950

That's why was fixing it at the other end :wink:

valeriupredoi:
ideally we should alter the data as little as possible, I reckon this will do it for now, could you test for your corner case pls? :beer:

valeriupredoi:
@Peter9192 we can put your suggested changes as per https://github.com/ESMValGroup/ESMValCore/pull/677#discussion_r442129625 in a new PR but as far as the yearly data and elimination of the wandering dates I am happy :beer:

valeriupredoi:
yeah was just saying in the comment from the issue - go for it :+1: :beer:

mattiarighi:
Can you fix the Codacy issue?

I will test and merge afterwards.

valeriupredoi:
cheers for testing and contributing @LisaBock and @Peter9192 - @mattiarighi that codacy issue can be ignored, if you could test when you haz time then this can go in then @Peter9192 can start on tuning the module more :beer:

bouweandela:
@valeriupredoi It would be great if you could fix the Codacy issue and help set a good example. If you think that we should change how we use Codacy I would be happy to discuss that over a beer at the next workshop :beers:  Anyway, it shouldn't be more than a minute or so it fix it.

valeriupredoi:
sorted @bouweandela :police_officer:

Peter9192:
> then @Peter9192 can start on tuning the module more

#685 work in progress

PR #678, opened 2020-06-17
Fix the syntax for adding multiple ensemble members from the same dataset

SarahAlidoost:

**Tasks**

* * *

### DIFF
--- a/doc/recipe/overview.rst
+++ b/doc/recipe/overview.rst
@@ -102,8 +102,8 @@ Here it is an example concatenating the `historical` experiment with `rcp85`
     datasets:
       - {dataset: CanESM2, project: CMIP5, exp: [historical, rcp85], ensemble: r1i1p1, start_year: 2001, end_year: 2004}
 
-It is also possible to define the ensemble as a list, although it is useful only
-case the two experiments have different ensemble names
+It is also possible to define the ensemble as a list when the two experiments have different ensemble names.
+In this case, the specified datasets are concatenated into a single cube:
 
 .. code-block:: yaml
 
@@ -129,9 +129,8 @@ for the ensemble members r1i1p1 to r5i1p1 and from r1i2p1 to r5i1p1 you can use:
       - {dataset: CanESM2, project: CMIP5, exp: historical, ensemble: "r(1:5)i(1:2)p1", start_year: 2001, end_year: 2004}
 
 Please, bear in mind that this syntax can only be used in the ensemble tag.
-Also, note that the combination of multiple experiments and ensembles has not been yet supported.
-For example running a recipe with exp: [historical, rcp85], ensemble: [r1i1p1, "r(2:3)i1p1"] may generate a dataset
-that contains only the r2i1p1 ensemble for rcp85, skipping the ensemble r3i1p1.
+Also, note that the combination of multiple experiments and ensembles, like
+exp: [historical, rcp85], ensemble: [r1i1p1, "r(2:3)i1p1"] has not been yet supported and will raise an error.
 
 Note that this section is not required, as datasets can also be provided in the
 Diagnostics_ section.
--- a/doc/recipe/overview.rst
+++ b/doc/recipe/overview.rst
@@ -38,7 +38,7 @@ the following:
     documentation:
       description: |
         Recipe to produce time series figures of the derived variable, the
-        Atlantic meriodinal overturning circulation (AMOC).
+        Atlantic meridional overturning circulation (AMOC).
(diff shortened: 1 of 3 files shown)


doc/recipe/overview.rst
```diff
@@ -113,22 +113,25 @@ case the two experiments have different ensemble names
...
+      - {dataset: CanESM2, project: CMIP5, exp: historical, ensemble: "r(1:10)i1p1", start_year: 2001, end_year: 2004}
 
 It can be included multiple times in one definition. For example, to generate the datasets definitions
 for the ensemble members r1i1p1 to r5i1p1 and from r1i2p1 to r5i1p1 you can use:
 
 .. code-block:: yaml
 
     datasets:
-      - {dataset: CanESM2, project: CMIP5, exp: historical, ensemble: r(1:5)i(1:2)p1, start_year: 2001, end_year: 2004}
+      - {dataset: CanESM2, project: CMIP5, exp: historical, ensemble: "r(1:5)i(1:2)p1", start_year: 2001, end_year: 2004}
```
bouweandela, line 129: This doesn't seem to work for me, have you tried this?

doc/recipe/overview.rst
```diff
@@ -113,22 +113,25 @@ case the two experiments have different ensemble names
...
 
 Please, bear in mind that this syntax can only be used in the ensemble tag.
+Also, note that the combination of multiple experiments and ensembles has not been yet supported.
+For example running a recipe with exp: [historical, rcp85], ensemble: [r1i1p1, "r(2:3)i1p1"] may generate a dataset
+that contains only the r2i1p1 ensemble for rcp85, skipping the ensemble r3i1p1.
```
bouweandela, line 134: The syntax with using a list concatenate the data from all the specified datasets into a single cube, so I don't think it makes sense to use it like this. This was implemented for cases where the names of the ensemble members were different for two `exp`s that you want to have concatenated, see the description here: https://github.com/ESMValGroup/ESMValCore/pull/207

doc/recipe/overview.rst
SarahAlidoost: @bouweandela thanks. It seems that neither r(1:5)i(1:2)p1 nor  “r(1:5)i(1:2)p1” works. I am wondering what is the use case for this syntax because it had been introduced here.

doc/recipe/overview.rst
SarahAlidoost: I think users can benefit from this additional information because currently, it is possible to make this combination. The esmvaltool doesn’t raise an error in this case and the recipe run is successful. However, users should know that the results are not reliable.

doc/recipe/overview.rst
bouweandela: Ok, I pushed some changes to this branch to make it work. I guess @jvegasbsc intended to implement this, but forgot to do it.

doc/recipe/overview.rst
bouweandela: I've pushed some changes to this branch so this will raise a RecipeError. Could you try if it works for you? Do you have a suggestion for how the documentation about this could be made more clear?

SarahAlidoost: @bouweandela thanks. Good, now it raises an error for that combination. I updated the documentation, accordingly.

bouweandela: Thanks!

doc/recipe/overview.rst
```diff
@@ -113,22 +113,24 @@ case the two experiments have different ensemble names
...
 
 .. code-block:: yaml
 
     datasets:
-      - {dataset: CanESM2, project: CMIP5, exp: historical, ensemble: r(1:5)i(1:2)p1, start_year: 2001, end_year: 2004}
+      - {dataset: CanESM2, project: CMIP5, exp: historical, ensemble: "r(1:5)i(1:2)p1", start_year: 2001, end_year: 2004}
 
 Please, bear in mind that this syntax can only be used in the ensemble tag.
+Also, note that the combination of multiple experiments and ensembles, like
+exp: [historical, rcp85], ensemble: [r1i1p1, "r(2:3)i1p1"] has not been yet supported and will raise an error.
```
bouweandela, line 133:
```suggestion
exp: [historical, rcp85], ensemble: [r1i1p1, "r(2:3)i1p1"] is not supported and will raise an error.
```
I would remove 'yet', because there are no plans to implement support for this in the future.

bouweandela:
Thanks @SarahAlidoost !

PR #679, opened 2020-06-17
Add percentiles to multi-model stats

### LINKED ISSUE #563: multi model statistics -- more statistics
@jvegasbsc How about adding standard deviation, min and max to the pool of available multi model statistics? It would be rather easy to implement. 
It would be nice to also have np.percentile in the mix, though this would require reforming the masked values to nans (numpy.nanpercentile), so idk if that's userful.

Peter9192:
Add the option to calculate percentiles in the multi-model preprocessor, using [`scipy.stats.mstats.mquantiles`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.mstats.mquantiles.html#scipy.stats.mstats.mquantiles). This is the same function as [used in Iris](https://scitools.org.uk/iris/docs/v1.9.0/html/iris/iris/analysis.html#iris.analysis.PERCENTILE).

It works by adding e.g. `p10` or `p99` to the list of operators in the recipe. `p99.5` will be interpreted as the 99.5 percentile, i.e. the 0.995 quantile, but the decimal point will be changed to a dash `-` in the output filename.

Unfortunately, the scipy function doesn't work directly on multidimensional arrays. Thus, I added a wrapper that flattens the extra dimensions before passing the array to scipy, similar to [what's done in Iris](https://github.com/SciTools/iris/pull/625).

**Tasks**

-   [x] Add unit tests

* * *

Closes #563


esmvalcore/preprocessor/_multimodel.py
```diff
@@ -68,6 +95,9 @@ def _compute_statistic(data, statistic_name):
         statistic_function = np.ma.max
     elif statistic_name == 'min':
         statistic_function = np.ma.min
+    elif statistic_name.startswith('p'):
```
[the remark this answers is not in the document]
Peter9192: Hmm.. I see the point, but this solution doesn't work. If you except it and pass, it will never reach the final `else`, and you will get `UnboundLocalError: local variable 'statistic_function' referenced before assignment`. So we should raise rather than pass. You could catch the `ValueError` and raise `NotImplementedError`, but then why not just keep the `ValueError`? That error message also provides a better pointer as to what's wrong. I'd suggest to just leave it for now, and sort it out once we get this statistic that starts with p.

doc/recipe/preprocessor.rst
```diff
@@ -647,8 +647,10 @@ to observational data, these biases have a significantly lower statistical
 impact when using a multi-model ensemble. ESMValTool has the capability of
 computing a number of multi-model statistical measures: using the preprocessor
 module ``multi_model_statistics`` will enable the user to ask for either a
-multi-model ``mean``, ``median``, ``max``, ``min`` and / or ``std`` with a set
-of argument parameters passed to ``multi_model_statistics``.
+multi-model ``mean``, ``median``, ``max``, ``min``, ``std``, and / or ``pXX``
+with a set of argument parameters passed to ``multi_model_statistics``.
+Percentiles can be specified like ``p10`` or ``p90``. ``p995`` will be
+interpreted as the 99.5-th percentile.
```
bouweandela, line 653: It would be better to include the period in the percentile, what if I want the 1.5th percentile? It will get interpreted as 15th with this notation. In the resulting filename you could then replace the '.' by a '-'.

esmvalcore/preprocessor/_multimodel.py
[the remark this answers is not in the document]
bouweandela: Maybe you could do something like this to make it more selective:
```suggestion
    elif re.match("^p\d+[\.]?\d*$", statistic):
```

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -328,13 +358,14 @@ def multi_model_statistics(products, span, statistics, output_products=None):
         computation;
         cube attribute of product is the data cube for computing the stats.
     span: str
-        overlap or full; if overlap stas are computed on common time-span;
-        if full stats are computed on full time spans.
+        overlap or full; if overlap stats are computed on common time-span;
```
bouweandela, line 361:
```suggestion
        overlap or full; if 'overlap' statistics are computed on common time-span;
```

esmvalcore/preprocessor/_multimodel.py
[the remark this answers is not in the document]
bouweandela: It would also be good to add a check that the values are valid in esmvalcore/_recipe_checks.py or in esmvalcore/recipe_schema.yml, so users don't spend ages running a recipe to regrid a 100 models only to find that they typed the wrong name for the statistic they would like to compute.

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -53,6 +54,32 @@ def _plev_fix(dataset, pl_idx):
...
+    """
+    # Ensure that the target axis is the last dimension.
+    data = np.rollaxis(data, axis, start=data.ndim)
+    shape = data.shape[:-1]
+    # Flatten any leading dimensions.
+    if shape:
+        data = data.reshape([np.prod(shape), data.shape[-1]])
+    # Perform the quantile calculation.
+    result = mstats.mquantiles(data, quantile, axis=-1, alphap=1, betap=1)
+    result = np.ma.MaskedArray(result)
```
bouweandela, line 72: According to the documentation [`scipy.stats.mstats.mquantiles`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.mstats.mquantiles.html#scipy.stats.mstats.mquantiles) returns a masked array, why does it need to be converted into another one?
```suggestion
```

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -53,6 +54,32 @@ def _plev_fix(dataset, pl_idx):
...
+    https://github.com/SciTools/iris/pull/625
+    """
+    # Ensure that the target axis is the last dimension.
+    data = np.rollaxis(data, axis, start=data.ndim)
+    shape = data.shape[:-1]
+    # Flatten any leading dimensions.
+    if shape:
+        data = data.reshape([np.prod(shape), data.shape[-1]])
+    # Perform the quantile calculation.
+    result = mstats.mquantiles(data, quantile, axis=-1, alphap=1, betap=1)
```
bouweandela, line 71:
```suggestion
    result = scipy.stats.mstats.mquantiles(data, quantile, axis=-1, alphap=1, betap=1)
```
would be a bit easier to read, because then you can see what function it is without scrolling to the top of the file to look at the import statements

esmvalcore/preprocessor/_multimodel.py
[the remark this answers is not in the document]
Peter9192: Okay, what about the following: `^(p\d{1,2})(\.\d*)?$`: starts with `p` followed by 1 or 2 digits, which may or may not be followed by a decimal part consisting of a period followed by 0 to &infin;  digits. Works for most cases like `p1`, `p1.`, `p1.1`, `p1.11`, `p11`, ..., and also for `p0`, but not for `p.1`, `p.11`, or `p100`.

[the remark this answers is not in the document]
bouweandela: You could also just accept anything starting with `p` and then try converting whatever you get to a valid percentile and raise a nice error message if that fails, it's the last option to try anyway.

doc/recipe/preprocessor.rst
Peter9192: I like the idea, so I tried to change `statistic_product.filename`, but it appears it is not possible to overwrite the filename of a `PreprocessorFile`. So how should I go about changing this?

bouweandela: Change this line
https://github.com/ESMValGroup/ESMValCore/blob/c5ffda0c66292a0a7d594b2e4b6e3478cde00075/esmvalcore/_recipe.py#L632-L633
to
```python
        attributes['dataset'] = attributes['alias'] = 'MultiModel{}'.format(
            statistic.title().replace('.', '_')
```

Peter9192:
Thanks for the suggestions @bouweandela! I think I got them all

PR #685, opened 2020-06-24
Simplify time handling in multi-model statistics preprocessor

Peter9192:
This PR simplifies the way time is handled in the multimodel statistics preprocessor. This should make it easier to understand what the code does, and hence to debug it. 

These changes are inspired by some recent unexpected behavior when cubes with different calendars were passed to the function (#665). The first important change, therefore, is to fix the calendar for the output cubes to a standard format. Simultaneously, I added a function that makes all input cubes also use this same standard calendar. 

Having consistent time arrays make time indexing much easier. Previously, complex mappings were needed to align the indexes representing the same time points in different cubes. Now, we can just get any time from any cube, without worrying too much about the index. Furthermore, I aligned the code between 'assemble overlap data' and 'assemble full data' to the extent that they could be merged into a single function. 

**Tasks**

-   [x] Create an issue: discussed [here](https://github.com/ESMValGroup/ESMValCore/pull/677#issuecomment-646095151)
-   [x] Add unit tests
-   [x] ~If writing a new/modified preprocessor function, please update the [documentation](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/preprocessor.html)~ Not necessary


Peter9192:
@valeriupredoi whenever you have time (I'm not in a rush)

valeriupredoi:
hey @Peter9192 great stuff mate! I am sorry I took a while to review, been on leave until today. A few questions:

- you are now forcing time points with units of `days since 1850-01-01...` from eg `new_times = reduce(np.intersect1d, time_spans)` for overlap; `_set_common_calendar()` is taking care of aligning all cubes time points with correct units (days from 1850...) and standard calendar - this is great! But I would rename the function to something like `_regrid_time_points_on_common_axis()` or something like that since the func does much more than setting a common calendar; 
- the reason why I wanted to use the actual `regrid_time()` from `_time` was that that one was doing a bit more than resetting time points - setting time bounds (which the current output of MM will not have) and correcting for the aux coords too (which shouldn't be a problem here I'm guessing); if you just add a guess bounds call I think it should be fine.

Anyways two small nicks, it's a very nice optimization you did imho! :beer:

valeriupredoi:
one more thing - can you pls add the info in the documentation that describes the time regridding within the module?

Peter9192:
Thanks @valeriupredoi !

> * the reason why I wanted to use the actual `regrid_time()`

Yeah, I wanted to use that as well. But then it got messy, since regrid_time doesn't change the calendar, so then we'd be modifying the calendar here and the actual time points there (and it's just 2 lines...). At first I copied all the stuff about aux-coords as well, but then I noticed this never makes it to `_put_in_cube()` anyway, so I left it out again. I reinstated the `guess_bounds` :beer:

valeriupredoi:
cool! ready to go, @mattiarighi could you give it a test please mate :beer:

valeriupredoi:
@Peter9192 could you pls fix the conflict from the new `master` mate, as much as I like the new changes I would feel comfortable to merge only after @mattiarighi has tested with his magic perfmetrics standards :beer:

mattiarighi:
> merge only after @mattiarighi has tested with his magic perfmetrics standards

Doing this right now... :hourglass:

mattiarighi:
It looks good to me. :+1:

Peter9192:
Please wait with merging, I think I can make it (much) faster. Testing now.

Peter9192:
> Please wait with merging, I think I can make it (much) faster. Testing now.

Okay never mind, I thought I had a quick fix. 

But I'd like to further discuss the efficiency of this function, for I noticed that for certain applications, the new functionality is a bit slower than before. Most of the time is taken by this line:
https://github.com/ESMValGroup/ESMValCore/blob/d3a4dddb1259fafa57267a69cdede9fb49a04cc8/esmvalcore/preprocessor/_multimodel.py#L238
In my use case this is where the cubes' data get realized (139 times, once for each time point). 

The existing implementation does this:
https://github.com/ESMValGroup/ESMValCore/blob/50525bbc923efca2e07ff24741a20418bc7fc2b6/esmvalcore/preprocessor/_multimodel.py#L286
i.e. it copies over the entire cubes' data and then (few lines later) indexes only a single time. 

The reason why the old implementation is faster in my use case, is that all the data is realized at once, rather than time by time. Indeed, adding 
```python
# Realize all cubes at once instead of time by time
[cube.data for cube in cubes]
```
before making the time slices significantly boosts the performance of the new implementation.

Now, my question is: do we want to intentionally realize all cubes in the new implementation before making the time slices?

Some thoughts:
- If it fits in memory, realizing all data at once is faster; but you'll lose the ability to perform out-of-core computations.
- The old function effectively also did this, so it couldn't have coped with out-of-core things anyway.
- Most of the applications will already have realized data anyway, since they need regridding before, which is not (yet) lazy.

My take would be: realize them for now. Once regridding becomes lazy we can think of a next optimization round where we can then think of a way to keep on working with cubes instead of numpy masked arrays, so that we can also use iris' statistics functions rather than our own.

valeriupredoi:
@Peter9192 can you please check the last commit (codacy seems to think there is a no effect statement). I would highly recommend we merge this now since @mattiarighi has tested and all is fine and concern ourselves with new implementations and optimiziations in a further separate PR :beer:

mattiarighi:
Do I need to test again?
There were two more commits after I tested.

Peter9192:
Let's not rush, the core v2 deadline has already expired anyway. I'm curious what @bouweandela thinks about whether or not to realize the cubes' data early on (those last 2 commits). If we proceed with that, it would be good to test again.

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -201,158 +185,88 @@ def _put_in_cube(template_cube, cube_data, statistic, t_axis):
...
+    Make sure all cubes' use the same standard calendar and time units.
 
+    Cubes may have different calendars. This function extracts the date
+    information from the cube and re-constructs a default calendar,
+    resetting the actual dates to the 15th of the month or 1st of july for
+    yearly data (consistent with `regrid_time`), so that there are no
+    mismatches in the time arrays.
 
-def _slice_cube(cube, t_1, t_2):
+    Doesn't work for (sub)daily data, because different calendars may have
```
bouweandela, line 198: This could work just fine if all calendars are the same already, in that case there would be no need to change the calendar or time points

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -201,158 +185,88 @@ def _put_in_cube(template_cube, cube_data, statistic, t_axis):
...
-    stats_cube = _put_in_cube(cubes[0][sl_1:sl_2 + 1],
-                              stats_dats,
-                              statistic,
-                              t_axis=None)
-    return stats_cube
+        # Update the cubes' time coordinate (both point values and the units!)
+        cube.coord('time').points = [t_unit.date2num(date) for date in dates]
+        cube.coord('time').units = t_unit
+        cube.coord('time').bounds = None
+        cube.coord('time').guess_bounds()
```
bouweandela, line 228: This doesn't work if there is only 1 time point. Related question: why is a time coordinate needed to compute a multimodel statistic?

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -201,158 +185,88 @@ def _put_in_cube(template_cube, cube_data, statistic, t_axis):
...
-        # hold time slices only
-        new_datas_array = _full_time_slice(cubes, empty_arr, indices_list,
-                                           new_arr, i)
-        # list to hold time slices
-        time_data = []
-        for j in range(len(cubes)):
-            time_data.append(new_datas_array[j])
-        stats_dats[i] = _compute_statistic(time_data, statistic)
-    stats_cube = _put_in_cube(cubes[0], stats_dats, statistic, time_axis)
+    template = cubes[0][:n_times]
```
bouweandela, line 268: why slice cube by time?

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -429,10 +349,7 @@ def multi_model_statistics(products, span, statistics, output_products=None):
 
     for statistic in statistics:
         # Compute statistic
-        if span == 'overlap':
-            statistic_cube = _assemble_overlap_data(cubes, interval, statistic)
-        elif span == 'full':
-            statistic_cube = _assemble_full_data(cubes, statistic)
+        statistic_cube = _assemble_data(cubes, statistic, span)
         statistic_cube.data = np.ma.array(statistic_cube.data,
```
bouweandela, line 353: This is not new in the pull request, but why is the array converted (again) to a masked array here? It would also be much more memory efficient to ensure that the data type is correct during the calculation already.

bouweandela:
> I'm curious what @bouweandela thinks about whether or not to realize the cubes' data early on (those last 2 commits). 

That would be fine, it's will keep the current behaviour and indeed realizing the data for a single time slice is likely to be inefficient.

bouweandela:
Question: when we spoke about this earlier, I thought you were planning to replace the entire statistic calculation part with using iris? i.e. add a model aux coord to each input cube, merge all input cubes into a single cube and then collapse it along the model dimension using an operator from the iris.analysis?

valeriupredoi:
as I said above - I think this is a good PR and should be merged at the level of code changes before @mattiarighi tested ([comment](https://github.com/ESMValGroup/ESMValCore/pull/685#issuecomment-652520617)) and then I would carry on overhauling in separate PR(s); I reckon piecemeal PR's are much better in cases where both the importance of the module and the code change volume are high :beer:

esmvalcore/preprocessor/_multimodel.py
Peter9192: The way the current function works is that a new cube is constructed by copying one of the input cubes and filling it with new data. This slice ensures that the length of the time dimension of the output cube corresponds with the length of the time overlap of all cubes.

Peter9192:
> Question: when we spoke about this earlier, I thought you were planning to replace the entire statistic calculation part with using iris? i.e. add a model aux coord to each input cube, merge all input cubes into a single cube and then collapse it along the model dimension using an operator from the iris.analysis?

I still think this should be the end goal, but that would be a huge leap from where the code was at. The changes in this PR are the first step in that direction. I'd be happy to continue with this, but I agree with @valeriupredoi that it might be better to take it step by step. The biggest hurdle is that you need to make sure that the dimensions of all input cubes are the same before you can merge them into a single cube.

> at the level of code changes before @mattiarighi tested

I really would like to keep the last commits as well, otherwise the PR might deteriorate performance for use cases like my own. It's a small change, but it would be good to test it again. I can also still address @bouweandela 's comments in this one.

esmvalcore/preprocessor/_multimodel.py
valeriupredoi: what do you mean why is a t-coord needed for a multimodel stat? :grin: you compute a time-dependent statistic `S(t) = \sum_i^N S_i(t) / N`

bouweandela: Created an issue for this: #738

esmvalcore/preprocessor/_multimodel.py
Peter9192: Fair point. I'll print a warning instead and attempt to continue.

esmvalcore/preprocessor/_multimodel.py
Peter9192: Let's solve this in a separate PR.

Peter9192:
I've addressed @bouweandela's comments. More work can be done in other PRs, but I'd like to keep the last few commits. @mattiarighi could you do another round of testing please (when you have time)?

esmvalcore/preprocessor/_multimodel.py
bouweandela: That's not quite what happens. Information from the template cube is used to construct a new cube, but its time coordinate is not used.

esmvalcore/preprocessor/_multimodel.py
bouweandela: It would be better not to change the calendars if no change is needed.

esmvalcore/preprocessor/_multimodel.py
Peter9192: I changed that in my last commit.

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -429,12 +349,9 @@ def multi_model_statistics(products, span, statistics, output_products=None):
...
         # Compute statistic
-        if span == 'overlap':
-            statistic_cube = _assemble_overlap_data(cubes, interval, statistic)
-        elif span == 'full':
-            statistic_cube = _assemble_full_data(cubes, statistic)
-        statistic_cube.data = np.ma.array(statistic_cube.data,
-                                          dtype=np.dtype('float32'))
+        statistic_cube = _assemble_data(cubes, statistic, span)
+        # statistic_cube.data = np.ma.array(statistic_cube.data,
+        #                                   dtype=np.dtype('float32'))
```
bouweandela, line 354:
```suggestion
```
Remove commented out stuff please.

esmvalcore/preprocessor/_multimodel.py
Peter9192: No, but the length of the time coordinate is what's important here. The actual points will be overwritten by the new time coordinate, which is the same length as `n_times`.

Peter9192: BTW, this is not new in this PR, so maybe we could improve it in a separate PR.

esmvalcore/preprocessor/_multimodel.py
Peter9192: Okay, I now made it such that it will first check if the calendars are consistent. If they are consistent, they will not be changed. Otherwise, they will be changed to a default calendar.

bouweandela: Much better, because now the function can hopefully be used to compute e.g. ensemble statistics without unexpected results. Maybe we should actually move the whole changing the time coordinate to the `regrid_time` preprocessor, that would be more consistent with what we do for the horizontal and vertical coordinates and would make the process more visible to users of the recipe and the implementation of the multi model statistics function much easier. I've created this issue  #744.

esmvalcore/preprocessor/_multimodel.py
bouweandela: The length of the time coordinate does not matter, just try with `template = cubes[0][:1]`.

Peter9192: You're right, my bad! Fixed now.

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -201,158 +185,102 @@ def _put_in_cube(template_cube, cube_data, statistic, t_axis):
...
-        if np.ma.is_masked(cube.data):
-            ndat.mask[indices[idx_cube]] = cube.data.mask
-        else:
-            ndat.mask[indices[idx_cube]] = False
-        ndatarr[idx_cube] = ndat[t_idx]
-
-    # return time slice
-    return ndatarr
-
+    t_unit = _get_consistent_time_unit(cubes)
```
bouweandela, line 211: It would be computationally more efficient to just return `cubes` at this point if no change to the calendars is needed.

bouweandela: And also more reliable, because if no change is needed, e.g. resetting the bounds and then guessing them might unexpectedly change them

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -201,158 +185,102 @@ def _put_in_cube(template_cube, cube_data, statistic, t_axis):
...
-    return stats_cube
-
-
-def _assemble_full_data(cubes, statistic):
-    """Get statistical data in iris cubes for FULL."""
-    # all times, new MONTHLY data time axis
-    time_axis = [float(fl) for fl in _monthly_t(cubes)]
+    for cube in cubes:
+        # Extract date info from cube
+        years = [cell.point.year for cell in cube.coord('time').cells()]
```
bouweandela, line 215:
```suggestion
        coord = cube.coord('time')
        years = [p.year for p in coord.units.num2date(coord.points)]
```
using the function [`cftime.num2date`](https://unidata.github.io/cftime/api.html#cftime.num2date) on a numpy array appears to be a lot faster than using the `cells()` method of the coordinate to iterate over the cells, also in other places

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -201,158 +185,102 @@ def _put_in_cube(template_cube, cube_data, statistic, t_axis):
...
+            # (sub)daily data
+            if cube.coord('time').units != t_unit:
+                logger.warning(
+                    "Multimodel encountered (sub)daily data and inconsistent "
+                    "time units or calendars. Attempting to continue, but "
+                    "might produce unexpected results.")
+            dates = [cell.point for cell in cube.coord('time').cells()]
+
+        # Update the cubes' time coordinate (both point values and the units!)
+        cube.coord('time').points = [t_unit.date2num(date) for date in dates]
```
bouweandela, line 239:
```suggestion
        cube.coord('time').points = t_unit.date2num(dates)
```
note that this function accepts an array of dates

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -201,158 +185,102 @@ def _put_in_cube(template_cube, cube_data, statistic, t_axis):
...
+
+
+def _get_time_slice(cubes, time):
+    """Fill time slice array with cubes' data if time in cube, else mask."""
+    time_slice = []
+    for cube in cubes:
+        cube_time = cube.coord('time').points
+        if time in cube_time:
+            idx = int(np.argwhere(cube_time == time))
+            subset = cube[idx].data
```
bouweandela, line 252:
```suggestion
            subset = cube.data[idx]
```
would be faster, because slicing a numpy array is less work than slicing an entire cube and you're only using the numpy array anyway

esmvalcore/preprocessor/_multimodel.py
Peter9192: That won't work, because the function will still regrid time for monthly and daily data, even if the calendars are the same. But we can do this after or as part of #744.

bouweandela: Ok, let's leave this for later.

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -201,158 +185,104 @@ def _put_in_cube(template_cube, cube_data, statistic, t_axis):
...
-    # data array for stats computation
-    stats_dats = np.ma.zeros(new_shape)
+    # Target array to populate with computed statistics
+    new_shape = [n_times] + list(cubes[0].shape[1:])
+    stats_data = np.ma.zeros(new_shape, dtype=np.dtype('float32'))
 
-    # assemble indices list to chop new_arr on
-    indices_list = []
+    # Realize all cubes at once instead of separately for each time slice
+    _ = [cube.data for cube in cubes]
```
bouweandela, line 277:
```suggestion
```
This is no longer needed, now that the cubes are realized in `_get_time_slice`

bouweandela: Anyway, let's leave it for later, it doesn't hurt either

Peter9192:
@bouweandela @mattiarighi how can we proceed with testing/merging this?

bouweandela:
We would need to run the current recipes that use the multi model statistics preprocessor function and check that that the results are unchanged.

Peter9192:
> Someone would need to run the current recipes that use the multi model statistics preprocessor function and check that that the results are unchanged.

@axel-lauer would you know who could do this?

Peter9192:
> Someone would need to run the current recipes that use the multi model statistics preprocessor function and check that that the results are unchanged.

That would be (@nielsdrost):

- [ ] esmvaltool/recipes/recipe_perfmetrics_CMIP5.yml
- [ ] esmvaltool/recipes/examples/recipe_python.yml
- [ ] esmvaltool/recipes/examples/recipe_python_object_oriented.yml
- [ ] esmvaltool/recipes/examples/recipe_preprocessor_test.yml
- [ ] esmvaltool/recipes/recipe_clouds_bias.yml
- [ ] esmvaltool/recipes/recipe_lauer13jclim.yml
- [ ] esmvaltool/recipes/recipe_smpi_4cds.yml
- [ ] esmvaltool/recipes/recipe_perfmetrics_CMIP5_4cds.yml
- [ ] esmvaltool/recipes/recipe_ocean_Landschuetzer2016.yml
- [ ] esmvaltool/recipes/recipe_flato13ipcc.yml
- [ ] esmvaltool/recipes/recipe_ocean_example.yml
- [ ] esmvaltool/recipes/recipe_ocean_bgc.yml
- [ ] esmvaltool/recipes/recipe_carvalhais14nat.yml
- [ ] esmvaltool/recipes/recipe_clouds_ipcc.yml
- [ ] esmvaltool/recipes/recipe_smpi.yml
- [ ] esmvaltool/recipes/recipe_perfmetrics_land_CMIP5.yml

valeriupredoi:
I hear some distant yodeling calling @mattiarighi 's name :grin:

axel-lauer:
I tested calculation of the **multi-model mean** with recipe_clouds_ipcc.yml. There are some small differences between this version and the preprocessor in the master branch in the order of 1e-8*(original value). These differences to not seem to average out over time. For recipe_clouds_ipcc.yml, such differences are not important, but it would be good to know where such differences are coming from. Here is an example for the differences in the multi-model mean of the derived variable "swcre" (time average over 20 model years, the range of swcre is about -150...0 W m-2):

swcre

mattiarighi:
I think the 1.e-8 difference is due to rounding errors, since the order in which the different dataset are averaged out is different in each run. If you want to get binary identical results, you should compare the multi-model median, which is not affected by this problem (the median is an "actual" value, whiel the mean results from an arithmetic calculation).

In the 1000 tests I've been running during the development of v2.0, I've always had this difference in the multi-model mean, but not in the median.

Hope this helps.

valeriupredoi:
oh but I think you are wrong @mattiarighi - you ran way more than 1000 tests :grin: @axel-lauer cheers for testing! Are those 1e-8 diffs in the prepocc-ed multimodel file (as outputted by the preprocessor) or in the final result of the diagnostic? :beer:

bouweandela:
Here is a thought: the fact that we think we need to run recipes that use this preprocessor to test this pull request shows that we have no confidence in our unit tests. Maybe we should think about improving those, so we do not have to manually test the preprocessor function again and again, with work on it already planned and happening in #673, #744, and #781.

valeriupredoi:
agreed with @bouweandela - we really need actual data to have a set of actual functional tests :+1:

bouweandela:
Would the CMIP6 license permit using some CMIP6 data?

Peter9192:
I think it's a good idea but it might require some more discussion. Should we create a new issue for that?

bouweandela:
> Should we create a new issue for that?

Yes, that's a good idea. However, I would prefer spending the time on fixing the automated testing rather than spending more time on manually testing this pull request.

Peter9192:
I agree @bouweandela that's a more efficient use of our precious time. But it will also take longer. Do you want to wait with merging this? I think it's been reviewed quite thoroughly and already partly tested. The only thing that recently came up is [this remark](https://github.com/ESMValGroup/ESMValCore/pull/673#issuecomment-700045740) by @sloosvel.
I would propose to address this (it's only a minor change) and then merge, otherwise it keeps blocking our progress.

stefsmeets:
Just out of curiousity, I merged master on this branch and ran `pytest -m "use_sample_data"` to see what would happen. It seems like there are a few regressions introduced with this PR.

```
test_multimodel.py::test_multimodel_regression_month[overlap] FAILED
test_multimodel.py::test_multimodel_regression_month[full] FAILED
test_multimodel.py::test_multimodel_regression_day[overlap-360_day] SKIPPED
test_multimodel.py::test_multimodel_regression_day[overlap-365_day] FAILED
test_multimodel.py::test_multimodel_regression_day[overlap-gregorian] FAILED
test_multimodel.py::test_multimodel_regression_day[overlap-proleptic_gregorian] FAILED
test_multimodel.py::test_multimodel_regression_day[overlap-julian] SKIPPED
test_multimodel.py::test_multimodel_regression_day[full-360_day] SKIPPED
test_multimodel.py::test_multimodel_regression_day[full-365_day] FAILED
test_multimodel.py::test_multimodel_regression_day[full-gregorian] FAILED
test_multimodel.py::test_multimodel_regression_day[full-proleptic_gregorian] FAILED
test_multimodel.py::test_multimodel_regression_day[full-julian] SKIPPED
test_multimodel.py::test_multimodel_no_vertical_dimension PASSED
test_multimodel.py::test_multimodel_no_horizontal_dimension XFAIL
test_multimodel.py::test_multimodel_only_time_dimension PASSED
test_multimodel.py::test_multimodel_no_time_dimension XFAIL
```

Peter9192:
Yes, I get the same errors. Good that you got the new tests to work! Now let's see what they say...

Peter9192:
So my first guess is that most of the errors are due to this: 

https://github.com/ESMValGroup/ESMValCore/blob/64bc2fe10f9abb77ddbff4f44b410e8e4f8ea171/esmvalcore/preprocessor/_multimodel.py#L207-L208

which is intentional, as reflected by the top post:

> These changes are inspired by some recent unexpected behavior when cubes with different calendars were passed to the function (#665). The first important change, therefore, is to fix the calendar for the output cubes to a standard format. Simultaneously, I added a function that makes all input cubes also use this same standard calendar.

From your test-develop perspective @stefsmeets, can you confirm that the test is supposed to fail for this change? If so, we can discuss whether this is desirable, or whether we should change the test.

Peter9192:
Okay I had a first look at the monthly data tests only. These are my findings:

- [ ] FAILED tests/sample_data/multimodel_statistics/test_multimodel.py::test_multimodel_regression_month[full]
  - Issue 1: Data are different. It is accurate to 4 digits, but not to 6. Could be similar to what's [discussed above](https://github.com/ESMValGroup/ESMValCore/pull/685#issuecomment-698253145). Changing the statistic from 'mean' to 'median' as suggested by Mattia solves the issue on the data.
  - Issue 2: Coordinates are different. This is largely on purpose:
    - This PR adds bounds (as discussed [here](https://github.com/ESMValGroup/ESMValCore/pull/673#issuecomment-700089313)) 
    - This PR sets the time to the 15th of each month instead of the first, consistent with `regrid_time`. 
    - The new cube is missing the 'var_name' attribute in the time coord though, I'll add this back.
- [ ] FAILED tests/sample_data/multimodel_statistics/test_multimodel.py::test_multimodel_regression_month[overlap]
  - Issue here is coords being different:
    - Different calendars (this PR resets all calendars to standard, old uses the first cube it encounters)
    - Different time points (this PR sets everything to the 15th, old ended up every 16th on 00 or 12 utc)
    - Different bounds (this should be addressed in https://github.com/ESMValGroup/ESMValCore/issues/803)
    - Lazy data (this PR has no chuncksizes attribute, old code did). This is ok.

Peter9192:
Differences in time coordinates in **daily** regression tests:

- Time points: daily reference data for the regression tests have time points at 12:00, whereas this PR now resets them to 00:00, consistent with `_regrid_time`. I propose to keep the current behaviour and update the reference cubes. Address this as part of #744 if another decision is preferred.
- Time attributus: daily reference data for the regression tests sometimes have legacy attributes like `_ChunkSizes` and `time_origin`. This PR doesn't reproduce that as the time coordinate is reconstructed from the ground up in all cases, instead of sometimes using one of the input cubes' time coord. I propose to keep the current behaviour and update the reference cubes. 
- Long name on time coordinate: in the reference it is sometimes "time", sometimes "Time axis". New behaviour: "time". I propose to update the reference cubes. 

Then there are some differences in the actual data (only for `span=full`) that I still need to look into.

Peter9192:
> Then there are some differences in the actual data (only for `span=full`) that I still need to look into.

This is due to a bug in the current implementation (on which the reference data was based). I made an issue #937. Reference data should be updated.

Peter9192:
>     * Issue 1: Data are different. It is accurate to 4 digits, but not to 6. Could be similar to what's [discussed above](https://github.com/ESMValGroup/ESMValCore/pull/685#issuecomment-698253145). Changing the statistic from 'mean' to 'median' as suggested by Mattia solves the issue on the data.

I also found out where this difference comes from: in the current master branch, the data is passed over from the original cubes into a new array here: https://github.com/ESMValGroup/ESMValCore/blob/4d52e4ac50483a98e946353a630427831da44f3a/esmvalcore/preprocessor/_multimodel.py#L331

The incoming data is `float32`, whereas the new data type is `float64`. Consequently, the statistics are calculated on the `float64` data, introducing an unrealistically high precision. Later on the dtype is converted again to float32, but at that point, the damage is already done. 

I've created a new PR to correct this change and to update the reference data. That way we can confirm that this PR doesn't break the preprocessor.

Peter9192:
Okay, so I think I've found all the quirks that surfaced after the introduction of the new regression tests. In summary:

- Data are different due to #937 and #940 
- Time points (and bounds) are different because this PR makes mmstats consistent with regrid time
- Calendars are different, this is one of the main intentions of this PR
- Some time coordinate attributes are different, there are inconsistencies and/or redundancies in the reference data

To do:
- [ ] Get #941 merged to address the precision problem
- [ ] Temporarily bypass the coordinate attributes check ([last commit](https://github.com/ESMValGroup/ESMValCore/pull/685/commits/2bd2d627491251389f81cf0076a133794dab52ad)) to confirm there are no other issues introduced by this PR.
- [ ] Update and commit the reference cubes with the code in this PR
- [ ] Revert the coordinate check bypass
- [ ] Merge

@stefsmeets can you check my diagnosis and approve the proposed solution(s)?

Peter9192:
Okay, so now you can see that the only remaining failures are the daily sample data tests:

FAILED tests/sample_data/multimodel_statistics/test_multimodel.py::test_multimodel_regression_day[full-365_day]
FAILED tests/sample_data/multimodel_statistics/test_multimodel.py::test_multimodel_regression_day[full-gregorian]
FAILED tests/sample_data/multimodel_statistics/test_multimodel.py::test_multimodel_regression_day[full-proleptic_gregorian]

This is because of #937. If you look at the test output, you can see that the shapes are inconsistent (2 (old) vs 62 (new) days in 2 months).

I will now revert the coordinate check. This will introduce more failures due to different time attributes etc. as discussed above.

Peter9192:
Indeed, now there are 8 failures. E.g. `AssertionError: assert DimCoord(arra...r_name='time') == DimCoord(arra...unkSizes': 1})` due the differences explained [here](https://github.com/ESMValGroup/ESMValCore/pull/685#issuecomment-756238112). 

I'll now update the reference data so that 
- The daily files are fixed
- The time points and bounds are consistent with regrid time
- The time coordinate attributes are cleaner and consistent

Peter9192:
@stefsmeets and/or @nielsdrost can you do a final check?

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -22,26 +21,14 @@
...
-
-
 def _plev_fix(dataset, pl_idx):
     """Extract valid plev data.
 
-    this function takes care of situations
-    in which certain plevs are completely
-    masked due to unavailable interpolation
-    boundaries.
+    this function takes care of situations in which certain plevs are
```
stefsmeets, line 30:
```suggestion
    This function takes care of situations in which certain plevs are
```

stefsmeets (commented):
I'm getting these failures on my machine, any idea what that is about?

```
FAILED tests/sample_data/multimodel_statistics/test_multimodel.py::test_multimodel_regression_month[full] - AssertionError:
FAILED tests/sample_data/multimodel_statistics/test_multimodel.py::test_multimodel_regression_day[full-365_day] - AssertionError:
FAILED tests/sample_data/multimodel_statistics/test_multimodel.py::test_multimodel_regression_day[full-gregorian] - AssertionError:
FAILED tests/sample_data/multimodel_statistics/test_multimodel.py::test_multimodel_regression_day[full-proleptic_gregorian] - AssertionError:
```

stefsmeets (approved):
Looks good, nice work!

nielsdrost:
Good work @Peter9192 and @stefsmeets and @bouweandela and @valeriupredoi !

PR #690, opened 2020-06-30
Suggested Documentation changes

ssmithClimate:
A few documentation suggestions, mentioned in:
[Issue #1684](https://github.com/ESMValGroup/ESMValTool/issues/1684)

Closes ESMValGroup/ESMValTool#1684

### DIFF
--- a/doc/recipe/overview.rst
+++ b/doc/recipe/overview.rst
@@ -184,7 +184,7 @@ arguments):
 Preprocessor operations will be applied using the default order
 as listed in :ref:`preprocessor_functions`.
 Preprocessor tasks can be set to run in the order they are listed in the recipe
-by adding `custom: true` to the preprocessor definition.
+by adding ``custom_order: true`` to the preprocessor definition.
 
 Recipe section: ``diagnostics``
 ===============================
--- a/doc/recipe/overview.rst
+++ b/doc/recipe/overview.rst
@@ -182,7 +182,7 @@ arguments):
 .. _Diagnostics:
 
 Preprocessor operations will be applied using the default order
-as `listed here <https://docs.esmvaltool.org/projects/esmvalcore/en/latest/api/esmvalcore.preprocessor.html#preprocessor-functions)`_.
+as listed in :ref:`preprocessor_functions`.
 Preprocessor tasks can be set to run in the order they are listed in the recipe
 by adding `custom: true` to the preprocessor definition.
 
--- a/doc/recipe/overview.rst
+++ b/doc/recipe/overview.rst
@@ -179,12 +179,12 @@ arguments):
    data, applying CMOR checks and fixes (:ref:`CMOR check and dataset-specific
    fixes`) and saving the data to disk.
 
-.. _Diagnostics:
-
 Preprocessor operations will be applied using the default order
 as listed in :ref:`preprocessor_functions`.
 Preprocessor tasks can be set to run in the order they are listed in the recipe
 by adding ``custom_order: true`` to the preprocessor definition.
+
+.. _Diagnostics:
 
 Recipe section: ``diagnostics``
 ===============================
--- a/doc/recipe/overview.rst
+++ b/doc/recipe/overview.rst
@@ -181,6 +181,11 @@ arguments):
 
 .. _Diagnostics:
 
+Preprocessor operations will be applied using the default order
+as `listed here <https://docs.esmvaltool.org/projects/esmvalcore/en/latest/api/esmvalcore.preprocessor.html#preprocessor-functions)`_.
+Preprocessor tasks can be set to run in the order they are listed in the recipe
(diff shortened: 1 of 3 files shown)


doc/recipe/overview.rst
```diff
@@ -181,6 +181,11 @@ arguments):
 
 .. _Diagnostics:
 
+Preprocessor operations will be applied using the default order
+as `listed here <https://docs.esmvaltool.org/projects/esmvalcore/en/latest/api/esmvalcore.preprocessor.html#preprocessor-functions)`_.
```
bouweandela, line 185:
```suggestion
as `listed here <https://docs.esmvaltool.org/projects/esmvalcore/en/latest/api/esmvalcore.preprocessor.html#preprocessor-functions>`_.
```

doc/recipe/overview.rst
```diff
@@ -181,6 +181,11 @@ arguments):
 
 .. _Diagnostics:
 
+Preprocessor operations will be applied using the default order
```
bouweandela, line 184: Maybe it would be nicer to add this text to the section above? https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/overview.html#recipe-section-preprocessors

ssmithClimate: Sounds good. 

One question however. Is there a difference between:
"adding `custom: true` to the preprocessor definition"
and "specified using the custom_order preprocessor function." (as stated now above?)

bouweandela: Thanks for pointing that out, I didn't know that text was there: it is wrong and should be removed, there is no `custom_order` preprocessor function.

bouweandela: To be more specific, what I meant by adding it to the section above, is moving the ``.. _Diagnostics:`` label of the diagnostics section below this new text.

doc/recipe/preprocessor.rst
```diff
@@ -59,7 +59,8 @@ Overview
 The ESMValTool preprocessor can be used to perform a broad range of operations
 on the input data before diagnostics or metrics are applied. The preprocessor
 performs these operations in a centralized, documented and efficient way, thus
-reducing the data processing load on the diagnostics side.
+reducing the data processing load on the diagnostics side.  For an overview of
+the preprocessor structure see the `recipe overview <https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/overview.html#recipe-section-preprocessors>`_.
```
bouweandela, line 63:
```suggestion
the preprocessor structure see the :ref:`Preprocessors`.
```
It would be better to use the label to create the link to the other section, that will work with multiple versions of the documentation.

doc/recipe/overview.rst
bouweandela: Sorry, I was confused about the exact name, indeed the only working solution is `custom_order: true`, just `custom: true` doesn't do anything. The wording above is a bit unfornate though, because it isn't really a preprocessor function, but just a way to change the order in which the preprocessor functions are executed.

doc/recipe/overview.rst
```diff
@@ -181,6 +181,11 @@ arguments):
...
+as `listed here <https://docs.esmvaltool.org/projects/esmvalcore/en/latest/api/esmvalcore.preprocessor.html#preprocessor-functions)`_.
+Preprocessor tasks can be set to run in the order they are listed in the recipe
+by adding `custom: true` to the preprocessor definition.
```
bouweandela, line 187:
```suggestion
by adding `custom_order: true` to the preprocessor definition.
```

doc/recipe/overview.rst
ssmithClimate: Ok

doc/recipe/overview.rst
ssmithClimate: Just removed previous text referencing custom_order as a function. Since this is a couple lines long seems to work well here instead of as a bullet point above.

ssmithClimate:
@bouweandela I believe all comments above are implemented and text is consistent now from what I can tell.

doc/recipe/overview.rst
```diff
@@ -181,6 +181,11 @@ arguments):
 
 .. _Diagnostics:
 
+Preprocessor operations will be applied using the default order
+as listed in :ref:`preprocessor_functions`.
+Preprocessor tasks can be set to run in the order they are listed in the recipe
+by adding ``custom_order: true`` to the preprocessor definition.
```
bouweandela, line 187:
```suggestion
Preprocessor operations will be applied using the default order
as listed in :ref:`preprocessor_functions`.
Preprocessor tasks can be set to run in the order they are listed in the recipe
by adding ``custom_order: true`` to the preprocessor definition.

.. _Diagnostics:
```
It would be good to keep the label of the diagnostics section just above the section title

PR #692, opened 2020-07-02
Add custom variable lweGrace

bascrezee:
Needed by https://github.com/ESMValGroup/ESMValTool/pull/1694. 

Remarks on `lweGrace`: 
- This variable is very specific to the GRACE dataset, given the unique nature of the product, I think it deserves it's own custom CMOR entry
- I don't feel authoritative to set valid min/max for this variable, they are set now, but can I take them out? Or are they required?


bascrezee:
@mattiarighi Do you have an answer to my question above? After that is resolved we can merge this, which is needed for https://github.com/ESMValGroup/ESMValTool/pull/1694.

bouweandela:
@axel-lauer Would you be able to answer the question, or do you know someone who could?

axel-lauer:
We already have a couple of custom variables for which no valid min/max are defined (e.g. rtnt, co2s, xch4). So I do not see a problem if these entries are missing for this new custom variable.

esmvalcore/cmor/tables/custom/CMOR_lweGrace.dat
```diff
@@ -0,0 +1,26 @@
...
+!----------------------------------
+! Additional variable information:
+!----------------------------------
+dimensions:        longitude latitude time
+out_name:          lweGrace
+type:              real
+valid_min:         -1000
+valid_max:         1000
+ok_min_mean_abs:   -1
+ok_max_mean_abs:   1
```
bouweandela, line 24: @bascrezee Looking at the other custom variables, I think this can just be removed then

bouweandela (approved):

PR #696, opened 2020-07-03
Add preprocessor function to resample time and compute x-hourly statistics

jvegasbsc:
@sloosvel found a use case in which we needed to convert 3-hourly data to 6-hourly and we found that this capability is missing in the core,  as we don't go futher than `daily_statistics`. This pull requests add support for x-hourly statistics

But in this case we are working with instantaneous data and so we don't really want to compute but just to extract some values. I added the functionality for this, so we can now convert data frequencies by resampling it.

**Tasks**

-   [x] Add unit tests


sloosvel:
Could they be added in the __init__ list to be called in the recipe? In this way you get rid of the data before continuing with the preprocessing. I guess the output product would need to have the frequency updated then.

esmvalcore/preprocessor/_time.py
```diff
@@ -711,3 +754,73 @@ def timeseries_filter(cube, window, span,
...
+    iris.cube.Cube
+        Cube with the new frequency.
+
+    Raises
+    ------
+    ValueError:
+        The specified freuqncy is not a divisor of 24.
+    """
+    if 24 % hours != 0:
+        raise ValueError('The number of hours must be a divisor of 24')
```
valeriupredoi, line 784: what happens if `hours = 24`?

valeriupredoi: also can you pls add a test that tests for the raised exception :beer:

jvegasbsc:
> Could they be added in the init list to be called in the recipe? In this way you get rid of the data before continuing with the preprocessing. I guess the output product would need to have the frequency updated then.

Added.

esmvalcore/preprocessor/_time.py
```diff
@@ -177,6 +177,49 @@ def get_time_weights(cube):
...
+    Chunks time in x hours periods and computes statistics over them;
+
+    Parameters
+    ----------
+    cube: iris.cube.Cube
+        input cube.
+
+    hours: int
+        Number of hours per period. Must be a divisor of 24
+        (1, 2, 3, 4, 6, 8, 12)
```
stefsmeets, line 193: I suggest adding this list to the documentation for clarity.

Also 24 is a divisor of 24 😁
```suggestion
        (1, 2, 3, 4, 6, 8, 12, 24)
```

esmvalcore/preprocessor/_time.py
```diff
@@ -177,6 +177,49 @@ def get_time_weights(cube):
     return time_weights
 
 
+def hourly_statistics(cube, hours, operator='mean'):
+    """
+    Compute hourly statistics.
+
+    Chunks time in x hours periods and computes statistics over them;
```
stefsmeets, line 184:
```suggestion
    Chunks time in x hours periods and computes statistics over them.
```

esmvalcore/preprocessor/_time.py
stefsmeets: Or what happens if `hours=0.5` or `hours=0.25`?

stefsmeets, line 784:
```suggestion
    allowed_hours = (1, 2, 3, 4, 6, 12)
    if hours not in allowed_hours:
        raise ValueError(f'The number of hours must be one of {allowed_hours}')
```

esmvalcore/preprocessor/_time.py
```diff
@@ -711,3 +754,73 @@ def timeseries_filter(cube, window, span,
...
+    """
+    if 24 % hours != 0:
+        raise ValueError('The number of hours must be a divisor of 24')
+    hours = range(0, 24, hours)
+    select_hours = iris.Constraint(
+        time=lambda cell: cell.point.hour in hours)
+    return cube.extract(select_hours)
+
+
+def resample_time(cube, month=None, day=None, hour=None):
```
stefsmeets, line 791: This is slightly confusing, how is this function different from `resample_hours`?

Does `resample_time(cube, hour=6)` give the same out put as `resample_hours(cube, hours=6)`

esmvalcore/preprocessor/_time.py
```diff
@@ -711,3 +754,73 @@ def timeseries_filter(cube, window, span,
                                weights=wgts)
 
     return cube
+
+
+def resample_hours(cube, hours):
```
stefsmeets, line 759: I'm thinking `interval` or `every_n_hours` would be a better variable name here instead of `hours`.

esmvalcore/preprocessor/_time.py
```diff
@@ -711,3 +754,73 @@ def timeseries_filter(cube, window, span,
...
+    """
+    Converts x-hourly data to y-hourly by eliminating extra timesteps.
+
+    Converts x-hourly data to y-hourly (y > x) by eliminating the extra
```
stefsmeets, line 763: Please add this (y > x) as a check or assertion

esmvalcore/preprocessor/_time.py
```diff
@@ -711,3 +754,73 @@ def timeseries_filter(cube, window, span,
...
+            return False
+        if day is not None and day != date.day:
+            return False
+        if hour is not None and hour != date.hour:
+            return False
+        return True
+
+    resample_constraint = iris.Constraint(
+        time=lambda cell: compare(cell.point))
+    return cube.extract(resample_constraint)
```
stefsmeets, line 826: Wouldn't this be a bit simpler?

```suggestion
    def compare(cell):
        date = cell.point
        if month is not None and month != date.month:
            return False
        if day is not None and day != date.day:
            return False
        if hour is not None and hour != date.hour:
            return False
        return True

    resample_constraint = iris.Constraint(
        time=compare)
    return cube.extract(resample_constraint)
```

stefsmeets (changes requested):
Hi @jvegasbsc , I just had a look at this pull request. It seems like it is very close to being ready to be merged. It just needs a bit of work to get it finished. Please see my review comments.

esmvalcore/preprocessor/_time.py
jvegasbsc: No, 

resample_time(cube, hour=6) -> Daily values taken at 6:00
resample_hour(cube, every_n_hours=6) -> Six hourly values at 0:00, 6:00, 12:00, 18:00

esmvalcore/preprocessor/_time.py
jvegasbsc: Changed to intervak

esmvalcore/preprocessor/_time.py
jvegasbsc: Done

esmvalcore/preprocessor/_time.py
```diff
@@ -182,9 +178,52 @@ def get_time_weights(cube):
...
+    Chunks time in x hours periods and computes statistics over them.
+
+    Parameters
+    ----------
+    cube: iris.cube.Cube
+        input cube.
+
+    hours: int
+        Number of hours per period. Must be a divisor of 24
+        (1, 2, 3, 4, 6, 8, 12, 24)
```
stefsmeets, line 193: This should probably not have the 24 then, since we only check up to 12.

```suggestion
        (1, 2, 3, 4, 6, 8, 12)
```

doc/recipe/preprocessor.rst
```diff
@@ -790,6 +793,35 @@ between 1 and 12 as the named month string will not be accepted.
...
+----------------------
+
+This function produces statistics for each month in the dataset.
+
+Parameters:
+    * operator: operation to apply. Accepted values are 'mean',
+      'median', 'std_dev', 'min', 'max' and 'sum'. Default is 'mean'
+
+See also :func:`esmvalcore.preprocessor.monthly_statistics`.
+
```
stefsmeets, line 824: Duplicate entry (See L838)

```suggestion
```

esmvalcore/preprocessor/_time.py
```diff
@@ -725,3 +755,81 @@ def timeseries_filter(cube, window, span,
                                weights=wgts)
 
     return cube
+
+
+def resample_hours(cube, interval, offset=0):
+    """Convert x-hourly data to y-hourly by eliminating extra timesteps.
+
+    Convert x-hourly data to y-hourly (y > x) by eliminating the extra
+    timesteps. This is intended to be used only with instantaneous values.
```
stefsmeets, line 764:
```suggestion
    timesteps. This is intended to be used only with instantaneous values.
    
    For example, `resample_hours(cube, interval=6)` -> Six-hourly intervals at 0:00, 6:00, 12:00, 18:00.
```

esmvalcore/preprocessor/_time.py
```diff
@@ -725,3 +755,81 @@ def timeseries_filter(cube, window, span,
...
+    select_hours = iris.Constraint(time=lambda cell: cell.point.hour in hours)
+    return cube.extract(select_hours)
+
+
+def resample_time(cube, month=None, day=None, hour=None):
+    """Change frequency of data by resampling it.
+
+    Converts data from one frequency to another by extracting the timesteps
+    that match the provided month, day and/or hour. This is meant to be used
+    with instantaneous values when computing statistics is not desired.
```
stefsmeets, line 807: I added the example from above, but maybe it is worth it to some more examples.

```suggestion
    with instantaneous values when computing statistics is not desired.
    
    For example, `resample_time(cube, hour=6)` -> Daily values taken at 6:00.
```

esmvalcore/preprocessor/_time.py
stefsmeets: Nice, I suggest adding these examples to the docstring to avoid confusion to make the intent clear.

stefsmeets (changes requested):

doc/recipe/preprocessor.rst
stefsmeets: Also causes CircleCI to complain:
```
Warning, treated as error:
/doc/doc/recipe/preprocessor.rst:838:Duplicate explicit target name: "monthly_statistics".
```

stefsmeets:
Nice work @jvegasbsc ! Can you fix the last CI errors:

commit/documentation:
```
Warning, treated as error:
/doc/doc/recipe/preprocessor.rst:974:Title underline too short.

``resample_hours``
-----------------
```

Codacy:
https://app.codacy.com/gh/ESMValGroup/ESMValCore/pullRequest?prid=5799969 

Then I think this is ready to be merged!

jvegasbsc:
Done!

stefsmeets (approved):

stefsmeets:
Hi @axel-lauer , this PR is finished from a technical point of view. We need a scientific review to get this merged, but I'm not entirely sure who to ask. Do you have any idea who we could ask?

axel-lauer:
Maybe you could take a look at this issue: https://github.com/ESMValGroup/ESMValTool/issues/1818. The list is still a draft and far from complete but I would guess we might ask e.g. someone listed under 'extreme events'.

doc/recipe/preprocessor.rst
```diff
@@ -790,6 +793,22 @@ between 1 and 12 as the named month string will not be accepted.
 
 See also :func:`esmvalcore.preprocessor.extract_month`.
 
+.. _hourly_statistics:
+
+`hourly_statistics``
```
stefsmeets, line 798:
```suggestion
``hourly_statistics``
```

stefsmeets:
Thanks @axel-lauer ! Let's see if we can do this 😉

Hi @jhardenberg @maritsandstad, could I ask one of you to do a scientific review of this PR? We already did a technical review and would like to get this merged. 

You can perhaps start by looking at the documentation (https://github.com/ESMValGroup/ESMValCore/blob/f8bfd90e93c30e0f4935dfbb4d7bc618274ac37d/doc/recipe/preprocessor.rst), and see if that makes sense and go from there. Thanks!

jhardenberg:
Hi, looks like a very useful addition indeed! The difference between `resample_time` and `resample_hours` is that the first one should be used to extract for example always 1PM from maybe all days, or to extract only 1PM on the 12th of each month, while `extract_hours` extracts data by resampling every 6h for example, correct?

Both seem useful, but would the exact same functionality of `extract_hours` not be covered by `extract_time` if it allowed to specify a **list** of hours, instead of only one value? Anyway this could maybe be a later change.

Isn't `extract_month` and (`resample_time` specifying only "month:") doing the same?

As you write this sampling makes sense only for instantaneous data and would be clearly wrong for cumulated data (or time-averaged) data (all mass or energy fluxes).  Would it maybe be better to add a check in the variable attributes to prevent users to apply it to such variables, like e.g. precipitation? Should the name  "extract_hours" maybe be changed to something like "extract_inst_hours" to make this clear and avoid abuse?

Actually it would be nice in the future to develop a similar functionality also for time-averaged variables (e.g. precipitation, radiative fluxes),  so that one can change from 3hr frequency to 6hr etc .... In this case, since the variables are in units like Kg/m^2/s or W/m^2, they should be averaged between timesteps. E.g. 3hr data would become 6hr data averaging the 1st and 2nd timestep, the 3rd and 4th etc. Also in this case an offset could be specified.

jvegasbsc:
> Hi, looks like a very useful addition indeed! The difference between resample_time and resample_hours is that the first one should be used to extract for example always 1PM from maybe all days, or to extract only 1PM on the 12th of each month, while resample_hours extracts data by resampling every 6h for example, correct?

Yes, that's the difference

> Both seem useful, but would the exact same functionality of resample_hours not be covered by resample_time if it allowed to specify a list of hours, instead of only one value? Anyway this could maybe be a later change.

Yes, but in that case I feel inclined to also allow lists on the other options and this will lead to frequencies that we are not currently supporting

> Isn't extract_month and (resample_time specifying only "month:") doing the same?

Yes, certain configurations of this function can do exactly what other functions already do. 

>As you write this sampling makes sense only for instantaneous data and would be clearly wrong for cumulated data (or time-averaged) data (all mass or energy fluxes). Would it maybe be better to add a check in the variable attributes to prevent users to apply it to such variables, like e.g. precipitation? Should the name "resample_hours" maybe be changed to something like "resample_inst_hours" to make this clear and avoid abuse?

A check in the cell methods to check if `time: mean` or `time:sum` can be done, but we are not checking those attributes. in the cmor checker so I am not really prone to take them as a reliable source of information, so I think we could not add more than a warning here 

> Actually it would be nice in the future to develop a similar functionality also for time-averaged variables (e.g. precipitation, radiative fluxes), so that one can change from 3hr frequency to 6hr etc .... In this case, since the variables are in units like Kg/m^2/s or W/m^2, they should be averaged between timesteps. E.g. 3hr data would become 6hr data averaging the 1st and 2nd timestep, the 3rd and 4th etc. Also in this case an offset could be specified.

The hourly statistics will do this, although an offset can not be specified

stefsmeets:
Hi @jhardenberg , if you are happy with the response, could you please approve this PR? Then I think this can be merged 😄

jhardenberg:
Hi, the responses are fine for me, thanks. As said, I would just add at least  a clear warning about  the fact that the sampling provided by this function should be used only for instantaneous data and not for fluxes (which should be averaged).

jhardenberg (approved):

stefsmeets:
Hi @ESMValGroup/esmvaltool-coreteam , this PR is ready to be merged!

jvegasbsc:
The pending errors are related to the documentation and they already exist on master

bouweandela:
> As said, I would just add at least a clear warning about the fact that the sampling provided by this function should be used only for instantaneous data and not for fluxes (which should be averaged).

@jvegasbsc It would have been nice if this comment by @jhardenberg could have been addressed before merging this, but it's too late now I guess.

doc/recipe/preprocessor.rst
```diff
@@ -798,6 +801,22 @@ between 1 and 12 as the named month string will not be accepted.
...
+This function produces statistics at a x-hourly frequency.
+
+Parameters:
+    * every_n_hours: frequency to use to compute the statistics. Must be a divisor of
+      24.
+
+    * operator: operation to apply. Accepted values are 'mean',
+      'median', 'std_dev', 'min', 'max' and 'sum'. Default is 'mean'
+
+See also :func:`esmvalcore.preprocessor.daily_statistics`.
```
lukruh, line 818: this should be:
See also :func:`esmvalcore.preprocessor.hourly_statistics`.
I guess

PR #700, opened 2020-07-06
Starting to test gitHub Actions

valeriupredoi:

**Tasks**

-   [ ] Add unit tests

* * *

Closes #issue_number


valeriupredoi:
@bouweandela @jvegasbsc have a look at this - Github Action tests allow us to test stuff (everything inc builds, installations, run tests etc) on different OS's and different Python envs - I still have to perfect the OSX and Windows tests, but for now I managed to verify the installation of @bouweandela 's new PyPi Core package :grin: :beer: 

I reckon these are very useful tests to run in different scenarios including when the package changes from PyPi/conda channel or when we change the environments - we can finally have control over both Python versions and OS support (for free)!

valeriupredoi:
yuss! made it work on OSX as well :beer: Some lessons:

- very important not to mix channels (ie `anaconda` mixed with `conda-forge`) for OSX
- `conda activate` doesn't play at all (despite many efforts of tricking `conda init` to accept the `bash` or `zsh` shells); `source activate` still works well though

jvegasbsc:
Related to the Windows installation, they are working on support for the WSL:
https://github.com/actions/virtual-environments/issues/50

We may have to wait a bit

nielsdrost:
Nice work @valeriupredoi! I guess this (for now) tests the packages on PyPi, not actually using the code in the repo, right?

To not overwhelm contributors in PR's with a large number of additional checks that do not actually get influenced by their PR we need some way to filter these out normally, I guess.

There seem to be ways to trigger actions by writing comments in the PR. This could be a fun way to do this, perhaps?

https://docs.github.com/en/actions/reference/events-that-trigger-workflows
https://github.com/marketplace/actions/pull-request-comment-trigger

valeriupredoi:
cheers @nielsdrost - I will tweak this accordingly - one thing I just noticed is that now the CircleCI tests appear under the `commit` tab and they get fully run with detailed output as if we run them normally. I removed the testing on Windows (cheers for the instructions in the tutorial!) since it is a nightmare, if not impossible, to test that install procedure via the GA testing environment, but the OSX and Debian tests pass nicely :beer:

valeriupredoi:
found a very convenient way to run the Github Actions locally 9saves the push an wait): https://github.com/nektos/act

valeriupredoi:
OK guys - I overhauled the testing frame and it's much nicer now - proves out you can create conda envs quite nicely in Github Actions and even change shells (we don't need that but just in case): I am now testing for installations from both PyPi and source and doing so for OSX and Linux for Py3.6-7-8. I also found out how to run the tests locally (on yer machine) w/o having to consume run time on the Github machine :beer:

valeriupredoi:
I will also do the same in ESMValTool, the big question mark I still have to figure out is when to trigger these tests - when the env file changes would be my idea - what say you @bouweandela @jvegasbsc @nielsdrost (I have no idea how to set the workflow trigger on a file change only, if you know... :beer: )

valeriupredoi:
OK guys, here we have us a complete set of tests, and they're already pretty useful:
- installation from PyPi works perfecto for all;
- installation from conda works for Linux (from miniconds);
- installation from conda fails for OSX (and Linux with anaconda, not miniconda);
- tests fail on Python 3.6-7 for Linux/OSX due to the issue raised by @bouweandela in https://github.com/ESMValGroup/ESMValCore/issues/719
I reckon this is quite ready - please comment and add suggestions :beer:

valeriupredoi:
another note of use: set `fail-fast: false` in the `strategy` under matrix testing so that the workflow doesn't get cancelled after the first job fails (this could be turned off for special cases, but we need each test to go) - this also saves us writing extra individual workflows :beer:

valeriupredoi:
about OSX and Python 3.8.5 - this is brand new (the Python version on OSX, that is - July 2020, I think a week ago) so we are not seeing a correct behaviour as it is the case with 3.6 and 3.7

bouweandela:
Nice work V, that's a lot of tests!

> when to trigger these tests

Would it be possible to run them once a day? On the `master` branch only (if a branch is tested)?

valeriupredoi:
yeah man, that's a good idea! So basically everytime we merge a branch into master make the Actions run?

bouweandela:
> So basically everytime we merge a branch into master make the Actions run? Or run them every day regardless of activity on master - like a cron job?

Running every day would be my preference, because then you notice when things break because dependencies were updated, especially for the tests that test installation of the package.

Issue #124, opened 2019-02-25
Add YAML schema for diagnostic_provenance.yml and improve error messages/checks


bouweandela:
At the moment diagnostic developers either get no error message at all, or various obscure error messages if they make a mistake in their diagnostic_provenance.yml file. It would be nicer if we could get a more clear error message by using a Yamale schema to check it.

bouweandela:
Other error messages and checks should be also be improved, e.g. when you provide a caption that is a list instead of a string, you will get this rather cryptic error message:

```
"""
Traceback (most recent call last):
  File "/mnt/lustre01/pf/b/b309057/SOFTWARE/miniconda3/envs/esmvaltool/lib/python3.7/multiprocessing/pool.py", line 121, in worker
    result = (True, func(*args, **kwds))
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_task.py", line 671, in _run_task
    output_files = task.run()
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_task.py", line 227, in run
    self.output_files = self._run(input_files)
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_task.py", line 496, in _run
    self._collect_provenance()
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_task.py", line 559, in _collect_provenance
    product.save_provenance()
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_provenance.py", line 252, in save_provenance
    self._include_provenance()
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_provenance.py", line 248, in _include_provenance
    write(filename, attributes)
  File "/mnt/lustre01/pf/b/b309057/ESMValTool/core/esmvalcore/_provenance.py", line 219, in _include_provenance_nc
    setattr(dataset, key, value)
  File "netCDF4/_netCDF4.pyx", line 2882, in netCDF4._netCDF4.Dataset.__setattr__
  File "netCDF4/_netCDF4.pyx", line 2812, in netCDF4._netCDF4.Dataset.setncattr
  File "netCDF4/_netCDF4.pyx", line 1568, in netCDF4._netCDF4._set_att
OSError: array string attributes can only be written with NETCDF4
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

PR #711, opened 2020-07-13
Produce a readable warning if ancestors are a string instead of a list.

### LINKED ISSUE #686: Warning for "Invalid ancestor" hard to read if ancestors is s string instead of a list
**Describe the bug**
In esmvalcore/_task.py a warning should be produced, if the ancestor file in attributes is not found. To identify the filename, attributes['ancestors'] should contain a list with the filename. If a string is found there instead of a list, the loop "for ancestor_file in ancestor_files:" runs over every character of this string and produces a cryptic warning, see below for an example. It would be better to warn if a string is in attributes['ancestors'].

**Please attach**

The diagnostics which issued these warnings have been fixed now, the main_log_debug.txt is attached for the non corrected version of the diagnostic  esmvaltool/diag_scripts/droughtindex/collect_drought_func.py used in esmvaltool/recipes/recipe_martin2018grl.yml
[main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/4844374/main_log_debug.txt)

See also  https://github.com/ESMValGroup/ESMValTool/pull/1689

katjaweigel:

**Tasks**

-   [ ] Add unit tests

Tested with recipe_hyint.yml (how do I make a unit test?).

Documentation is not effected because it should just change a warning? 
* * *

Closes  #686 and helps to create better readable warnings for some of the warnings reported in e.g https://github.com/ESMValGroup/ESMValTool/issues/1716

### DIFF
--- a/esmvalcore/_task.py
+++ b/esmvalcore/_task.py
@@ -583,8 +583,8 @@ class DiagnosticTask(BaseTask):
                 logger.warning(
                     "Ancestor file(s) %s specified for recording provenance "
                     "of %s, created by diagnostic script %s in task %s is "
-                    "a string instead of a list", ancestor_files, filename,
-                    self.script, self.name)
+                    "a string but should be a list of strings", ancestor_files,
+                    filename, self.script, self.name)
                 ancestor_files = [ancestor_files]
             for ancestor_file in ancestor_files:
                 if ancestor_file in ancestor_products:
--- a/esmvalcore/_task.py
+++ b/esmvalcore/_task.py
@@ -579,6 +579,14 @@ class DiagnosticTask(BaseTask):
                     self.script, self.name)
                 valid = False
             ancestors = set()
+            if isinstance(ancestor_files, str):
+                valid = False
+                logger.warning(
+                    "Ancestor file(s) %s specified for recording provenance "
+                    "of %s, created by diagnostic script %s in task %s is "
+                    "a string instead of a list", ancestor_files, filename,
+                     self.script, self.name)
+                ancestor_files = [ancestor_files] 
             for ancestor_file in ancestor_files:
                 if ancestor_file in ancestor_products:
                     ancestors.add(ancestor_products[ancestor_file])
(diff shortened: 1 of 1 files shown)


mattiarighi:
Can you remove the draft status?

katjaweigel:
Yes, as soon as the test with the last small change runs through
 (should take about 10 mintutes).

katjaweigel:
Did run successfully so I removed the draft

koldunovn:
Hi @katjaweigel @mattiarighi 
It looks like the additional problem with this part of the code, might be that it assumes `ancestor_files` always exist. First there is a check `if not ancestor_files:` that will issue a warning, but later it still tries to do `for ancestor_file in ancestor_files:`.

I don't know if we whant to try to solve it here or in another PR. I would just probably do something like:

```python
            if not ancestor_files:
                logger.warning(
                    "No ancestor files specified for recording provenance of "
                    "%s, created by diagnostic script %s in task %s", filename,
                    self.script, self.name)
                valid = False
            ancestors = set()
            if isinstance(ancestor_files, str):
                logger.warning(
                    "Ancestor file(s) %s specified for recording provenance "
                    "of %s, created by diagnostic script %s in task %s is "
                    "a string instead of a list", ancestor_files, filename,
                    self.script, self.name)
                ancestor_files = [ancestor_files]
            if valid:
                for ancestor_file in ancestor_files:
                    if ancestor_file in ancestor_products:
                        ancestors.add(ancestor_products[ancestor_file])
                    else:
                        valid = False
                        logger.warning(
                            "Invalid ancestor file %s specified for recording "
                            "provenance of %s, created by diagnostic script %s "
                            "in task %s", ancestor_file, filename, self.script,
                            self.name)
```
But this adds another if :)

bouweandela:
> It looks like the additional problem with this part of the code, might be that it assumes ancestor_files always exist. First there is a check if not ancestor_files: that will issue a warning, but later it still tries to do for ancestor_file in ancestor_files:

The only assumption is that the value associated with the `ancestors` key is of type `list`. If no `ancestors` key is present, an empty list will be used and the for loop will work as expected (nothing happens if you loop over an empty list). Unfortunately it turned out that even that was too strong an assumption.

For the next release I'm planning to add a check that the type is correct, see #124, but I wanted to start with just some warnings so not all recipes are suddenly broken.

bouweandela:
cherry-picked this into v2.0.0 branch

PR #709, opened 2020-07-11
Options arg in read_config_user_file now optional

jvegasbsc:
To fix and issue with the cmorizers

**Tasks**

-   [x] [Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do, if you haven't done so already (and add the link at the bottom)
-   [x] This pull request has a descriptive title that can be used in a changelog
-   [x] Add unit tests
-   [x] Public functions should have a numpy-style docstring so they appear properly in the [API documentation](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/api/esmvalcore.html). For all other functions a one line docstring is sufficient.
-   [x] If writing a new/modified preprocessor function, please update the [documentation](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/preprocessor.html)
-   [x] Circle/CI tests pass. Status can be seen below your pull request. If the tests are failing, click the link to find out why.
-   [x] Codacy code quality checks pass. Status can be seen below your pull request. If there is an error, click the link to find out why. If you suspect Codacy may be wrong, please ask by commenting.
-   [x] Please use `yamllint` to check that your YAML files do not contain mistakes
-   [x] If you make backward incompatible changes to the recipe format, make a new pull request in the [ESMValTool repository](https://github.com/ESMValGroup/ESMValTool) and add the link below

If you need help with any of the tasks above, please do not hesitate to ask by commenting in the issue or pull request.

* * *

Closes #issue_number

PR #716, opened 2020-07-16
Options arg in read_config_user_file now optional (#709)

jvegasbsc:
As requested by @schlunma

### DIFF
--- a/esmvalcore/_config.py
+++ b/esmvalcore/_config.py
@@ -31,7 +31,7 @@ def find_diagnostics():
 DIAGNOSTICS_PATH = find_diagnostics()
 
 
-def read_config_user_file(config_file, folder_name, options):
+def read_config_user_file(config_file, folder_name, options=None):
     """Read config user file and store settings in a dictionary."""
     config_file = os.path.abspath(
         os.path.expandvars(os.path.expanduser(config_file)))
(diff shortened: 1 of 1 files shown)


schlunma:
Perfect, thanks @jvegasbsc! In #709 there are 9 files changed, here it is only 1. Are these changes already in included in `master`? I think every `v2.0.0` should be a subset of `master`, so every change in `v2.0.0` should also be in `master`.

jvegasbsc:
Those changes are already on master. I cherry picked the commit

schlunma:
All right, but doesn't that mean that there are changes in `v2.0.0` which should not be there?

bouweandela:
> Those changes are already on master. I cherry picked the commit

@jvegasbsc @mattiarighi Sorry if this was not clear. Commits should be cherry picked from the master branch into the v2.0.0 release branch, not the other way around. The way to get something in the release is to make a pull request to `master` and then ask me or @mattiarighi to cherry-pick that particular commit into the `v2.0.0` release branch. Now the branches have a different history and every pull request that has been merged since the `v2.0.0` release branch forked off the `master` branch has also been included in it when #709 was merged.

mattiarighi:
Sorry @bouweandela, hope this does not cause too much trouble...

Is there a way to revert this?
I think we only had 2 PR merged in the last week: #709 and #716

bouweandela:
It's a few more, but nothing big. I think it's fine to leave it like this for now. Let's be more careful next time.

PR #723, opened 2020-07-21
Reduce size of docker image

jvegasbsc:
This is a small refactor of the docker file that reduces the size of the docker image from 1.1 GB to ~800 MB (as stated in Docker Hub). 

**Tasks**

-   [ ] Add unit tests

* * *

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,21 +2,11 @@
 # docker build -t ${TAG_OF_YOUR_CHOICE} . -f docker/Dockerfile
 FROM continuumio/miniconda3
 
-# update the conda packages
-RUN conda update -y conda pip
-
-
-# install environment packages
-COPY . /home/root/source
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -2,21 +2,10 @@
...
-
-
-# install environment packages
-COPY . /home/root/source
-WORKDIR /home/root/source
-RUN ls
-RUN conda env update --name base --file environment.yml
-RUN pip install .
-RUN conda clean --all -y
-RUN rm -r /home/root/source
```
bouweandela, line 16: Would it still make sense to remove the source code directory after installation to reduce the size?

jvegasbsc: I am not sure this is having a great impact as Docke stores the containers by layers and the source is in the previous layers. 

What I have seen is critical is reducing the number of layers and do the cleaning at the same time.

PR #725, opened 2020-07-22
Update package description in setup.py

mattiarighi:


bouweandela:
@mattiarighi I changed this a bit, so it reads the information from the same source as Zenodo, that will help keep things in sync.

PR #727, opened 2020-07-22
Set unit=1 if anomalies are standardized

### LINKED ISSUE #726: Standardized anomalies should be unitless.
**Describe the bug**
The title says it all. Found by @hb326 while testing https://github.com/ESMValGroup/ESMValTool/issues/1687.

bascrezee:
Closes #726. @hb326 Can you test in combination with the MPQB framework?


bascrezee:
@hb326 Can you test?

bascrezee:
@hb326 Did you test? I can already tell that we run into a few issues that have to be solved.

hb326:
@bascrezee: I did test it just now. Does not work for the MPQB.
See the following error message:
`ValueError: Unable to convert from 'Unit('1')' to 'Unit('kg m-2 s-1')'.`
`2020-08-06 09:41:58,509 UTC [44220] INFO    If you suspect this is a bug or need help, please open an issue on https://github.com/ESMValGroup/ESMValTool/issues and attach the run/recipe_*.yml and run/main_log_debug.txt files from the output directory.`

Seems to me it is the same error as you reported in #741

hb326:
I ran the MPQB recipe again, with the inlcuded changes to the preprocessor.
Good news: it ran through!
Not so good news: I am missing the units at the y-axis... Are there supposed to be units?
lineplot_pr_1983-2016

bascrezee:
Can you check the units of the preprocessor file?  `ncdump -h ` I would expect them to be `1`.

hb326:
> 
> 
> Can you check the units of the preprocessor output? `ncdump -h ` I would expect them to be `1`.

@bascrezee: yes, the unit is '1' in the preprocessor output.

bascrezee:
Good. Than all is fine I would say, the units not being shown at the y-axis there is Iris behaviour (`iris.quickplot.plot`). Apparently it defaults to not showing in this case, which is fine. I am now working on the tests as requested by @jvegasbsc.

bascrezee:
Can this be merged?

Issue #731, opened 2020-07-28
Installation and testing of esmvalcore package


valeriupredoi:
#732  and its Github Actions give us insight into the portability of `esmvalcore` as package:

**Installation**:
- installation from PyPi works fine for both Linux (Python>=3.6, <=3.8) and OSX (Python>=3.6, <=3.7)
- installation from conda:
  - works fine via miniconda on Linux (Python>=3.6, <=3.8)
  - works fine via anaconda on Linux (Python>=3.6, <=3.8) **only** if inside an empty virtual env (`conda create -n esmvalcore`), otherwise, it fails with unsolvable environment
  - works fine via miniconda on OSX (Python=3.6 only) **only** if inside an empty virtual env (`conda create -n esmvalcore`), otherwise, it fails with unsolvable environment, also fails for Python<=3.7 in any case
- installation from source works well via miniconda for both Linux and OSX

**Testing**:
testing is easier:
- it works for all bar a couple failed tests for Python<3.8 (https://github.com/ESMValGroup/ESMValCore/issues/719)

We should definitely add the instruction for users to create an empty conda env if their installation fails; we should also look into maybe pinning some deps for the conda env, I am going to remove `libunwind` since I know for sure it's not needed no more, see what that does.

EDIT: spoke too soon, we still need `libunwind` bah

valeriupredoi:
**UPDATE** on testing: oddly enough both Circle and Github Actions CI's fail all Python tests with the system-specific `INTERNALERROR> AttributeError: 'tuple' object has no attribute 'rsplit'` but I am not seeing this on my local machine

valeriupredoi:
> **UPDATE** on testing: oddly enough both Circle and Github Actions CI's fail all Python tests with the system-specific `INTERNALERROR> AttributeError: 'tuple' object has no attribute 'rsplit'` but I am not seeing this on my local machine

solved by merging `master`

valeriupredoi:
OK @bouweandela last PR https://github.com/ESMValGroup/ESMValCore/pull/735 addresses the instructions specifically encouraging the use of miniconda and telling users to create an empty env if the env cannot be solved in `(base)`. There still are a few outstanding issues though:

Changed the PR to https://github.com/ESMValGroup/ESMValCore/pull/732 and reporting this from there:

[...]

Issue #719, opened 2020-07-20
Failing unit tests with Python 3.6, =<3.7.3


bouweandela:
These tests are failing on CircleCI with Python 3.7.3:
```
FAILED tests/integration/cmor/_fixes/cmip6/test_cesm2.py::test_cl_fix_file - TypeError: 'NonCallableMagicMock' object is not callable
FAILED tests/integration/cmor/_fixes/cmip6/test_cesm2_waccm.py::test_cl_fix_file - TypeError: 'NonCallableMagicMock' object is not callable
```

The is probably caused by this bug: https://bugs.python.org/issue23078

valeriupredoi:
this happens both on Linux/Python3.6 and OSX/Python3.6 and 3.7 as well, Bouwe, see https://github.com/ESMValGroup/ESMValCore/pull/700 (note that 3.8 for OSX is in fact 3.7, they haven't upgraded the machine's OSX Python yet, 3.8.5 is a week old)

bouweandela:
Updating the docker container that we use for running our unit tests on CircleCI fixed this.

PR #732, opened 2020-07-30
Github Actions Tests

valeriupredoi:
**ESMValCore Github Actions**

This is the main PR that implements Github Actions testing for ESMValCore. Issue: #731. A few explanations:

**When are tests triggered**

Triggering is activated on pushes to `master` and this branch (that should stay as parallel development branch even after merges into `master`); the entry in each of the action files responsible for triggering looks like this
```yaml
# runs on a push on master and at the end of every day
on:
  # triggering on push without branch name will run tests everytime
  # there is a push on any branch
  # turn it on only if needed
  push:
    branches:
    - master
    - github-actions2
  # run the test only if the PR is to master
  # turn it on if required
  #pull_request:
  #  branches:
  #  - master
  schedule:
    - cron: '0 0 * * *'
```
and it means that any `push` on `master` and `github-actions2` branches will trigger the test runs; also tests are being run on `master` (only) at midnight every night.

**What sort of tests are done**

We have four types of tests: test installation from conda, from PyPi and from source and actual unit+intergation tests; each of these are run for Linux (Ubuntu) and OSX and for each of the three supported Python versions: 3.6, 3.7 and 3.8, 24 tests in total.

**How are the tests run**

- Each test is run part of a workflow, but each test is run irrespective if other tests have failed in the workflow, this is set by: `fail-fast: false` in the workflow `strategy`; anywhere where conda is used it is via **miniconda3** and uses the latest version and `conda-forge` as channel for deps:
```yaml
          miniconda-version: "latest"
          channels: conda-forge
```
- Each test is run either inside a conda environment or not, depending on the type of test:
    - PyPi installation: the flow creates a conda environment where the package is installed;
    - Conda installation: linux: installation is done straight into the `(base)` channel; OSX: installation is done inside an empty `esmvalcore` environment (`activate-environment: esmvalcore`);
    - Source installation: installation is done inside a complete conda environment;
    - Unit/integration tests are done in a complete conda environment and are run as

```bash
python setup.py test --addopts '-m "not installation"'
```


valeriupredoi:
OK guys @bouweandela @jvegasbsc we are in a good position to put these GA tests on `master` - I have set up cron testing (and the commented out line is running it every day at midnight) but I don't have rights to run on `master` and these cron actions run only on master; I have a test PR at https://github.com/ESMValGroup/ESMValCore/pull/734 to see it run every 10min and if you approve that then we can have a look at the cron jobs running (hopefully they will!); testing is fairly comprehensive so a nightly test will tell us quite a bit about the status of our package. BTW I closed #700 because there was already too much fluff there plus some dude said that if you start a fresh PR the cron jobs for GA will run on another branch too other than `master` (he was wrong) :beer:

valeriupredoi:
@bouweandela have a look at [this test fail](https://github.com/ESMValGroup/ESMValCore/pull/732/checks?check_run_id=928409430)

nielsdrost:
As already suggested by @valeriupredoi, I propose we merge #734 first to debug any basic GA problems there, and then continue with this one.

valeriupredoi:
cheers muchly @nielsdrost :beer: On it now!

valeriupredoi:
good news! the cron scheduled test ran last night (yeah I completely forgot to check until now) :grin: https://github.com/ESMValGroup/ESMValCore/actions/runs/206335719

nielsdrost:
I guess the next step is to get the failing tests fixed, right? Anything else still needed?

valeriupredoi:
we need to decide what Python versions we don't support no more - @bouweandela @jvegasbsc @nielsdrost can we agree on that? Then I can just remove the versions tested against from the configs and hopefully all will be smooth as a baby's arse afterwards :grin:

valeriupredoi:
OK so all tests that are failing are failing coz of that issue with Python<3.7.6. I am a bit concerned with the fact that conda installation fails on OSX on Python 3.8.5 - I would have expected dependency conflicts to have been solved by now, given Python 3.8.5 is almost a month and a half old on OSX :confused:

nielsdrost:
> we need to decide what Python versions we don't support no more - @bouweandela @jvegasbsc @nielsdrost can we agree on that? Then I can just remove the versions tested against from the configs and hopefully all will be smooth as a baby's arse afterwards

Ideally we would support all officially supported Python versions (https://en.wikipedia.org/wiki/History_of_Python) but I guess  Python<3.7 are out due to an issue? So, 3.7 and 3.8?

bouweandela:
I would like to keep supporting 3.6 at least until 3.9 is out (planned for October). As far as I know using the tool works OK, but there is a small bug in the unittest standard librarary that will not be fixed in Python 3.6 (#719).

nielsdrost:
> I would like to keep supporting 3.6 at least until 3.9 is out (planned for October). As far as I know using the tool works OK, but there is a small bug in the unittest standard librarary that will not be fixed in Python 3.6 (#719). This bug has no consequences for the tool, only that two unit tests fail.

We could also keep them, but in that case I'm not sure how to get this PR merged, as it is now failing. Do you want to skip these two specific tests? Or, skip them only for Python=3.6?

bouweandela:
> Or, skip them only for Python=3.6?

Yes, that's probably the best idea.

valeriupredoi:
sounds like a plan! am gonna plug the skip for < 3.7.6 and have another look at the OSX with 3.8, then we can merge :+1:

valeriupredoi:
OK @nielsdrost @bouweandela I fixed those tests and actually had to remove Python 3.8 from conda installation on OSX - it simply doesn't work yet (not sure if it will until the switch to 3.9). Anyway, this is ready for merge, can I get any final remarks and/or approvals pls :beer:

.github/workflows/action-install-from-conda.yml
```diff
@@ -0,0 +1,108 @@
...
+
+# runs on a push on master and at the end of every day
+on:
+  # triggering on push without branch name will run tests everytime
+  # there is a push on any branch
+  # turn it on only if needed
+  push:
+    branches:
+    - master
+    - github-actions2
```
nielsdrost, line 25: This can probably be removed, I guess

.github/workflows/action-install-from-pypi.yml
```diff
@@ -22,6 +22,8 @@ on:
   push:
     branches:
     - master
+    - github-actions2
```
nielsdrost, line 25: This can now be removed, I guess

.github/workflows/action-install-from-source.yml
```diff
@@ -0,0 +1,103 @@
...
+
+# runs on a push on master and at the end of every day
+on:
+  # triggering on push without branch name will run tests everytime
+  # there is a push on any branch
+  # turn it on only if needed
+  push:
+    branches:
+    - master
+    - github-actions2
```
nielsdrost, line 25: This can now be removed, I guess

nielsdrost (approved):
Looks good to me, thanks @valeriupredoi !

.github/workflows/action-install-from-conda.yml
valeriupredoi: I actually want to keep it and the branch up as well - saves us time in case we need to make any tweaks in the future

valeriupredoi:
cheers @nielsdrost :beer:

valeriupredoi:
@bouweandela u wanna have a looksee too? :+1:

valeriupredoi:
OK merged, cheers guys for the feedback! I am keeping the branch for easy testing w/o having to resort to test merges to `master` :beer:

PR #754, opened 2020-08-19
Fix units after derivation

### LINKED ISSUE #750: Wrong units after derivation of variable
After the derivation script of a variable is executed, the units of the resulting cube are set to the ones given by the CMOR table without converting the data / checking if the units are correct:

https://github.com/ESMValGroup/ESMValCore/blob/594cb65498f92e6c08586e86d05933538be09de1/esmvalcore/preprocessor/_derive/__init__.py#L106

This is particularly problematic if the units after the derivation of the variable depend on the project (e.g. CMIP5 vs. CMIP6) and is really hard to debug, since it does **not** result in an exception.

**Solution**

I will add a check if the units are correct, and if not convert the units.

@hb326

schlunma:
Adds checks if the units after executing a derivation script are compatible to the CMOR table and converts the units if necessary. Since not all derivation scripts have unit tests, I do not know if all of these still work. However, the changes in this PR only raise an exception if the units are not convertible (e.g. `m` and `kg`) and in this case the tool should definitely not silently set the units to the excepted ones like it does right now.

**Tasks**

-   [x] Add unit tests

* * *

Closes #750


bouweandela:
It would be good to test if the currently implemented derivation scripts actually work, instead of just adding some warnings and leaving the problems for our users. Maybe you could update the recipe `examples/recipe_preprocessor_derive_test.yml` so it includes all currently implemented derivation functions and then we can at least test it for for both a CMIP5 and a CMIP6 dataset.

schlunma:
Good idea! I totally forgot that this recipe existed. I will update it and test all variables for CMIP5 and CMIP6 models.

bouweandela:
@axel-lauer Do you think it would be possible to find someone to do a scientific review of this pull request? That would require running the recipe in https://github.com/ESMValGroup/ESMValTool/pull/1790 and having a look at the results.

It would be great if this could still be included in v2.1, but the deadline is already this Monday, so maybe it will have to wait for the next release.

schlunma:
Change the derivation of `vegfrac` and `ohc` accoring to [this](https://github.com/ESMValGroup/ESMValTool/pull/1790#issuecomment-702718719) comment, but did not test though.

bouweandela:
@axel-lauer Would you be able to do a final test of the changes you proposed, so we can still include this in v2.1? We're doing the feature freeze today, but since this is a bugfix, we could still include it if it gets approved soon.

axel-lauer:
I'll test the changes and report back soon...

axel-lauer:
There were two corrections introduced: och and vegfrac. The changes to ohc look good, the values for CMIP6 models are now in the expected range. vegfrac still looks weird (e.g. vegfrac=100% in Antarctica). This is not related to this PR or the corrections introduced but I think it would be good to get this fixed now.

vegfrac is currently calculated as 100.0 - baresoilfrac (in %)

The CMOR tables (CMIP5) say:
baresoilFrac = fraction of entire grid cell  that is covered by bare soil.
residualFrac = Fraction of Grid Cell that is Land but Neither Vegetation-Covered nor Bare Soil

From this, I would presume vegfrac *should* actually be calculated as follows:

vegfrac = 100.0 - baresoilfrac - residualfrac

@ledm , I think you introduced vegfrac. Can you confirm that this would be correct? Please let us know what you think s soon as possible.

axel-lauer:
@ledm do you hav an opinion on correcting vegfrac? If not, I propose to change the calculation of vegfrac to what I think is correct:

vegfrac = 100.0 - baresoilfrac - residualfrac

@schlunma could you implement this so we can get this merged?

schlunma:
Fixed `vegfrac` (I also adapted the recipe in ESMValGroup/ESMValTool#1790). The output looks better now, but there are weird values at the coast of Antarctica and Greenland:

vegfrac


While testing this I encountered an issue with the CMOR checker. `residualFrac` has the coordinate `typeresidual` which has no standard name in the [CMOR table](https://github.com/ESMValGroup/ESMValCore/blob/29611ac05a10408706ddb90381555b5bcd1c97d0/esmvalcore/cmor/tables/cmip6/Tables/CMIP6_coordinate.json#L2742-L2764), i.e. `''`. However, since `iris` gives `None` for an empty standard name, the checker tool fails with

```
type: standard_name should be , not None
```

I added a fix for this:

```diff
                 try:
                     cube_coord = self._cube.coord(var_name=coordinate.out_name)
-                    if cube_coord.standard_name != coordinate.standard_name:
+                    if (cube_coord.standard_name is None and
+                            coordinate.standard_name == ''):
+                        pass
+                    elif cube_coord.standard_name != coordinate.standard_name:
                         self.report_critical(
                             self._attr_msg,
                             coordinate.out_name,
```

This isn't really nice but does the job. The optimal solution here would be when `iris.cube.Cube.standard_name == ''` evaluates `True` when the standard name is not set (i.e. is `None`).

bouweandela:
> This isn't really nice but does the job. The optimal solution here would be when `iris.cube.Cube.standard_name == ''` evaluates `True` when the standard name is not set (i.e. is `None`).

I think the best solution would be to change the way we read the cmor tables so we read this in as `None` instead of as `''`, but let's do that for the next release.

@axel-lauer Would you be able to test the latest changes?

axel-lauer:
I tested the latest changes and it works. As @schlunma already noticed, however, there are spurious values along the coast lines. I presume these are caused by us assuming a land fraction of 100% for all land grid cells, i.e.:

vegfrac = 100% - baresoil - residualfrac

I think to get rid of those spurious coastal values, we would have to use the "real" land fraction and calculate vegfrac as:

vegfrac = landfrac - baresoil - residualfrac

Here is the "corrected" example from @schlunma. This now looks fine to me.

vegfrac

As "landfrac" I used the atmospheric fx file "sftlf". I guess to generalize this, a check would be needed if baresoil / residualfrac and sftlf are on the same grid.

axel-lauer:
Thanks @schlunma! I tested this and it works. I would leave it like this for now. Let's get this merged then!

valeriupredoi:
good stuff, guys! we have just cherry-picked this PR in the 2.1 release :beer:

Issue #505, opened 2020-02-20
Units reported in metadata.yml may differ from actual units


bouweandela:
Because the units reported in metadata.yml are read from the CMOR table, preprocessing operations that change the units of the data will make this information invalid.

Maybe we could just remove the units entry from metadata.yml/varname_info.ncl files, or we should read it from the cube before saving.

bouweandela:
@mattiarighi @schlunma Any idea if diagnostics actually use the units attribute from the interface instead of from the data?

schlunma:
> @mattiarighi @schlunma Any idea if diagnostics actually use the units attribute from the interface instead of from the data?

I have several examples where I check that both are equal (so it would be fine to just use once). I think it would be nice to drop the attributes in `metadata.yml`, but we need to check a lot of diagnostics for that.

Alternatively, we could set the attribute to `cube.units` in last preprocessor step (assuming that `cube.units` was modified correctly by the preprocessors).

mattiarighi:
I tried to grep quickly through the NCL scripts and there are a lot of occurrence of `@units`. My feeling is that this is read from the data, but it's hard to tell without looking at every script in detail.

Checking all the affected diags might be lot of work, it could be more effective: 
- to remove the units entry from metadata.yml/varname_info.ncl as suggested
- to run the affected recipes / diagnostics
- to fix the errors which may occur

Such a change would require testing all the affected recipes anyway.

schlunma:
Note that this might also apply to other attributes such as `short_name`, `long_name`, etc. as discussed in #759. A possible solution is given [here](https://github.com/ESMValGroup/ESMValCore/pull/759#issuecomment-706020763).

bouweandela:
@bettina-gier Since you reported this issue in #866, do you have an opinion on if having `units` in metadata.yml/var_info.ncl is a useful feature? I have a slight preference for just removing it, because the information is already available in the data.

bettina-gier:
In ncl it's nice because a lot of diagnostics set up their arrays to save data in before loading it, and then they can already set the units in the same spot. But it's not a problem to do it later and probably better to read it from the data as that is more likely to be correct in the current state.
So unless we make sure that all preprocessor functions would reflect correct units and add this to the units attribute in those files, it's better to remove them to not give misleading (or plain wrong) information.

schlunma:
[...]

PR #759, opened 2020-08-25
Add trend preprocessor

### LINKED ISSUE #757: Trend preprocessor
It would be nice to have a preprocessor that calculates the (linear) trend of a field along a given coordinate.

Yes, I will implement this.

schlunma:
Added preprocessor that calculates the linear trend of a dataset along a given coordinate.

**Tasks**

-   [x] Add unit tests

* * *

Closes #757.


esmvalcore/preprocessor/_trend.py
```diff
@@ -0,0 +1,181 @@
...
+    trend_arr = da.apply_along_axis(
+        _get_slope, axis, cube.data, coord.points, dtype=cube.dtype,
+        shape=())
+    trend_arr = da.ma.masked_invalid(trend_arr)
+
+    # Create dummy cube by collapsing along dimension and add trend data
+    aggregator = iris.analysis.Aggregator('trend', _remove_axis_np,
+                                          lazy_func=_remove_axis_da)
+    cube = cube.collapsed(coord, aggregator)
+    cube.data = trend_arr
```
bouweandela, line 120: Why are you not constructing the aggregator you actually need and then using that to compute the result?

doc/recipe/preprocessor.rst
```diff
@@ -1312,6 +1313,39 @@ use ``coords: [year, day_of_year]``.
 See also :func:`esmvalcore.preprocessor.amplitude`.
 
 
+.. _trend:
+
+Trend
+=====
+
+The ``_trend.py`` module contains the following preprocessor functions:
```
bouweandela, line 1321:
```suggestion
The trend module contains the following preprocessor functions:
```
It would be better not to mention private Python modules that only esmvalcore developers need to know about in documentation aimed at the general public (who may not even know Python, because they're developing an R or NCL diagnostic).

esmvalcore/preprocessor/_trend.py
schlunma: Good question -- I tried it first and saw that the `call_func` of [`iris.analysis.Aggregator`](https://scitools.org.uk/iris/docs/latest/iris/iris/analysis.html#iris.analysis.Aggregator) does only take the cube's `data` as argument, not the coordinate points which I need. Now I realize that I can probably provide the coordinate's points via `kwargs`. Let me check if that works.

bouweandela (approved):
Code looks good, though I did not check that the formulas were implemented correctly. @mattiarighi Could you please do a scientific review and run some tests?

ruthlorenz:
this would be super helpful for this recipe: https://github.com/ESMValGroup/ESMValTool/pull/1648
so I can test this.
The resulting unit is weird: 1.15740740740741e-05 s-1.K
Is this intended?

valeriupredoi:
K/s - that looks like a warming rate to me - close to the rate at which my living room warms up after I turn the heat on and the boiler hasn't run the whole summer :grin:

schlunma:
Awesome @ruthlorenz, thanks for testing!

Yes, the weird unit is intended (by the `cf_units` package): If you have something like `Unit('K')/Unit('day')` this is automatically converted to `some number K s-1`. If you want a nicer unit you can use the `convert_units` preprocessor.

ruthlorenz:
Great, thanks!
So I convert to K/years which I can compare to our version of ClimWip which does its own preprocessing (using xarray and scipy.stats.linregress for trend calculation)

so the preprocessor I used is:
```
annual_trend:
    regrid:
      target_grid: 2.5x2.5
      scheme: linear
    mask_landsea:
      mask_out: sea
    extract_region:
      start_longitude: 0.0
      end_longitude: 39.0
      start_latitude: 30.0
      end_latitude: 76.25
    annual_statistics:
      operator: mean
    linear_trend:
      coordinate: time
    convert_units:
      units: "K year-1"
```

Differences are small (something 1e-6) BUT they do not seem to be random.
[diff_ACCESS1-0_ClimWIP_ESMVal_TREND.pdf](https://github.com/ESMValGroup/ESMValCore/files/5341431/diff_ACCESS1-0_ClimWIP_ESMVal_TREND.pdf)
[diff_BNU-ESM_ClimWIP_ESMVal_TREND.pdf](https://github.com/ESMValGroup/ESMValCore/files/5341432/diff_BNU-ESM_ClimWIP_ESMVal_TREND.pdf)
[diff_CCSM4_ClimWIP_ESMVal_TREND.pdf](https://github.com/ESMValGroup/ESMValCore/files/5341434/diff_CCSM4_ClimWIP_ESMVal_TREND.pdf)

when I compare ClimWIP to CDO it looks random:
[diff_ACCESS1-0_ClimWIP_CDO_TREND.pdf](https://github.com/ESMValGroup/ESMValCore/files/5341438/diff_ACCESS1-0_ClimWIP_CDO_TREND.pdf)
while for ESMValTool there is some pattern:

schlunma:
That sounds weird. I compared the code to [`scipy.stats.linregress`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.linregress.html) and didn't find large discrepancies (note that using `linregress` directly in the code does not work due to issues with `dask`).

Are these absolute error or relative errors? I honestly do not know where this is coming from...

bouweandela:
Maybe you could have a look at the [CDO source code](https://code.mpimet.mpg.de/attachments/20826/cdo-1.9.8.tar.gz) to see how they calculate it, you can find the code in the file `src/Trend.cc`? I just had a quick look and my impression is that they have a different formula for computing the slope:
```python
xy_sum = (x_arr * y_arr).sum()
xx_sum = (x_arr * x_arr).sum()
y_sum = y.sum()
x_sum = x.sum()
x_mean = x.mean()
slope = (xy_sum - x_mean * y_sum) / (xx_sum - x_mean * x_sum)
```
This is also what is suggested as the correct formula by [this post](https://math.stackexchange.com/questions/204020/what-is-the-equation-used-to-calculate-a-linear-trendline)

schlunma:
> Maybe you could have a look at the [CDO source code](https://code.mpimet.mpg.de/attachments/20826/cdo-1.9.8.tar.gz) to see how they calculate it, you can find the code in the file `src/Trend.cc`? I just had a quick look and my impression is that they have a different formula for computing the slope:
> 
> ```python
> xy_sum = (x_arr * y_arr).sum()
> xx_sum = (x_arr * x_arr).sum()
> y_sum = y.sum()
> x_sum = x.sum()
> x_mean = x.mean()
> slope = (xy_sum - x_mean * y_sum) / (xx_sum - x_mean * x_sum)
> ```
> 
> This is also what is suggested as the correct formula by [this post](https://math.stackexchange.com/questions/204020/what-is-the-equation-used-to-calculate-a-linear-trendline)

This is mathematically similar to the current implementation (I just took it from [Wikiepedia](https://en.wikipedia.org/wiki/Simple_linear_regression)). I can implement this one and check if this works better, maybe there are numerical reasons for this discrepancy?

ruthlorenz:
there is a difference between estimating a liner trend for a timeseries (where x is time and NOT uncertain) versus estimating a trendline for some x,y data where x and y are uncertain. the wikipedia article seems to do the latter, while CDO does the timeseries.
Could that be the difference?

schlunma:
No, both formulas (Wikipedia and the CDO code that Bouwe) use ordinary least squares regression which assumes **no** error in X and i.i.d. errors in Y. I really think that these are numerical differences since mathematically these formulas are identical (I just checked that myself). I will implement the CDO formula and let you know when I'm done.

valeriupredoi:
LS method - straightforward and not involving any step sizes or number of iterations in the parameter space - this only means numerical differences resulting from either the machine or the numerical library or both, if you say they are order 1e-6 (what's the actual value of the slope?) that sounds to me this is the problem as @schlunma points out - note that these should be consistent if running on the same machine, if you switch machines they will differ

schlunma:
@ruthlorenz I changed the formula, can you test again? The tests still pass, so (at least for the test cases) the two formulas give identical results.

ruthlorenz:
will do, just nee to wait until another ESMVal recipe is finished (with different core). I let you know

valeriupredoi:
you guys not reading me comments - it's both machine precision (which is usually smaller than 1e-6) and numerical library diffs - CDO uses some C/Fortran(?) library, scipy another C one, they don't give identical results :beer:

schlunma:
> you guys not reading me comments - it's both machine precision (which is usually smaller than 1e-6) and numerical library diffs - CDO uses some C/Fortran(?) library, scipy another C one, they don't give identical results beer

I fully agree with you V., I also think these are numerical issues :smile:

valeriupredoi:
> > you guys not reading me comments - it's both machine precision (which is usually smaller than 1e-6) and numerical library diffs - CDO uses some C/Fortran(?) library, scipy another C one, they don't give identical results beer
> 
> I fully agree with you V., I also think these are numerical issues smile

finally, I get appreciation today :rofl: No but seriously, 1e-6 should be fine, but I still don't know what the absolute value of the `slope` is, if it's 1e-5 you guys are hosed :grin:

ruthlorenz:
thats why i uploaded the second figure with the trend itself ;-) luckily the absolute trend is larger, so all good.

one more thing, the preprocessor changes the variable name and adds _trend.
I understand why, but because other preprocessors do not do this, at least for our recipe where the idea is to use climatologies, trends, or standard deviations it will cause issues.

schlunma:
> one more thing, the preprocessor changes the variable name and adds _trend.
> I understand why, but because other preprocessors do not do this, at least for our recipe where the idea is to use climatologies, trends, or standard deviations it will cause issues.

Hmmm...yeah I can see that this is a problem. What is your opinion here @bouweandela @valeriupredoi? I thought changing the name here makes sense since in contrast to a simple mean we are actually calculating a new variable with new units.

bouweandela:
My feeling is that it makes sense to change the names, because you're also changing the units. However, it would be good to make sure that you then also change these things in metadata.yml, or things will become inconsistent, see #505. This should also be clearly documented.

It should be possible to write diagnostics in such a way that they are able to deal with different variables without too much effort for the diagnostics where it makes sense to do so.

bouweandela:
Regarding the precision issue, we could probably make it possible to run the tool in 64-bit floating point number precision by converting the data from 32-bit floats to 64-bit floats at load time and convert back to 32-bit at save time. That would take the data through all calculations with a higher precision. Recipes that need such high precision could then request this by providing a keyword argument to the preprocessor [`load`](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/api/esmvalcore.preprocessor.html#esmvalcore.preprocessor.load) function. This would come at the cost of slower computations and higher memory use of course.

schlunma:
> My feeling is that it makes sense to change the names, because you're also changing the units. However, it would be good to make sure that you then also change these things in metadata.yml, or things will become inconsistent, see #505. This should also be clearly documented.

That makes sense. Can you point me to the location where `metadata.yml` is written? We cannot change `metadata.yml` in the preprocessor itself since it only has access to `cube`. Would it maybe make sense to update `metadata.yml` in the `save` step with the actual metadata of the cubes so that this would be done automatically?

valeriupredoi:
> Regarding the precision issue, we could probably make it possible to run the tool in 64-bit floating point number precision by converting the data from 32-bit floats to 64-bit floats at load time and convert back to 32-bit at save time. That would take the data through all calculations with a higher precision. Recipes that need such high precision could then request this by providing a keyword argument to the preprocessor [`load`](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/api/esmvalcore.preprocessor.html#esmvalcore.preprocessor.load) function. This would come at the cost of slower computations and higher memory use of course.

I'll eat my boots if we can find any diagnostic needing 64-bit precision - if there is one it means we should probably talk to the CERN guys and ask them to stop trolling us :switzerland:

bouweandela:
Maybe you could update the `product.attributes` `dict` after this line of code?
https://github.com/ESMValGroup/ESMValCore/blob/29611ac05a10408706ddb90381555b5bcd1c97d0/esmvalcore/preprocessor/__init__.py#L296

Keys to be updated with the information from the cubes would be `short_name`, `standard_name`, `long_name` and `units`, but only if they are the same for all cubes.

esmvalcore/preprocessor/_trend.py
```diff
@@ -0,0 +1,185 @@
...
+        return trend_arr
+
+    aggregator = iris.analysis.Aggregator('trend', call_func,
+                                          lazy_func=lazy_func,
+                                          x_data=coord.points)
+    cube = cube.collapsed(coord, aggregator)
+
+    # Adapt metadata
+    if cube.var_name is not None:
+        cube.var_name += '_trend'
```
bouweandela, line 121: I'm not sure if an `_` is a valid character for a `short_name`, because that would break splitting the filenames on `_` to get the facets of the DRS.

schlunma:
I think I will simply not modify the names and leave this to another PR, as this is clearly out of scope of this PR. I will add a comment of #505 with that.

schlunma:
@ruthlorenz I removed the adaption of the variable names, so it shouldn't cause any issues anymore in your recipe!

schlunma:
@ruthlorenz If you are happy with these changes please approve them :+1:

ruthlorenz (approved):
now I also checked the formula for the standard error estimate, looks good to me.

Issue #345, opened 2019-10-30
Handling of time ranges (mostly in datafinder)


zklaus:
There are a couple of issues with the current handling of time ranges in the data finder:

- #311 deals with an emac requirement to have time ranges in the middle of the filename
- as we start dealing with sub-monthly frequencies we might need higher resolution than the current `start_year` `end_year` pair permits
- fixed frequency data (`fx` variables) don't have a time range
- #342 shows a case where ideally the time range for some datasets would be computed from the metadata of other datasets

The intuitive choice seems to be to specify time ranges as a single string item per dataset, and adopting the CMIP convention for the format (that has stayed consistent from CMIP5 to CMIP6) that is also used for obs4MIPs these days.

This could also open up the possibility of having special values like `corresponding` to determine the correct range from other datasets and would allow the explicit use of the `-clim` suffix that appears in climatological variables such as those in the `monC` table.

Challenges might include older obs4MIPs datasets that separate the two parts of their timerange with `_` instead of `-` and non-standard datasets.

So to sum up the idea is:

- remove `start_year` and `end_year` in favor of a `timerange` item in the dataset specs
- adopt [the CMIP convention for specifying time ranges (p 14-16)](http://goo.gl/v1drZl) including `-clim` suffix
- from this information build a `{timerange}` tag for use in the `config-developer.yml` file

What do you guys think?

bouweandela:
I think this would mostly be applicable to the way we specify times in the recipe. What do you think about this, @mattiarighi ?

mattiarighi:
Sounds convenient. 
How would the new dataset key look like?
```
timerange: 1996-2001
```
?

How many formats are we going to support?
I can think about the following ones:
```
timerange: 1996-2001
timerange: 199603-200110
timerange: 19960105-20010509
```

zklaus:
Ideally, I would like to go completely with CMIP, ie quoting from the abovementioned document:
> The <time_range> is a string generated consistent with the following:
If frequency = “fx” then
                  <time_range>=””
else
<time_range> = N1-N2[suffix] where N1 and N2 are integers of the form ‘yyyy[MM[dd[hh[mm[ss]]]]]’ (expressed as a string, where where ‘yyyy’, ‘MM’, ‘dd’, ‘hh’ ‘mm’ and ‘ss’ are integer year, month, day, hour, minute, and second, respectively)
endif
 
> where suffix is defined as follows:
[...]

PR #771, opened 2020-09-08
Find files for CMIP6 DCPP startdates

### LINKED ISSUE #632: Dealing with startdates in filenames
Some experiments (DCPP, etc) contain startdates in the filenames and this is currently not supported.

Started working in branch dev_dccp_startdates, in which startdates are dealt with the same way as ensemble members by getting expanded. 
The requirement for start_year and end_year is relaxed if the startdate is provided, because in this way the number of chunks per startdate can be flexible. I am not sure if there would be a way to determine how many chunks per startdate will there with parameters from the metadata?

sloosvel:

**Tasks**

-   [x] Add unit tests

* * *

A 'startdate' tag, that gets expanded in the same way as ensembles, can be added when calling some CMIP6 DCPP experiments that follow the scheme:

{short_name}_{mip}_{dataset}_{exp}_{startdate}-{ensemble}_{grid}*

The paths sometimes include the startdate, but in other cases they do not, so both cases are considered:

- ../{exp}/{startdate}-{ensemble}/... (as in Jasmin)
- ../{exp}/{ensemble}/... 

@jvegasbsc is there any other that should be considered?

Closes #632


jvegasbsc (approved):
Nice work! I just added a couple of tests for the replace_tags function

jvegasbsc:
@remi-kazeroni, can you please test it in DKRZ to see if we need some extra work to support it. If you or any other volunter don't have time  to test it, we can merge it so at least we can read DCPP data in Jasmin

remi-kazeroni:
> @remi-kazeroni, can you please test it in DKRZ to see if we need some extra work to support it. If you or any other volunter don't have time to test it, we can merge it so at least we can read DCPP data in Jasmin

Yes sure, I'll have a look at this PR this week

zklaus:
Note that what you call startdate here is actually part of the CMIP6 DRS as `sub_experiment`. See [CMIP6 Global Attributes, DRS, Filenames, Directory Structure, and CV’s](http://goo.gl/v1drZl), particularly p. 13, DRS components and p. 14, filename template. This also makes reference to the applicable CV in [CMIP6_sub_experiment.json](https://github.com/WCRP-CMIP/CMIP6_CVs/blob/master/CMIP6_sub_experiment_id.json).

Perhaps it would be prudent to adopt the same terminology. That would also open the functionality up to other MIPs that might have sub-experiments in the future.

The first linked document also tells us what the directory structure should be (p.17, directory structure template), but of course we can't force every data center to conform.

remi-kazeroni (approved):
Nice work! I have tried on various combinations of datasets with and without `sub_experiments` and all of them can be read as expected. I just have a few general comments, feel free to ignore if not relevant:

1. Do you think it would be useful to document these changes? The `sub_experiment` key could be added [here](https://github.com/ESMValGroup/ESMValCore/blob/master/doc/recipe/overview.rst#recipe-section-datasets)
2. The `start_year` and `end_year` are not needed when a `sub_experiment` is provided in the recipe. Instead, the decadal time bounds are used (e.g. `start_year`: 2000 and `end_year`: 2010 for `sub_experiment`: s2000-r1i1p1f1). One can still consider a shorter time range (e.g. 2005-2008) but in this case filenames contain the whole decadal time range instead of the shorter range (e.g. `plots/*_dcppA-hindcast_s2000-r1i1p1f1_tas_2000-2010.png` and not  `plots/*_dcppA-hindcast_s2000-r1i1p1f1_tas_2005-2008.png`). Is that expected? Perhaps this example is not relevant in practice...

jvegasbsc:
> The start_year and end_year are not needed when a sub_experiment is provided in the recipe. Instead, the decadal time bounds are used (e.g. start_year: 2000 and end_year: 2010 for sub_experiment: s2000-r1i1p1f1). One can still consider a shorter time range (e.g. 2005-2008) but in this case filenames contain the whole decadal time range instead of the shorter range (e.g. plots/*_dcppA-hindcast_s2000-r1i1p1f1_tas_2000-2010.png and not plots/*_dcppA-hindcast_s2000-r1i1p1f1_tas_2005-2008.png). Is that expected? Perhaps this example is not relevant in practice...

Without changing the way we manage the time range, handling the start and end year will be a nightmare for the full hindcasts, as we will have to specify it manually for all years from 1979 to 2010 or so. The ability to use `s(1979:2010)` will be useless. 

We have a pending issue to deal with the time range #345 and I remember another suggestion asking for a way yo specify relative ranges (first 20 years, last 50, all available...) I think we should do this rework in a separated pull request.

zklaus:
The sub-experiment should really be completely independent from the startrange and endrange. It only tells you something about the initialization of the experiment. Every run is 10 years long, but there is a new run started for every year. This means that there are many different sub-experiments covering the same years and comparing these is one of the main interests of the DCPP mip.

Consequently, we should expect it to be common to find dataset blocks like
```yaml
datasets:
  - {sub-experiment: s1960, start_year: 1965, end_year: 1970}
  - {sub-experiment: s1962, start_year: 1965, end_year: 1970}
  - {sub-experiment: s1963, start_year: 1965, end_year: 1970}
  - {sub-experiment: s1964, start_year: 1965, end_year: 1970}
```

sloosvel:
>The sub-experiment should really be completely independent from the startrange and endrange.

I think it already is. This finds and loads all the files available for a certain sub-experiment value. And later it adds the overall `start_year` and `end_year`. It's true that you cannot subset it as @remi-kazeroni says, but I think people normally load all years available anyway.

>Consequently, we should expect it to be common to find dataset blocks like

datasets:
  - {sub-experiment: s1960, start_year: 1965, end_year: 1970}
  - {sub-experiment: s1962, start_year: 1965, end_year: 1970}
  - {sub-experiment: s1963, start_year: 1965, end_year: 1970}
  - {sub-experiment: s1964, start_year: 1965, end_year: 1970}
 
I am a bit lost because that's what I was trying to avoid: having 50 lines in a recipe to call a single experiment with multiple subexperiments.

zklaus:
> I am a bit lost because that's what I was trying to avoid: having 50 lines in a recipe to call a single experiment with multiple subexperiments.

Why?

sloosvel:
Because everything belongs to the same dataset-experiment pair, which is what every new line in a recipe represents. It also makes the recipe easier to read. 

And the end goal will be to be able to deal with all those sub-experiments  as one in a single cube with a new dimension. So it's like a single dataset.

zklaus:
I think in the context of the DCPP mip, every sub-experiment should be seen as its own experiment. You don't always (or even generally) want to look at all sub-experiments at the same time.

remi-kazeroni:
> datasets:
> 
>     * {sub-experiment: s1960, start_year: 1965, end_year: 1970}
> 
>     * {sub-experiment: s1962, start_year: 1965, end_year: 1970}
> 
>     * {sub-experiment: s1963, start_year: 1965, end_year: 1970}
> 
>     * {sub-experiment: s1964, start_year: 1965, end_year: 1970}
> 
> 
> I am a bit lost because that's what I was trying to avoid: having 50 lines in a recipe to call a single experiment with multiple subexperiments.

But I think this can already be done in one line with this PR using: ` {sub-experiment: s(1960:1964), start_year: 1965, end_year: 1970}` (assuming 1961 is used as well). 
I tested that with the `examples/recipe_python.yml` and only the data between `start_year` and `end_year` are used and plotted. But the filenames in the `plots` dir are in the form `*dcppA-hindcast_s1964-r1i1p1f1_tas_1964-1974.png` while one would expect `*dcppA-hindcast_s1964-r1i1p1f1_tas_1965-1970.png`

zklaus:
Sure, there are many ways to specify these. The point is neither should we manipulate start and end because of sub-experiment, nor sub-experiment because of start and end.

sloosvel:
So which changes in this pull request do you not approve of?

esmvalcore/_data_finder.py
```diff
@@ -229,6 +236,17 @@ def _get_filenames_glob(variable, drs):
     return filenames_glob
 
 
+def _update_output_file(variable, files):
+    intervals = [get_start_end_year(name) for name in files]
+    variable.update({'start_year': min(intervals)[0]})
+    variable.update({'end_year': max(intervals)[1]})
+    filename = variable['filename'].replace(
+        '.nc', '_{start_year}-{end_year}.nc'.format(**variable)
+    )
```
zklaus, line 245: Why are these manipulations necessary?

esmvalcore/_data_finder.py
```diff
@@ -268,7 +288,7 @@ def get_output_file(variable, preproc_dir):
         variable['variable_group'],
         _replace_tags(cfg['output_file'], variable)[0],
     )
-    if variable['frequency'] != 'fx':
+    if variable['frequency'] != 'fx' and 'sub_experiment' not in variable:
```
zklaus, line 291: Imho, `sub_experiments` still need start and end year.

esmvalcore/_recipe.py
```diff
@@ -1074,8 +1075,19 @@ def _initialize_variables(self, raw_variable, raw_datasets):
...
-            check.variable(variable, required_keys)
-        variables = self._expand_ensemble(variables)
+            if 'sub_experiment' in variable:
+                subexperiment_keys = deepcopy(required_keys)
+                subexperiment_keys.update({'sub_experiment'})
+                try:
+                    subexperiment_keys.remove('start_year')
+                    subexperiment_keys.remove('end_year')
+                except KeyError:
+                    continue
```
zklaus, line 1085: Why should be remove this? (Side note: [`discard`](https://docs.python.org/3/library/stdtypes.html#frozenset.discard) might be preferable to `remove`.)

esmvalcore/_recipe_checks.py
```diff
@@ -127,7 +127,7 @@ def data_availability(input_files, var, dirnames, filenames):
             f"Missing data for {var['alias']}: {var['short_name']}")
 
     # check time avail only for non-fx variables
-    if var['frequency'] == 'fx':
+    if var['frequency'] == 'fx' or 'sub_experiment' in var:
```
zklaus, line 130: I think `sub_experiments` should still require start and end year.

zklaus (changes requested):

zklaus:
> So which changes in this pull request do you not approve of?

In the review above I tried to mark all the places that this touches. All of those remarks really are about the same issue.

sloosvel:
Good to know! I'll try to see what can be done about them.

zklaus:
There is one comment from the previous review left to address.

Now, this PR seems to introduce new functionality in terms of `all_years`. That appears to be completely unrelated to DCPP, so I suggest putting it into a different PR. That way we will be able to wrap this one up more quickly and the functionality (that will be appreciated by others for other experiments!) will get better visibility.

sloosvel:
>  That appears to be completely unrelated to DCPP,

The DCPP functionality that we need at the department requires it. This pull request without it does not help much our use case. 

Would it be that bad to include it as a first step, and then refine it for other experiments in another pull request?

esmvalcore/_data_finder.py
```diff
@@ -268,7 +276,7 @@ def get_output_file(variable, preproc_dir):
         variable['variable_group'],
         _replace_tags(cfg['output_file'], variable)[0],
     )
-    if variable['frequency'] != 'fx':
+    if variable['frequency'] != 'fx' not in variable:
```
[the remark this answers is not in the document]
sloosvel: No, no, my bad. But this appears as outdated to me.

zklaus:
Ok to merge for me. Nice work!

sloosvel:
Great! @ESMValGroup/esmvaltool-coreteam anyone with the time to merge it?

valeriupredoi:
good stuff @sloosvel :beer:

Issue #724, opened 2020-07-21
Regrid hangs for ocean variable sea water salinity for standard 2x2 grid


valeriupredoi:
```bash
2020-07-21 16:22:42,761 UTC [7554] DEBUG   Running regrid(sea_water_salinity                         (time: 252; depth: 75; cell index along second dimension: 330; cell index along first dimension: 360)
     Dimension coordinates:
          time                                  x           -                                      -                                      -
          depth                                 -           x                                      -                                      -
          cell index along second dimension     -           -                                      x                                      -
          cell index along first dimension      -           -                                      -                                      x
     Auxiliary coordinates:
          latitude                              -           -                                      x                                      x
          longitude                             -           -                                      x                                      x
     Attributes:
          Conventions: CF-1.7 CMIP-6.2
          activity_id: CMIP
          branch_method: standard
          branch_time_in_child: 0.0
          branch_time_in_parent: 144000.0
          cmor_version: 3.4.0
          comment: mo: This variable is reported using a z* coordinate system rather than...
          cv_version: 6.2.20.1
          data_specs_version: 01.00.29
          experiment: all-forcing simulation of the recent past
          experiment_id: historical
          external_variables: areacello volcello
          forcing_index: 2
          frequency: mon
          further_info_url: https://furtherinfo.es-doc.org/CMIP6.MOHC.UKESM1-0-LL.historical.none....
          grid: Native eORCA1 tripolar primarily 1 deg with meridional refinement down...
          grid_label: gn
          initialization_index: 1
          institution: Met Office Hadley Centre, Fitzroy Road, Exeter, Devon, EX1 3PB, UK
          institution_id: MOHC
          license: CMIP6 model data produced by the Met Office Hadley Centre is licensed under...
          mip_era: CMIP6
          mo_runid: u-bc179
          nominal_resolution: 100 km
          original_name: mo: (variable_name: so)
          original_units: 1e-3
          parent_activity_id: CMIP
          parent_experiment_id: piControl
          parent_mip_era: CMIP6
          parent_source_id: UKESM1-0-LL
          parent_time_units: days since 1850-1-1 00:00:00
          parent_variant_label: r1i1p1f2
          physics_index: 1
          product: model-output
          realization_index: 1
          realm: ocean
          source: UKESM1.0-LL (2018): 
aerosol: UKCA-GLOMAP-mode
atmos: MetUM-HadGEM3-GA7.1...
          source_id: UKESM1-0-LL
[...]

PR #773, opened 2020-09-09
Introduce a parallel loop to do esmpy regridding for many levels

valeriupredoi:

**Tasks**

-   [ ] Add unit tests

* * *

Closes #issue_number #724 

This PR introduces a parallel loop that assembles the esmpy regridders from multiple processes, this is needed because in the case of eg 75 levels one needs to wait forever and a half. This speeds up the regridder assembly from 400s to 1s :beer:

### DIFF
--- a/esmvalcore/preprocessor/_regrid_esmpy.py
+++ b/esmvalcore/preprocessor/_regrid_esmpy.py
@@ -3,7 +3,6 @@
 
 import concurrent.futures
 import itertools
-import time as tr
 
 import ESMF
 import iris
@@ -227,7 +226,6 @@ def build_regridder_3d(src_rep, dst_rep, regrid_method, mask_threshold):
 
 def build_regridder(src_rep, dst_rep, method, mask_threshold=.99):
     """Build regridders from representants."""
-    t1 = tr.time()
     regrid_method = ESMF_REGRID_METHODS[method]
     if src_rep.ndim == 2:
         regridder = build_regridder_2d((src_rep, dst_rep), regrid_method,
@@ -235,8 +233,6 @@ def build_regridder(src_rep, dst_rep, method, mask_threshold=.99):
     elif src_rep.ndim == 3:
         regridder = build_regridder_3d(src_rep, dst_rep, regrid_method,
                                        mask_threshold)
-    t2 = tr.time()
-    print("ESMF REGRIDDER TIME for DIM", str(src_rep.ndim), t2 - t1)
     return regridder
 
 
--- a/esmvalcore/preprocessor/_regrid_esmpy.py
+++ b/esmvalcore/preprocessor/_regrid_esmpy.py
@@ -210,7 +210,7 @@ def build_regridder_3d(src_rep, dst_rep, regrid_method, mask_threshold):
                                       iter_pack,
                                       itertools.repeat(regrid_method),
                                       itertools.repeat(mask_threshold),
-                                      chunksize=100000000):
+                                      chunksize=100):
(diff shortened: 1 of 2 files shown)


valeriupredoi:
OK found a bug - was printing a list from a generator (and that was eliminating the generator, no wonder that was so fast hahah). Here's a dilemma: even if this is faster by a factor of 5-6 for eg 8 processes, this means we can't use `max_parallel_tasks` more than 1 since daemonic processes can't have children (I love this error message :grin: ) so we'll take a serious hit with the other tasks that will have to stay serial; any way the 3d regridding can be made faster w/o parallelization @zklaus ? :beer:

valeriupredoi:
OK so the reason why this is still a Draft is that it doesn't actually work - the parallel loop stalls and I've managed to identify where it stalls:

```python
def build_regridder_2d(iter_pack, regrid_method, mask_threshold):
    """Build regridder for 2d regridding."""
    src_rep, dst_rep = iter_pack
    dst_field = cube_to_empty_field(dst_rep)
    src_field = cube_to_empty_field(src_rep)
    regridding_arguments = {
        'srcfield': src_field,
        'dstfield': dst_field,
        'regrid_method': regrid_method,
        'unmapped_action': ESMF.UnmappedAction.IGNORE,
        'ignore_degenerate': True,
    }
```
-> this block runs fine in parallel!
```python
    if np.ma.is_masked(src_rep.data):
        src_field.data[...] = ~src_rep.data.mask.T
        src_mask = src_field.grid.get_item(ESMF.GridItem.MASK,
                                           ESMF.StaggerLoc.CENTER)
        src_mask[...] = src_rep.data.mask.T
        center_mask = dst_field.grid.get_item(ESMF.GridItem.MASK,
                                              ESMF.StaggerLoc.CENTER)
        center_mask[...] = 0
        mask_regridder = ESMF.Regrid(
            src_mask_values=MASK_REGRIDDING_MASK_VALUE[regrid_method],
            dst_mask_values=np.array([]),
            **regridding_arguments)
        regr_field = mask_regridder(src_field, dst_field)
        dst_mask = regr_field.data[...].T < mask_threshold
        center_mask[...] = dst_mask.T
    else:
        dst_mask = False
```
-> there is something in this block that triggers a wait/acquire call that stalls all the processes; if you comment it out (well, of course it's silly, but for prototyping purposes) it goes on and builds the `field_regridder` fine, only to stop at:
```
    def regridder(src):
        """Regrid 2d for irregular grids."""
        res = get_empty_data(dst_rep.shape, src.dtype)
        data = src.data
        if np.ma.is_masked(data):
            data = data.data
        src_field.data[...] = data.T
        regr_field = field_regridder(src_field, dst_field)
        res.data[...] = regr_field.data[...].T
        res.mask[...] = dst_mask
        return res
```
with reducing (pickling) issues - this is expected since it's very hard to pickle a function within a function - if that is converted to normal excution block rather than a function all goes through nicely in parallel. Thing is I am not confident I can make the necessary code changes since I am not that familiar with the esmpy regridding, that's why @bouweandela @zklaus and meself we should probably have a chat about it :beer:

valeriupredoi:
OK I've dug even deeper in this - things like 2dim `lat` objects are hard to pass around to processes eg `lat.ndim` or `lat.points` are fine and accessed instantly by the child process but stuff like `lat.bounds` result in deadlocks; no such problems with 1dim coords. Hmmm this is harder to parallelize than I first thought it'd be :man_facepalming:

valeriupredoi:
this is not working and is a bit of a dead end so let's keep the discussion going in #775

Issue #775, opened 2020-09-10
Optimization needed for regrid (including esmpy) and level selection


valeriupredoi:
This is an amalgamation of multiple issues that were open at separate times, initially believed to be caused by other problems; issues closed and squashed in here #724 #774 and draft PR #773 ; in short:

- regridding large 3d datasets (like ocean data) takes a long time; this is not due to poor implementation but rather it is the limit of any serial run w/o parallelization; @bouweandela has implemented lazy regridding that will be available in `iris` (well done man!) but that will not make things much faster if we don't parallelize the regridding at least on vertical levels;
- level selection currently consumes a lot of memory - it is not lazy and is done via loading the full data object into memory several times, as many levels there are;

I think we need to start thinking about this and talk about it; I have already tried to parallelize the esmpy 3d regridding but it's difficult due to the amount of complexity that needs to be passed to the parallel process. Anyway, I will have a look at the level selection imminently, but for now this is a placeholder for @rswamina @omeuriot @ledm and others working on ocean data that things will improve, just bear with us :beer:

valeriupredoi:
@bouweandela `_vertical_interpolate()` is not:
```python
    if np.any(mask):
        # Ensure that the data is masked appropriately.
        new_data = np.ma.array(new_data, mask=mask, fill_value=_MDI)

    # Construct the resulting cube with the interpolated data.
    return _create_cube(cube, new_data, src_levels, levels.astype(float))
```

valeriupredoi:
also that function keeps two copies of the data from `cube.data` and `new_data` which are of similar sizes and dereferencing happens only at the very end so they stay in memory

valeriupredoi:
OK partially addressed memory issue from `vertical_interpolate` in #776 (saving about 30% memory) but the main problem is `stratify` gets all the stuff done unlazied, did we have a fix to remove stratify? :beer:

bouweandela:
Some time ago, I tried to use ESMPy with the function [dask.array.map_blocks](https://docs.dask.org/en/latest/array-api.html#dask.array.map_blocks) to make regridding for grids with 2D latitude/longitude coordinates lazy, but the problem is that ESMPy cannot be used from multiple Python threads.

Here is a fairly minimal example demonstrating the problem:
```python
import threading

import ESMF
import numpy as np


def run():
    ESMF.Grid(np.array((10, 20)),
              num_peri_dims=1,
              staggerloc=[ESMF.StaggerLoc.CENTER])


[...]

PR #776, opened 2020-09-11
Use dask to reduce memory consumption of extract_levels for masked data

valeriupredoi:

**Tasks**

-   [ ] Add unit tests

* * *

**partially addressing** #775 

Mostly appreciated by ocean people, I noticed a memory consumption reduction by about 30% for level selection.

### DIFF
--- a/esmvalcore/preprocessor/_regrid.py
+++ b/esmvalcore/preprocessor/_regrid.py
@@ -436,7 +436,7 @@ def _vertical_interpolate(cube, src_levels, levels, interpolation,
     # force mask onto data as nan's
     if da.ma.getmaskarray(cube.core_data()).any():
         if cube.has_lazy_data():
-            cube.core_data()[da.ma.getmaskarray(cube.core_data())] = np.nan
+            cube.data = da.ma.filled(cube.core_data(), np.nan)
         else:
             cube.data[cube.data.mask] = np.nan
 
--- a/esmvalcore/preprocessor/_regrid.py
+++ b/esmvalcore/preprocessor/_regrid.py
@@ -434,11 +434,7 @@ def _vertical_interpolate(cube, src_levels, levels, interpolation,
         src_levels.points, cube.shape, cube.coord_dims(src_levels))
 
     # force mask onto data as nan's
-    if da.ma.getmaskarray(cube.core_data()).any():
-        if cube.has_lazy_data():
-            cube.data = da.ma.filled(cube.core_data(), np.nan)
-        else:
-            cube.data[cube.data.mask] = np.nan
+    cube.data = da.ma.filled(cube.core_data(), np.nan)
 
     # Now perform the actual vertical interpolation.
     new_data = stratify.interpolate(
--- a/esmvalcore/preprocessor/_regrid.py
+++ b/esmvalcore/preprocessor/_regrid.py
@@ -435,7 +435,7 @@ def _vertical_interpolate(cube, src_levels, levels, interpolation,
 
     # force mask onto data as nan's
     if da.ma.getmaskarray(cube.core_data()).any():
-        cube.core_data()[da.ma.getmaskarray(cube.core_data())] = np.nan
+        cube.core_data()[tuple(da.ma.getmaskarray(cube.core_data()))] = np.nan
 
     # Now perform the actual vertical interpolation.
     new_data = stratify.interpolate(
--- a/esmvalcore/preprocessor/_regrid.py
+++ b/esmvalcore/preprocessor/_regrid.py
@@ -435,7 +435,10 @@ def _vertical_interpolate(cube, src_levels, levels, interpolation,
 
     # force mask onto data as nan's
     if da.ma.getmaskarray(cube.core_data()).any():
-        cube.core_data()[da.ma.getmaskarray(cube.core_data())] = np.nan
+        if cube.has_lazy_data():
+            cube.core_data()[da.ma.getmaskarray(cube.core_data())] = np.nan
+        else:
+            cube.data[cube.data.mask] = np.nan
 
     # Now perform the actual vertical interpolation.
     new_data = stratify.interpolate(
--- a/esmvalcore/preprocessor/_regrid.py
+++ b/esmvalcore/preprocessor/_regrid.py
@@ -7,6 +7,7 @@ from copy import deepcopy
 import numpy as np
 import stratify
 import iris
+from dask import array as da
 from iris.analysis import AreaWeighted, Linear, Nearest, UnstructuredNearest
 from iris.util import broadcast_to_shape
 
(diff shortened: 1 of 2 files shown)


valeriupredoi:
can one of you @bouweandela @jvegasbsc @schlunma pls review this, it'd be useful to have it in the release :+1: :beer:

bouweandela (changes requested):
Could you add a test like this one, but with lazy input and output data?
https://github.com/ESMValGroup/ESMValCore/blob/a6d4bbcc887057dc1de1eea2871727e69b6c2c58/tests/integration/preprocessor/_regrid/test_extract_levels.py#L48-L57

bouweandela:
Oh wait, that won't work, because this doesn't make the process completely lazy, right?

esmvalcore/preprocessor/_regrid.py
```diff
@@ -433,14 +434,17 @@ def _vertical_interpolate(cube, src_levels, levels, interpolation,
         src_levels.points, cube.shape, cube.coord_dims(src_levels))
 
     # force mask onto data as nan's
-    if np.ma.is_masked(cube.data):
-        cube.data[cube.data.mask] = np.nan
+    if da.ma.getmaskarray(cube.core_data()).any():
+        if cube.has_lazy_data():
+            cube.core_data()[da.ma.getmaskarray(cube.core_data())] = np.nan
+        else:
+            cube.data[cube.data.mask] = np.nan
```
bouweandela, line 441: Maybe this could be simplified to
```python
cube.data = da.ma.filled(cube.core_data(), np.nan)
```
by using [dask.array.ma.filled](https://docs.dask.org/en/latest/array-api.html#dask.array.ma.filled)?

valeriupredoi:
> Oh wait, that won't work, because this doesn't make the process completely lazy, right?

yer, it's a half-mule process due to vinterp not returning a lazy object

esmvalcore/preprocessor/_regrid.py
valeriupredoi: interesting, didn't know about this - I like it! Let me try it out :beer:

esmvalcore/preprocessor/_regrid.py
```diff
@@ -433,14 +434,17 @@ def _vertical_interpolate(cube, src_levels, levels, interpolation,
...
-    if np.ma.is_masked(cube.data):
-        cube.data[cube.data.mask] = np.nan
+    if da.ma.getmaskarray(cube.core_data()).any():
+        if cube.has_lazy_data():
+            cube.data = da.ma.filled(cube.core_data(), np.nan)
+        else:
+            cube.data[cube.data.mask] = np.nan
 
     # Now perform the actual vertical interpolation.
     new_data = stratify.interpolate(
```
schlunma, line 444: Does `stratify.interpolate` work with lazy data (i.e. when given a `dask` array instead of a `numpy` array)? I used it some time ago and I think it always realized the data. If that is the case I think replacing `cube.data` with `cube.core_data()` has no effect.

valeriupredoi: it can ingest either - lazy or not - but it realizes the data inside its bellows - you are right about that! But there is improvement in memory from constructing the nan mask with lazy data and that is visible (30% improvement overall in function) - not bad for cases of big data, also the big impovement will be from the lazy implement of stratify but that's above my paygrade :grin:

schlunma: Awesome, then I am happy with it :smile:

schlunma (approved):

esmvalcore/preprocessor/_regrid.py
valeriupredoi: danke schoen :beer:

valeriupredoi:
@bouweandela :+1: or :-1:

esmvalcore/preprocessor/_regrid.py
bouweandela: What I meant is that you can replace the whole block, there should be no need to check if the data is masked first, the function will do that for you, according to its documentation

bouweandela:
> @bouweandela +1 or -1

Please see https://github.com/ESMValGroup/ESMValCore/pull/776#discussion_r498755758

esmvalcore/preprocessor/_regrid.py
valeriupredoi: you need to check for the existence of a mask for the realized, non-lazy case because otherwise it will fail with AttributeError, man

bouweandela: Can you implement a unit test that shows that? I tried changing it and the unit tests in pass just fine

valeriupredoi: well that's because there is a test I already added. You need :eyeglasses: B-man? :grin:

bouweandela: That test passes just fine with the code change I'm proposing, no AttributeError occurs?

valeriupredoi: Bouwe man, can you post the code you wrote here, we may be talking about the same thing but thinking otherwise

esmvalcore/preprocessor/_regrid.py
```diff
@@ -433,14 +434,17 @@ def _vertical_interpolate(cube, src_levels, levels, interpolation,
         src_levels.points, cube.shape, cube.coord_dims(src_levels))
 
     # force mask onto data as nan's
-    if np.ma.is_masked(cube.data):
-        cube.data[cube.data.mask] = np.nan
+    if da.ma.getmaskarray(cube.core_data()).any():
+        if cube.has_lazy_data():
+            cube.data = da.ma.filled(cube.core_data(), np.nan)
+        else:
+            cube.data[cube.data.mask] = np.nan
```
bouweandela, line 441:
```suggestion
cube.data = da.ma.filled(cube.core_data(), np.nan)
```
This is what I mean, it looks like this works and I think it's better because we leave all the if/else checks to `da.ma.filled`

valeriupredoi: ahaa! now I get it - didn't know that thing was powerful enough to account for the realized data and fill it with nans if needed, cheers dude!

tests/unit/preprocessor/_regrid/test_extract_levels.py
```diff
@@ -145,8 +149,15 @@ def test_interpolation__masked(self):
         masked = ma.empty(self.shape)
         masked.mask = mask
         cube = _make_cube(masked, dtype=self.dtype)
+        # save cube to test the lazy data interpolation too
+        iris.save(cube, self.filename)
         with mock.patch(
                 'stratify.interpolate', return_value=new_data) as mocker:
+            # first test lazy
+            loaded_cube = iris.load_cube(self.filename)
```
bouweandela, line 157: To make a cube with realized data lazy, you can do
```
cube.data = cube.lazy_data()
```
or
```
cube.data = da.asarray(cube.data, chunks=(1, 2, 3, 4))
```
if you want more control over how the lazy array is created. There is no need to first save to file.

valeriupredoi: cool! I didn't know that, but I wanted to replicate the actual conditions the function runs in anyway

bouweandela (approved):

Issue #498, opened 2020-02-19
ESMValCore (and ESMValTool) Python/Jupyter Notebook API


bouweandela:
For interactive data exploration, it would be great if we could make it possible to use ESMValCore and the diagnostics included in ESMValTool accessible from a Jupyter notebook.

Here is an example of what such a notebook could look like:  [ESMValTool Notebook API example.ipynb](https://gist.github.com/bouweandela/c60fb23ad84bcc1569237d79608e9c02)

And here is a rough mapping from current functionality to the proposed Python API
- `Config` -  `esmvalcore._config.read_config_user_file`/`esmvalcore._config.CFG`
- `Variable` -  `esmvalcore.cmor.table.VariableInfo`
- `Dataset` -  `esmvalcore.preprocessor.PreprocessorFile`
- `Dataset.find_files` - `esmvalcore._data_finder.get_input_filelist`
- `Dataset.filename` - `esmvalcore._data_finder.get_output_file`
- `DatasetList` -  `esmvalcore.preprocessor.PreprocessingTask`
- `Diagnostic` -  `esmvalcore._task.DiagnosticTask`
- `compute` - `esmvalcore._task.run_tasks`
- `load` -  `esmvalcore._recipe.read_recipe_file`

@ESMValGroup/esmvaltool-coreteam Any comments?

Peter9192:
This would also be nice:
`DatasetList.select(model='EC-EARTH', experiment='rcp85')`

zklaus:
This looks very nice!

I wonder how strong the notebook link really is. It looks like a [dataclasses](https://docs.python.org/3/library/dataclasses.html) or [attrs](https://www.attrs.org) inspired interface that could be nice for us to have with or without notebooks, no?

For the notebook part, could you put a pointer on how this connects to any jupyter magic? Say inline dataset printing and plotting like iris cubes or similar?

bouweandela:
Related to #30

bouweandela:
@zklaus Indeed there is nothing specific to a Jupyter notebook in this proposal, you could just as well use the interface from a Python script, I changed the title accordingly. The reason I mentioned the Jupyter notebook is that people typically use this kind of interface from a notebook.

stefsmeets:
[...]

PR #785, opened 2020-09-16
Implement importable config object

stefsmeets:
Hi all, we started working on an importable config object. This PR can be used to track the progress. For more discussion see this issue: https://github.com/ESMValGroup/ESMValCore/issues/498

**Tasks**

-   [X] [Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do, if you haven't done so already (and add the link at the bottom)
-   [x] Add unit tests
~~-   [ ] If writing a new/modified preprocessor function, please update the [documentation](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/preprocessor.html)~~
~~-   [ ] If you make backward incompatible changes to the recipe format, make a new pull request in the [ESMValTool repository](https://github.com/ESMValGroup/ESMValTool) and add the link below~~

The development in this PR is quite complete. Tests for all new functionality has been added. Here is a list of things still to do:

**To do**

- [ ] Documentation
- [ ] Merge master 😬
- [ ] Fix Codacy errors
- [ ] Fix CI errors
- [ ] Check `config-default.yml` for which variables we want to keep (would be nice if someone could have a look at this)
- [ ] Consider renaming `config_user` variable to `session` to fit the config API


stefsmeets:
```python
from esmvalcore import config

# optional, this file is loaded by default
config.load('~/.esmvaltool/config-user.yml')

config.plot_dir
>> PosixPath('/home/stef/r/esmvalcore/esmvaltool_output/session_20200916_095523/plots')
```

valeriupredoi:
careful with this - heavy object construction may lead to the objects being un-picklable and the mpi will either have a hard time sending data to processes or it will just not be able to pickle - configuration is done only once but if the `config` object is sent over to multiple processes and accessed by them it's definitely something to be careful with (a dictionary is easier sent than a class)

stefsmeets:
Thanks for the heads-up @valeriupredoi , that is good to keep in mind. I'm not too worried about that at the moment, because most of the state is stored as a dictionary and many python objects are easily pickled/unpickled.

valeriupredoi:
yeah sure - me neither, but just something we should keep in mind before we make too many objects and too many recursive calls, I have just went through the hoops with another code that was 100% recursive objects that was impossible to parallelize w/o major code changes :grin:

stefsmeets:
Just an update on this, I have been looking at the matplotlib source code to get some idea of how to approach this. Essentially, they define their config (`rcParams`) as follows:

```python
configDefault = load_config_from_file(`default_config.yaml`)
configDefault.update(_hardcoded_defaults)

config = Config()  # first initialize as an empty mapping
config.update(configDefault)
config.update(load_config_from_file(`config.yaml`)

configOrig= rcParams.copy()  # keep a reserve copy
```

Config is defined as a MutableMapping (https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/__init__.py#L639):

```python
class Config(MutableMapping, dict):  # mapping to the configuration file.
    validators = _validators
    def __setitem__(self, item)  # validate items on the way in
    def __getitem__(self, item)  # return items from the mapping

def load_config_from_file(path='path_to.yaml') -> Config:
    ...
```

Each item is validated through `__setitem__`, where `_validators` is a look-up table with the function names to validate (https://github.com/matplotlib/matplotlib/blob/a4128a71374a51eee6a2fac226e3ac09fcc2acd9/lib/matplotlib/rcsetup.py#L1008):

```python
_validators = {
   'variable_name`: callable,
   'nested.variable_name`: callable,
}
```

Nested variable names are a bit of an issue that I'm not sure how to handle. I think we could simply flatten them for internal use.

Additionally, I think we should consider implementing another class named `Locations` or something to generate things such as the work/plot/session/run/etc directories and make the Config object solely responsible for the configuration.

stefsmeets:
I've been trying to see how far I get with this idea of having an importable config object (`esmvalcore.config`). 

At the moment I have based my implementation on a mutable mapping mimicking `rcParams` in matplotlib. Every item added to the mapping is validated/cleaned before the variable is updated (`esmvalcore._config_validators`). This really cleans up the config loading, because it offloads all the validation to the config object. 

The nice thing is that the CLI parameters can go through the same interface, making use of the same validators if necessary.

I have also added a separate importable locations class (`esmvalcore.locations`), making use of the `pathlib` library. The idea is that the user locations are isolated from the config, and can change if a new session is initiated.

stefsmeets:
I added the options from the CLI to the config. Having the validators is prevents you from making mistakes if you use a wrong command line option or variable. For example, `esmvaltool run recipe.yml --asdf` now results in `KeyError`.

stefsmeets:
Got the code to a working state with a basic recipe. The main difficulty was in the `rootpath` that is being passed around when looking for files. I added a function to recreate the `rootpath` dictionary group to get it to work.

Peter9192:
Perhaps [confuse](https://pypi.org/project/confuse/) could be interesting as well. It also takes care of defaults, type checking, etc. It can also parse command line options, but the drawback is that it uses argparse for that (which we've just abandoned in favour of fire).

stefsmeets:
@Peter9192 , based on our discussion this morning I have added a new keyword: `site` (which can be one of jasmin/dkrz/ethz). These are linked to a config each (`config-jasmin.yml`), which include the site-specific rootpaths/drs. It is also possible to specify a path to a yaml directly using the `include` keyword`. The settings read from those files will override the ones already defined in `config-user.yml`. The idea is that this will unclutter the `user-config.yml` a bit

Peter9192:
> @Peter9192 , based on our discussion this morning I have added a new keyword: `site` (which can be one of jasmin/dkrz/ethz). These are linked to a config each (e.g. `config-jasmin.yml`), which include the site-specific rootpaths/drs. These settings will override the ones already defined in the current config. The idea is that this will unclutter the `user-config.yml`, and make it easier to run on a specific machine.

Nice! Just a few things to think about:
- What if I'm on Jasmin, but I'd like to add an additional data path, being my home directory with some extra data I downloaded?
- Could the DRS be moved from the config-developer to these files?
- Could we make it possible to specify a path-specific drs? This currently doesn't work, I think:

```yaml
rootpath:
 CMIP6: [/badc/cmip6/data/CMIP6, /home/peter/custom_data/]

drs:
 CMIP6: [BADC, 'flat']
```

stefsmeets:
Alright, it has been quite a bit of work to untangle all the different configs which work at different levels. I have split off the cmor settings (to initialize the CMOR_TABLES) into its own config file. Institutes did not really belong there, so I have moved that to its own file as well. This makes the `get_institutes` function a bit simpler. 

The main issue is how to organize the 'data reference syntax'. We want to be able to have multiple drs specs per project (to combine user/site inputs). But, the 'output_file' can only be defined once. So I have updated the `config-default.yml` with this in mind. The tricky part is to get the `_data_finder.py` to play along. 

To handle the data in the code, I'm playing around with a `ProjectData` class, which handles the data locations, and can be extended to handle data finding. Currently, the code is working, but very much WIP, which I will continue with after the weekend 🎉

stefsmeets:
Some nice progress today. Everything is working as intended.

I started going through the tests. Obviously, many of them are still failing, but most so far were easy to fix.

```
pytest tests/integration/test_recipe.py -k test_default_preprocessor
pytest tests/integration/test_recipe.py -k test_default_fx_preprocessor
```
Both give `assert preproc_dir.startswith(str(tmp_path))`. I'm just wondering about the intended functionality here. Should the output dir (`esmvaltool_output`) be in the directory where esmvaltool is invoked or in the same place as the recipe?

Also failing is `pytest tests/integration/test_recipe.py -k test_simple_cordex_recipe` with `esmvalcore._recipe_checks.RecipeError: Unable to load CMOR table (project) 'CORDEX' for variable 'tas' with mip 'mon'`. I'm not quite sure why.

The rest of the tests that fail are mostly in the data_finder, because of the changed config structure, but those should be not too bad to fix.

Peter9192:
@stefsmeets you could use git blame to find out who wrote or last modified those tests, and maybe tag them to get their attention? Also it would be good to merge recent changes from master and resolve the merge conflicts that have arisen.

stefsmeets:
I also came accross two 'hidden' default values.

```yaml
max_years: null  # set in main
run_diagnostic: true  # set in read_config_user
```

I have added both to the default config file for now, because ESMValTool will crash without them.

stefsmeets:
I think #794 should also be addressed by this PR.

stefsmeets:
I fixed all the data_finder tests and a few more. Only 6 more failing tests to go from the existing ones 🎉

esmvalcore/_exceptions.py
```diff
@@ -0,0 +1,23 @@
+import sys
+
+
+class SuppressedError(Exception):
+    """Errors subclassed from SuppressedError hide the full traceback.
+
+    This can be used for simple user-facing errors that do not need the
+    full traceback.
+    """
+    pass
```
bouweandela, line 10:
```suggestion
```

esmvalcore/configuration/_config_validators.py
```diff
@@ -0,0 +1,298 @@
...
+        for pattern in diagnostics or ()
+    }
+
+
+class ESMValToolDeprecationWarning(UserWarning):
+    # Custom warning, because DeprecationWarning is hidden by default
+    pass
+
+
+def deprecate(func, variable, version: str = None):
```
bouweandela, line 184: This package might also be nice, or do you think it's too much?
https://pypi.org/project/Deprecated/

stefsmeets: Seems like a bit much, and it solves the problem in the same way (using decorators). Is there anything that you think is missing from what we have now?

bouweandela: What I liked about the package is that it supports mentioning stuff in the documentation

stefsmeets: This seems to be specific for functions. We are not really deprecating functions though, only keyword arguments.

jvegasbsc:
> Actually, since all the command line variables are added to the config, should we set their defaults in the config file directly? This maybe more elegant than checking if a certain variable exists in the config. We could instead just set their defaults False or None

Good idea. It also goes in line with the old idea of having a `esmvaltool debug` command that is just a run with different flags.

stefsmeets:
The last commit fixes the remaining failing tests (mostly path issues). Thanks @bouweandela !

@bouweandela had a nice idea how to distiguish between the `config` and a `session`. The difference is that `config` is a global and thus importable. A `session` object inherits all the variables from the config, but in addition it has a name (i.e. that of the recipe) and has additional information about the session directories (i.e. `run_dir` / `preproc_dir`).

The config can be used to start a new session:

```python
session = config.start_session('recipe_name')

session == config  # True, since both inherit from dict

session['remove_preproc_dir'] = False
session == config  # False
```

This means that many session objects can exist, but they must all be started from the global config. This has the advantage that changes to a session variables can be independent from the global config, i.e. when running multiple recipes at once.

stefsmeets:
Some good progress on this PR, it is nearly complete. I added tests for the new functionality, and failing tests are now passing on my PC.

To do:
[ ] Documentation
[ ] Merge master 😬
[ ] Fix Codacy errors
[ ] Fix CI errors
[ ] Check `config-default.yml` for which variables we want to keep (would be nice if someone could have a look at this)
[ ] Consider renaming `config_user` variable to `session` to fit the config API

Peter9192:
I noticed that ESMValTool prints the config file it uses to the command line as the first log message when you run the tool. It would be nice if this would be the full path, for the default config-user.yml may be located in several different places. Would it be possible to change this here?

stefsmeets:
> I noticed that ESMValTool prints the config file it uses to the command line as the first log message when you run the tool. It would be nice if this would be the full path, for the default config-user.yml may be located in several different places. Would it be possible to change this here?

It does not do this for me. Could you explain what you mean or copy the output here?

bouweandela:
Hi @stefsmeets, I tried to review this, but the number of changes is so large that it becomes very difficult to review:
```
43 files changed, 2338 insertions(+), 1238 deletions(-)
```
Do you think it would be possible to implement this with a smaller number of changes?

I noticed that there are a lot of formatting changes too, so maybe you could first make a pull request to just reformat the files you plan to change in this pull request, to at least get rid of all the cosmetic changes?

stefsmeets:
> 
> 
> Hi @stefsmeets, I tried to review this, but the number of changes is so large that it becomes very difficult to review:
> 
> ```
> 43 files changed, 2338 insertions(+), 1238 deletions(-)
> ```
> 
> Do you think it would be possible to implement this with a smaller number of changes?
> 
> I noticed that there are a lot of formatting changes too, so maybe you could first make a pull request to just reformat the files you plan to change in this pull request, to at least get rid of all the cosmetic changes?

I was a little bit worries about that too, but I think we missed that opportunity to implement this in smaller steps once we started working on the config file parameters. Many things are so tightly coupled, that it is difficult to split this up.

With cosmetic changes, do you mean the auto-formatting? There may have been some 'collateral damage' from yapf/isort, is this what you mean?

I can split it up in several pull requests, but it would take me a couple of days. It would have to be then at least 4 pull requests, probably in this order:
- Reformat relevant files
- Refactor data finder / drs structure
- Add validated config object
- Update config file parameters

Peter9192 (commented):
Hey @stefsmeets here's my "high-level" feedback to the PR :-)

I very much like the improved DRS handling!!

One suggestion would be to split the input and output data in the DRS. Then you could change 'ProjectData' to 'InputData', which I think is more intuitive. `output_file` could have a single default for all projects, or a dict with editable output_names for different projects (not sure whether that's necessary/desired functionality).

The API works nicely, but it does require some documentation in order to know how to use it. I think it would be nice if we could make it a little bit more intuitive. Here's some things I noticed when playing around with the API:

- If you print the configuration, it would be nice if all ProjectData objects are grouped together.
- It think if the ProjectData objects would have some str or repr method, that would make it more intuitive to understand what they are, and how to use/update them.
- Should we expose all the methods of the cfg object? E.g. `clear`. I used this, and now I cannot get the default cfg back. Why would someone want to use `pop` or `popitem`? And what is the use case for `find_all`? I suppose it would be better to add things later than to have to remove them again.
- I'm not so keen on the `from _projects import searchlocation`. Perhaps we could add a method to projectdata like `cmip6.searchlocations.add()` that does the same thing?
- Why does `get_input_filelists` return 3 lists?
- It might also be nice if it could return smt like: 'No input data found for ... in ....'
- The projectdata has a method called `get_cmor_table`, but then you get CMIPInfo object, which again has a method `get_table`. That's not very intuitive. Perhaps we can make this an attribute or property instead of a method?

And here's some questions/comments related to the changes in the code:

- I like how you split up the cfg developer. But considering that some people apparently used to change all of these things sometimes, would it make sense to move cmor_config.yml (and perhaps also institutes.yml) from `cmor` to the new `config` directory?
- `_exceptions.py`: since this came up in a recent discussion, do you think this can help to get cleaner error messages for missing data?
- `esmvalcore._main.py`: here you update the session with a couple of 'standard' command line options. But when we started using `fire`, I think the idea was that you could pass *any* parameter that's in the config file also via the command line, and it will be updated accordingly. Does that still work?
- Related: I can't help but feeling that some of the config options are really very specific runtime arguments. If the session is very tightly coupled to a specific run, would it make sense to pass those runtime options only to the session, and not to the config? 
- I wonder how the refactored data finder functions impact cases like https://github.com/ESMValGroup/ESMValTool/issues/1884

One final remark: I really like the overview of the main changes in this PR that you provided in a comment on your example notebook. Perhaps we can change the top post of this PR to include that summary? Then it will be easier to find how/why this changed for people who are tracing the changes back to this PR.

stefsmeets:
Just a note to keep an eye on https://github.com/ESMValGroup/ESMValTool/pull/1707 , which may benefit from this PR

stefsmeets:
This PR is going nowhere, so I'm closing it and moving forward with a different attempt. Closing this in favour of #868.
I may split other useful bits in different PRs.

PR #786, opened 2020-09-16
set the multiprocessing distribution method from getgoing

valeriupredoi:
This is needed because on OSX on Python 3.8 the start method is either `None` or is set to `spawn` (see https://github.com/saltstack/salt/issues/55847) - and tests are failing (well, one can not actually run with more than one process) see https://github.com/ESMValGroup/ESMValCore/runs/1116923215?check_suite_focus=true

**Tasks**

-   [x] Add unit tests

* * *

Closes #issue_number

### DIFF
--- a/esmvalcore/_task.py
+++ b/esmvalcore/_task.py
@@ -4,6 +4,7 @@ import contextlib
 import datetime
 import errno
 import logging
+import multiprocessing as mp
 import numbers
 import os
 import pprint
@@ -22,6 +23,12 @@ import yaml
 from ._citation import _write_citation_files
 from ._config import DIAGNOSTICS_PATH, TAGS, replace_tags
 from ._provenance import TrackedFile, get_task_provenance
+
+
+# set the start method for multiprocessing
+# on OSX Darwin 10.14.6 it is None, leading to no process being executed
+if mp.get_start_method(allow_none=True) is None:
+    mp.set_start_method('fork', force=True)
 
 logger = logging.getLogger(__name__)
 
--- a/esmvalcore/_task.py
+++ b/esmvalcore/_task.py
@@ -24,7 +24,6 @@ from ._citation import _write_citation_files
 from ._config import DIAGNOSTICS_PATH, TAGS, replace_tags
 from ._provenance import TrackedFile, get_task_provenance
 
-
 # set the start method for multiprocessing
 # on OSX Darwin 10.14.6 it is None, leading to no process being executed
 if mp.get_start_method(allow_none=True) is None:


valeriupredoi:
cheers for merging the `vmpfrof -> vprof` PR @bouweandela - let's now get this one in as well so we can have a fully working environment on OSX with Python 3.8 - see https://github.com/ESMValGroup/ESMValCore/runs/1175425522?check_suite_focus=true :beer:

bouweandela:
I can see the failing tests, but I do not quite understand what the problem is here. Why is it a problem that the default multiprocessing start method is 'spawn' on Mac OS? I don't think we are using the `salt` package that you linked an issue from above?

valeriupredoi:
you are right, I should have edited my comment when I found out the real cause - the method is not set on OSX (returns `None`) for Python 3.8 - I don't know why but I see this both on the GA machine and on the Mac I was using for tests myself :beer:

bouweandela:
I would rather like to know what the actual bug is, but if  you cannot find an issue, we'll just have to try and work around it.

Could you try what happens if you just insert
```python
multiprocessing.get_start_method()
```
here (just before creating the `Pool` instance)
https://github.com/ESMValGroup/ESMValCore/blob/fdf1d059d1ec52efd6dab72d7523daf87bfb0fe2/esmvalcore/_task.py#L685-L686
instead of the changes in this pull request?
According to the [get_start_method documentation](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.get_start_method) that should fix the start method if it hasn't been fixed yet.

valeriupredoi:
yep I agree with you, but sadly I don't have access to the Mac no more - I also *believe* I tried that and it didn't work, maybe you can try plugging that into `github-actions2` branch and see if the test runs to completion? :beer:

valeriupredoi:
and yeps - it don't work https://github.com/ESMValGroup/ESMValCore/runs/1181768916?check_suite_focus=true

valeriupredoi:
@bouweandela if you don't plan on allowing this through we will know for sure the tool doesn't work on OSX with Python 3.8 (possibly 3.9+ too - not 100% sure of that yet, 3.9 is still too young for OSX) (see repeated [failed tests](https://github.com/ESMValGroup/ESMValCore/runs/1678480314?check_suite_focus=true)) so we either allow the patch through or we tell users about the OS+Python version restrictions :+1:

bouweandela:
I'm not convinced this 'solution' is the right approach, they changed the default way of starting new processes on Mac OS for a reason, i.e. fork didn't work: see https://bugs.python.org/issue33725. Just forcing the old behaviour is unlikely to solve the problem.

We have the following note on Mac OS X in the [tutorial installation instructions](https://esmvalgroup.github.io/ESMValTool_Tutorial/02-installation/index.html):

> On MacOSX, installation only works fine for esmvaltool-python and esmvaltool-ncl with Python 3.7.

Maybe we could add something similar in the common installation issues section I added in ESMValGroup/ESMValTool#1971, but until at least one active developer of @ESMValGroup/esmvaltool-coreteam gets a Mac and is willing to invest the time to support Mac OS I think the experience of using ESMValTool on Mac OS will remain quite bumpy.

jvegasbsc:
> until at least one active developer of @ESMValGroup/esmvaltool-coreteam gets a Mac and is willing to invest the time to support Mac OS I think the experience of using ESMValTool on Mac OS will remain quite bumpy.

It was usually quite smooth. Maybe I can borrow the old one I was using and take a look at this

valeriupredoi:
OK sounds good - Javi, make sure you are using Python 3.8, older versions work fine. Maybe we can setup a docker container or a VM somewhere where we have OSX latest installed?

bvreede:
FWIW I ran tests on my machine (OSX with Python 3.8) using this branch, and all was fine. This fixes the compatibility of ESMValTool with Python 3.8, so it could be worth merging this PR. I discussed this with @stefsmeets and he agrees :)

valeriupredoi:
cheers @bvreede - testing much appreciated! You'll have to convince @bouweandela to have this merged, he's quite suspicious of its functionality, I have given up trying to convince him :laughing:

stefsmeets:
> I'm not convinced this 'solution' is the right approach, they changed the default way of starting new processes on Mac OS for a reason, i.e. fork didn't work: see https://bugs.python.org/issue33725. Just forcing the old behaviour is unlikely to solve the problem.

Seems like a nice fix to get it to work. If you go through the discussion, you will see that 'fork' has the same issues on 3.7 and 3.8 on OSX 10.13 and above. It is that on 3.8 they recognized this and marked the behaviour as unsafe by defaulting to 'spawn'. The cause is unexpected behaviour from code calling into Apple system frameworks (which we don't use). There is no reason to believe that this patch from @valeriupredoi will cause any new issues in esmvaltool.

I suggest we apply this patch, *or* make it very clear that we do not support esmvaltool on Mac.

bouweandela:
The failing tests were fixed in #1003.

PR #796, opened 2020-09-28
Added clip_start_end_year preprocessor

### LINKED ISSUE #762: Preprocessor order: positioning of extract_time
I have a question about the preprocessor ordering:

Why are the preprocessors `extract_time`, `extract_season` and `extract_month` placed before `cmor_check_data`?

```python
...
    'cmor_check_metadata',
    # Time extraction
    'extract_time',
    'extract_season',
    'extract_month',
    # Data reformatting/CMORization
    'fix_data',
    'cmor_check_data',
    # Level extraction
    'extract_levels',
...
```

That way, they cannot be shuffled around by `custom_order = true`, since they belong to the initial steps:

https://github.com/ESMValGroup/ESMValCore/blob/594cb65498f92e6c08586e86d05933538be09de1/esmvalcore/preprocessor/__init__.py#L133

I guess the reason for this is that `fix_data` is executed after the data is reduced by those 3 preprocessors, but with that reasoning we could also move `fix_data` after `extract_levels`, `extract_region`, etc.

Would it be possible to move those 3 preprocessors further down the preprocessor pipeline (after `cmor_check_data`)? In my case, I want to use `extract_time` after `anomalies`, which is currently not possible.

schlunma:
This PR adds an `clip_start_end_year` preprocessor which is necessary to move the `extract_time`, `extract_month` and `extract_season` preprocessors down in the preprocessor order which allows their use with `custom_order: true`.

**Tasks**

-   [x] Add unit tests

* * *

Closes #762


bouweandela:
It would be good to take #345 into account, at least for the naming of the new preprocessor function.

schlunma:
Ready for review

valeriupredoi:
wait - am thoroughly confused - now we have `extract_timerange` which is a more specific case of `extract_time` and they are both preprocessor functions and only `extract_timerange` is used in the recipe for data filtering, that doesn't support filtering on months, which is needed in cases when we have data ending in weird months like, say, October - @schlunma can you pls tell me the logic behind this - I see both a duplicated functionality and the use of a less performant function at the same time (yeah, am cranky today :grin: ) :beer: - also, have you changed the doumentation?

valeriupredoi (changes requested):
see me comment above :beer:

schlunma:
Hey V., sorry I somehow didn't post my answer on your question above (which I already typed :grimacing:).

So here is the thing:
`extract_timerange` should **NOT** be used by the user, it's only an internal function for extracting the `start_year` and `end_year` given by the recipe (similar to other "hidden" preprocessor functions like `load` or `concatenate`). That's the reason why we don't need documentation for it. `extract_time` can now be used by the user and behaves like a "regular" preprocessor now (i.e. it can be shuffled around with `custom_order=true` and it is not used by default anymore).

I really think that we should merge this soon, other people ran into this issue as well: https://github.com/ESMValGroup/ESMValCore/issues/762#issuecomment-705012864 and https://github.com/ESMValGroup/ESMValTool/pull/1864#issuecomment-709996187.

valeriupredoi (approved):
ok I got persuaded by @schlunma - looks like something more flexible and frontend-y than the internal backend function :+1:

bouweandela:
@bascrezee @lukasbrunner @Peter9192 This looks like something you need https://github.com/ESMValGroup/ESMValCore/pull/796#issuecomment-724594430, would one of you be willing to try if it works for your use case?

Peter9192:
Anyone currently reviewing this? Otherwise I can have a go at it

schlunma:
Yes, please have a go at it @Peter9192. Thanks! :relaxed:

tests/integration/test_recipe.py
```diff
@@ -826,6 +867,8 @@ def test_custom_preproc_order(tmp_path, patched_datafinder, config_user):
           custom:
```
Peter9192, line 867: Shouldn't we test here for a custom order that has both extract_timerange (by means of the dataset keys) and extract_time (by means of the standard preprocessor)?

tests/integration/test_recipe.py
```diff
@@ -341,64 +400,46 @@ def test_default_preprocessor(tmp_path, patched_datafinder, config_user):
...
-        'save': {
-            'compress': False,
-            'filename': product.filename,
-        }
-    }
+    defaults = _get_default_settings_for_chl(fix_dir, product.filename)
+    assert product.settings == defaults
+
+
+def test_default_preprocessor_custom_order(tmp_path, patched_datafinder,
```
Peter9192, line 407: I'm not sure if I understand what is being tested here. Could you add a short docstring?

tests/unit/preprocessor/_time/test_time.py
```diff
@@ -149,6 +149,51 @@ def test_extract_time_no_time(self):
...
+        sliced = extract_timerange(self.cube, 1950, 1950)
+        iris.coord_categorisation.add_month_number(sliced, 'time')
+        iris.coord_categorisation.add_year(sliced, 'time')
+        assert_array_equal(np.arange(1, 13, 1),
+                           sliced.coord('month_number').points)
+        assert_array_equal(np.full(12, 1950), sliced.coord('year').points)
+
+    def test_extract_timerange_4_years(self):
+        """Test extract_timerange with 4 years."""
+        sliced = extract_timerange(self.cube, 1949, 1951)
```
Peter9192, line 170: This is not 4 years, is it?

Peter9192 (commented):
Code looks good @schlunma ! I left a few comments, will now proceed with some offline testing.
One general question: is `extract_timerange` really the most suitable name for this preprocessor, or can we come up with something slightly more distinguishing? `clip_start_endyear` is the best I could come up with so far.

Peter9192:
Pytest locally ran fine after merging origin/master into this branch, and I tested with a custom recipe, which also produced the expected results. So I'm quite happy with the code. If you could just address my comments/questions above, then I think we're good to go. 

<details><summary>test diagnostic</summary>

```yaml
documentation:
  description: test PR
  authors:
    - kalverla_peter

preprocessors:
  tas_anomalies:
    custom_order: true
    area_statistics:
      operator: mean
    anomalies:
      period: full
      reference:
        start_year: 1981
        start_month: 1
        start_day: 1
        end_year: 2010
        end_month: 12
        end_day: 31
      standardize: false
    extract_time:
      start_year: 2081
      start_month: 1
      start_day: 1
      end_day: 31
      end_month: 12
      end_year: 2099

diagnostics:
  test1:
    variables:
      tas:
        start_year: 1965
        end_year: 2099
        mip: Amon
        preprocessor: tas_anomalies
        additional_datasets:
          - {dataset: ACCESS1-0, project: CMIP5, exp: [historical, rcp85], ensemble: r1i1p1}
    scripts:
      quickplot:
        quickplot:
          plot_type: plot
```

</details>

tests/integration/test_recipe.py
schlunma: Yes, I added a test for that!

tests/integration/test_recipe.py
schlunma: Added docstring.

tests/unit/preprocessor/_time/test_time.py
schlunma: Yes, counting to 3 is hard :smile:

schlunma:
Thanks @Peter9192 for your review! :+1: I included all your comments, so this PR should be ready now.

tests/unit/preprocessor/_time/test_time.py
```diff
@@ -149,6 +156,51 @@ def test_extract_time_no_time(self):
         assert cube == sliced
 
 
+class TestExtractTimerange(tests.Test):
```
Peter9192, line 159:
```suggestion
class TestClipStartEndYear(tests.Test):
```
Should we rename this one as well?

Peter9192 (approved):
Thanks @schlunma I think you missed one rename, but otherwise I'm happy to see this merged

PR #798, opened 2020-09-30
Fix area_statistics fx files loading

jvegasbsc:
When some of the datasets do not have fx_files, the current function expects them to have their paths set to None but it is actually the empty string. With this fix, the function will work as expected with both options.

**Tasks**

-   [ ] Add unit tests

### DIFF
--- a/esmvalcore/preprocessor/_area.py
+++ b/esmvalcore/preprocessor/_area.py
@@ -1,8 +1,7 @@
-"""
-Area operations on data cubes.
+"""Area operations on data cubes.
 
-Allows for selecting data subsets using certain latitude and longitude bounds;
-selecting geographical regions; constructing area averages; etc.
+Allows for selecting data subsets using certain latitude and longitude
+bounds; selecting geographical regions; constructing area averages; etc.
 """
 import logging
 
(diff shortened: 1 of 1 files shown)


esmvalcore/preprocessor/_area.py
```diff
@@ -177,7 +175,7 @@ def tile_grid_areas(cube, fx_files):
     grid_areas = None
     if fx_files:
         for key, fx_file in fx_files.items():
-            if fx_file is None:
+            if not fx_file:
                 continue
```
valeriupredoi, line 179: @jvegasbsc can you pls briefly explain the logic? where would one get a `''` for a file variable?

jvegasbsc: After the resolve, fx_files contains `{areacello: ""}` when a specific dataset does not have fx files. Ideally, we may want to changes the resolve so it returns None, but this will suffice in the meantime

valeriupredoi: right! good catch!

valeriupredoi (approved):

valeriupredoi:
one annoying mention - when you make a one-liner code change pls don't run `yapf` or fix docstrings manually, it makes it hard to spot the actual functional change :grin:

jvegasbsc:
> one annoying mention - when you make a one-liner code change pls don't run yapf or fix docstrings manually, it makes it hard to spot the actual functional change 😁

Blame @bouweandela pre-commit hooks

Issue #799, opened 2020-09-28
Cube coord bounds make cube intersection misbehave


lukasbrunner:
**Issue in a nutshell**

- operation: `_area.py/extract_region()`
- task at hand: func needs to perfrom cube intersection on 0-360 lons
- comment: `extract_region` works well when run on a cube with negative longitude points and run as a stand alone preprocessor step, see [comment](https://github.com/ESMValGroup/ESMValCore/issues/799#issuecomment-701301097) but the intersection operation misbehaves when a `regrid` step has been run before, see [comment](https://github.com/ESMValGroup/ESMValCore/issues/799#issuecomment-701313532)

**Original Post**

Hi all!

I am currently testing the new recipe ClimWIP, which, among other things, calls the pre-processor to extract the following region:
https://github.com/ESMValGroup/ESMValTool/blob/414d7d2e84a62a777dfada79297c0ed400b6bbd3/esmvaltool/recipes/recipe_climwip.yml#L52-L56

For me the resulting longitude values look like this:
` lon = -1.25, 1.25, 3.75, 6.25, 8.75, 11.25, 13.75, 16.25, 18.75, 21.25, 
    23.75, 26.25, 28.75, 31.25, 33.75, 36.25, 38.75, 351.25, 353.75, 356.25 ;
`

Can anyone recreate this behaviour? Is it even expected? To me this looks wrong. My understanding is that the longitude should either be in [-180, 180) OR in [0, 360) but neither case is true here. 

Many thanks,
Lukas

[main_log_debug.txt](https://github.com/ESMValGroup/ESMValTool/files/5292272/main_log_debug.txt)

Peter9192:
Good! At least we get the same results now, that helps.

valeriupredoi:
it's the `bounds` that fool `intersection`, here's a minimal example:
```python
import iris
import numpy as np

lon_points = np.array([-8.75, -6.25, -3.75, -1.25,  1.25,  3.75,  6.25,  8.75, 11.25,
                       13.75, 16.25, 18.75, 21.25, 23.75, 26.25, 28.75, 31.25, 33.75,
                       36.25, 38.75])
lon_bounds = np.array([
    [ -2.5,   0. ],
    [  0. ,   2.5],
    [  2.5,   5. ],
    [  5. ,   7.5],
    [  7.5,  10. ],
    [ 10. ,  12.5],
    [ 12.5,  15. ],
    [ 15. ,  17.5],
[...]

PR #800, opened 2020-09-30
temporary fix to an issue with cube intersection when cube has bounds

valeriupredoi:

**Tasks**

-   [ ] Add unit tests

* * *

Closes #issue_number #799 and addresses an inherently iris issue firt posted in https://github.com/SciTools/iris/issues/3391

### DIFF
--- a/esmvalcore/preprocessor/_area.py
+++ b/esmvalcore/preprocessor/_area.py
@@ -61,7 +61,7 @@ def extract_region(cube, start_longitude, end_longitude, start_latitude,
             latitude=(start_latitude, end_latitude),
             ignore_bounds=True,
         )
-        # delete bounds: kluge for issue 799
+        # delete bounds: kluge for https://github.com/ESMValGroup/ESMValCore/issues/799
         # TODO remove the bounds handling after iris fix
         region_subset.coord("longitude").bounds = None
         region_subset = region_subset.intersection(longitude=(0., 360.))
--- a/esmvalcore/preprocessor/_area.py
+++ b/esmvalcore/preprocessor/_area.py
@@ -61,7 +61,8 @@ def extract_region(cube, start_longitude, end_longitude, start_latitude,
             latitude=(start_latitude, end_latitude),
             ignore_bounds=True,
         )
-        # delete bounds: kluge for https://github.com/ESMValGroup/ESMValCore/issues/799
+        # delete bounds: kluge for
+        # issue here https://github.com/ESMValGroup/ESMValCore/issues/799
         # TODO remove the bounds handling after iris fix
         region_subset.coord("longitude").bounds = None
         region_subset = region_subset.intersection(longitude=(0., 360.))
--- a/esmvalcore/preprocessor/_area.py
+++ b/esmvalcore/preprocessor/_area.py
@@ -61,7 +61,11 @@ def extract_region(cube, start_longitude, end_longitude, start_latitude,
             latitude=(start_latitude, end_latitude),
             ignore_bounds=True,
         )
+        # delete bounds: kluge for issue 799
+        # TODO remove the bounds handling after iris fix
+        region_subset.coord("longitude").bounds = None
         region_subset = region_subset.intersection(longitude=(0., 360.))
+        region_subset.coord("longitude").guess_bounds()
         return region_subset
     # Irregular grids
     lats = cube.coord('latitude').points


esmvalcore/preprocessor/_area.py
```diff
@@ -61,7 +61,11 @@ def extract_region(cube, start_longitude, end_longitude, start_latitude,
             latitude=(start_latitude, end_latitude),
             ignore_bounds=True,
         )
+        # delete bounds: kluge for issue 799
```
Peter9192, line 64:
```suggestion
        # delete bounds: kluge for https://github.com/ESMValGroup/ESMValCore/issues/799
```

Peter9192 (approved):
I can confirm that this works, but I'm not keen on workarounds like this. Let's also try to get the issue fixed in iris.

Peter9192:
I don't mind about codacy (all the more reason to get back to this at some point), but we should solve the flake8 error.

jvegasbsc:
> I can confirm that this works, but I'm not keen on workarounds like this. Let's also try to get the issue fixed in iris.

Neither am I. It will be possible to, at least, remove the bounds only if we need it and keep them unchanged in all the other cases? I am not keen on dumping metadata so easily

valeriupredoi:
yeah we could kill the bounds and re-guess them only if `longitude` has negative points, as we've seen the case, but that'll only add to the confusing nature of this kluge - I believe we should just have it as it is and immediately get rid of it when @bjlittle and the Iris team have fixed this in Iris :beer:

ruthlorenz:
Are you guys sure this solves the issue?
the case was a region to extract from -10 to 39 longitude.
Without this fix the values in the preprocessed file I obtain are 

lon = -1.25, 1.25, 3.75, 6.25, 8.75, 11.25, 13.75, 16.25, 18.75, 21.25,  23.75, 26.25, 28.75, 31.25, 33.75, 36.25, 38.75, 351.25, 353.75, 356.25

with the fix its
lon = 1.25, 3.75, 6.25, 8.75, 11.25, 13.75, 16.25, 18.75, 21.25, 23.75, 26.25, 28.75, 31.25, 33.75, 36.25, 38.75, 351.25, 353.75, 356.25, 358.75

it gets rid of the -1.25 which is good, but at the end it still jumps from 38.75 to 351.25.
So extract_region ends up adding the region west of the 0 meridian at the eastern boundary.

Or is this a second issue? that extract_region does not work properly for regions crossing the 0 meridian?

jvegasbsc:
This is a known behaviour of Iris (I think @mwjury also experimented this). Although it is quite of weird when you check the file itself, it should not interfere with anything, I have seen it been interpreted correctly in diagnostics, but if it break yours we can look for a solution.

mwjury:
Yes, at first I was very confused when looking at the maps I produced, though it made sense looking at the data in ncview. But aside from that everything has been working correctly further downstream. Iris has it's problem with plotting such data, so the only solution was to center lon on zero in the diagnostics.

ruthlorenz:
Thanks for the clarification. I do not think it breaks anything, just looks weird and makes the comparison to another implementation more difficult. We will keep this feature in mind ;-)

valeriupredoi:
just tested for the unwanted behavor with iris=3.0.1 (current version we use) and it's still there, reported in this [comment](https://github.com/SciTools/iris/issues/3391#issuecomment-788838200) - do we want to go ahead with this patch here (more of a kluge rather than a patch) or wait for the iris folk to sort it out? Asking since we need to do some maintenance on this branch if we want to go ahead with it :beer:

Peter9192:
I think it's a good idea to have this as a workaround, but we should be a bit more sophisticated about how to implement in. For example, we could add a test that triggers this error and mark it with xfail, linking to the iris issue or this one. Then, if it's fixed in iris, we should see it because the xfail test suddenly passes. I also still like the suggestion to do this only if strictly necessary.

rcomer:
I think you could get the same effect using `ignore_bounds=True` in your call to `intersection`.  It would save having to delete them.

bouweandela:
Looks like this may be fixed in iris soon now https://github.com/SciTools/iris/pull/4046, nice work @rcomer!

valeriupredoi:
> Looks like this may be fixed in iris soon now [SciTools/iris#4046](https://github.com/SciTools/iris/pull/4046), nice work @rcomer!

yes, my apologies @rcomer - I got distracted and forgot to thank you for your work! :beer:

bjlittle:
We're going to pull @rcomer's fix into `v3.0.x` and make it available in the forthcoming `iris` v3.0.2 :+1:

rcomer:
Due to some terrifying conflicts, the fix has moved to a new PR: https://github.com/SciTools/iris/pull/4059

valeriupredoi:
fixed (and tested by meself) by the good people at Iris in https://github.com/SciTools/iris/pull/4059

PR #808, opened 2020-10-02
Deprecate defining write_plots and write_netcdf in config-user file

bouweandela:
Deprecate use of write_plots, write_netcdf, output_file_type config-user file

Please define those settings in the diagnostic script setting of the recipe instead, for those diagnostic scripts that support these settings

See #93

**Tasks**

~Add unit tests~ Skipped because all this code is going to be replaced in #785 soon
-   [x] Update documentation


bouweandela:
@stefsmeets Do you think it still makes sense to implement unit tests for this, given that you're going to change everything anyway in #785?

esmvalcore/_recipe.py
```diff
@@ -1242,15 +1242,22 @@ def _initialize_scripts(self, diagnostic_name, raw_scripts,
             if self._cfg['write_ncl_interface']:
                 settings['exit_on_ncl_warning'] = self._cfg['exit_on_warning']
             for key in (
-                    'output_file_type',
```
stefsmeets, line 1245: I think it is enough to just mark the parameters as deprecated in the config and raise a warning. I think it would be better to keep this part of the code `as-is` for now. The checks to re-add the keys below adds clutter. Save these for a later patch, and apply it to remove the parameters from the entire codebase for 2.3.

esmvalcore/_task.py
```diff
@@ -367,10 +367,7 @@ def _write_ncl_settings(self):
             'run_dir',
             'plot_dir',
             'work_dir',
-            'output_file_type',
```
stefsmeets, line 370: Same here, don't delete these variables just yet and remove the block to add them back below.

esmvalcore/_task.py
```diff
@@ -549,15 +553,12 @@ def _collect_provenance(self):
             'exit_on_ncl_warning',
             'input_files',
             'log_level',
-            'output_file_type',
```
stefsmeets, line 552: Same here.

esmvalcore/_config.py
```diff
@@ -44,18 +45,28 @@ def read_config_user_file(config_file, folder_name, options=None):
...
+            warnings.warn(DeprecationWarning(msg))
+
     if options is None:
         options = dict()
     for key, value in options.items():
         cfg[key] = value
 
     # set defaults
     defaults = {
-        'write_plots': True,
```
stefsmeets, line 54: I think we should keep the defaults for now to ensure no diagnostic code breaks until 2.3

stefsmeets (changes requested):
Looks good. I was thinking just marking the config parameters is enough (see comments).

esmvalcore/_recipe.py
bouweandela: The problem with keeping this as is, is that the parameters from config-user.yml then overwrite what it says in the recipe, and people might get confused that it doesn't work if they add the setting in the recipe

esmvalcore/_task.py
bouweandela: some change is needed here, because otherwise these settings will not be available as variable `diag_script_info@write_plots` etc, making it impossible to adjust diagnostic scripts to the new situtation before they are removed in 2.3. As an intermediate step, I thought I would make them available as both `diag_script_info@write_plots` and `config_user_info@write_plots`.

esmvalcore/_task.py
bouweandela: That would ignore recording their value in the provenance, so it would break the idea that all diagnostic script settings are recorded in the provenance

esmvalcore/_config.py
bouweandela: Removing the defaults would not break any diagnostic code, but it does break some recipes if you do not have the parameters in your config-user file. The downside of not removing them is that there might be not enough incentive to update any recipe. Maybe you're right and this is too strong though. Shall I add them back in?

bouweandela:
> Looks good. I was thinking just marking the config parameters is enough (see comments).

Thanks for reviewing!

esmvalcore/_config.py
valeriupredoi: yeah I reckon the defaults should be kept, I can see a reason why keeping them even after 2.3 since we may want to revisit some in the future

valeriupredoi (changes requested):
am not actually requesting changes, I am just thinking these options are indeed not needed anymore but the functional infrastructure they offer is useful - am thinking for the future when we may get to a point where we use a web based interface that runs diagnostics in the backend and the user gets a nice output that they can immediately use for a paper or presentation  etc - the `write_netcdf` and `output_format` would be nice to be a tunable button

esmvalcore/_recipe.py
stefsmeets: But if the warning is given for the config, then that should be enough in my opinion. Maybe it would be then a good idea to do a double check in the recipe runner and re-issue the warning there.

schlunma:
I vote strongly against removing `output_file_type`. Especially having the command line option `output-file-type` to quickly change the file format for plots is really convenient. For example, I often switch between PDF (used for TeX or for seperate files of plots sent to journals) and PNG (used for Word).

I don't have a strong opinion about the other two options (more in favor of removing them), but I would really like to keep `output_file_type`.

valeriupredoi:
> I vote strongly against removing `output_file_type`. Especially having the command line option `output-file-type` to quickly change the file format for plots is really convenient. For example, I often switch between PDF (used for TeX or for seperate files of plots sent to journals) and PNG (used for Word).
> 
> I don't have a strong opinion about the other two options (more in favor of removing them), but I would really like to keep `output_file_type`.

albeit never used by me (but then again I don't do science), am glad to see this being used and with a good potential for the future quick-run WPS user. Cheers for commenting @schlunma :beer:

bouweandela:
Thanks for the comments @valeriupredoi and @schlunma! Did you also have a look at issue #93 for the reasoning behind removing these settings?

> I vote strongly against removing output_file_type. Especially having the command line option output-file-type to quickly change the file format for plots is really convenient. For example, I often switch between PDF (used for TeX or for seperate files of plots sent to journals) and PNG (used for Word).

@schlunma It would be even more convenient if the diagnostic simply wrote both file formats, because then you wouldn't even need to re-run or change a setting to get the plots in a different the format. Since plotting is not a very time-consuming process, I think the time saving is very minimal.  The arguments against having the feature are that it provides a poor user experience because many of the diagnostics do not respect the setting and the user needs to select the right output type before running. Also, it might not be so easy to implement this feature consistently across all diagnostics, for example: I doubt whether the R plotting library supports the same output file types of the Python plotting library or NCL plotting library. I can add a small plotting function to the ESMValTool diag_scripts/shared function that plots several types by default and use it from the example recipe for demonstration purposes, would that make sense? Do you think `.png` and `.pdf` would be enough, or would other file formats also make sense?

> am thinking for the future when we may get to a point where we use a web based interface that runs diagnostics in the backend and the user gets a nice output that they can immediately use for a paper or presentation etc - the `write_netcdf` and `output_format` would be nice to be a tunable button

@valeriupredoi `write_netcdf` might reduce the runtime and disk space usage of diagnostics that write very large output files, but the majority of diagnostics write very small netcdf files, because there is only so much information that can go into a plot. Therefore it's much more convenient to simply always write all netcdf files. For diagnostic developers who really want to be able to disable this, we can recommend in the documentation that they implement this setting in the recipe. For a web based interface it would probably be nicest to have the netcdf files, because then you can implement an interactive viewer so that the users of the website can zoom in, etc. Regarding configuring the output file type: it's much simpler to always write all plot types, the extra compute time won't be very much.

schlunma:
> @schlunma It would be even more convenient if the diagnostic simply wrote both file formats, because then you wouldn't even need to re-run or change a setting to get the plots in a different the format. Since plotting is not a very time-consuming process, I think the time saving is very minimal. The arguments against having the feature are that it provides a poor user experience because many of the diagnostics do not respect the setting and the user needs to select the right output type before running. Also, it might not be so easy to implement this feature consistently across all diagnostics, for example: I doubt whether the R plotting library supports the same output file types as the Python plotting library or the NCL plotting library. I can add a small plotting function to the ESMValTool diag_scripts/shared function that plots several types by default and use it from the example recipe for demonstration purposes, would that make sense? Do you think `.png` and `.pdf` would be enough, or would other file formats also make sense?

Alright, that does make sense for me. A small function that writes `.png` and `.pdf` would be sufficient for me. Maybe something like this (to consider `matplotlib.pyplot.savefig` and `matplotlib.figure.Figure.savefig`)?

```python
import matplotlib.pyplot as plt

def savefig(basename, cfg, fig=None, ext=None **kwargs):
    if ext is None:
        ext = ['png', 'pdf']
    paths = ...  # get correct paths with basename, ext and cfg
    plot_object = plt if fig is None else fig
    for path in paths:
        plot_object.savefig(path, **kwargs)
```

Just as a small side note: Plotting might be time-consuming in rare cases, e.g. I have an example where I plot a scatterplot with ~200'000 points. In this case it's also not useful to save the file as '*.pdf', since this has more than 100 MB and takes 10min to open on my local machine. However, I think is a rare exception.

esmvalcore/_config.py
bouweandela: Restored the defaults!

esmvalcore/_recipe.py
bouweandela: >  Maybe it would be then a good idea to do a double check in the recipe runner and re-issue the warning there.

Do you mean here by the 'recipe runner'? Wouldn't that result in the same amount of code clutter, but less user friendly?

valeriupredoi:
cheers for the debate, guys - always useful to discuss things this way :+1: I agree with you all - I am happy the defaults are back and as for the plots file types I think there's more than two right? - eps, png, jpg, pdf, gif (yeah we should implement a gif, we can make climate change memes :grin: - no but seriously, for looking at moving stuff in plots) - so I think that should be tunable since I don't see the need to output all these out. I reckon it's best kept under the hood now (hence remove the optionality from config) but keep the functionality :beer:

valeriupredoi:
@schlunma 

> However, I think is a rare exception.

might not be as rare as David Attenborough would say - the Autoassess diags output hundreds of plots :beer:

esmvalcore/_recipe.py
stefsmeets: Well, I was thinking if that is the problem, we can also warn users as late as possible -- when the recipe is being executed by what I called the recipe runner 😅

bouweandela:
This needs to be adjusted to take into account #868 and the feature will now be removed in v2.4, because it didn't make it for the 2.1 release.

stefsmeets:
> This needs to be adjusted to take into account #868 and the feature will now be removed in v2.4, because it didn't make it for the 2.1 release.

Do we have a deprecation plan somewhere to keep track of which variables are deprecated / removed in which version? Maybe in #93?

bouweandela:
> Do we have a deprecation plan somewhere

No

> Maybe this can be added in #93?

I updated the issue

jvegasbsc:
Can you please update these, so we can do it for this release?

esmvalcore/_config.py
```diff
@@ -44,18 +45,37 @@ def read_config_user_file(config_file, folder_name, options=None):
     with open(config_file, 'r') as file:
         cfg = yaml.safe_load(file)
 
+    # DEPRECATED: remove in v2.4
+    for setting in ('write_plots', 'write_netcdf', 'output_file_type'):
+        if setting in cfg:
+            msg = (
+                f"Using '{setting}' in {config_file} is deprecated and will "
+                "be removed in ESMValCore version 2.3. For diagnostics "
```
jvegasbsc, line 53:
```suggestion
                "be removed in ESMValCore version 2.4. For diagnostics "
```

esmvalcore/_config.py
```diff
@@ -44,18 +45,37 @@ def read_config_user_file(config_file, folder_name, options=None):
...
+
     if options is None:
         options = dict()
     for key, value in options.items():
         cfg[key] = value
+        # DEPRECATED: remove in v2.4
+        if key in ('write_plots', 'write_netcdf', 'output_file_type'):
+            msg = (
+                f"Setting '{key}' from the command line is deprecated and "
+                "will be removed in ESMValCore version 2.3. For diagnostics "
```
jvegasbsc, line 69:
```suggestion
                "will be removed in ESMValCore version 2.4. For diagnostics "
```

jvegasbsc (changes requested):
You missed a couple of 2.3, but nothing really serious

stefsmeets:
Just wondering if these are still up-to-date:
https://github.com/ESMValGroup/ESMValCore/blob/9ac52c6a97e898227d7592cf246657d768ee407d/esmvalcore/experimental/config/_config_validators.py#L253 

I marked them for 2.2, but I think it should be 2.3 or 2.4 now also?

bouweandela:
I think they're fine, deprecate in 2.2 (the upcoming release) and remove in 2.4 (in autumn).

bouweandela:
After looking through the code of ESMValTool to see what it would take to remove `output_file_type` from config-user.yml, I think that may be a lot of work. I think it's probably not worth the effort.

bouweandela:
@jvegasbsc @valeriupredoi @stefsmeets Could you have another look at this please to see if anything remains to be done before this can be merged? It would be nice if we could include it in the v2.2 release.

valeriupredoi (approved):
looks good, cheers @bouweandela - do we have an issue in ESMValTool about this (to remove the bits in the diags that are deprecated)?

bouweandela:
Good point! See https://github.com/ESMValGroup/ESMValTool/issues/2005

valeriupredoi:
> Good point! See [ESMValGroup/ESMValTool#2005](https://github.com/ESMValGroup/ESMValTool/issues/2005)

noice! I added a "future release" label there because it's cool to have one :grin:

esmvalcore/experimental/config/_config_validators.py
```diff
@@ -253,8 +253,6 @@ def deprecate(func, variable, version: str = None):
     # deprecate in 2.2.0
     'write_plots': deprecate(validate_bool, 'write_plots', '2.2.0'),
```
jvegasbsc, line 254:
```suggestion
    'write_plots': deprecate(validate_bool, 'write_plots', '2.4.0'),
```

esmvalcore/experimental/config/_config_validators.py
```diff
@@ -253,8 +253,6 @@ def deprecate(func, variable, version: str = None):
...
     'write_netcdf': deprecate(validate_bool, 'write_netcdf', '2.2.0'),
```
jvegasbsc, line 255:
```suggestion
    'write_netcdf': deprecate(validate_bool, 'write_netcdf', '2.4.0'),
```

jvegasbsc (changes requested):
This two should point to 2.4, no?

Issue #814, opened 2020-10-05
Upload packages to esmvalgroup conda channel and PyPI using GitHub actions


bouweandela:
Here is an action for deploying to PyPI: https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

Here is an action to upload a conda package to a custom conda channel: https://github.com/marketplace/actions/anaconda-action by my colleague @fdiblen

EDIT: The action above is not quite ready, best to copy/paste/modify the actions defined in https://github.com/matchms/matchms/tree/master/.github/workflows for now. They do build the Conda package, deploy to Conda channel and test the deployed packages as well.

valeriupredoi:
good finds @bouweandela - will have a look at these later on today, currently installing ESMValTool centrally on Jasmin :partying_face: (won't be available until it gets linked in a module by the CEDA folk but this is the first step)

PR #820, opened 2020-10-08
Create Github Actions workflow to build and deploy on Test PyPi and PyPi

valeriupredoi:
Automate the deployment to PyPi 🐍📦

See  #814

**Tasks**

### DIFF
--- a/.github/workflows/action-pypi-build-and-deploy.yml
+++ b/.github/workflows/action-pypi-build-and-deploy.yml
@@ -16,6 +16,9 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: 3.8
+    - name: Export CWD to PYTHONPATH
+      run: >-
+        export PYTHONPATH="$PWD:$PYTHONPATH"
     - name: Install pep517
       run: >-
         python -m
--- a/.github/workflows/action-pypi-build-and-deploy.yml
+++ b/.github/workflows/action-pypi-build-and-deploy.yml
@@ -16,20 +16,20 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: 3.8
+    - name: Set PYTHONPATH
+      run: >-
+        echo "PYTHONPATH=$PWD" >> $GITHUB_ENV
     - name: Check PYTHONPATH
       run: >-
-        echo "::set-env name=PYTHONPATH::$PWD"
         echo $PYTHONPATH
     - name: Install pep517
       run: >-
-        echo "::set-env name=PYTHONPATH::$PWD"
         python -m
         pip install
         pep517
         --user
     - name: Build a binary wheel and a source tarball
       run: >-
-        echo "::set-env name=PYTHONPATH::$PWD"
         python -m
         pep517.build
         --source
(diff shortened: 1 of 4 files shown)


valeriupredoi:
@bouweandela do you have a `pyproject.toml` hanging around from previous PyPi builds? I am a bit lazy and not 100% sure of creating one around; do we need to use pep517 to create the wheel or we just use the usual `python3 setup.py sdist bdist_wheel`?

bouweandela:
It would probably be good to use that, since it looks like that's the future of Python packaging, see also this blog: https://snarky.ca/what-the-heck-is-pyproject-toml/

From the blog, it looks like the following content would be enough:
```
[build-system]
requires = ["setuptools >= 40.6.0", "wheel"]
build-backend = "setuptools.build_meta"
```

valeriupredoi:
I'll try that in a jiffy, cheers dude :beer:

valeriupredoi:
OK good news! I got the darn thing to build the wheel automatically (cheers for the pyproject file suggestion, @bouweandela ) but it looks like I have no permission to upload to Test PyPi - see [test](https://github.com/ESMValGroup/ESMValCore/runs/1230941927?check_suite_focus=true) - this means that either @bouweandela hasnt added me as collaborator on Test PyPi or the Secrets have not been set correctly on ESMValCore/Settings @mattiarighi - can you guys check this by running the test yourselves pls? To trigger the test just do a dummy push to this branch

valeriupredoi:
> OK good news! I got the darn thing to build the wheel automatically (cheers for the pyproject file suggestion, @bouweandela ) but it looks like I have no permission to upload to Test PyPi - see [test](https://github.com/ESMValGroup/ESMValCore/runs/1230941927?check_suite_focus=true) - this means that either @bouweandela hasnt added me as collaborator on Test PyPi or the Secrets have not been set correctly on ESMValCore/Settings @mattiarighi - can you guys check this by running the test yourselves pls? To trigger the test just do a dummy push to this branch

problem solved by @bouweandela adding me to PyPi/Test PyPi :beer:

valeriupredoi:
here is a (most probable) explanation for the failed circleCI test: https://github.com/pypa/pip/issues/8437

bouweandela:
From the blog:

> which is why the current directory is not put on sys.path automatically; you will need sys.path.insert(0, os.path.dirname(__file__)) or equivalent if you're importing local files

So you'll probably need to insert that before trying to import esmvalcore._version in setup.py. Alternatively, I think you could just write a small function to read the version number from the file without importing it.

valeriupredoi:
@bouweandela you Dutch genius! How did I miss that? Yeh, does the job, but two things: the package already being on Test PyPi the upload fails (should I remove it from Test PyPi? - am scared) and PEP8 complains about the module import not being at top of file - bs if you ask me but hey, it's PEP8

bouweandela:
You can't remove packages from (test) PyPI, because that's just a really bad idea. Just use a new version number for the next try, e.g. `2.1.0.post0`, `2.1.0.post1` etc.

bouweandela (approved):
Nice job!

PR #818, opened 2020-10-08
add conda build Github Actions workflow

valeriupredoi:
Tested off the `github-actions2` test branch, it builds the conda package nicely; we need to add an upload to channel option (hence installing `anaconda-client`) but maybe we do that later. test off test branch [here](https://github.com/ESMValGroup/ESMValCore/actions/runs/295505663)

Before you start, please read our [contribution guidelines](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html).

**Tasks**

-   [x] [Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do, if you haven't done so already (and add the link at the bottom)
-   [ ] This pull request has a descriptive title that can be used in a changelog
-   [ ] Add unit tests
-   [ ] Public functions should have a numpy-style docstring so they appear properly in the [API documentation](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/api/esmvalcore.html). For all other functions a one line docstring is sufficient.
-   [ ] If writing a new/modified preprocessor function, please update the [documentation](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/preprocessor.html)
-   [x] Circle/CI tests pass. Status can be seen below your pull request. If the tests are failing, click the link to find out why.
-   [x] Codacy code quality checks pass. Status can be seen below your pull request. If there is an error, click the link to find out why. If you suspect Codacy may be wrong, please ask by commenting.
-   [ ] Please use `yamllint` to check that your YAML files do not contain mistakes
-   [ ] If you make backward incompatible changes to the recipe format, make a new pull request in the [ESMValTool repository](https://github.com/ESMValGroup/ESMValTool) and add the link below

If you need help with any of the tasks above, please do not hesitate to ask by commenting in the issue or pull request.

* * *

See #814

PR #825, opened 2020-10-12
Build and publish the esmvalcore package to conda via Github Actions workflow

valeriupredoi:
This builds and publishes the esmvalcore package to conda - **currently untested**; the conda build alone step is done via PR #818 so should work OK but the publishing is brand new territory for me. @bouweandela and @fdiblen could you guys have a look please. @fdiblen cheers for the pointer to the test GA workflow you have, I used that one! :beer: BTW the base branch is `release_2.1` so we can easily merge this in the release branch if you guys think it's okay :+1:

### DIFF
--- a/.github/workflows/action-conda-publish.yml
+++ b/.github/workflows/action-conda-publish.yml
@@ -48,7 +48,7 @@ jobs:
         shell: bash -l {0}
         run: |
           export BUILD_FOLDER=/tmp/esmvalcore/_build
-          anaconda --token ${{ secrets.ANACONDA_TOKEN }} upload --user esmvalgroup --labels test $BUILD_FOLDER/noarch/*.tar.bz2
+          anaconda --token ${{ secrets.ANACONDA_TOKEN }} upload --user esmvalgroup $BUILD_FOLDER/noarch/*.tar.bz2 --label test
   verify:
     name: verify / python-${{ matrix.python-version }} / ${{ matrix.os }}
     strategy:
--- a/.github/workflows/action-conda-publish.yml
+++ b/.github/workflows/action-conda-publish.yml
@@ -3,6 +3,9 @@ name: Anaconda Publish
 on:
   release:
     types: [published]
+  push:
+    branches:
+    - conda_publish_GA
 
 jobs:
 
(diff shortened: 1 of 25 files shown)


.github/workflows/action-conda-publish.yml
```diff
@@ -0,0 +1,78 @@
...
+  publish:
+    name: publish / python-3.8 / ubuntu-latest
+    runs-on: 'ubuntu-latest'
+    steps:
+      - uses: actions/checkout@v2
+      - uses: goanpeca/setup-miniconda@v1
+        with:
+          activate-environment: esmvaltool
+          auto-update-conda: true
+          environment-file: environment.yml
```
bouweandela, line 18: The dependencies from environment.yml are not needed to build the package, so I would remove this

.github/workflows/action-conda-publish.yml
```diff
@@ -0,0 +1,78 @@
...
+          export BUILD_FOLDER=/tmp/esmvalcore/_build
+          mkdir -p $BUILD_FOLDER
+          conda build package \
+            --channel esmvalgroup --channel conda-forge \
+            --croot $BUILD_FOLDER \
+      - name: Push the package to anaconda cloud
+        shell: bash -l {0}
+        run: |
+          export BUILD_FOLDER=/tmp/esmvalcore/_build
+          anaconda --token ${{ secrets.ANACONDA_TOKEN }} upload --user esmvalgroup --force $BUILD_FOLDER/noarch/*.tar.bz2
```
bouweandela, line 48:
```suggestion
          anaconda --token ${{ secrets.ANACONDA_TOKEN }} upload --user esmvalgroup $BUILD_FOLDER/noarch/*.tar.bz2
```
I'm not sure if we want to force uploading, because we do not want to upload if there are errors

bouweandela, line 48: For testing, you could probably use the labels flag to upload to a `test` label, I think something like this: `--labels test`

bouweandela (commented):
Looks like a good start! Could you merge master into this branch, so the diff on GitHub shows which files are actually changed?

.github/workflows/action-conda-publish.yml
bouweandela: Did you create an ANACONDA_TOKEN already and ask @mattiarighi to set it?

valeriupredoi: no, I was about to, good call! But isn't there one already from previous uploads?

bouweandela: No, I always manually uploaded packages so far. See here for how to create a token: https://docs.anaconda.com/anaconda-cloud/user-guide/tasks/work-with-accounts/#creating-access-tokens

valeriupredoi: yes I have now generated one, emailed @mattiarighi to add it :beer:

valeriupredoi: and it works well thanks to @mattiarighi adding it to them secrets :beer:

valeriupredoi:
we have liftoff of the Conda Package space shuttle :rocket: ie all works fine in test mode see {GA tests](https://github.com/ESMValGroup/ESMValCore/runs/1242057658?check_suite_focus=true)

bouweandela (approved):
Looks great!

PR #838, opened 2020-10-19
Fix numbering of steps in release instructions

bouweandela:
Just noticed that the numbering of the steps was wrong, this pull request fixes it.


valeriupredoi (approved):
I'll teach you to count next time we have a Zoom work session, Bouwe :grin:

PR #844, opened 2020-10-23
Remove code needed for Python <3.6

bouweandela:
This pull request removes code needed for Python versions we no longer support: `dict`s are ordered since Python 3.6, so there is no need to use an `OrderedDict` anymore.

It also removes the function `esmvalcore._recipe._get_value`, which was not used for anything anymore.

### DIFF
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -3,7 +3,6 @@ import fnmatch
 import logging
 import os
 import re
-from collections import OrderedDict
 from copy import deepcopy
 from pprint import pformat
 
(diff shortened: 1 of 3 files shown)


valeriupredoi (approved):
good stuff, @bouweandela :beer: We sure have this specified in the documentation - we support only Python>=3.6?

bouweandela:
Yes, it's here, first line: https://docs.esmvaltool.org/en/latest/quickstart/installation.html

valeriupredoi:
good to go in me books!

stefsmeets:
I have had good experience with pyupgrade (https://github.com/asottile/pyupgrade) for this in the past.

PR #848, opened 2020-10-27
Update install.rst

bascrezee:
Using the https protocol here is more user-friendly, right? No need for a user to set-up the ssh protocol with git.

### DIFF
--- a/doc/quickstart/install.rst
+++ b/doc/quickstart/install.rst
@@ -176,7 +176,7 @@ To install from source for development, follow these instructions.
    ``.cshrc``/``.tcshrc`` file instead.
 -  Update conda: ``conda update -y conda``
 -  Clone the ESMValCore Git repository:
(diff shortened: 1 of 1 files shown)


bouweandela:
Personally I would recommend using ssh, because it saves you from having to type your password for every commit, which I think is not user friendly. But tastes differ, so maybe we could mention both options?

bascrezee:
That's true. But in this particular case, for `git clone`, the https protocol would work without having anything setup, whereas the `ssh` protocol fails if you didn't configure it correctly.

stefsmeets:
Looks good, I also ran into this issue today on a different repo. +1 for recommending https because it does not need a github login.

jvegasbsc (approved):

valeriupredoi (approved):
I too approve of this (not because I found any issues with git+ssh, but because I do things like in the 1980s :grin: )

PR #850, opened 2020-10-28
Add requests as a dependency

### LINKED ISSUE #851: Nightly 'install' test on CircleCI is failing
Since a few days, the nightly test that installation from source works is failing on CircleCI, see e.g. [here](https://app.circleci.com/pipelines/github/ESMValGroup/ESMValCore/3420/workflows/e63bbe2d-c55f-4328-8d57-3b3e42b65e0e/jobs/15254).

It looks like this happens because we make use of the `requests` library, but did not add it as a dependency.

bouweandela:
Add requests as a dependency, because it's missing from the list of dependencies and it's used:

https://github.com/ESMValGroup/ESMValCore/blob/02b8f668e89cabae656978b44faa1bb27f7158be/esmvalcore/_citation.py#L8

**Tasks**

* * *

Closes #851

### DIFF
--- a/package/meta.yaml
+++ b/package/meta.yaml
@@ -53,6 +53,7 @@ requirements:
     - psutil
     - pydot
     - pyyaml
+    - requests
     - shapely
     - yamale==2.*  # in esmvalgroup channel
 
--- a/setup.py
+++ b/setup.py
@@ -41,6 +41,7 @@ REQUIREMENTS = {
         'prov[dot]',
         'psutil',
         'pyyaml',
+        'requests',
         'scitools-iris>=2.2',
         'shapely[vectorized]',
         'stratify',


valeriupredoi (approved):
what's it do?

bouweandela:
It fixes the failing tests, see #851

valeriupredoi:
yeah, sorry, I commented before I looked :grin: - but still, it seems that package should be installed automatically without us ever asking for it (the two envs I built today have it), I wonder why the CI env is ignoring it

valeriupredoi:
we should try to avoid requesting deps manually, they may be not needed anymore in the future and we're left with conflicts from packages we don't actually need

bouweandela:
> we should try to avoid requesting deps manually, they may be not needed anymore in the future and we're left with conflicts from packages we don't actually need

That's all fine, but we do need this package because we do use it, see the top post.

> I wonder why the CI env is ignoring it

Maybe some other package that we have as a dependency used to have it as a dependency, but doesn't anymore.

valeriupredoi:
> Maybe some other package that we have as a dependency used to have it as a dependency, but doesn't anymore.

That can't be - I built two envs from scratch today, with the latest `environment.yml` - but with the latest conda 4.9.1 - so it might be worth keeping an eye. Whatevs, let's get this one in, but I'll think of a test (kind of a reverse test) to periodically test the dependncy list see if we have any unused deps

PR #852, opened 2020-10-29
Fix path settings for DKRZ/Mistral

bouweandela:
Several paths were not set correctly for finding data on Mistral, this pull request improves that.

### DIFF
--- a/esmvalcore/config-user.yml
+++ b/esmvalcore/config-user.yml
@@ -85,17 +85,20 @@ profile_diagnostic: false
 #  CORDEX: /mnt/lustre02/work/ik1017/C3SCORDEX/data/c3s-cordex/output
 #  OBS: /mnt/lustre02/work/bd0854/DATA/ESMValTool2/OBS
 #  OBS6: /mnt/lustre02/work/bd0854/DATA/ESMValTool2/OBS
-#  obs4mips: /mnt/lustre02/work/bd0854/DATA/ESMValTool2/OBS 
-#  ana4mips: /mnt/lustre02/work/bd0854/DATA/ESMValTool2/OBS 
+#  obs4mips: /mnt/lustre02/work/bd0854/DATA/ESMValTool2/OBS
+#  ana4mips: /mnt/lustre02/work/bd0854/DATA/ESMValTool2/OBS
+#  native6: /mnt/lustre02/work/bd0854/DATA/ESMValTool2/RAWOBS
+#  RAWOBS: /mnt/lustre02/work/bd0854/DATA/ESMValTool2/RAWOBS
 #drs:
 #  CMIP6: DKRZ
 #  CMIP5: DKRZ
 #  CMIP3: DKRZ
-#  CORDEX: DKRZ
-#  OBS: DKRZ
-#  OBS6: DKRZ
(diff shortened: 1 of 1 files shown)


bouweandela:
@bjoernbroetz @remi-kazeroni If one of you could check that the settings are now correct would be great!

esmvalcore/config-user.yml
```diff
@@ -85,17 +85,20 @@ profile_diagnostic: false
...
 #  CMIP6: DKRZ
 #  CMIP5: DKRZ
 #  CMIP3: DKRZ
-#  CORDEX: DKRZ
-#  OBS: DKRZ
-#  OBS6: DKRZ
-#  obs4mips: DKRZ
-#  ana4mips: DKRZ
+#  CORDEX: BADC
+#  obs4mips: default
```
remi-kazeroni, line 97: `# CORDEX: default` no?

esmvalcore/config-user.yml
```diff
@@ -85,17 +85,20 @@ profile_diagnostic: false
...
 #drs:
 #  CMIP6: DKRZ
 #  CMIP5: DKRZ
 #  CMIP3: DKRZ
-#  CORDEX: DKRZ
-#  OBS: DKRZ
-#  OBS6: DKRZ
-#  obs4mips: DKRZ
-#  ana4mips: DKRZ
+#  CORDEX: BADC
```
remi-kazeroni, line 96: Shouldn't it be: `# CORDEX: default`?

remi-kazeroni (commented):
I think there is one `drs` to change (`CORDEX`). The rest is good.
But I am not too sure about: `CMIP5` at DKRZ. Is there a difference between `/mnt/lustre02/work/bd0854/DATA/ESMValTool2/CMIP5_DKRZ` and `/mnt/lustre01/work/kd0956/CMIP5/data/cmip5/output1` @hb326 @axel-lauer ?

bettina-gier:
The first one is a symbolic link to the second one, so it's all the same.

esmvalcore/config-user.yml
bouweandela: Here is an example filename:
```
EUR-11/KNMI/MPI-M-MPI-ESM-LR/historical/r1i1p1/KNMI-RACMO22E/v1/mon/tas/v20190619/tas_EUR-11_MPI-M-MPI-ESM-LR_historical_r1i1p1_KNMI-RACMO22E_v1_mon_195001-195012.nc
```
so the directories are named according to
https://github.com/ESMValGroup/ESMValCore/blob/5c094eecac2f2ff652bb34bbf883c8c5ea4ad419/esmvalcore/config-developer.yml#L233
while `default` is a structure with no subdirectories:
https://github.com/ESMValGroup/ESMValCore/blob/5c094eecac2f2ff652bb34bbf883c8c5ea4ad419/esmvalcore/config-developer.yml#L231

remi-kazeroni: I see. Thanks for pointing this out

remi-kazeroni (approved):

remi-kazeroni:
Good to merge in my opinion

PR #865, opened 2020-11-19
Refactored mask regridding for irregular grids (fixes #772)

### LINKED ISSUE #772: Unexpected behaviour when regridding / extracting a region
Hello, I'm attempting to use ESMValTool with data from the Met Office HadGEM3-RA regional model.
[ESMValTool_issue_files.zip](https://github.com/ESMValGroup/ESMValCore/files/5188809/ESMValTool_issue_files.zip)

I have successfully run my model output through an in house CMORization / mip_convert suite and added project lines to read the files to a custom `config-developer.yml ` file (using the same CMOR tables as for CORDEX). My simple recipe and diagnostic runs without any errors, I've also not noticed any relevant warnings.

However, something is going wrong with the regridding steps (and if I remove regridding steps from the recipe, I get problems with area extraction, as I'll explain later).

I set `save_intermediary_cubes` to true in my `config_user.yml` in order to examine the state of the data as it moves through the pre processor. Regardless of regridding scheme chosen, the data created immediately after the regridding step (in `05_regrid.nc`) has all been set to zero, across every single grid point. (The data prior to this step is still intact)

I wondered if this is perhaps related to this open pull request https://github.com/ESMValGroup/ESMValCore/pull/185 (If so, is the code under review likely to fix my problem?) 

I also tried removing the regridding steps from my recipe entirely. In this case although again my recipe ran without any warnings or errors, the output was not as expected and I noticed something goes wrong at the point that the region extraction is carried out. This time the data immediately after the region extraction has all been set to 1e20 (masking?). Again, the preprocessed data immediately prior to this step is fine.

Any help appreciated, I'm hoping to use ESMValTool to perform analysis of some downscaled climate projections for the Caribbean that we have  recently completed.

Thanks

zklaus:
**Tasks**

-   [x] [Create an issue](https://github.com/ESMValGroup/ESMValCore/issues). This PR addresses issue #772.
-   [ ] Add unit tests
-   [x] If writing a new/modified preprocessor function, please update the [documentation](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/preprocessor.html). This is only an implementation change.
-   [x] Please use `yamllint` to check that your YAML files do not contain mistakes. No yaml files added or changed.
-   [x] If you make backward incompatible changes to the recipe format, make a new pull request in the [ESMValTool repository](https://github.com/ESMValGroup/ESMValTool) and add the link below. No changes to the recipe format.

* * *

Closes #772

### DIFF
--- a/esmvalcore/preprocessor/_regrid_esmpy.py
+++ b/esmvalcore/preprocessor/_regrid_esmpy.py
@@ -144,6 +144,29 @@ def get_representant(cube, ref_to_slice):
     return cube[rep_ind]
 
 
+def regrid_mask_2d(src_field, src_data, dst_field,
+                   regridding_arguments, regrid_method, mask_threshold):
+    """Regrid the mask from the source field to the destination."""
+    original_src_mask = np.ma.getmaskarray(src_data)
+    src_field.data[...] = ~original_src_mask.T
+    src_mask = src_field.grid.get_item(ESMF.GridItem.MASK,
+                                       ESMF.StaggerLoc.CENTER)
+    src_mask[...] = original_src_mask.T
+    center_mask = dst_field.grid.get_item(ESMF.GridItem.MASK,
+                                          ESMF.StaggerLoc.CENTER)
+    center_mask[...] = 0
+    mask_regridder = ESMF.Regrid(
+        src_mask_values=MASK_REGRIDDING_MASK_VALUE[regrid_method],
+        dst_mask_values=np.array([]),
+        **regridding_arguments)
+    regr_field = mask_regridder(src_field, dst_field)
+    dst_mask = regr_field.data[...].T < mask_threshold
+    center_mask[...] = dst_mask.T
+    if not dst_mask.any():
+        dst_mask = np.ma.nomask
+    return dst_mask
+
+
 def build_regridder_2d(src_rep, dst_rep, regrid_method, mask_threshold):
     """Build regridder for 2d regridding."""
     dst_field = cube_to_empty_field(dst_rep)
@@ -155,23 +178,9 @@ def build_regridder_2d(src_rep, dst_rep, regrid_method, mask_threshold):
         'unmapped_action': ESMF.UnmappedAction.IGNORE,
         'ignore_degenerate': True,
     }
-    if np.ma.is_masked(src_rep.data):
-        src_field.data[...] = ~src_rep.data.mask.T
-        src_mask = src_field.grid.get_item(ESMF.GridItem.MASK,
-                                           ESMF.StaggerLoc.CENTER)
-        src_mask[...] = src_rep.data.mask.T
-        center_mask = dst_field.grid.get_item(ESMF.GridItem.MASK,
-                                              ESMF.StaggerLoc.CENTER)
-        center_mask[...] = 0
-        mask_regridder = ESMF.Regrid(
-            src_mask_values=MASK_REGRIDDING_MASK_VALUE[regrid_method],
-            dst_mask_values=np.array([]),
-            **regridding_arguments)
-        regr_field = mask_regridder(src_field, dst_field)
-        dst_mask = regr_field.data[...].T < mask_threshold
-        center_mask[...] = dst_mask.T
-    else:
-        dst_mask = False
+    dst_mask = regrid_mask_2d(src_field, src_rep.data, dst_field,
+                              regridding_arguments,
+                              regrid_method, mask_threshold)
     field_regridder = ESMF.Regrid(src_mask_values=np.array([1]),
                                   dst_mask_values=np.array([1]),
                                   **regridding_arguments)


bouweandela (commented):
Looks good. Did you want to include a unit test that shows that the bug has been solved?

Does this pull request completely solve #772, or are there remaining issues to be tackled?

valeriupredoi:
cheers @zklaus :beer: Unfortunately I have gotten rid of my test data set and I think it's prob best if @thomascrocker tested it anyway since he'll pay more attention to the issues he's already seen - @thomascrocker could you give it one more test please mate? I agree with @bouweandela in [comment](https://github.com/ESMValGroup/ESMValCore/pull/865#pullrequestreview-548047293) - nice fix!

valeriupredoi (approved):
approval pending testing and unit test inclusion (if Klaus deems it necessary, I think it'd be good to have one but am not too fussed)

thomascrocker:
> cheers @zklaus 🍺 Unfortunately I have gotten rid of my test data set and I think it's prob best if @thomascrocker tested it anyway since he'll pay more attention to the issues he's already seen - @thomascrocker could you give it one more test please mate? I agree with @bouweandela in [comment](https://github.com/ESMValGroup/ESMValCore/pull/865#pullrequestreview-548047293) - nice fix!

Thanks for sorting the fix out. I'm up against a Christmas deadline at the moment, but I'll make sure to test this with my data in the new year.
Cheers

thomascrocker:
Happy new year all!
The fix solves the regridding, but there are still problems with variable attributes that need to be addressed. See https://github.com/ESMValGroup/ESMValCore/issues/772#issuecomment-730410103

valeriupredoi:
hey fellas, what's the status here: will the work needed to address the attributes be done in a separate PR? I'd advocate we merge this now and do the remainder of the work in a separate PR if it's not inter-dependent with the code changes here :beer:

thomascrocker:
I was just looking into regridding again with CORDEX and had a go at using this branch. There are a couple of problems..
This https://github.com/ESMValGroup/ESMValCore/issues/772#issuecomment-730410103 problem with the coord attributes is still outstanding..
And a potentially bigger (dataset) problem...

Some (but not all) of the CORDEX data on ESGF, for example:
http://esg-dn1.nsc.liu.se/thredds/fileServer/esg_dataroot4/cordexdata/cordex/output/EUR-11/SMHI/NCC-NorESM1-M/historical/r1i1p1/SMHI-RCA4/v1/mon/tas/v20180820/tas_EUR-11_NCC-NorESM1-M_historical_r1i1p1_SMHI-RCA4_v1_mon_198101-199012.nc
Does not have any bounds on it's latitude and longitude co-ordinates. The regridding needs bounds on the grid in order to work. Since these are multidimensional coordinates (due to the rotated grid) they can't be easily added to the data as part of automated fixes using the `guess_bounds` iris function.

thomascrocker:
> Thanks @thomascrocker. You are right. We are dealing with this precise problem at the moment at our institute. I think the best approach is probably to store known-good latitude-longitude coords including bounds under a hash of the rlat, rlon coords. In principle it should be straight-forward to calculate the bounds from the 1d bounds using [`iris.analysis.cartography`](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/analysis/cartography.html?highlight=rotate#iris.analysis.cartography.unrotate_pole) or [`cartopy.crs.RotatedPole`](https://scitools.org.uk/cartopy/docs/latest/crs/projections.html?highlight=rotate#cartopy.crs.RotatedPole), but storing known good coords might save us some grieve down the line because the CORDEX data tends to have some very small difference in coords in the file on the order of picometers.
> 
> Unfortunately, this might have to work until after the June release.

Thanks, that's a good idea. The MOHC EUR-11 domain files for example do have the correct lat and lon coords, with bounds, so I may see if I can hack together a dataset fix that reads that information and applies it to all CORDEX data on that domain.

bouweandela:
Is the problem with the bounds somehow overlapping with #184?

thomascrocker:
> Is the problem with the bounds somehow overlapping with #184?

Yes you're right. Good spot, it's the same problem. Although as I've mentioned not all CORDEX data is missing bounds, data from some institutes does have it.

@zklaus is correct though that despite the fact that CORDEX runs over the same domain should all be on the same grid, there are often very small rounding error style differences in the grid specification from one institute to another, so the approach of storing template grid specifications for each domain somewhere and then applying them as a fix to CORDEX data on load is a good idea to save the inevitable errors that would result from doing any multi model statistics or arithmetic.

bouweandela:
@zklaus Would that solution then replace #184?

zklaus:
I think this PR and #184 solve different aspects of regridding CORDEX data. This one is really addressing a bug in the regridding, #184 is addressing a common problem in CORDEX data.

bouweandela:
I did not mean this pull request, I meant: would the solution proposed in https://github.com/ESMValGroup/ESMValCore/pull/865#issuecomment-838034466 replace #184?

zklaus:
Well, yes, though it was basically already proposed in https://github.com/ESMValGroup/ESMValCore/pull/184#issuecomment-626573621 and seems to have been picked up subsequently by @mwjury, but in a slightly unorthodox way by attaching some textfiles containing python code in comments on the ticket.

Can we perhaps move forward like this: Let's finish this PR ignoring the issue of missing bounds on cordex. Let's attack that in #184. For that purpose, could you, @thomascrocker, head over there and sort out with @mwjury how to solve it?

thomascrocker:
> Well, yes, though it was basically already proposed in [#184 (comment)](https://github.com/ESMValGroup/ESMValCore/pull/184#issuecomment-626573621) and seems to have been picked up subsequently by @mwjury, but in a slightly unorthodox way by attaching some textfiles containing python code in comments on the ticket.
> 
> Can we perhaps move forward like this: Let's finish this PR ignoring the issue of missing bounds on cordex. Let's attack that in #184. For that purpose, could you, @thomascrocker, head over there and sort out with @mwjury how to solve it?

Makes sense to keep this PR as fixing the specific regridding issue, and keep #184 to deal with coord bounds. I will put any further work on that problem there.

valeriupredoi:
this is looking good @zklaus :beer:

PR #868, opened 2020-11-20
Implement importable config object in experimental API submodule

stefsmeets:
Hi all, we started working on an importable config object. This PR can be used to track the progress. For more discussion see this issue: https://github.com/ESMValGroup/ESMValCore/issues/498

This PR is a second attempt at implementing an importable config object, since #785 became 'too big to review'.

For this PR, I have kept the implementation completely separate from the current codebase in the `future` directory. In that case we can move faster, as the API will be somewhat experimentaland likely to change. Any useful bits can easily be moved down once we start to depen.

- To start, I have added an importable config. A key feature is that all variables are validated. That means that unknown input will be rejected, and wrong values are easy to catch.
- The config is strictly meant for interactive use at the moment.
- I have made `config-user` valid by default. At present, users will have to update the config-user in order to do any work.
- If `.esmvaltool/config-user.yml` does not exist, it will default to the one from the codebase.
- Unlike the previous PR, the goal of this PR is NOT to change any config parameters. That will be done in a follow-up PR if still deemed necessary.

For a demonstration of the implementation, see this notebook here: https://gist.github.com/stefsmeets/dccd59d8c1f11768a9a5f297d7faeaab 

**Tasks**

-   [x] Add unit tests
-  [x] [Write documentation](https://esmvaltool--868.org.readthedocs.build/projects/ESMValCore/en/868/api/esmvalcore.api.html)

* * *

Described by #498 
Successor of #785 
Documentation: https://esmvaltool--868.org.readthedocs.build/projects/ESMValCore/en/868/api/esmvalcore.api.html


stefsmeets:
Some points of feedback that is still relevant from @Peter9192 from #785 :
- Should we expose all the methods of the cfg object? E.g. clear. I used this, and now I cannot get the default cfg back. Why would someone want to use pop or popitem? And what is the use case for find_all? I suppose it would be better to add things later than to have to remove them again.
- Related: I can't help but feeling that some of the config options are really very specific runtime arguments. If the session is very tightly coupled to a specific run, would it make sense to pass those runtime options only to the session, and not to the config?

esmvalcore/future/config/_config_object.py
```diff
@@ -0,0 +1,189 @@
...
+
+import yaml
+
+import esmvalcore
+
+from ._config_validators import _validators
+from ._validated_config import ValidatedConfig
+
+
+class ESMValCoreConfig(ValidatedConfig):
```
nielsdrost, line 15: I'm not sure we should use the term "Core" in these user facing objects. How about ESMValToolConfig, or perhaps even ESMValConfig? :-)

esmvalcore/future/config/_config_object.py
```diff
@@ -0,0 +1,189 @@
...
+    @staticmethod
+    def load_from_file(filename):
+        """Reload user configuration from the given file."""
+        path = Path(filename).expanduser()
+        if not path.exists():
+            try_path = USER_CONFIG_DIR / filename
+            if try_path.exists():
+                path = try_path
+            else:
+                raise FileNotFoundError(f'No such file: `{filename}`')
```
nielsdrost, line 29: Should we also mention we tried the try_path here?

esmvalcore/future/config/_config_object.py
```diff
@@ -0,0 +1,189 @@
...
+
+        _load_user_config(path)
+
+    def start_session(self, name: str):
```
nielsdrost, line 33: nice, I like this clean interface :-)

nielsdrost (commented):
Looks good to me, thanks @stefsmeets!

esmvalcore/future/config/_config_object.py
stefsmeets: I'm struggling a lot with the naming conventions, because I find them a bit inconsistent at the moment. I think that if we import something from ESMValCore, it should be named ESMValCoreSomething. Also for better decoupling, because ESMValTool also exists. `ESMValConfig` is a good compromise that avoids the problem altogether.

esmvalcore/future/config/_config_object.py
```diff
@@ -0,0 +1,189 @@
...
+        mapping = {}
+
+    global CFG
+    global CFG_ORIG
+
+    CFG.clear()
+    CFG.update(CFG_DEFAULT)
+    CFG.update(mapping)
+
+    CFG_ORIG = ESMValCoreConfig(CFG.copy())
```
Peter9192, line 159: I'm not sure if I understand what this is for

Peter9192 (approved):
Like :+1:

esmvalcore/future/config/_config_object.py
stefsmeets: I think I avoided using Config to avoid naming conflicts with other things called (some variation of) config.

esmvalcore/future/config/_config_object.py
stefsmeets: > 
> 
> I'm not sure if I understand what this is for

This is to ensure we can restore the config if someone deletes it by accident 😅 Basically I took this from `.matplotlibrc`. Maybe an idea to get rid of it if we do not foresee a use for it.

esmvalcore/future/config/_config_object.py
stefsmeets: Yes, good point!

Peter9192:
One more thing: maybe we can print (on import?) that this is still in development and subject to frequent changes?

esmvalcore/future/config/_config_object.py
stefsmeets: What about `UserConfig`?

bouweandela: I like `Config` as suggested by @Peter9192 , if someone as a naming conflict they can always write
```python
from esmvalcore import Config as ESMValCoreConfig
```
to avoid it

bouweandela:
Would it be an idea to call the new module `esmvalcore.experimental` instead of `esmvalcore.future`? That's what they call it in [iris](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/experimental.html) too and I think it covers the content a bit better because possibly not everything from this module will move to the main tool.

doc/api/esmvalcore.api.rst
```diff
@@ -0,0 +1,139 @@
...
+
+A ``Session`` is very similar to the config.
+It is also a dictionary, and copies all the keys from the ``Config``.
+At this moment, ``session`` is essentially a copy of ``CFG``:
+
+.. code-block:: python
+
+    >>> print(session == CFG)
+    True
+    >>> session['output_dir'] = '~/session_output'
```
bouweandela, line 107:
```suggestion
    >>> session['output_dir'] = '~/esmvaltool_output'
```

bouweandela: It might be nice to change session['output dir']` to what is stored in the `session_dir` attribute below, to avoid confusion: the output of the session will be written to there. I also noticed that if I change session['output_dir'], the session_dir attribute is not changed

doc/api/esmvalcore.api.rst
```diff
@@ -0,0 +1,139 @@
...
+    >>> session.run_dir
+    /home/user/esmvaltool_output/my_session_20201203_155821/run
+    >>> session.work_dir
+    /home/user/esmvaltool_output/my_session_20201203_155821/work
+    >>> session.preproc_dir
+    /home/user/esmvaltool_output/my_session_20201203_155821/preproc
+    >>> session.plot_dir
+    /home/user/esmvaltool_output/my_session_20201203_155821/plots
+
+``Session`` objects are persistent, so multiple sessions can be initiated from the ``Config``.
```
bouweandela, line 127: What do you mean by persistent?

doc/api/esmvalcore.api.rst
```diff
@@ -0,0 +1,139 @@
...
+.. code-block:: python
+
+    >>> print(session == CFG)
+    True
+    >>> session['output_dir'] = '~/session_output'
+    >>> print(session == CFG)  # False
+    False
+
+A ``Session`` also knows about the directories where the data will stored.
+The session name is used to previx the directories.
```
bouweandela, line 112:
```suggestion
The session name is used to prefix the directories.
```

esmvalcore/experimental/config/_config_object.py
```diff
@@ -0,0 +1,191 @@
...
+            try_path = USER_CONFIG_DIR / filename
+            if try_path.exists():
+                path = try_path
+            else:
+                raise FileNotFoundError(f'Cannot find: `{filename}`'
+                                        f'locally or in `{try_path}`')
+
+        _load_user_config(path)
+
+    def reload(self):
```
bouweandela, line 34: Maybe `load_from_file` and `reload` could be combined into a single function, e.g. called `load`? I don't really see why we need two public functions for loading the configuration file

esmvalcore/experimental/config/_config_object.py
```diff
@@ -0,0 +1,191 @@
...
+import esmvalcore
+
+from ._config_validators import _validators
+from ._validated_config import ValidatedConfig
+
+
+class Config(ValidatedConfig):
+    """ESMValTool configuration object."""
+
+    validate = _validators
```
bouweandela, line 18: This could probably be made private, I think this will only confuse users
```suggestion
    _validate = _validators
```

esmvalcore/experimental/config/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""ESMValTool config module."""
+
+from ._config_object import CFG
+
+__all__ = [
+    'CFG',
+]
```
bouweandela, line 7:
```suggestion
    'Config',
    'Session',
]
It would be nice to include these here
```

doc/api/esmvalcore.api.rst
```diff
@@ -0,0 +1,139 @@
...
+    >>> session.plot_dir
+    /home/user/esmvaltool_output/my_session_20201203_155821/plots
+
+``Session`` objects are persistent, so multiple sessions can be initiated from the ``Config``.
+
+
+API reference
+*************
+
+.. autoclass:: esmvalcore.experimental.config._config_object.Config
```
bouweandela, line 133:
```suggestion
.. autoclass:: esmvalcore.experimental.config.Config
```
It would be better to only show public parts of the API. If the `Config` class is interesting to users, it should be part of the public API. Same for the `Session` class below.

esmvalcore/experimental/config/_config_object.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+def get_user_config_location():
+    """Get the user config location by looking in the expected places.
+
+    Check if environment variable `ESMVALTOOL_CONFIG` exists, otherwise
+    use the default location in the user config dir (`~/.esmvaltool`).
+    """
+    try:
+        config_location = Path(os.environ['ESMVALTOOL_CONFIG'])
```
bouweandela, line 172: I'm not too keen on supporting environmental variables, because they lead to very hard to debug situations, e.g. if people set them and then run into trouble a few years later because there was something set in their bashrc file that they completely forgot about

esmvalcore/experimental/config/_config_object.py
```diff
@@ -0,0 +1,191 @@
...
+    except IOError:
+        if raise_exception:
+            raise
+        mapping = {}
+
+    global CFG
+
+    CFG.clear()
+    CFG.update(CFG_DEFAULT)
+    CFG.update(mapping)
```
bouweandela, line 162: These functions could be methods on the `Config` class, that would be cleaner and also avoid the need for the `global` use

esmvalcore/experimental/__init__.py
```diff
@@ -0,0 +1,14 @@
+"""ESMValCore experimental API module."""
+
+from ._warnings import warnings
+
+warnings.warn(
+    '\n  Thank you for trying out the new ESMValCore API.'
+    '\n  Note that this API is experimental and may be subject to change.'
+    '\n  More info: https://github.com/ESMValGroup/ESMValCore/issues/498', )
+
+from .config import CFG  # noqa: E402
```
bouweandela, line 10:
```suggestion
from .config import CFG
```
is `noqa` needed here? I think flake8 recognizes that `CFG` is used in `__all__`

bouweandela (commented):
Looks like a good start, though I do have some minor comments on the functionality and implementation. Thanks a lot for making the pull request smaller, now I could review it in just a bit more than an hour!

Regarding the exceptions and hiding the stack trace, the idea would be to hide the stack trace from standard out (and if applicable from main_log.txt, but not from main_log_debug.txt) if a well-known error occurs, so the user gets an easy to understand error message. At the moment this code raises many different exceptions, it would be good to reduce this to a single exception called e.g. `ConfigError` that inherits from the `SuppressedError` that we can then hide the stack trace of.

esmvalcore/config-user.yml
```diff
@@ -41,19 +41,19 @@ profile_diagnostic: false
...
 # Directory structure for input data: [default]/BADC/DKRZ/ETHZ/etc
 # See config-developer.yml for definitions.
-#drs:
-#  CMIP5: default
-#  CORDEX: default
-#  OBS: default
+drs:
+  CMIP5: default
+  CORDEX: default
+  OBS: default
```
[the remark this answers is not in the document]
stefsmeets: We cannot do that if we want to work in an interactive environment. We must have defaults for drs/rootpath (even if they are empty).

esmvalcore/experimental/_exceptions.py
```diff
@@ -0,0 +1,24 @@
...
+
+class SuppressedError(Exception):
+    """Errors subclassed from SuppressedError hide the full traceback.
+
+    This can be used for simple user-facing errors that do not need the
+    full traceback.
+    """
+
+
+def _suppressed_hook(error, message, traceback):
```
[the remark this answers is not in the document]
stefsmeets: This is not an issue yet, but it is definitely possible.

esmvalcore/experimental/config/_config_object.py
stefsmeets: As static methods you mean? I am not sure, as they are really two different things. The globals are necessary to make the configuration importable.

esmvalcore/experimental/__init__.py
stefsmeets: E402 is about imports not being at the top of the file. I want to have the warning as the very first thing on import of `esmvalcore.experimental`, so it must come first. This is why I added the noqa.

esmvalcore/experimental/config/_config_object.py
stefsmeets: I have found them super useful for all sorts of scenarios. I can remove them for now.

doc/api/esmvalcore.api.rst
stefsmeets: Agreed. Although the `Session`/`Config` are not meant to be directly instantiated by users.

esmvalcore/experimental/config/__init__.py
stefsmeets: Disagreed, they are not meant to be called directly by users.

esmvalcore/experimental/config/_config_object.py
stefsmeets: Yeah, good point. I was wondering why `validate` showed up on the docs all of a sudden...

esmvalcore/experimental/config/_config_object.py
stefsmeets: Because they do two different things semantically. One loads specifically from a file, and one reloads the current configuration file. What does `load` do with no argument?

doc/api/esmvalcore.api.rst
stefsmeets: Yep, not a great choice of words...

esmvalcore/experimental/config/_config_validators.py
```diff
@@ -0,0 +1,284 @@
...
+    if not version:
+        version = 'a future version'
+
+    if current_version >= version:
+        warnings.warn(f"`{variable}` has been deprecated in {version}",
+                      ESMValToolDeprecationWarning)
+    else:
+        warnings.warn(f"`{variable}` will be deprecated in {version}.",
+                      ESMValToolDeprecationWarning,
+                      stacklevel=2)
```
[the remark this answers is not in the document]
stefsmeets: Yeah, perhaps better to use `removed` instead.

doc/api/esmvalcore.api.rst
stefsmeets: Good find, I changed it now so it works as expected. Session itself must be a dict so it can be passed to the recipe runner, otherwise I would have stored a copy of the config on the Session itself, but for now I wanted to keep it strictly compatible with the user config. I suggest we tackle this in a follow-up item once we can better chart out what we need this to be.

esmvalcore/config-user.yml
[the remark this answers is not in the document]
stefsmeets: > The reason that this is commented out, is that users complained that having this as the default configuration does not work/make sense for anyone.

Do users not complain that the current user config breaks ESMValTool by default? That, in my opinion, not a good user experience. I'd rather have some default that works out of the box.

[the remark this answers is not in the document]
bouweandela: The point was that esmvaltool would just start running and then crash complaining that the data was not present in a seemingly random place that the user never specified. I agree that the error message if you don't have the section or if it's empty should be improved though.

esmvalcore/experimental/config/_config_object.py
bouweandela: Without a filename it would just do what this function does now. If you do provide the filename, it would set `self['config_file'] = filename` first.

esmvalcore/experimental/config/_config_object.py
bouweandela: Thanks

esmvalcore/experimental/config/_config_object.py
bouweandela: > As static methods you mean?

Yes, static if needed. It would provide a better grouping of functionality

> The globals are necessary to make the configuration importable.

I meant the use of `global` statement, it's [discouraged by pylint](https://stackoverflow.com/questions/14503973/python-global-keyword-vs-pylint-w0603)

bouweandela:
It would also be great if you could have a look at the Codacy warnings

esmvalcore/experimental/config/_config_object.py
stefsmeets: I think I can actually get rid of the globals at this point.

esmvalcore/experimental/config/_config_validators.py
```diff
@@ -0,0 +1,284 @@
...
+            # The condition on this list comprehension will preserve the
+            # behavior of filtering out any empty strings (behavior was
+            # from the original validate_stringlist()), while allowing
+            # any non-string/text scalar values such as numbers and arrays.
+            inp = [
+                scalar_validator(val) for val in inp
+                if not isinstance(val, str) or val
+            ]
+        else:
+            raise ValueError(
```
stefsmeets, line 81: Should we raise custom error instead of generic ValueError here (and the rest of this code)?

esmvalcore/config-user.yml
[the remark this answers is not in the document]
stefsmeets: What about if we leave `config-user.yml` as it is, but raise a warning if `rootpath`/`drs` is not defined.

stefsmeets:
> Regarding the exceptions and hiding the stack trace, the idea would be to hide the stack trace from standard out (and if applicable from main_log.txt, but not from main_log_debug.txt) if a well-known error occurs, so the user gets an easy to understand error message. At the moment this code raises many different exceptions, it would be good to reduce this to a single exception called e.g. `ConfigError` that inherits from the `SuppressedError` that we can then hide the stack trace of.

This works well... although that said, IPython uses a different mechanism altogether 🤦‍♂️

bouweandela (approved):

Issue #869, opened 2020-11-20
Numba causing installation issues with python=3.9


valeriupredoi:
conda=4.9.2 env builds well but installing doesn't:
```
    Running setup.py install for llvmlite ... error
    ERROR: Command errored out with exit status 1:
     command: /home/valeriu/miniconda3/envs/esmvalcore-492/bin/python3.9 -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-elou0qm8/llvmlite/setup.py'"'"'; __file__='"'"'/tmp/pip-install-elou0qm8/llvmlite/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-woct46sr/install-record.txt --single-version-externally-managed --compile --install-headers /home/valeriu/miniconda3/envs/esmvalcore-492/include/python3.9/llvmlite
         cwd: /tmp/pip-install-elou0qm8/llvmlite/
    Complete output (29 lines):
    running install
    running build
    got version from file /tmp/pip-install-elou0qm8/llvmlite/llvmlite/_version.py {'version': '0.34.0', 'full': 'c5889c9e98c6b19d5d85ebdd982d64a03931f8e2'}
    running build_ext
    /home/valeriu/miniconda3/envs/esmvalcore-492/bin/python3.9 /tmp/pip-install-elou0qm8/llvmlite/ffi/build.py
    LLVM version... Traceback (most recent call last):
      File "/tmp/pip-install-elou0qm8/llvmlite/ffi/build.py", line 105, in main_posix
        out = subprocess.check_output([llvm_config, '--version'])
      File "/home/valeriu/miniconda3/envs/esmvalcore-492/lib/python3.9/subprocess.py", line 420, in check_output
        return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
      File "/home/valeriu/miniconda3/envs/esmvalcore-492/lib/python3.9/subprocess.py", line 501, in run
        with Popen(*popenargs, **kwargs) as process:
      File "/home/valeriu/miniconda3/envs/esmvalcore-492/lib/python3.9/subprocess.py", line 947, in __init__
        self._execute_child(args, executable, preexec_fn, close_fds,
[...]

PR #870, opened 2020-11-20
Pin Python to less than 3.9

valeriupredoi:
https://github.com/ESMValGroup/ESMValCore/issues/869 - we're in the bogs with `python=3.9` and we need to fix that, until then we need a functional package; note that the latest environment throws 3 test fails as well, as seen from the GA tests [here](https://github.com/ESMValGroup/ESMValCore/runs/1431151389?check_suite_focus=true) - those need to be fixed, preferably in a different branch, let's get this one merged quickly before users start crying around. I'll have a look at those tests on Monday :+1:  - OK I figured it out - the tests are dying coz of #876 and are fixed in #878 :beer:

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -8,5 +8,5 @@ dependencies:
   - esmpy
   - iris>=2.2.1
   - graphviz
-  - python>=3.6  # if 3.7 lxml will not import correctly if <4.5.0
+  - python>=3.6,<3.9  # if 3.7 lxml will not import correctly if <4.5.0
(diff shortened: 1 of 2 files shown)


bouweandela:
Please also pin the Python version in `package/meta.yml`

valeriupredoi:
ya the CI setup is always escaping me for some reason, cheers dude, just did! :beer:

bouweandela (approved):

Issue #859, opened 2020-11-09
thetao CESM2 z-level ("olevel") needs fix to convert units from cm to m


valeriupredoi:
CESM2 (eg `/badc/cmip6/data/CMIP6/CMIP/NCAR/CESM2/historical/r1i1p1f1/Omon/thetao/gn/latest/thetao_Omon_CESM2_historical_r1i1p1f1_gn_185001-201412.nc` on Jasmin) has depth (generic vertical level, `olevel`) units in `centimeters` (and its bounds in meters!) but the cmor checks in the preprocessor are not picking this up). Run test:
```yaml
---
documentation:
  description: |
    Test

  authors:
    - righi_mattia

diagnostics:
  BerkeleyEarth:
    description: BerkeleyEarth check
    variables:
      thetao:
    additional_datasets:
#/badc/cmip6/data/CMIP6/CMIP/NCAR/CESM2/historical/r1i1p1f1/Omon/thetao/gn/latest/thetao_Omon_CESM2_historical_r1i1p1f1_gn_185001-201412.nc
      - {dataset: CESM2, project: CMIP6, exp: historical, mip: Omon, ensemble: r1i1p1f1, grid: gn, start_year: 1990, end_year: 1995}
    scripts: null
```
I have forgotten - are we expecting checks on `olevel` or this is something we should fix via a fix file? seems a bit odd since it's  a valid dimensional coordinate and not even `--check_level strict` picks up the weird units and wonky bounds :beer:

valeriupredoi:
amigos ping @jvegasbsc

jvegasbsc:
> I have forgotten - are we expecting checks on olevel or this is something we should fix via a fix file? seems a bit odd since it's a valid dimensional coordinate and not even --check_level strict picks up the weird units and wonky bounds 🍺

@sloosvel has an open PR for generic levels: #598

Let's get it merged before this

valeriupredoi:
sweet - can I reviews it?

valeriupredoi:
arite, #598 is now in da `master` and we can go ahead and write a fix for that CESM2 broken data, I'll do it today hopefully

zklaus:
@valeriupredoi are you still up for this? In that case I'll change the assignment.

valeriupredoi:
yeah, this totally fell off my radar scope - yes, I'll assign meself

Issue #840, opened 2020-10-20
Dataset problem: Hybrid sigma pressure coordinate points in CESM models


sloosvel:
**Describe the dataset issue**
Due to #598 , the CMOR checker will be able to detect issues in the hybrid sigma pressure coordinate for models that so far were passing the checks. 

Some fixes are already implemented for CESM2 models, but with the new checks it turns out that the level coordinate has wrong negative values as well as wrong units ('hPa' instead of '1') and the already implemented fixes are not enough:

``` 
netcdf cl_Amon_CESM2_historical_r1i1p1f1_gn_185001-201412 {

...

lev = -3.64346569404006, -7.59481964632869, -14.3566322512925,
    -24.6122200042009, -35.9232500195503, -43.1937500834465,
    -51.6774989664555, -61.5204982459545, -73.7509578466415,
    -87.8212302923203, -103.317126631737, -121.547240763903,
    -142.994038760662, -168.225079774857, -197.908086702228,
    -232.828618958592, -273.910816758871, -322.241902351379,
    -379.100903868675, -445.992574095726, -524.687174707651,
    -609.778694808483, -691.389430314302, -763.404481112957,
    -820.858368650079, -859.53476652503, -887.020248919725,
    -912.644546944648, -936.198398470879, -957.485479535535,
    -976.325407391414, -992.556095123291 ; 
}
```
According to [cfconventions](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.0/build/apd.html):
`The hybrid sigma-pressure coordinate for level k is defined as a(k)+b(k) or ap(k)/p0+b(k), as appropriate.`

I guess the points should be recalculated using one of those formulas? I do not have much experience with these types of coordinates though. 

- Project (CMIP5/CMIP6/OBS/obs4mips/etc): CMIP6
- Full dataset description (fill out as many as you know, please):
  - dataset: CESM2
  - experiment: historical
  - mip: Amon
  - grid: gn
  - version: v20190308
  - variable used: cl
- Problems encountered (please describe what the actual problems are: e.g. wrong standard name, issue with dimensions): Explained above
- Pointer to existing copy of data on ESGF node (it would be very useful if you could provide a physical
[...]

Issue #515, opened 2020-02-26
Dataset problem: CESM2 marine depth units are cm instead of meters (gn grid)


ledm:
**Describe the dataset issue**
also see https://github.com/ESMValGroup/ESMValCore/issues/859 (apparent duplicate)

The CMIP6 model CESM2 seems to have all its depth dimension units set to Unit('centimeters') instead of the standard meters. I've seen it in every marine variables that I've looked at in the `gn` grid. The `gr` grid seems to be working fine though.

For instance, on jasmin, some files that don't work are:
```
/badc/cmip6/data/CMIP6/CMIP/NCAR/CESM2/piControl/r1i1p1f1/Omon/thetao/gn/latest/thetao_Omon_CESM2_piControl_r1i1p1f1_gn_070001-079912.nc
/badc/cmip6/data/CMIP6/CMIP/NCAR/CESM2/piControl/r1i1p1f1/Omon/ph/gn/latest/ph_Omon_CESM2_piControl_r1i1p1f1_gn_000101-009912.nc
```

- Project: CMIP6
- Full dataset decription (fill out as many as you know, please):
  - dataset: all marine datasets (that I've checked)
  - experiment: piControl, hist, ssp* 
  - mip: Oyr, Omon, Oday
  - grid: gn
  - version: latest
  - variable used:all marine datasets (that I've checked)

zklaus:
Clarification request: Are the values of the depth consistent with the wrong units or not?

ledm:
It's a strange one. The cube's `lev` units are wrong, they should be recorded as `Unit(m)`. However, the depth coordinate `points` appear to be in centimetres, but the `bounds` appear to be in meters: ie:

```
 c.coords('lev')
[DimCoord(array([5.00000000e+02, 1.50000000e+03, 2.50000000e+03, 3.50000000e+03,
       4.50000000e+03, 5.50000000e+03, 6.50000000e+03, 7.50000000e+03,
       8.50000000e+03, 9.50000000e+03, 1.05000000e+04, 1.15000000e+04,
       1.25000000e+04, 1.35000000e+04, 1.45000000e+04, 1.55000000e+04,
       1.65098398e+04, 1.75479043e+04, 1.86291270e+04, 1.97660273e+04,
       2.09711387e+04, 2.22578281e+04, 2.36408828e+04, 2.51370156e+04,
       2.67654199e+04, 2.85483652e+04, 3.05119219e+04, 3.26867988e+04,
[...]

PR #882, opened 2020-11-25
Added fixes for hybrid level coordinates of CESM2 models

schlunma:
This PR adds fixes for `cl`, `cli` and `clw` for the CESM2 models.

**Tasks**

-   [x] Add unit tests

* * *

Closes ESMValGroup/ESMValTool#1903.

### DIFF
--- a/esmvalcore/cmor/_fixes/cmip6/cesm2.py
+++ b/esmvalcore/cmor/_fixes/cmip6/cesm2.py
@@ -1,12 +1,16 @@
 """Fixes for CESM2 model."""
 from shutil import copyfile
 
-from netCDF4 import Dataset
 import numpy as np
+from netCDF4 import Dataset
 
 from ..fix import Fix
-from ..shared import (add_scalar_depth_coord, add_scalar_height_coord,
-                      add_scalar_typeland_coord, add_scalar_typesea_coord)
+from ..shared import (
+    add_scalar_depth_coord,
+    add_scalar_height_coord,
+    add_scalar_typeland_coord,
+    add_scalar_typesea_coord,
+)
 from .gfdl_esm4 import Siconc as Addtypesi
 
 
@@ -23,27 +27,6 @@ class Cl(Fix):
             'atmosphere_hybrid_sigma_pressure_coordinate')
         dataset.close()
         return new_path
-
-    def fix_data(self, cube):
-        """Fix data.
-
-        Fixed ordering of vertical coordinate.
-
-        Parameters
-        ----------
-        cube: iris.cube.Cube
-            Input cube to fix.
-
-        Returns
-        -------
-        iris.cube.Cube
-
-        """
-        (z_axis,) = cube.coord_dims(cube.coord(axis='Z', dim_coords=True))
-        indices = [slice(None)] * cube.ndim
-        indices[z_axis] = slice(None, None, -1)
-        cube = cube[tuple(indices)]
-        return cube
 
     def fix_file(self, filepath, output_dir):
         """Fix hybrid pressure coordinate.
@@ -76,6 +59,30 @@ class Cl(Fix):
         dataset.variables['b_bnds'][:] = dataset.variables['b_bnds'][::-1, :]
         dataset.close()
         return new_path
+
+    def fix_metadata(self, cubes):
+        """Fix ``atmosphere_hybrid_sigma_pressure_coordinate``.
+
+        See discussion in #882 for more details on that.
+
+        Parameters
+        ----------
+        cubes : iris.cube.CubeList
+            Input cubes.
+
+        Returns
+        -------
+        iris.cube.CubeList
+
+        """
+        cube = self.get_cube_from_list(cubes)
+        lev_coord = cube.coord(var_name='lev')
+        a_coord = cube.coord(var_name='a')
+        b_coord = cube.coord(var_name='b')
+        lev_coord.points = a_coord.core_points() + b_coord.core_points()
+        lev_coord.bounds = a_coord.core_bounds() + b_coord.core_bounds()
+        lev_coord.units = '1'
+        return cubes
 
 
 Cli = Cl
--- a/tests/integration/cmor/_fixes/cmip6/test_cesm2.py
+++ b/tests/integration/cmor/_fixes/cmip6/test_cesm2.py
@@ -97,15 +97,21 @@ def test_cl_fix_file(mock_get_filepath, tmp_path, test_data_path):
 
 
 @pytest.fixture
-def cl_cube():
+def cl_cubes():
     """``cl`` cube."""
     time_coord = iris.coords.DimCoord(
         [0.0, 1.0], var_name='time', standard_name='time',
         units='days since 1850-01-01 00:00:00')
+    a_coord = iris.coords.AuxCoord(
+        [0.1, 0.2, 0.1], bounds=[[0.0, 0.15], [0.15, 0.25], [0.25, 0.0]],
+        var_name='a', units='1')
+    b_coord = iris.coords.AuxCoord(
+        [0.9, 0.3, 0.1], bounds=[[1.0, 0.8], [0.8, 0.25], [0.25, 0.0]],
+        var_name='b', units='1')
     lev_coord = iris.coords.DimCoord(
-        [0.0, 1.0, 2.0], var_name='lev',
-        standard_name='atmosphere_hybrid_sigma_pressure_coordinate', units='1',
-        attributes={'positive': 'up'})
+        [999.0, 99.0, 9.0], var_name='lev',
+        standard_name='atmosphere_hybrid_sigma_pressure_coordinate',
+        units='hPa', attributes={'positive': 'up'})
     lat_coord = iris.coords.DimCoord(
         [0.0, 1.0], var_name='lat', standard_name='latitude', units='degrees')
     lon_coord = iris.coords.DimCoord(
@@ -122,30 +128,23 @@ def cl_cube():
         standard_name='cloud_area_fraction_in_atmosphere_layer',
         units='%',
         dim_coords_and_dims=coord_specs,
+        aux_coords_and_dims=[(a_coord, 1), (b_coord, 1)],
     )
-    return cube
+    return iris.cube.CubeList([cube])
 
 
-def test_cl_fix_data(cl_cube):
-    """Test ``fix_data`` for ``cl``."""
-    fix = Cl(None)
-    out_cube = fix.fix_data(cl_cube)
-    assert out_cube.shape == cl_cube.shape
-    np.testing.assert_allclose(out_cube.data,
-                               [[[[8, 9],
-                                  [10, 11]],
-                                 [[4, 5],
-                                  [6, 7]],
-                                 [[0, 1],
-                                  [2, 3]]],
-                                [[[20, 21],
-                                  [22, 23]],
-                                 [[16, 17],
-                                  [18, 19]],
-                                 [[12, 13],
-                                  [14, 15]]]])
-    np.testing.assert_allclose(out_cube.coord(var_name='lev').points,
-                               [2.0, 1.0, 0.0])
+def test_cl_fix_metadata(cl_cubes):
+    """Test ``fix_metadata`` for ``cl``."""
+    vardef = get_var_info('CMIP6', 'Amon', 'cl')
+    fix = Cl(vardef)
+    out_cubes = fix.fix_metadata(cl_cubes)
+    out_cube = out_cubes.extract_cube(
+        'cloud_area_fraction_in_atmosphere_layer')
+    lev_coord = out_cube.coord(var_name='lev')
+    assert lev_coord.units == '1'
+    np.testing.assert_allclose(lev_coord.points, [1.0, 0.5, 0.2])
+    np.testing.assert_allclose(lev_coord.bounds,
+                               [[1.0, 0.95], [0.95, 0.5], [0.5, 0.0]])
 
 
 def test_get_cli_fix():
--- a/tests/integration/cmor/_fixes/cmip6/test_cesm2_waccm.py
+++ b/tests/integration/cmor/_fixes/cmip6/test_cesm2_waccm.py
@@ -7,6 +7,7 @@ import iris
 import numpy as np
 import pytest
 
+from esmvalcore.cmor._fixes.cmip6.cesm2 import Cl as BaseCl
 from esmvalcore.cmor._fixes.cmip6.cesm2 import Tas as BaseTas
 from esmvalcore.cmor._fixes.cmip6.cesm2_waccm import Cl, Cli, Clw, Tas
 from esmvalcore.cmor.fix import Fix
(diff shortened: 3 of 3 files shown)


zklaus:
I don't think the diagnosis is correct. I explained this in some detail [in the issue](https://github.com/ESMValGroup/ESMValCore/issues/883#issuecomment-734214000). The gist is: Thanks to the `formula_terms` attribute `lev` **is** `a` and thus is absolutely used for the calculation.

schlunma:
Yes @zklaus, for the datasets shown in #883 (with hybrid height) you're absolutely right, apologies for that.

I jumped to conclusions here: I only analyzed datasets with **hybrid pressure** coordinates, which show the behavior I explain in the description of this PR: for example, CESM2 fails with

```
esmvalcore.cmor.check.CMORCheckError: There were errors in variable cl:
lev: units should be 1, not hPa
 lev: has values < valid_min = 0.0
```

The dataset looks like this:

```
netcdf cl_Amon_CESM2_historical_r10i1p1f1_gn_185001-189912 {
dimensions:
        lat = 192 ;
        nbnd = 2 ;
        lon = 288 ;
        lev = 32 ;
        time = UNLIMITED ; // (600 currently)
variables:
        float cl(time, lev, lat, lon) ;
                cl:_FillValue = 1.e+20f ;
                cl:cell_measures = "area: areacella" ;
                cl:cell_methods = "area: time: mean" ;
                cl:comment = "Includes both large-scale and convective cloud." ;
                cl:coordinates = "time lev lat lon" ;
                cl:description = "Includes both large-scale and convective cloud." ;
                cl:frequency = "mon" ;
                cl:id = "cl" ;
                cl:long_name = "Percentage Cloud Cover" ;
                cl:mipTable = "Amon" ;
                cl:missing_value = 1.e+20 ;
                cl:out_name = "cl" ;
                cl:prov = "Amon ((isd.003))" ;
                cl:realm = "atmos" ;
                cl:standard_name = "cloud_area_fraction_in_atmosphere_layer" ;
                cl:time = "time" ;
                cl:time_label = "time-mean" ;
                cl:time_title = "Temporal mean" ;
                cl:title = "Percentage Cloud Cover" ;
                cl:type = "real" ;
                cl:units = "%" ;
                cl:variable_id = "cl" ;
        float ps(time, lat, lon) ;
                ps:long_name = "surface_air_pressure" ;
                ps:units = "Pa" ;
        float p0 ;
                p0:long_name = "vertical coordinate formula term: reference pressure" ;
                p0:units = "Pa" ;
        double a(lev) ;
                a:bounds = "a_bnds" ;
                a:long_name = "vertical coordinate formula term: a(k)" ;
        double b(lev) ;
                b:bounds = "b_bnds" ;
                b:long_name = "vertical coordinate formula term: b(k)" ;
        double lat(lat) ;
                lat:axis = "Y" ;
                lat:bounds = "lat_bnds" ;
                lat:standard_name = "latitude" ;
                lat:title = "Latitude" ;
                lat:type = "double" ;
                lat:units = "degrees_north" ;
                lat:valid_max = 90. ;
                lat:valid_min = -90. ;
        double lev(lev) ;
                lev:axis = "Z" ;
                lev:bounds = "lev_bnds" ;
                lev:positive = "up" ;
                lev:standard_name = "alevel" ;
                lev:title = "atmospheric model level" ;
                lev:type = "double" ;
                lev:units = "hPa" ;
        double lon(lon) ;
                lon:axis = "X" ;
                lon:bounds = "lon_bnds" ;
                lon:standard_name = "longitude" ;
                lon:title = "Longitude" ;
                lon:type = "double" ;
                lon:units = "degrees_east" ;
                lon:valid_max = 360. ;
                lon:valid_min = 0. ;
        double time(time) ;
                time:axis = "T" ;
                time:bounds = "time_bnds" ;
                time:standard_name = "time" ;
                time:title = "time" ;
                time:type = "double" ;
                time:units = "days since 0001-01-01 00:00:00" ;
                time:calendar = "noleap" ;
        double time_bnds(time, nbnd) ;
                time_bnds:calendar = "noleap" ;
                time_bnds:units = "days since 0001-01-01 00:00:00" ;
        double b_bnds(lev, nbnd) ;
                b_bnds:long_name = "vertical coordinate formula term: b(k+1/2)" ;
        double lat_bnds(lat, nbnd) ;
                lat_bnds:units = "degrees_north" ;
        double lon_bnds(lon, nbnd) ;
                lon_bnds:units = "degrees_east" ;
        double lev_bnds(lev, nbnd) ;
                lev_bnds:formula = "p = a*p0 + b*ps" ;
                lev_bnds:formula_terms = "p0: p0 a: a_bnds b: b_bnds ps: ps" ;
                lev_bnds:standard_name = "atmosphere_hybrid_sigma_pressure_coordinate" ;
                lev_bnds:units = "hPa" ;
        double a_bnds(lev, nbnd) ;
                a_bnds:long_name = "vertical coordinate formula term: a(k+1/2)" ;
```

Please note that the `formula_terms` attribute is missing for `lev`, which we add through a fix. However, for the bounds it reads: `p0: p0 a: a_bnds b: b_bnds ps: ps`, so `lev` is **not** used.

So I think the issue I describe here only applies for the the hybrid pressure coordinate, not for the hybrid height coordinate, so I will remove this from the PR here. However, as described in #883 there is still an issue with these hybrid height coordinates, since the recipe describe there works when only a single dataset is used, but **may** fail when 5 of them are used together.

(Sorry for the slow response, since mistral is down today I had to download the corresponding datasets to my local machine).

schlunma:
When #883 is solved then maybe we can solve this with a fix after all, because I think only the 4 CESM models are really suffering from this issue (due to #883 I saw this for 4 additional models, which were apparently false positives).

Let's wait for @sloosvel to fix #883. After that I will have a look which datasets are really affected.

valeriupredoi:
OK adding #859 and its sister #515 to this to be closed - can I confirm we **do need** fixes for those CESM2 olevs in cm? :beer:

schlunma:
> OK adding #859 and its sister #515 to this to be closed - can I confirm we **do need** fixes for those CESM2 olevs in cm? beer

I only found theses issues I mentioned here in the atmosphere coordinate, but I would not at all be surprised if the ocean coordinates are messed up, too :smile:

valeriupredoi:
aye, what's the worst it can happen if I write a fix - make the fishes and ocean people happy :+1:

schlunma:
Okay, with @sloosvel's fix (#886) I can now confirm that at least 5 CMIP6 models suffer from the issue described here:

- CAS-ESM2-0 (`/mnt/lustre02/work/bd0854/DATA/ESMValTool2/CMIP6_DKRZ/CMIP/CAS/CAS-ESM2-0/historical/r1i1p1f1/Amon/cl/gn/v20200302/cl_Amon_CAS-ESM2-0_historical_r1i1p1f1_gn_185001-201412.nc`)
- CESM2 (`/mnt/lustre02/work/bd0854/DATA/ESMValTool2/CMIP6_DKRZ/CMIP/NCAR/CESM2/historical/r1i1p1f1/Amon/cl/gn/v20190308/cl_Amon_CESM2_historical_r1i1p1f1_gn_185001-201412.nc`)
- CESM2-FV2 (`/mnt/lustre02/work/bd0854/DATA/ESMValTool2/CMIP6_DKRZ/CMIP/NCAR/CESM2-FV2/historical/r1i1p1f1/Amon/cl/gn/v20191120/cl_Amon_CESM2-FV2_historical_r1i1p1f1_gn_200001-201412.nc`)
- CESM2-WACCM (`/mnt/lustre02/work/bd0854/DATA/ESMValTool2/CMIP6_DKRZ/CMIP/NCAR/CESM2-WACCM/historical/r1i1p1f1/Amon/cl/gn/v20190415/cl_Amon_CESM2-WACCM_historical_r1i1p1f1_gn_185001-201412.nc`)
- CESM2-WACCM-FV2 (`/mnt/lustre02/work/bd0854/DATA/ESMValTool2/CMIP6_DKRZ/CMIP/NCAR/CESM2-WACCM-FV2/historical/r1i1p1f1/Amon/cl/gn/v20191120/cl_Amon_CESM2-WACCM-FV2_historical_r1i1p1f1_gn_200001-201412.nc`)

I can confirm that the `lev` coordinate values are not used in any of these datasets (`lev:formula_terms = "p0: p0 a: a b: b ps: ps"`) and take values from -1000 to 1000 hPa (but they should be between 0 and 1 and have units of 1).

How should we proceed here? Since I still have no idea how to fix these datasets correctly, I propose to do it similarly to the changes I made in this PR. I think it is safe to exclude the values of coordinate `atmosphere_hybrid_sigma_pressure_coordinate` from the checks. Additional support for that is given by the CMOR table, in which `lev` is not used for the creation of `air_pressure`:

https://github.com/ESMValGroup/ESMValCore/blob/c07e280274ef00aba2ad12969e22ca43660d5d5d/esmvalcore/cmor/tables/cmip6/Tables/CMIP6_coordinate.json#L2019

In contrast, for `atmosphere_hybrid_height_coordinate`, `lev` is explicitely used:

https://github.com/ESMValGroup/ESMValCore/blob/c07e280274ef00aba2ad12969e22ca43660d5d5d/esmvalcore/cmor/tables/cmip6/Tables/CMIP6_coordinate.json#L722

axel-lauer (approved):
Looks fine to me. This is a good approach to be able to work with the affected model datasets as we are not able to provide "fixes" for this issue for the affected models. Without this fix, some variables from these models could not be processed.

zklaus (changes requested):
I am still in the process of evaluation. My findings so far: The data from the CESM models are all broken. The `lev` coordinate has an invalid standard name that does not refer to a parametric coordinate, the `lev` bounds use a different coordinate system from the `lev` coordinate itself.

As such, those models need to be fixed independently from any action suggested in this PR.

That leaves us with only the CAS model. I have not finished looking into that.

As such, at the moment it is unclear if this PR will address an actual problem.

axel-lauer:
@zklaus In addition to the CESM models and CAS mentioned here, I also encountered this problem with the following models when processing variable clw: UKESM1-0-LL, KACE-1-0-G, ACCESS-ESM1-5, ACCESS-CM2, HadGEM3-GC31-LL. All of these models (including the CESM models and CAS) could be processed correctly before introducing the "lev check".

schlunma:
@axel-lauer These models have hybrid height coordinates and should work now after the merge of #886.

However, I do not agree with you here @zklaus. This PR addresses a problem that does not allow the processing of datasets anymore that have been successfully used before. We are aware the the CESM2 models need fixes (we already implemented them) and successfully used `cl` data from CAS-ESM2-0, CESM2, CESM2-FV2 and CESM2-WACCM for an accepted publication.

Since the `lev` coord is not used in the final datasets, see

https://github.com/ESMValGroup/ESMValCore/blob/c07e280274ef00aba2ad12969e22ca43660d5d5d/esmvalcore/cmor/tables/cmip6/Tables/CMIP6_coordinate.json#L2019

we did not address it in the fixes.

I could of course provide a "fix" for this coordinate by simply linearly mapping the values `[-999, 0]` to the range `[0, 1]`, but would this be more correct than before? What's the point of fixing something that is not used on the final output and we also don't know how to fix?

zklaus:
Following @schlunma's comment, I assume that only CESM and CAS are open. If this is not correct, please mention any other models that remain problematic here.

The point is not to fix something that isn't used. The point is that we have two potential sources of height information and we seem not to be able to reconcile them with one another. If that is true, we don't know if they have been used successfully before. We only know that they have been used before and that the tests have done exactly what they are supposed to do: Uncover problematic data. If it is not true, and we can reconcile the two sources of information, then we can write a simple fix. For the CESM models, there is nothing in the files that ties the parametric coordinates to the data field and several of the coordinate variables are defective.

So whatever the correct resolution of these problems is (and it may well be that you used the correct height information in the end), this seems to be a problem with this specific data and as such, I don't think we should deactivate the check generally.

ruthlorenz:
If this is an issue with the data which could (potentially) be solved by the modelling group it should be reported to them. We have done this in the past and the modelling groups are generally happy to be informed and try to fix things if they can. They then also add the information here [https://errata.es-doc.org/static/index.html](url) so other people have a chance of knowing these issues.

schlunma:
I think I found a solution to this problem:

As stated in [Appendix D](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#parametric-v-coord) of the CF conventions, "the hybrid sigma-pressure coordinate for level k is defined as a(k)+b(k) or ap(k)/p0+b(k), as appropriate".

I checked this, and it is true for most CMIP6 models (the ones differing are the CESM2 models, the CAS model and 1 other model where there are numerical inconsistencies). This should allow us to calculate the correct values for the `atmosphere_hybrid_sigma_pressure_coordinate` for CESM2 and CAS in a fix, which will solve this problem. I will open a PR on Monday.

However, I really think that in some very specific cases it might not be useful or user-friendly to check all nitty gritty aspects of the CMOR table when there is a scientific justification. In this specific case, the height information was always given (by the formula which does **NOT** depend on `atmosphere_hybrid_sigma_pressure_coordinate`) and **unique**, so checking the values of the `atmosphere_hybrid_sigma_pressure_coordinate` is really not useful. It probably took me now a day now to get an idea how to "fix" a dataset that is not wrong from a scientific point of view, which is kind of annoying.

sloosvel:
>As stated in Appendix D of the CF conventions, "the hybrid sigma-pressure coordinate for level k is defined as a(k)+b(k) or ap(k)/p0+b(k), as appropriate".

That was pointed out in #840. Back in the day I tried summing a+b and I got coordinate values that match those of other models, so if that works for you, we can add this type of fix instead of skipping them. Copying the values from a model with the right coordinate as a fix would not be that bad either.

schlunma:
Will re-open soon, this was automatically closed by force-pushing the current master to it (so I can have a fresh start on this branch).

schlunma:
I added the fix for the CESM2 models according to the link that @sloosvel (I didn't see that back then, sorry). Unfortunately the CAS model is not available on mistral anymore, so I cannot test this.

@sloosvel can you do a technical review, please?

@axel-lauer can you do a scientific review, please? I used this recipe for testing:

[recipe_test_cl_cli_clw.yml.txt](https://github.com/ESMValGroup/ESMValCore/files/6063010/recipe_test_cl_cli_clw.yml.txt)

sloosvel (approved):
I could only run the CESM2 dataset, as the others are missing on jasmin, but the fix looks good to me.

axel-lauer (changes requested):
I have tried this fix and it produces results that look good when using no preprocessor. When using e.g. the following preprocessor to convert the data to altitude levels (same happens for pressure levels), only the lowermost level contains valid values while all other levels contain only missing values:
```  ppa40:
    extract_levels:
      levels: {cmor_table: CMIP6, coordinate: alt40}
      coordinate: altitude
      scheme: linear
```
This behavior differs from the results obtained using the current master branch (with `--check_level=ignore` to get the tool to process cl, cli or clw from CESM2). With the master branch, all altitude / pressure levels contain valid values.

I currently do not understand why this happens, but I think this might need some more investigation before this fix can be merged.

schlunma:
That is really strange....having a look at this right now!

axel-lauer (approved):
After talking to @schlunma I redid the tests with a clean version of this branch (fix_hybrid_lev_coords). This solved the problem, the results now look good and as expected. It seems the problem I reported yesterday was caused by a somewhat screwed up version of my local repository copy. Sorry about that. This fix can be merged now!

schlunma:
@zklaus can you either approve the changes or dismiss you review? Merging is blocked because you requested changes on the original version of this PR which is completely outdated now.

axel-lauer:
We have positive reviews and we need this PR now for further work. Since the PR changed in a way that the changes requested by @zklaus for an earlier version do not apply any more, I think merging this now is fine.

PR #635, opened 2020-05-13
New / updated model fixes

axel-lauer:
This PR updates the existing fixes for the CMIP6 models ACCESS-CM2 and ACCESS-ESM1-5 and adds fixes for the CMIP6 models CIESM and CESM2-WACCM-FV2 for 3-dim cloud variables (cl, clw, cli).

The existing fixes for ACCESS-CM2 and ACCESS-ESM1-5 produce wrong results when converting the hybrid-height coordinates to height or pressure levels because the coefficients "b" needed for calculating "z = a + b * orog" are wrongly defined in the input files. This fix replaces "b" with correct values.

The integration tests for both ACCESS models have been updated, tests for CIESM and CESM2-WACCM-FV2 have been added.

PR #892, opened 2020-12-02
Add plev/altitude conversion to extract_levels

axel-lauer:
This pull requests extends the existing preprocessor function _extract_levels_. The possibility to convert pressure levels (air_pressure) into height (altitude) and vice versa is added. For the conversion, existing code in cmor/_fixes/shared.py is used that uses the US standard atmosphere to convert from height to pressure. A similar function has been implemented to also allow conversion from pressure to height.

**Tasks**

### DIFF
--- a/esmvalcore/cmor/_fixes/shared.py
+++ b/esmvalcore/cmor/_fixes/shared.py
@@ -2,6 +2,7 @@
 import logging
 import os
 import warnings
+from functools import lru_cache
 
 import dask.array as da
 import iris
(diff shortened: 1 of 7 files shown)


schlunma:
Is there a specific reason why you don't calculate the altitude `bounds`?

axel-lauer:
Calculating altitude bounds did not work for the MPI-ESM1-2-LR model (variable clw)...

schlunma:
I guess that model also needs a fix similar to [`MPI-ESM1-2-HR`](https://github.com/ESMValGroup/ESMValCore/blob/master/esmvalcore/cmor/_fixes/cmip6/mpi_esm1_2_hr.py) to get correct bounds. Can you maybe add that to #635? I will add an `if` statement and only add the bounds if possible.

axel-lauer:
@schlunma Sounds good!

schlunma (commented):
Scientific review: Successfully tested the new function on real data (about 20 CMIP6 models) and on synthetic data (see tests) :+1: 

Technical review: Fine from my side, but since I wrote parts of the code myself maybe @valeriupredoi can have a brief look on the changes?

@axel-lauer I think it would be nice to have one sentence about this functionality in the documentation? And maybe also the docstring of `extract_levels`?

axel-lauer:
@schlunma good point! I just added some documentation.

schlunma (approved):

esmvalcore/cmor/_fixes/shared.py
```diff
@@ -29,6 +29,21 @@ def _get_altitude_to_pressure_func():
...
+    source_file = os.path.join(base_dir, 'us_standard_atmosphere.csv')
+    data_frame = pd.read_csv(source_file, comment='#')
+    func = interp1d(data_frame['Pressure [Pa]'],
+                    data_frame['Altitude [m]'],
+                    kind='cubic',
+                    fill_value='extrapolate')
+    return func
+
+
+pressure_to_altitude = _get_pressure_to_altitude_func()  # noqa
```
bouweandela, line 44: You could also use the standard library decorator [`lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache) to cache the the result. That would have the advantage that the module loads faster because the table only needs to be read when the function is called.

bouweandela (commented):
Looks good to me, apart from the minor technical improvement mentioned in the comment that could be made.

bouweandela:
It would also be nice to mention the functionality here: https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/preprocessor.html#vertical-interpolation

axel-lauer:
I thought the information in https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/preprocessor.html#vertical-interpolation is taken from the docstrings. These have been updated accordingly...

bouweandela:
No, that's the info here: https://docs.esmvaltool.org/projects/esmvalcore/en/latest/api/esmvalcore.preprocessor.html#esmvalcore.preprocessor.extract_levels

The preprocessor as it can be used in the recipe has it's own documentation page:
https://github.com/ESMValGroup/ESMValCore/blob/master/doc/recipe/preprocessor.rst#vertical-interpolation

esmvalcore/cmor/_fixes/shared.py
schlunma: So I just have to add `@lru_cache` to `_get_pressure_to_altitude_func` and then use `_get_pressure_to_altitude_func()` instead of `pressure_to_altitude`?

axel-lauer:
@bouweandela Thanks for pointing me to the documentation page. I added a brief description of the new functionality.

bouweandela (approved):

PR #900, opened 2020-12-09
Refer to ESMValTool GitHub discussions page in the error message

bouweandela:
Refer users to the new [discussions](https://github.com/ESMValGroup/ESMValTool/discussions?discussions_q=category%3AQ%26A) page if they have questions.


valeriupredoi:
@bouweandela pls fix the annoying blank lines needed by Lord Codacy, but more generally - can you pls start a template discussion on the discussions page so people can see how to approach it (I've never used discussions so am new to this) :beer:

bouweandela:
The blank lines are annoying indeed, they are caused by using `docformatter` and it has this issue https://github.com/myint/docformatter/issues/43. I'll add them again manually.

esmvalcore/_main.py
```diff
@@ -436,11 +432,19 @@ def display(lines, out):
...
         logger.info(
-            "If you suspect this is a bug or need help, please open an issue "
-            "on https://github.com/ESMValGroup/ESMValTool/issues and attach "
-            "the run/recipe_*.yml and run/main_log_debug.txt files from the "
-            "output directory.")
+            "\n"
+            "If you have a question or need help, please start a new "
+            "discussion on "
+            "https://github.com/ESMValGroup/ESMValTool/discussions?"
+            "discussions_q=category%3AQ%26A"
```
nielsdrost, line 442:
```suggestion
            "https://github.com/ESMValGroup/ESMValTool/discussions"
```

nielsdrost: I suggest we point to the top level url for discussions for now since I'm not sure if we will be making any tweaks to the setup and/or github may change it.

nielsdrost (approved):
Good idea @bouweandela, thanks!

esmvalcore/_main.py
bouweandela: I thought it might be nice to use a single place for questions for help, but indeed if you think that they might change the available labels that's not very useful now.

bouweandela: Removed the category

bouweandela:
The failing tests are unrelated to this pull request and fixed in #921.

PR #935, opened 2021-01-11
remove redundant test line from test_utils.py

valeriupredoi:
No need for that, also added maintainers in the recipes that are tested here see https://github.com/ESMValGroup/ESMValTool/pull/1970 - please run the tests and make sure all tests pass OK before committing and merging into `master` - two tests were failing due to lack of maintainers in the recipe!


stefsmeets (approved):
👍

Issue #781, opened 2020-09-15
Making the multi-model statistics preprocessor lazy


Peter9192:
The multi-model statistics preprocessor is currently sub-optimal. It operates on Numpy arrays because Iris cubes are quite strict in checking inconsistencies between the input cubes. This workaround impedes lazy evaluation and introduces some vulnerabilites (e.g. #267, #665) and a trail of ad-hoc fixes.

Inconsistent time coordinates were probably the biggest obstacle. However, starting from #677, and continuing in #685, we have already drastically reduced the complexity of time handling. Additionally, #744 proposes to move everything related to 'homogenizing' the time coordinate to the regrid_time preprocessor altogether. Moreover, Iris will soon support lazy regridding (https://github.com/SciTools/iris/pull/3701), so in that case we could really benefit from lazy evaluation in multi-model statistics as well.

In #673, we have implemented a new function that computes 'multi-cube' statistics using Iris build-in functions. This already works for ensemble statistics, because in that case, the cubes are already very homogeneous. 

It would be really nice if we could use this Iris 'engine' for the regular multi-model statistics preprocessor as well.

**Would you be able to help out?**
I might have some time, but it would be good to wait for #685, #744 and #673. In the meantime, this is a placeholder to keep the helicopter view of everything that's happening with the multi-model statistics preprocessor.

stefsmeets:
https://github.com/ESMValGroup/ESMValCore/issues/674

valeriupredoi:
good call! we need to liaise with `iris` folk - @bjlittle - because a basic multimodel statistic (and lazy altogether) would best sit in `iris`. We can do it here first but migration to `iris` is something both us and them would benefit from :beer:

Peter9192:
So what we did in #673 is basically: `iris.cube.cubeList(...).merge_cube().collapse()` with some fiddling with auxcoords and equalizing attributes. I'm not sure if that warrants a dedicated function in `iris` or whether it's just a good application of existing functionality. So indeed, I'm curious what @bjlittle has to say about that.

valeriupredoi:
no, I'm talking about a dedicated `iris.analysis.multimodel_statistics(args=(cubes, statistic, ...), **kwargs)` function

Peter9192:
I know. But what I'm saying is that you could also see it as a use case that is already supported by existing functionality in `iris`. So I'm not sure if really needs to be a dedicated function in iris. But of course I would be happy to see it migrate there eventually. 

[...]

PR #949, opened 2021-01-18
Refactor multi-model statistics code to facilitate ensemble stats and lazy evaluation

Peter9192:
## Description
This PR refactors the multimodel statistics code. Specifically, it separates the core functionality (computing statistics across cubes) from the intermediate layer that deals with the handling of ESMValCore products. 

This facilitates re-use of the core function in e.g. an ensemble statistics function (see #673) and also it makes it easier to make the core functionality lazy (see #781 and #950). 

- Needed for #781
- Needed for #673

***

## Before you get started

## Checklist

-   [x] N/A [Documentation](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#documentation) is available for new functionality

-   [x] N/A Update [ESMValTool](https://github.com/esmvalgroup/esmvaltool) and link the pull request(s) in the description

***


stefsmeets (approved):
Nice work @Peter9192 , all tests are passing, so it looks good to me! Separating how products are processed from cubes makes the code much easier to work with.

doc/recipe/preprocessor.rst
```diff
@@ -1512,14 +1509,14 @@ In the most general case, we can set upper limits on the maximum memory the
 analysis will require:
 
 
-``Ms = (R + N) x F_eff - F_eff`` - when no multimodel analysis is performed;
+``Ms = (R + N) x F_eff - F_eff`` - when no multi-model analysis is performed;
 
-``Mm = (2R + N) x F_eff - 2F_eff`` - when multimodel analysis is performed;
+``Mm = (2R + N) x F_eff - 2F_eff`` - when multi-model analysis is performed;
```
valeriupredoi, line 1514: have you tested these or are they theoretical assumptions? why is there now a factor 2 for the computational efficiency?

doc/recipe/preprocessor.rst
```diff
@@ -681,7 +680,7 @@ days in a year may vary between calendars, (sub-)daily data are not supported.
 .. code-block:: yaml
 
     preprocessors:
-      multimodel_preprocessor:
+      multi_model_preprocessor:
```
valeriupredoi, line 683: why change the preprocessor name?

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -1,14 +1,10 @@
-"""multimodel statistics.
+"""Statistics across cubes.
 
-Functions for multi-model operations
-supports a multitude of multimodel statistics
-computations; the only requisite is the ingested
-cubes have (TIME-LAT-LON) or (TIME-PLEV-LAT-LON)
-dimensions; and obviously consistent units.
+This module contains functions to compute statistics across multiple cubes or
+products.
```
valeriupredoi, line 5: not quite - across means one applies a certain operation (statistic) to each cube or product individually from a set of cubes or products, this is different, it's a multicube/multiproduct statistic, and in the lingo of a climate modeller - multimodel statistic. Let's not reinvent the wheel :grin:

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -1,14 +1,10 @@
...
 
-It operates on different (time) spans:
-- full: computes stats on full dataset time;
-- overlap: computes common time overlap between datasets;
+Wrapper functions separate esmvalcore internals, operating on products, from
+generalized functions that operate on iris cubes.
```
valeriupredoi, line 7: this is very confusing

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -295,74 +291,55 @@ def _assemble_data(cubes, statistic, span='overlap'):
...
+    consistent shapes.
 
-    Restrictions needed by the input data:
-    - model datasets must have consistent shapes,
-    - higher dimensional data is not supported (ie dims higher than four:
-    time, vertical axis, two horizontal axes).
+    This function operates directly on numpy (masked) arrays and rebuilds the
+    resulting cubes from scratch. Therefore, it is not suitable for lazy
+    evaluation. This function is restricted to maximum four-dimensional data:
+    time, vertical axis, two horizontal axes.
```
valeriupredoi, line 305: I think the old docstring was explaining the functionality very well using terms that a climate scientist would understand better than this changed one, I don't get why make things very technical when there is no need to, remember this is read by not so technical people via the API documentation

valeriupredoi (changes requested):
cheers @Peter9192 for this! I am still to read the changes to the documentation, the functionality seems to be nicely restructured, but am not quite happy with the docstring changes - they have become too technical and I don't think this is a good idea, given who we cater for :+1: **EDIT** I have now gone through the documentation changes and I like'em, job well done there - but still confused as to why `multimodel` -> `multi-model` :beer:

doc/recipe/preprocessor.rst
Peter9192: I don't know where this comes from. Just added the hyphen

esmvalcore/preprocessor/_multimodel.py
Peter9192: Okay, I'm not a native english speaker, hence the confusion with 'across'. What word do you use for applying a function like this? 'Over'?

The reason I want to get rid of multi-model, is that we will use it for ensemble stats as well (which could be the same model). Multi-model is just a specific case of the more general multi-cube operations.

esmvalcore/preprocessor/_multimodel.py
Peter9192: Good point about the intended audience. 

This is in anticipation of #950, where I implement an altnerative function that does exactly the same thing in a different way. But I'll rephrase it for this PR.

doc/recipe/preprocessor.rst
Peter9192: For consistency. Some parts of the documentation used 'multimodel' and others 'multi-model'. So I just search and replaced all instances of multimodel. For the preprocessor name it doesn't really matter what it's called, but this now more consistent with the rest. Do you prefer multimodel without the hyphen?

Peter9192:
Thanks @valeriupredoi I'll look into your suggestions! I appreciate the point about the audience for the docstrings.

doc/recipe/preprocessor.rst
valeriupredoi: (I tested and gathered data and got these down empirically) - also, my bad, I saw double, and didn't notice the only change was the dash :eyeglasses:

doc/recipe/preprocessor.rst
valeriupredoi: aha! good point! no, I don't give a hoots, but I was wondering why the change, all good by me then :+1:

esmvalcore/preprocessor/_multimodel.py
valeriupredoi: just use multi-cube/multi-product statistics?

esmvalcore/preprocessor/_multimodel.py
valeriupredoi: cheers :beer:

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -372,22 +349,89 @@ def multi_model_statistics(products, span, statistics, output_products=None):
...
+        for product in products:
+            statistics_product.wasderivedfrom(product)
+
+        logger.info("Generated %s", statistics_product)
+        statistics_products.add(statistics_product)
+
+    return statistics_products
+
+
+def multi_model_statistics(products: set,
```
bouweandela, line 418: Note that this function is part of the public API. Changing it so it no longer accepts cubes is a breaking change that will probably break diagnostic code. Would it be possible to make it work again that if you put cubes in, you get cubes out? And of course the docstring needs to be good as @valeriupredoi already pointed out.

valeriupredoi: shucks! I totally missed that - very good point @bouweandela - yes, @Peter9192 please see this and revise the approach :+1:

Peter9192: Yes I figured that out as well but I'm struggling to find an appropriate solution.

stefsmeets: @Peter9192 Didn't we talk about making `multi_model_statistics` cube only (not to mess with the public API) and simply pass the arguments to `multi_cube_statistics`. Then use `_multi_model_statistics` with products for internal use? Or am I confusing 2 PRs now?

esmvalcore/preprocessor/_multimodel.py
Peter9192: The old docstring is back in the public API function below.

esmvalcore/preprocessor/_multimodel.py
Peter9192: Done

esmvalcore/preprocessor/_multimodel.py
Peter9192: Undone

esmvalcore/preprocessor/_multimodel.py
Peter9192: After a lot of contemplating I have accepted that the public API function will have to accept both products and cubes as possible inputs. I had rather hidden those products from the public altogether, but this becomes very ugly very quickly. So I have now restored the public `multi_model statistics` as the single entry point that returns different functions based on its input type.

Peter9192:
All green! @valeriupredoi I think I got everything covered, could you please have another look? Thanks :beers:

valeriupredoi (approved):
looks good, development tests pass too (not cached ones on Circle, actual branch and ESMValTool dev ones), cheers @Peter9192 :beer: @bouweandela did you want to have another look yerself?

tests/unit/preprocessor/_multimodel/test_multimodel.py
```diff
@@ -7,17 +7,19 @@
 from cf_units import Unit
 
 import tests
-from esmvalcore.preprocessor import multi_model_statistics
-from esmvalcore.preprocessor._multimodel import (_assemble_data,
-                                                 _compute_statistic,
-                                                 _get_time_slice, _plev_fix,
-                                                 _put_in_cube,
-                                                 _unify_time_coordinates)
+from esmvalcore.preprocessor._multimodel import (
```
bouweandela, line 10: It would be good to restore the test that `multi_model_statistics` is imported from `esmvalcore.preprocessor` (instead of from `esmvalcore.preprocessor._multimodel`), because without this test it may be possible to remove the function from the public API without a test breaking

bouweandela (commented):
Looks good to me, more clean. Just a minor comment on the unit tests.

tests/unit/preprocessor/_multimodel/test_multimodel.py
Peter9192: Done!

Peter9192:
Now the test fails because of iris docs change or smt like that 😞

bouweandela:
> Now the test fails because of iris docs change or smt like that disappointed

That should be fixed by #964

PR #901, opened 2020-12-09
Add Listing and finding recipes to the experimental notebook API

stefsmeets:
Hi all, this code implements some new functionality for the experimental API we are building. It introduces some functions to list recipes, find recipes using some search patterns, and display nicely formatted recipe metadata in a jupyter notebook.

**Tasks**

-   [x] [Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do, if you haven't done so already (and add the link at the bottom)
-   [x] This pull request has a descriptive title that can be used in a changelog
-   [x] Add unit tests
-   [x] Public functions should have a numpy-style docstring so they appear properly in the [API documentation](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/api/esmvalcore.html). For all other functions a one line docstring is sufficient.
-   [x] If writing a new/modified preprocessor function, please update the [documentation](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/preprocessor.html)
-   [x] Circle/CI tests pass. Status can be seen below your pull request. If the tests are failing, click the link to find out why.
-   [x] Codacy code quality checks pass. Status can be seen below your pull request. If there is an error, click the link to find out why. If you suspect Codacy may be wrong, please ask by commenting.

If you need help with any of the tasks above, please do not hesitate to ask by commenting in the issue or pull request.

* * *

Follows up on #868 
For more discussion, see issue #498 
Documentation: https://esmvaltool--901.org.readthedocs.build/projects/ESMValCore/en/901/api/esmvalcore.api.html 
Demo notebook: https://gist.github.com/stefsmeets/f4b7fc7fffe78cecb895b12bfb8ea671

PR #957, opened 2021-01-21
Add displaying and loading of recipe output to the notebook API

stefsmeets:
## Description

Hi everyone, this PR extends the API to run recipes in an interactive notebook.

It adds functionality to obtain the output data (data or plots), and display or load them directly in the notebook.

-   For more discussion, see issue #498 
-   Follows up on #907, #901, #868 
-   Link to documentation: https://esmvaltool--957.org.readthedocs.build/projects/ESMValCore/en/907/api/esmvalcore.api.html 
-   Demo notebook: https://gist.github.com/stefsmeets/064c86a83a0490589203eee6cc51e31b 

## Todo

- [x] Wrap `recipe_output` dict and add convenience methods
- [x] Add shorthand to get items of a type, i.e. `recipe_output.get_output_of_type('data')`
- [x] Elaborate/update on documentation
- [x] Fine-tune `TaskOutput` usage (`__repr__` / rename `.filename` attribute)

***

## Before you get started

## Checklist

***


stefsmeets:
I just realized we do not depend on `xarray` in ESMValCore. Just wondering if having a `.load_xarray` method warrants adding it to the list of dependencies. Afaik `ESMValTool` has it in its dependencies after all.

esmvalcore/experimental/recipe_output.py
```diff
@@ -0,0 +1,127 @@
...
+
+    kind = 'image'
+
+    def _repr_png_(self):
+        """Render png as image in Jupyter Notebook."""
+        from IPython.display import Image
+        return Image(self.filename).data
+
+
+class DataItem(OutputItem):
```
Peter9192, line 132: Similarly:
```suggestion
class OutputData(OutputItem):
```

or `OutputFile` or `OutputDataset`?

esmvalcore/experimental/recipe_output.py
```diff
@@ -0,0 +1,127 @@
...
+
+        return item_class(filename=filename, attributes=attributes)
+
+
+class ImageItem(OutputItem):
+    """Container for image output."""
+
+    kind = 'image'
+
+    def _repr_png_(self):
```
Peter9192: Why is this not used in the example notebook?

Peter9192 (changes requested):
Very cool stuff @stefsmeets, this is really impressive already! Here's a few general comments:

- The documentation (before the API reference) is a bit limited, and I'm not sure if I could readily use it. The notebook helped me to understand how to the items from the output. A slightly more elaborate example would be nice.
- Would it be simple enough to make shorthand methods for these intimidating lines in the example notebook like `data_items = [item for tasks in recipe_output.values() for item in tasks if item.kind == 'data']`? Then I think that's worth adding.
- How much effort would it be to have a nice markdown rendering of the output, like you did for the Recipe Info? I think that could potentially be awesome as well.

esmvalcore/experimental/recipe_output.py
stefsmeets: I'm not sure I don't understand this comment. IPython looks for `_repr_xxx_` methods. This is what renders the images in the notebook, it is never called by a user directly.

esmvalcore/experimental/recipe_output.py
```diff
@@ -0,0 +1,127 @@
+from pathlib import Path
+from pprint import pformat
+
+from .recipe_metadata import Contributor, Reference
+
+
+class OutputItem():
```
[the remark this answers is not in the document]
stefsmeets: What do you propose?

stefsmeets:
>     * The documentation (before the API reference) is a bit limited, and I'm not sure if I could readily use it. The notebook helped me to understand how to the items from the output. A slightly more elaborate example would be nice.

Yeah, I agree. I am collecting the notebooks I'm developing so that they can be used to support the documentation. I will try to see what I can do.

>     * Would it be simple enough to make shorthand methods for these intimidating lines in the example notebook like `data_items = [item for tasks in recipe_output.values() for item in tasks if item.kind == 'data']`? Then I think that's worth adding.

This would mean turning recipe output into a class, and then add a method like `.get_output_of_type('data')`. I can do that 👍

>     * How much effort would it be to have a nice markdown rendering of the output, like you did for the Recipe Info? I think that could potentially be awesome as well.

Technically it is not too difficult. I'm already using HTML so that I can render the caption as well as the image. This can be extended with other attributes. I think you can do inline html in markdown, which may be a another way to do it.

I actually started adding the authors, but I stopped because I was worried of losing scope for this PR. It would require going through the attributes and presenting them in a nice way, including stuff like authors and references (which we already have), but also realms and domains. I think this PR puts a good baseline to continue from.

esmvalcore/experimental/recipe_output.py
```diff
@@ -0,0 +1,127 @@
...
+    kind = None
+
+    def __init__(self, filename, attributes={}):
+        self.attributes = attributes
+        self.filename = Path(filename)
+
+        self._authors = None
+        self._references = None
+
+    def __repr__(self):
```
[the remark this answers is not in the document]
stefsmeets: Good idea, I added wrappers for recipe and task output to make this possible 👍

esmvalcore/experimental/recipe_output.py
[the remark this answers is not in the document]
Peter9192: Make it hidden? _OutputItem()

esmvalcore/experimental/recipe_output.py
Peter9192: Why then in the example notebook do you also need to import and use the Display method?

stefsmeets:
It was a bit of work, but I added wrappers for the task and recipe output to make it a bit easier to work with these objects. This also lays the foundation to do stuff like `recipe_output.report()` and get a nice html representation. I think this is something we may want in the future (but not in this PR!).

doc/api/esmvalcore.api.recipe.rst
```diff
@@ -59,7 +59,37 @@ Sessions can also be explicitly specified.
...
+
+After running a recipe, output can also be retrieved using:
+
+.. code-block:: python
+
+    >>> output = recipe.get_output()
+
+Output comes in two kinds, :py:class:`DataOutput`, which correspond to data files in `.nc` format and :py:class:`ImageOutput` which correspond to plots in `.png` format.
+
+Image output uses IPython magic to plot themselves in a notebook environment, whereas data output have special methods to load the data using :py:mod:`xarray`,
```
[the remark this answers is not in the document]
stefsmeets: 👍

doc/api/esmvalcore.api.recipe.rst
```diff
@@ -59,7 +59,37 @@ Sessions can also be explicitly specified.
...
+
+.. code-block:: python
+
+    >>> data = item.load_xarray()
```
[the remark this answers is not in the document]
stefsmeets: Solved with updated documentation

esmvalcore/experimental/recipe_output.py
stefsmeets: I use that in the notebook to show multiple images in a row (`display` is like `print` for images). The `repr` does not trigger otherwise. The method was cleaned up to not depend on `IPython`, and displaying in the notebook is now achieved using `_repr_html_`.

stefsmeets:
Hi @Peter9192 , I implemented all your feedback. If you approve, I think this PR is ready to be merged!

doc/api/esmvalcore.api.recipe.rst
```diff
@@ -59,7 +59,12 @@ Sessions can also be explicitly specified.
     >>> output = recipe.run(session)
     <log messages>
 
-:py:meth:`esmvalcore.experimental.Recipe.run` returns an object that contains the locations of the data and figures (not implemented yet).
+:py:meth:`esmvalcore.experimental.Recipe.run` returns an dictionary of objects that can be used to inspect
+the output of the recipe. The output is an instance of :py:class:`esmvalcore.experimental.recipe_output.ImageItem` or
+:py:class:`esmvalcore.experimental.recipe_output.ImageItem` depending on its type.
+
+For working with recipe output, see: ref:`_api_recipe_output`.
```
Peter9192, line 66:
```suggestion
For working with recipe output, see: :ref:`_api_recipe_output`.
```

doc/api/esmvalcore.api.recipe.rst
```diff
@@ -59,7 +59,12 @@ Sessions can also be explicitly specified.
     >>> output = recipe.run(session)
     <log messages>
 
-:py:meth:`esmvalcore.experimental.Recipe.run` returns an object that contains the locations of the data and figures (not implemented yet).
+:py:meth:`esmvalcore.experimental.Recipe.run` returns an dictionary of objects that can be used to inspect
+the output of the recipe. The output is an instance of :py:class:`esmvalcore.experimental.recipe_output.ImageItem` or
+:py:class:`esmvalcore.experimental.recipe_output.ImageItem` depending on its type.
```
Peter9192, line 64: I think these also need to be updated right?

doc/api/esmvalcore.api.recipe_output.rst
```diff
@@ -0,0 +1,138 @@
...
+Or ``iris``:
+
+.. code:: python
+
+    >>> cube = data_file.load_iris()
+    >>> type(cube)
+    iris.cube.CubeList
+
+
+API reference
```
Peter9192, line 135: This page is so much better already! I just wonder if we should still add something about citation info or dataset attributes, etc. Since these attributes are listed in the API reference below, it might be nice to provide a very short example to show how they can be used.

And should we include a statement about the output of the preprocessing tasks? Since they are removed unless `remove_preproc_dir` is set to false?

tests/sample_data/experimental/test_run_recipe.py
```diff
@@ -23,11 +29,26 @@ def recipe():
...
 
     output = recipe.run()
 
-    assert not output  # output is not yet defined
+    assert len(output) > 0
+    assert isinstance(output, RecipeOutput)
+
+    for task, task_output in output.items():
+        assert isinstance(task_output, TaskOutput)
+        assert len(task_output) > 0
```
Peter9192, line 47: If this recipe doesn't contain diagnostics, shouldn't the task output be empty?

Peter9192 (approved):
Hey Stef! This is starting to be very very cool! I have a couple of follow-up comments/questions, but all-in-all I'm very happy with this PR. 

On codacy: shall we just move the imports to the top?

tests/sample_data/experimental/test_run_recipe.py
stefsmeets: Apparently it returns the preprocessor files.

doc/api/esmvalcore.api.recipe_output.rst
stefsmeets: > This page is so much better already! I just wonder if we should still add something about citation info or dataset attributes, etc. Since these attributes are listed in the API reference below, it might be nice to provide a very short example to show how they can be used.

I will add a few lines about this. 👍

> And should we include a statement about the output of the preprocessing tasks? Since they are removed unless `remove_preproc_dir` is set to false? (PS: I just realized the whole preprocessor file parsing still needs to be added. Nevertheless, a statement about this might be helpful.)

Good point! I will look into how handle preprocessor output / specific task output / edge cases in a separate PR (including documentation) and keep this PR limited to recipe output only.

stefsmeets:
> On codacy: shall we just move the imports to the top?

Can't do that, `esmvaltool` is imported to find the `diagnostics` path, and `xarray` is imported as a convenience function, but we don't have it in our dependencies (see https://github.com/ESMValGroup/ESMValCore/pull/957#issuecomment-764655790).

doc/api/esmvalcore.api.recipe.rst
stefsmeets: Apparently the label should not include the underscore... 🙄

    :ref:`api_recipe_output`

stefsmeets:
This PR is ready to be merged @nielsdrost !

esmvalcore/_task.py
```diff
@@ -625,12 +629,12 @@ def _collect_provenance(self):
...
-            self.__class__.__name__,
-            self.script,
-            pprint.pformat(self.settings, indent=2),
-            super(DiagnosticTask, self).str(),
-        )
+        settings_string = pprint.pformat(self.settings, indent=2)
+        txt = (f"{self.__class__.__name__}:\n"
+               f"script: {self.script}\n"
+               f"settings:\n{settings_string}\n"
+               f"{super(DiagnosticTask, self)}\n")
```
nielsdrost, line 636: Any reason why you use arguments to super() here?

nielsdrost (changes requested):

esmvalcore/_task.py
stefsmeets: This was already there, I don't know why so I left it.

nielsdrost: @stefsmeets promises to fix this  in followup PR ;-)

nielsdrost (approved):

esmvalcore/_task.py
bouweandela: > Any reason why you use arguments to super() here?

It's Python 2 syntax

Issue #956, opened 2021-01-20
Multimodel regression tests don't fail, even though plev in sample data are inconsistent.


Peter9192:
**Describe the bug**
In #950 I implemented regression tests for a new/alternative multi-cube statistics engine. It fails because the sample data has differering coordinates on plev: some of the cubes have many levels, most only two. But most importantly, there are 2 cubes that have values of `[100000.00000001,  92500.00000001]` for the `air_pressure` coordinate.

To my surprise, this *doesn't* get picked up by the current implementation, which happily uses the plev coordinate of the first input cube for the result. In this case it's safe to assume that the levels are supposed to be the same, but this may not always be the case. I would sleep better if ESMValTool would catch these kind of errors.

stefsmeets:
Good to see that this is picked up by the new multi-model tests! Initially I had an `extract_levels` function in the preprocessor for these tests, but I took them out, because apparently these are set by the CMOR standards. So I assumed all to be the same.

I would rather note code too many work-arounds in the tests. One way to solve it would be to ignore the problematic datasets here: `https://github.com/ESMValGroup/ESMValTool_sample_data/blob/master/datasets.yml`.

stefsmeets:
Made an issue here:
https://github.com/ESMValGroup/ESMValTool_sample_data/issues/19

Peter9192:
I wonder if the cmor checks/fixes pick this up if you run it normally in a recipe. In that case we might need to pull the sample data through the checks before using it in the tests.

stefsmeets:
I don't know if the fixes work with this, but the risk is that we are re-implementing the entire tool just to get the tests to run.

valeriupredoi:
@Peter9192 can you pls point me to the branch and test that displays this behaviour mate? :beer:

Peter9192:
> can you pls point me to the branch and test that displays this behaviour mate? 🍺

It's already merged: https://github.com/ESMValGroup/ESMValCore/blob/a285ee6e4a6b423105cd9e86fc6ddb1565cf177e/tests/sample_data/multimodel_statistics/test_multimodel.py#L168 and my question is why doesn't this fail, and should it?

See this PR where the test was first implemented (https://github.com/ESMValGroup/ESMValCore/pull/856)

valeriupredoi:
cheers, will have a look now!

valeriupredoi:
[...]

Issue #744, opened 2020-08-07
Move temporal 'regridding' from multi_model_statistics preprocessor to regrid_time preprocessor


bouweandela:
At the moment, the preprocessor `multi_model_statistics` will also attempt to automatically 'regrid' the time coordinate of the incoming datasets. Other coordinates, such as horizontal and vertical coordinates, need to be explicitly regridded by defining a preprocesor in the recipe. It would be good to be consistent and make the temporal regridding more visible and move the temporal regridding to the `regrid_time` preprocessor. If the datasets have a time coordinate, users would then need to specify the `regrid_time` preprocessor in the recipe as well as as the `regrid` and `extract_levels`.

This would also fix the problem described in #738, that `multi_model_statistics` does not work for variables that do not have a time coordinate.

Peter9192:
It would be really useful if regrid_time could then also check the calendar (as does mmstats, also see #685). I keep running into incompatible time coordinates because the calendars differ.

schlunma:
> It would be really useful if regrid_time could then also check the calendar (as does mmstats, also see #685). I keep running into incompatible time coordinates because the calendars differ.

Yes please!

I (and @hb326) encountered problems with [`iris.quickplot.plot`](https://scitools.org.uk/iris/docs/latest/iris/iris/quickplot.html#iris.quickplot.plot) when the time coordinates of the different cubes have different calendars. As a fix we use the following code:

```python
def _unify_time_coord(cube):
    """Unify time coordinate of cube."""
    if not cube.coords('time', dim_coords=True):
        return
    time_coord = cube.coord('time')
    dates_points = time_coord.units.num2date(time_coord.points)
    dates_bounds = time_coord.units.num2date(time_coord.bounds)
    new_units = Unit('days since 1850-01-01 00:00:00')
    new_time_coord = iris.coords.DimCoord(
        new_units.date2num(dates_points),
        bounds=new_units.date2num(dates_bounds),
        var_name='time',
        standard_name='time',
        long_name='time',
        units=new_units,
    )
    coord_dims = cube.coord_dims('time')
    cube.remove_coord('time')
    cube.add_dim_coord(new_time_coord, coord_dims)
```

It think it would make sense to include this in `regrid_time`.

zklaus:
I support the idea, but careful with the suggested fix above. That may be appropriate in a visualization script where you know the input, but I see two problems with it:
[...]

PR #968, opened 2021-01-28
First attempt at replacing multimodel stats with a native iris altern…

### LINKED ISSUE #781: Making the multi-model statistics preprocessor lazy

Peter9192:
## Description
This PR re-implements the multi-model statistics preprocessor. This implementation delegates more of the work to native iris functions. This enables, among others, lazy evaluation for some common operators.

-   Closes #781 
-   Link to documentation:

***

## Before you get started

## Checklist

-   [ ] PR has a descriptive title for the [changelog](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#branches-pull-requests-and-code-review)
-   [ ] Code follows the [style guide](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#code-style)

***

### DIFF
--- a/esmvalcore/preprocessor/_multimodel.py
+++ b/esmvalcore/preprocessor/_multimodel.py
@@ -1,180 +1,15 @@
 """Functions to compute multi-cube statistics."""
 
 import logging
-import re
 from datetime import datetime
-from functools import partial, reduce
+from functools import reduce
 
 import cf_units
 import iris
 import numpy as np
-import scipy
+from iris.experimental.equalise_cubes import equalise_attributes
 
 logger = logging.getLogger(__name__)
-
-
-def _plev_fix(dataset, pl_idx):
-    """Extract valid plev data.
-
-    This function takes care of situations in which certain plevs are
-    completely masked due to unavailable interpolation boundaries.
-    """
-    if np.ma.is_masked(dataset):
-        # keep only the valid plevs
-        if not np.all(dataset.mask[pl_idx]):
-            statj = np.ma.array(dataset[pl_idx], mask=dataset.mask[pl_idx])
-        else:
-            logger.debug('All vals in plev are masked, ignoring.')
-            statj = None
-    else:
-        mask = np.zeros_like(dataset[pl_idx], bool)
-        statj = np.ma.array(dataset[pl_idx], mask=mask)
-
-    return statj
-
-
-def _quantile(data, axis, quantile):
-    """Calculate quantile.
-
-    Workaround for calling scipy's mquantiles with arrays of >2
-    dimensions Similar to iris' _percentiles function, see their
-    discussion: https://github.com/SciTools/iris/pull/625
-    """
-    # Ensure that the target axis is the last dimension.
-    data = np.rollaxis(data, axis, start=data.ndim)
-    shape = data.shape[:-1]
-    # Flatten any leading dimensions.
-    if shape:
-        data = data.reshape([np.prod(shape), data.shape[-1]])
-    # Perform the quantile calculation.
-    result = scipy.stats.mstats.mquantiles(data,
-                                           quantile,
-                                           axis=-1,
-                                           alphap=1,
-                                           betap=1)
-    # Ensure to unflatten any leading dimensions.
-    if shape:
-        result = result.reshape(shape)
-    # Check whether to reduce to a scalar result
-    if result.shape == (1, ):
-        result = result[0]
-
-    return result
-
-
-def _compute_statistic(data, statistic_name):
-    """Compute multimodel statistic."""
-    data = np.ma.array(data)
-    statistic = data[0]
-
-    if statistic_name == 'median':
-        statistic_function = np.ma.median
-    elif statistic_name == 'mean':
-        statistic_function = np.ma.mean
-    elif statistic_name == 'std':
-        statistic_function = np.ma.std
-    elif statistic_name == 'max':
-        statistic_function = np.ma.max
-    elif statistic_name == 'min':
-        statistic_function = np.ma.min
-    elif re.match(r"^(p\d{1,2})(\.\d*)?$", statistic_name):
-        # percentiles between p0 and p99.99999...
-        quantile = float(statistic_name[1:]) / 100
-        statistic_function = partial(_quantile, quantile=quantile)
-    else:
-        raise ValueError(f'No such statistic: `{statistic_name}`')
-
-    # no plevs
-    if len(data[0].shape) < 3:
-        # get all NOT fully masked data - u_data
-        # data is per time point
-        # so we can safely NOT compute stats for single points
-        if data.ndim == 1:
-            u_datas = data
-        else:
-            u_datas = [d for d in data if not np.all(d.mask)]
-        if len(u_datas) > 1:
-            statistic = statistic_function(data, axis=0)
-        else:
-            statistic.mask = True
-        return statistic
-
-    # plevs
-    for j in range(statistic.shape[0]):
-        plev_check = []
-        for cdata in data:
-            fixed_data = _plev_fix(cdata, j)
-            if fixed_data is not None:
-                plev_check.append(fixed_data)
-
-        # check for nr datasets
-        if len(plev_check) > 1:
-            plev_check = np.ma.array(plev_check)
-            statistic[j] = statistic_function(plev_check, axis=0)
-        else:
-            statistic.mask[j] = True
-
-    return statistic
-
-
-def _put_in_cube(template_cube, cube_data, statistic, t_axis):
-    """Quick cube building and saving."""
-    tunits = template_cube.coord('time').units
-    times = iris.coords.DimCoord(t_axis,
-                                 standard_name='time',
-                                 units=tunits,
-                                 var_name='time',
-                                 long_name='time')
-    times.bounds = None
-    times.guess_bounds()
-
-    coord_names = [c.long_name for c in template_cube.coords()]
-    coord_names.extend([c.standard_name for c in template_cube.coords()])
-    if 'latitude' in coord_names:
-        lats = template_cube.coord('latitude')
-    else:
-        lats = None
-    if 'longitude' in coord_names:
-        lons = template_cube.coord('longitude')
-    else:
-        lons = None
-
-    # no plevs
-    if len(template_cube.shape) == 3:
-        cspec = [(times, 0), (lats, 1), (lons, 2)]
-    # plevs
-    elif len(template_cube.shape) == 4:
-        plev = template_cube.coord('air_pressure')
-        cspec = [(times, 0), (plev, 1), (lats, 2), (lons, 3)]
-    elif len(template_cube.shape) == 1:
-        cspec = [
-            (times, 0),
-        ]
-    elif len(template_cube.shape) == 2:
-        # If you're going to hardwire air_pressure into this,
-        # might as well have depth here too.
-        plev = template_cube.coord('depth')
-        cspec = [
-            (times, 0),
-            (plev, 1),
-        ]
-
-    # correct dspec if necessary
-    fixed_dspec = np.ma.fix_invalid(cube_data, copy=False, fill_value=1e+20)
-    # put in cube
-    stats_cube = iris.cube.Cube(fixed_dspec,
-                                dim_coords_and_dims=cspec,
-                                long_name=statistic)
-    coord_names = [coord.name() for coord in template_cube.coords()]
-    if 'air_pressure' in coord_names:
-        if len(template_cube.shape) == 3:
-            stats_cube.add_aux_coord(template_cube.coord('air_pressure'))
-
-    stats_cube.var_name = template_cube.var_name
-    stats_cube.long_name = template_cube.long_name
-    stats_cube.standard_name = template_cube.standard_name
-    stats_cube.units = template_cube.units
-    return stats_cube
 
 
 def _get_consistent_time_unit(cubes):
@@ -212,7 +47,6 @@ def _unify_time_coordinates(cubes):
         if 0 not in np.diff(years):
             # yearly data
             dates = [datetime(year, 7, 1, 0, 0, 0) for year in years]
-
         elif 0 not in np.diff(months):
             # monthly data
             dates = [
@@ -242,115 +76,110 @@ def _unify_time_coordinates(cubes):
         cube.coord('time').guess_bounds()
 
 
-def _get_time_slice(cubes, time):
-    """Fill time slice array with cubes' data if time in cube, else mask."""
-    time_slice = []
-    for cube in cubes:
-        cube_time = cube.coord('time').points
-        if time in cube_time:
-            idx = int(np.argwhere(cube_time == time))
-            subset = cube.data[idx]
-        else:
-            subset = np.ma.empty(list(cube.shape[1:]))
-            subset.mask = True
-        time_slice.append(subset)
-    return time_slice
-
-
-def _assemble_data(cubes, statistic, span='overlap'):
-    """Get statistical data in iris cubes."""
-    # New time array representing the union or intersection of all cubes
+def _interpolate(cubes, func):
+    """Expand or subset cubes so they share a common time span."""
+    _unify_time_coordinates(cubes)
     time_spans = [cube.coord('time').points for cube in cubes]
+    new_times = reduce(func, time_spans)
+    # new_times = cubes[0].coord('time').units.num2date(new_times)
+    sample_points = [('time', new_times)]
+    scheme = iris.analysis.Nearest(extrapolation_mode='nan')
+    return [cube.interpolate(sample_points, scheme) for cube in cubes]
+
+
+def _extend(cubes):
+    return _interpolate(cubes, np.intersect1d)
+
+
+def _subset(cubes):
+    """Only keep the times that are present in all cubes."""
+    return _interpolate(cubes, np.union1d)
+
+
+def _align(cubes, span):
+    """Expand or subset cubes so they share a common time span."""
+
     if span == 'overlap':
-        new_times = reduce(np.intersect1d, time_spans)
+        new_cubes = _subset(cubes)
     elif span == 'full':
-        new_times = reduce(np.union1d, time_spans)
-    n_times = len(new_times)
+        new_cubes = _extend(cubes)
+    else:
+        raise ValueError("Unknown value for span. Expected 'full' or 'overlap'"
+                         "got {}".format(span))
 
-    # Target array to populate with computed statistics
-    new_shape = [n_times] + list(cubes[0].shape[1:])
-    stats_data = np.ma.zeros(new_shape, dtype=np.dtype('float32'))
+    for cube in new_cubes:
+        cube.coord('time').guess_bounds()
+    return new_cubes
 
-    # Realize all cubes at once instead of separately for each time slice
-    _ = [cube.data for cube in cubes]
 
-    # Make time slices and compute stats
-    for i, time in enumerate(new_times):
-        time_data = _get_time_slice(cubes, time)
-        stats_data[i] = _compute_statistic(time_data, statistic)
+def _combine(cubes, dim='new_dim'):
+    """Merge iris cubes into a single big cube with new dimension."""
+    equalise_attributes(cubes)
 
-    template = cubes[0]
-    stats_cube = _put_in_cube(template, stats_data, statistic, new_times)
-    return stats_cube
+    for i, cube in enumerate(cubes):
+        concat_dim = iris.coords.AuxCoord(i, var_name=dim)
+        cube.add_aux_coord(concat_dim)
+
+    cubes = iris.cube.CubeList(cubes)
+    return cubes.merge_cube()
+
+
+def _compute(cube, statistic, dim='new_dim'):
+    """Compute statistic."""
+    operators = vars(iris.analysis)
+
+    try:
+        operator = operators[statistic.upper()]
+    except KeyError as err:
+        raise ValueError(
+            f'Statistic `{statistic}` not supported in',
+            '`ensemble_statistics`. Choose supported operator from',
+            '`iris.analysis package`.') from err
+
+    # This will always return a masked array
+    return cube.collapsed('concat_dim', operator)
 
 
 def _multicube_statistics(cubes, statistics, span):
-    """Compute statistics over multiple cubes.
+    """Compute multi-cube statistics.
 
-    Can be used e.g. for ensemble or multi-model statistics.
+    Cubes are merged and subsequently collapsed along a new auxiliary
+    coordinate. Inconsistent attributes will be removed.
 
-    This function was designed to work on (max) four-dimensional data:
-    time, vertical axis, two horizontal axes.
+    This function deals with non-homogeneous cubes by taking the time union
+    computed across a common overlap in time (set span: overlap) or across the
+    full length in time of each model (set span: full). Apart from the time
+    coordinate, cubes must have consistent shapes.
 
-    Apart from the time coordinate, cubes must have consistent shapes. There
-    are two options to combine time coordinates of different lengths, see
-    the `span` argument.
+    This method uses iris' built in functions, exposing the operators in
+    iris.analysis and supporting lazy evaluation.
+
+    Note: some of the operators in iris.analysis require additional
+    arguments, such as percentiles or weights. These operators are
+    currently not supported.
 
     Parameters
     ----------
     cubes: list
-        list of cubes over which the statistics will be computed;
+        list of cubes to be used in multimodel stat computation;
     statistics: list
-        statistical metrics to be computed. Available options: mean, median,
-        max, min, std, or pXX.YY (for percentile XX.YY; decimal part optional).
+        statistical measures to be computed. Choose from the
+        operators listed in the iris.analysis package.
     span: str
-        overlap or full; if overlap, statitsticss are computed on common time-
-        span; if full, statistics are computed on full time spans, ignoring
-        missing data.
+        'overlap' or 'full'. If 'overlap', statitsticss are computed on common
+        time span; if 'full', statistics are computed on full time spans,
+        ignoring missing data.
 
     Returns
     -------
     dict
         dictionary of statistics cubes with statistics' names as keys.
-
-    Raises
-    ------
-    ValueError
-        If span is neither overlap nor full.
     """
-    if len(cubes) < 2:
-        logger.info('Found only 1 cube; no statistics computed for %r',
-                    list(cubes)[0])
-        return {statistic: cubes[0] for statistic in statistics}
-
-    logger.debug('Multicube statistics: computing: %s', statistics)
-
-    # Reset time coordinates and make cubes share the same calendar
-    _unify_time_coordinates(cubes)
-
-    # Check whether input is valid
-    if span == 'overlap':
-        # check if we have any time overlap
-        times = [cube.coord('time').points for cube in cubes]
-        overlap = reduce(np.intersect1d, times)
-        if len(overlap) <= 1:
-            logger.info("Time overlap between cubes is none or a single point."
-                        "check datasets: will not compute statistics.")
-            return cubes
-        logger.debug("Using common time overlap between "
-                     "datasets to compute statistics.")
-    elif span == 'full':
-        logger.debug("Using full time spans to compute statistics.")
-    else:
-        raise ValueError(
-            "Unexpected value for span {}, choose from 'overlap', 'full'".
-            format(span))
-
-    # Compute statistics
+    aligned_cubes = _align(cubes, span=span)
+    big_cube = _combine(aligned_cubes)
     statistics_cubes = {}
     for statistic in statistics:
-        statistic_cube = _assemble_data(cubes, statistic, span)
-        statistics_cubes[statistic] = statistic_cube
+        statistics_cubes[statistic] = _compute(big_cube, statistic)
 
     return statistics_cubes
 


stefsmeets:
Hi @Peter9192 , I just noticed https://github.com/ESMValGroup/ESMValCore/issues/1009 got fixed in a separate PR which is causing a merge conflict. Just wanted to check with you if this is still an issue in the changes we are making here (meaning can we ignore it)?

Peter9192:
> 
> 
> Hi @Peter9192 , I just noticed #1009 got fixed in a separate PR (#1010) which is causing a merge conflict. Just wanted to check with you if this is still an issue in the changes we are making here (meaning can we ignore it)?

If this is a feature that's needed than there should be a test for it. I think the new implementation will work only if either no or all cubes have this 'altitude' coordinate, not sure what the current master does...

stefsmeets:
Multimodel regression tests with sample data show that the new implementation reproduces the old implementation with an accuracy of 5 to 6 decimal places, which is pretty nice!

Some tests cannot be run because of https://github.com/ESMValGroup/ESMValCore/issues/956, and these have been marked with `xfail`.

Peter9192:
> Multimodel regression tests with sample data show that the new implementation reproduces the old implementation with an accuracy of 5 to 6 decimal places, which is pretty good!

That sounds similar to my findings here: https://github.com/ESMValGroup/ESMValCore/pull/685#issuecomment-758783758
Note that numpy all_close's `decimals` argument is not the same as the precision you'd expect based on a `float32` dtype, this confused me back then.

stefsmeets:
> That sounds similar to my findings here: [#685 (comment)](https://github.com/ESMValGroup/ESMValCore/pull/685#issuecomment-758783758)
> Note that numpy all_close's `decimals` argument is not the same as the precision you'd expect based on a `float32` dtype, this confused me back then.

Yeah, it is something different. I'm just reporting what I see in the tests. It reports the maximum absolute difference. Part of this seems to be rounding errors too. Four tests fail with something like:

```
E       AssertionError: 
E       Arrays are not almost equal to 6 decimals
E       
E       Mismatched elements: 452 / 744 (60.8%)
E       Max absolute difference: 1.01725261e-05
E       Max relative difference: 3.97164548e-08
E        x: MaskedArray([[[[247.463582, 247.496257],
E                      [247.35909 , 247.368022],
E                      [247.756348, 247.765279]],...
E        y: MaskedArray([[[[247.46358, 247.49626],
E                      [247.35909, 247.36803],
E                      [247.75635, 247.76527]],...
```

Peter9192:
Right, that seems to be in the order of numerical precision.

stefsmeets:
Hi @Peter9192 , I think I'm more or less finished with this PR. Could you have a look at the implementation and let me know if it fits with what you had in mind before we send it off for review?

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -3,184 +3,34 @@
...
+    'hmean': iris.analysis.HMEAN,  # not lazy in iris
+    'max': iris.analysis.MAX,
+    'median': iris.analysis.MEDIAN,  # not lazy in iris
+    'min': iris.analysis.MIN,
+    'rms': iris.analysis.RMS,
+    'sum': iris.analysis.SUM,
+    'mean': iris.analysis.MEAN,
+    'std_dev': iris.analysis.STD_DEV,
+    'variance': iris.analysis.VARIANCE,
+    # not directly supported
```
Peter9192, line 27:
```suggestion
    # The following require extra kwargs. ATM this is only supported for percentiles
```

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -248,115 +97,206 @@ def _unify_time_coordinates(cubes):
...
+        new_times = reduce(np.intersect1d, all_times)
+        if new_times.size < 1:
+            raise ValueError("No time overlap found between input cubes.")
+    else:
+        raise ValueError("Unknown value for span. Expected 'full' or 'overlap'"
+                         "got {}".format(span))
+
+    return new_times
+
+
```
Peter9192, line 118:
```suggestion
```

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -248,115 +97,206 @@ def _unify_time_coordinates(cubes):
...
+def _subset(cube, time_points):
+    """Subset cube to given time range."""
+    begin = cube.coord('time').units.num2date(time_points[0])
+    end = cube.coord('time').units.num2date(time_points[-1])
+    constraint = iris.Constraint(time=lambda cell: begin <= cell.point <= end)
+    return cube.extract(constraint)
+
+
+def _extend(cube, time_points):
+    """Extend cube to a specified time range."""
```
Peter9192, line 140: expand on the mechanism and considerations?

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -248,115 +97,206 @@ def _unify_time_coordinates(cubes):
...
+    cube.coord('time').bounds = None
+    cube_points = cube.coord('time').points
+
+    if not np.all(np.diff(cube_points) > 0):
+        raise ValueError('Time points are not monotonic')
+
```
Peter9192, line 146:
```suggestion
```

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -248,115 +97,206 @@ def _unify_time_coordinates(cubes):
...
-    This function was designed to work on (max) four-dimensional data:
-    time, vertical axis, two horizontal axes.
+        cube.cell_methods = None
 
-    Apart from the time coordinate, cubes must have consistent shapes. There
-    are two options to combine time coordinates of different lengths, see
-    the `span` argument.
+        for coord in cube.coords():
+            coord.long_name = None
+            coord.attributes = None
```
Peter9192, line 231: use iris.resolve.resolve?

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -248,115 +97,206 @@ def _unify_time_coordinates(cubes):
...
-    # Make time slices and compute stats
-    for i, time in enumerate(new_times):
-        time_data = _get_time_slice(cubes, time)
-        stats_data[i] = _compute_statistic(time_data, statistic)
 
-    template = cubes[0]
-    stats_cube = _put_in_cube(template, stats_data, statistic, new_times)
-    return stats_cube
+def _combine(cubes, dim='new_dim'):
+    """Merge iris cubes into a single big cube with new dimension."""
```
Peter9192, line 210: add assumption all same shape?

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -248,115 +97,206 @@ def _unify_time_coordinates(cubes):
...
+
+def _multicube_statistics(cubes, statistics, span):
+    """Compute statistics over multiple cubes.
+
+    Can be used e.g. for ensemble or multi-model statistics.
+
+    Cubes are merged and subsequently collapsed along a new auxiliary
+    coordinate. Inconsistent attributes will be removed.
+    """
+    aligned_cubes = _align(cubes, span=span)
```
Peter9192, line 295: Maybe copy the original cubes so we don't modify them inplace?

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -248,115 +97,206 @@ def _unify_time_coordinates(cubes):
...
+def _combine(cubes, dim='new_dim'):
+    """Merge iris cubes into a single big cube with new dimension."""
+    equalise_attributes(cubes)  # in-place
 
+    for i, cube in enumerate(cubes):
+        try:
+            # since cubes are updated in-place, new coord may already exist
+            cube.remove_coord(dim)
+        except iris.exceptions.CoordinateNotFoundError:
+            pass
```
Peter9192, line 218:
```suggestion
    for i, cube in enumerate(cubes):
```

stefsmeets: Couldn't apply the change here, so I did it locally 👍

esmvalcore/preprocessor/_multimodel.py
stefsmeets: Good point! Do you think this will have performance implications for large non-lazy datasets?

esmvalcore/preprocessor/_multimodel.py
stefsmeets: [Resolve](https://scitools-iris.readthedocs.io/en/stable/generated/api/iris/common/resolve.html#iris.common.resolve.Resolve) seems to be for resolving two cubes, not a list of cubes. I don't think using this here will be simpler than stripping the metadata.

stefsmeets:
Hi @sloosvel @schlunma @valeriupredoi , this PR is ready for review. Can I ask (one of) you have a look and let me know what you think?

sloosvel:
Sure, I can/will take a look.

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -248,115 +98,190 @@ def _unify_time_coordinates(cubes):
...
+        raise ValueError(f"Invalid argument for span: {span!r}"
+                         "Must be one of 'overlap', 'full'.")
 
-    # Target array to populate with computed statistics
-    new_shape = [n_times] + list(cubes[0].shape[1:])
-    stats_data = np.ma.zeros(new_shape, dtype=np.dtype('float32'))
+    for cube in new_cubes:
+        # Make sure bounds exist and are consistent
+        cube.coord('time').bounds = None
+        cube.coord('time').guess_bounds()
```
sloosvel, line 189: Maybe you can use the `get_time_bounds` function in `_time.py`. Sometimes the bounds that you get using `guess_bounds` are shifted or have negative values if the point starts at 0.

stefsmeets: Thanks for making me aware of this Is this an issue you are running into with the current implementation? Because in `master` we also use `guess_bounds` for the time coordinate. 

Looking at this, we would have to infer the frequency argument ourselves. Afaik it is not trivial to 'guess' the time frequency (it is not stored as an attribute on the cube), and it would have to work for all possible frequencies. I think the best way to go about it is to make the `frequency` argument optional and guess it in the `_get_time_bounds` function with `frequency=None`. 

I think that is a bit out of scope for this PR. And considering that this issue is present in the current `master` branch, could you make an issue so that we follow-up in a separate PR?

stefsmeets: Thanks @Peter9192 , good point, I overlooked that. I was looking at the `_align` function, where this may be more difficult.

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -248,115 +98,190 @@ def _unify_time_coordinates(cubes):
...
+    """Merge iris cubes into a single big cube with new dimension.
 
+    This assumes that all input cubes have the same shape.
+    """
+    equalise_attributes(cubes)  # in-place
 
-def _multicube_statistics(cubes, statistics, span):
-    """Compute statistics over multiple cubes.
+    for i, cube in enumerate(cubes):
+        concat_dim = iris.coords.AuxCoord(i, var_name=dim)
```
valeriupredoi, line 202: why are you adding a scalar aux coord here? `new_dim` is in fact the dataset index

valeriupredoi (changes requested):
nice work guys! I have a few questions that I'd appreciate your answers. As a general look at this - as I see it - is we are transitioning from a mechanical approack of slicing over time and levs and throwing all that slicing in a couple time-consuming but memory-light for loops to creating lazy data arrays and renouncing the for loops in favour of working directly with the overlap/padded lazy arrays - neat! But - you are relying heavily on cube conncatenation and merging  which are known to be brittle and you are copying a lot of data - so before I can approve this I would be very interested in seeing a performance comparison with the old version - for some really hefty datasets :beer:

esmvalcore/preprocessor/_multimodel.py
stefsmeets: Just had a go at trying to see if there is an easy way to do this, but I do not see a way to do this cleanly. @valeriupredoi @sloosvel , would you be OK if we handle this in a separate PR or make it part of #744?

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -248,115 +98,190 @@ def _unify_time_coordinates(cubes):
...
+        new_slice = template_cube.copy()
+        new_slice.coord('time').points = float(time_point)
+        cube_list.append(new_slice)
+
+    cube_list.append(cube)
+
+    for time_point in pad_end:
+        new_slice = template_cube.copy()
+        new_slice.coord('time').points = float(time_point)
+        cube_list.append(new_slice)
```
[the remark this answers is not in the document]
stefsmeets: Clearly not my best work, but I do not know a better way to do this with iris that is still lazy. 😅

I tried going the route you described by generating the data object directly, but I ran into the issue that a _lot_ of code is needed to cleanly copy over the attributes and dimensional coordinates (see iris.extrapolate. So, we went for the solution that is easy to read.

Note that this happens only for span='full' (which is used by 3 recipes), in the case that the time coordinates are not already aligned (an edge case to an edge case if you will).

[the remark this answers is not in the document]
stefsmeets: If you have a code snippet to generate the left/right cubes, that would be super helpful! 💪

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -248,115 +98,190 @@ def _unify_time_coordinates(cubes):
...
+    begin = cube_points[0]
+    end = cube_points[-1]
+
+    pad_begin = time_points[time_points < begin]
+    pad_end = time_points[time_points > end]
+
+    if (len(pad_begin)) == 0 and (len(pad_end) == 0):
+        return cube
+
+    template_cube = cube[:1].copy()
```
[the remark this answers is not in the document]
stefsmeets: Good question! We want to get a cube with the time axis intact. `cube[0]` returns a scalar cube.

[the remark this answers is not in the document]
stefsmeets: Just tested this to make sure that `.copy` is also a lazy operation if the input cube is lazy.

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -248,115 +98,190 @@ def _unify_time_coordinates(cubes):
...
+    all_time_arrays = [cube.coord('time').points for cube in cubes]
 
-def _assemble_data(cubes, statistic, span='overlap'):
-    """Get statistical data in iris cubes."""
-    # New time array representing the union or intersection of all cubes
-    time_spans = [cube.coord('time').points for cube in cubes]
     if span == 'overlap':
-        new_times = reduce(np.intersect1d, time_spans)
+        common_time_points = reduce(np.intersect1d, all_time_arrays)
+        new_cubes = [_subset(cube, common_time_points) for cube in cubes]
```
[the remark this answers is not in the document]
stefsmeets: Yes, I added unit tests specifically to make sure that the entire chain is fully lazy.

https://github.com/ESMValGroup/ESMValCore/blob/0bd079b4336366ce8c1d0f2fdc8fc1d10c7c649f/tests/unit/preprocessor/_multimodel/test_multimodel.py#L168
https://github.com/ESMValGroup/ESMValCore/blob/0bd079b4336366ce8c1d0f2fdc8fc1d10c7c649f/tests/unit/preprocessor/_multimodel/test_multimodel.py#L189

esmvalcore/preprocessor/_multimodel.py
stefsmeets: I think this is to make sure we have a dimension to collapse the cube against, for the statistics calculation. 
@Peter9192 do you have something to add to this?

esmvalcore/preprocessor/_multimodel.py
valeriupredoi: either way works for me, cheers :+1:

esmvalcore/preprocessor/_multimodel.py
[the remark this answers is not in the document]
valeriupredoi: OK about indexing, I thought you were selecting the 0th time cube, if you do `cube[0,]` or `cube[0]` that shouldn't return a scalar cube, but the 0th time cube of the original cube - yes, the time axis would be scalar or better said, vanish so if you still need it yes, do the [:1] indexing :+1: About copy - yes, a copy doesn't realize the data but even if unrealized, it still can be a hefty cube in memory so that's why I am on about trying not to make too many copies

esmvalcore/preprocessor/_multimodel.py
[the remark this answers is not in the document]
valeriupredoi: terrific! cheers :beer:

esmvalcore/preprocessor/_multimodel.py
sloosvel: Sure, it's not that big of an issue anyway.

esmvalcore/preprocessor/_multimodel.py
valeriupredoi: aha got it! nice trick, but be 100% sure that the collapse is happening along the ordered time axis and not along any of the spatial dimensional axes, Also what if there's only one cube - all will work fine but the collapse itself doesn't make sense statistically ie it will return values from the only cube as statistics which is statistically incorrect and pointless. I don't see any more warnings about stats being computed on < 2 data arrays like we had before

esmvalcore/preprocessor/_multimodel.py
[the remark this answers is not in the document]
valeriupredoi: actually - you're right, just noticed it is indeed only for `full` (the previous implementation was terrible so this one's definitely better, not even gonna bother you with cube pads :grin: )

esmvalcore/preprocessor/_multimodel.py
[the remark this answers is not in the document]
valeriupredoi: the problem is that weighted collapse is still not lazy in iris AFASK

esmvalcore/preprocessor/_multimodel.py
Peter9192: Good point about single cube inputs, we should probably get that check back indeed. 

On the new dim: actually we are *not* collapsing them along the time axis or any of the spatial dimensions. We're introducing a temporary "multicube" dimension, (`new_dim`), stack the cubes along that new dimension, and then collapse them again to calculate the multicube statistics. This only works if all cubes have consistent shapes and an auxiliary scalar indexing coord, see the [documentation on `iris.merge()`](https://scitools-iris.readthedocs.io/en/latest/userguide/merge_and_concat.html#using-cubelist-merge)

valeriupredoi:
OK bad news guys - this implementation is running out of memory on JASMIN sci 2 - see the trace at the end of this post:
I ran the original implementation (which I can't believe it's 3 years old already ha!) and on my sample data I got an average (single-process) run time of about 85s and a max mem intake of 4.7GB
```
params
======
stats: mean, median
span: overlap

master
======
start: 13:16:09,035
end:   13:17:30,259
time: 81s
mem: 4.6GB

start: 13:25:23,065
end: 13:26:46,256
time: 83s
mem: 4.7GB

branch
======
start: 13:33:26
end(fail): 13:33:58
mem: 6.9GB
fails at: median

second run got killed outright, at median again
```

trace:
```
2021-03-10 13:33:58,609 UTC [9017] INFO    Maximum memory used (estimate): 6.9 GB
2021-03-10 13:33:58,610 UTC [9017] INFO    Sampled every second. It may be inaccurate if short but high spikes in memory consumption occur.
2021-03-10 13:33:58,611 UTC [9017] ERROR   Program terminated abnormally, see stack trace below for more information:
  File "/home/users/valeriu/anaconda3/envs/release220/lib/python3.9/site-packages/iris/_data_manager.py", line 204, in data
    result = as_concrete_data(self._lazy_array)
  File "/home/users/valeriu/anaconda3/envs/release220/lib/python3.9/site-packages/iris/_lazy_data.py", line 251, in as_concrete_data
    (data,) = _co_realise_lazy_arrays([data])
  File "/home/users/valeriu/anaconda3/envs/release220/lib/python3.9/site-packages/iris/_lazy_data.py", line 214, in _co_realise_lazy_arrays
    computed_arrays = da.compute(*arrays)
  File "/home/users/valeriu/anaconda3/envs/release220/lib/python3.9/site-packages/dask/base.py", line 564, in compute
    return repack([f(r, *a) for r, (f, a) in zip(results, postcomputes)])
  File "/home/users/valeriu/anaconda3/envs/release220/lib/python3.9/site-packages/dask/base.py", line 564, in <listcomp>
    return repack([f(r, *a) for r, (f, a) in zip(results, postcomputes)])
  File "/home/users/valeriu/anaconda3/envs/release220/lib/python3.9/site-packages/dask/array/core.py", line 1063, in finalize
    return concatenate3(results)
  File "/home/users/valeriu/anaconda3/envs/release220/lib/python3.9/site-packages/dask/array/core.py", line 4696, in concatenate3
    return _concatenate2(arrays, axes=list(range(x.ndim)))
  File "/home/users/valeriu/anaconda3/envs/release220/lib/python3.9/site-packages/dask/array/core.py", line 340, in _concatenate2
    return concatenate(arrays, axis=axes[0])
  File "/home/users/valeriu/anaconda3/envs/release220/lib/python3.9/site-packages/dask/array/ma.py", line 27, in _concatenate
    out = np.ma.concatenate(arrays, axis=axis)
  File "/home/users/valeriu/anaconda3/envs/release220/lib/python3.9/site-packages/numpy/ma/core.py", line 6964, in concatenate
    d = np.concatenate([getdata(a) for a in arrays], axis)
  File "<__array_function__ internals>", line 5, in concatenate
numpy.core._exceptions._ArrayMemoryError: Unable to allocate 2.74 GiB for an array with shape (9, 1260, 180, 360) and data type float32

  File "/home/users/valeriu/esmvalcore/esmvalcore/_main.py", line 433, in run
    fire.Fire(ESMValTool())
  File "/home/users/valeriu/anaconda3/envs/release220/lib/python3.9/site-packages/fire/core.py", line 141, in Fire
  File "/home/users/valeriu/anaconda3/envs/release220/lib/python3.9/site-packages/fire/core.py", line 466, in _Fire
  File "/home/users/valeriu/anaconda3/envs/release220/lib/python3.9/site-packages/fire/core.py", line 681, in _CallAndUpdateTrace
  File "/home/users/valeriu/esmvalcore/esmvalcore/_main.py", line 410, in run
  File "/home/users/valeriu/esmvalcore/esmvalcore/_main.py", line 104, in process_recipe
    recipe.run()
  File "/home/users/valeriu/esmvalcore/esmvalcore/_recipe.py", line 1366, in run
    self.tasks.run(max_parallel_tasks=self._cfg['max_parallel_tasks'])
  File "/home/users/valeriu/esmvalcore/esmvalcore/_task.py", line 672, in run
    self._run_sequential()
  File "/home/users/valeriu/esmvalcore/esmvalcore/_task.py", line 683, in _run_sequential
    task.run()
  File "/home/users/valeriu/esmvalcore/esmvalcore/_task.py", line 250, in run
  File "/home/users/valeriu/esmvalcore/esmvalcore/preprocessor/__init__.py", line 466, in _run
    self.products = _apply_multimodel(self.products, step,
  File "/home/users/valeriu/esmvalcore/esmvalcore/preprocessor/__init__.py", line 408, in _apply_multimodel
    result = preprocess(products - exclude, step, **settings)
  File "/home/users/valeriu/esmvalcore/esmvalcore/preprocessor/__init__.py", line 286, in preprocess
  File "/home/users/valeriu/esmvalcore/esmvalcore/preprocessor/__init__.py", line 272, in _run_preproc_function
  File "/home/users/valeriu/esmvalcore/esmvalcore/preprocessor/_multimodel.py", line 392, in multi_model_statistics
    return _multiproduct_statistics(
  File "/home/users/valeriu/esmvalcore/esmvalcore/preprocessor/_multimodel.py", line 300, in _multiproduct_statistics
    statistics_cubes = _multicube_statistics(cubes=cubes,
  File "/home/users/valeriu/esmvalcore/esmvalcore/preprocessor/_multimodel.py", line 284, in _multicube_statistics
    statistics_cubes[statistic] = _compute(big_cube, statistic)
  File "/home/users/valeriu/esmvalcore/esmvalcore/preprocessor/_multimodel.py", line 266, in _compute
    return cube.collapsed(dim, operator, **kwargs)
  File "/home/users/valeriu/anaconda3/envs/release220/lib/python3.9/site-packages/iris/cube.py", line 3946, in collapsed
    unrolled_data = np.transpose(self.data, dims).reshape(new_shape)
  File "/home/users/valeriu/anaconda3/envs/release220/lib/python3.9/site-packages/iris/cube.py", line 2100, in data
    return self._data_manager.data
  File "/home/users/valeriu/anaconda3/envs/release220/lib/python3.9/site-packages/iris/_data_manager.py", line 217, in data
    raise MemoryError(emsg.format(self.shape, self.dtype))
MemoryError: Failed to realise the lazy data as there was not enough memory available.
The data shape would have been (9, 1260, 180, 360) with dtype('float32').
 Consider freeing up variables or indexing the data before trying again.
2021-03-10 13:33:58,643 UTC [9017] INFO    
If you have a question or need help, please start a new discussion on https://github.com/ESMValGroup/ESMValTool/discussions
If you suspect this is a bug, please open an issue on https://github.com/ESMValGroup/ESMValTool/issues
To make it easier to find out what the problem is, please consider attaching the files run/recipe_*.yml and run/main_log_debug.txt from the output directory.
```

Test with this recipe:
```yaml
preprocessors:
  ppNOLEV1thr10:
    regrid:
      target_grid: 1x1
      scheme: linear
    multi_model_statistics:
      span: overlap
      statistics: [mean, median]

diagnostics:
  sm:
    description: Soil moisture
    variables:
      tas:
        preprocessor: ppNOLEV1thr10
        mip: Amon
        project: CMIP5
        exp: historical
        ensemble: r1i1p1
        start_year: 1900
        end_year: 2004
    additional_datasets:
      - {dataset: ACCESS1-0}
      - {dataset: ACCESS1-3}
      - {dataset: bcc-csm1-1}
      - {dataset: CanESM2}
      - {dataset: GFDL-ESM2G}
      - {dataset: HadGEM2-ES}
      - {dataset: inmcm4}
      - {dataset: IPSL-CM5A-LR}
      - {dataset: NorESM1-M}
    scripts: null
    ```

**Comment by valeriupredoi** [comment]
yeah I had somewhat of a recollection that `median` is not lazy - kinda hard to make it, given that one needs to count how many cows in a herd to get a median. But I am bit concerned with the other one, `mean` is lazy and I still see a 20-30% increase in memory intake - if at all, we should try make this module memory-efficient first, then faster, also I am not sure how the uptake scales with data sizes, I'll do more tests. But great work speeding it up though! :beer:

**Comment by stefsmeets** [comment]
Can you re-run it without median and use min or max or something? If median is not lazy, then the new implementation will not be more memory conserving that the current implementation in `master`. Afaik the `master` implementation attempts to be somewhat memory conserving, whereas this branch relies on the lazy implementation in iris.

**Comment by Peter9192** [comment]
Okay here's my suggestions to addressing V's concerns on memory and laziness:

1. To get lazy percentiles (and median which is just the 50 percentile): we should get https://github.com/SciTools/iris/pull/3901 across the finish line. We can at least test this PR and maybe @bjlittle can see if it is possible to prioritize this feature (and other lazy aggregators) in iris?
2. To address the memory concerns, we should come up with a strategy to control chunk sizes in the preprocessor. For the time being we could implement something along the lines of https://github.com/ESMValGroup/ESMValTool/blob/master/esmvaltool/diag_scripts/hydrology/compute_chunks.py.

**Comment by Peter9192** [comment]
@valeriupredoi I just added the option the rechunk the cube before computing the multimodel stats. This should keep the memory use within acceptable limits. With your example recipe I can compute the mean in 3:10 minutes and 4.8 GB peak memory on my work laptop.

Performance can further be tuned by controlling the block size for the chunks. By default it's set to 128MiB in Dask, but I made it flexible so you can play around with it if you like. Let me know what you find.

I also tested the lazy percentiles implementation from https://github.com/SciTools/iris/pull/3901. This seems to need a bit more work in order to make it just as efficient.

**Comment by Peter9192** [comment]
We can get 50x faster percentiles if we don't need masked arrays: https://github.com/SciTools/iris/pull/2687. That's a worthwhile optimization to keep in mind!

**Comment by valeriupredoi** [comment]
good progress, guys, I'll run a couple more tests (note that 3min10s is rather slow, but your PC processor might be a Pentium III for all I know :rofl: ). Will update wityh test results in a jiffy!

**Comment by valeriupredoi** [comment]
OK - excellent work rechunking! The `mean` in my test is now taking 4.8G of memory and running in sub-second times. I also checked the data - identical to the data from the old implementation :+1: I did notice new dims and attrs in the cube though:
```
(old implementation MM cube)
air_temperature / (K)               (time: 1260; latitude: 180; longitude: 360)
     Dimension coordinates:
          time                           x               -               -
          latitude                       -               x               -
          longitude                      -               -               x
     Attributes:
          Conventions: CF-1.7

(new MM cube)
air_temperature / (K)               (time: 1260; latitude: 180; longitude: 360)
     Dimension coordinates:
          time                           x               -               -
          latitude                       -               x               -
          longitude                      -               -               x
     Scalar coordinates:
          height: 1.75 m, bound=(1.5, 2.0) m
          new_dim: 4, bound=(0, 8)
     Attributes:
          Conventions: CF-1.7
          experiment_id: historical
          frequency: mon
          initialization_method: 1
          modeling_realm: atmos
          physics_version: 1
          project_id: CMIP5
          realization: 1
     Cell methods:
          mean: new_dim
```
Note that: `new_dim` which is the MM cubes index axis should not be a cell measure and the experiment is meaningless since it's taken from the template cube metadata - if we have different experiments cubes chucked in the MM, that'll not be accurate; not sure what the `physics_version` is - I believe that's a model-specific parameter, again, that's not needed since it may be inaccurate and introduce junk in provenance

**Comment by valeriupredoi** [comment]
unfortunately `median` is a bomb - on the machine I am running the memory reaches 14.4G at which point it gets killed flat out :bomb:

**Comment by Peter9192** [comment]
> unfortunately `median` is a bomb - on the machine I am running the memory reaches 14.4G at which point it gets killed flat out 💣

True, for that we'll need the lazy implementation in iris. See https://github.com/SciTools/iris/pull/3901

**Comment by Peter9192** [comment]
Thanks for testing @valeriupredoi , we can work on getting those attributes and dimensions right. Depending on the progress in iris, we can try to make a temporary workaround for median

**Comment by valeriupredoi** [comment]
OK guys first hints abouts scalability: I increased the data length (in time) with about 50% nad am noticing some worrying signs: this implementation uses about 7-7.2GB of memory (increase of 50% roughly) and the time it runs with the new data is 10s - that's about 10x slower than with 50% less data - so it seems that the memory scaling is linear but not the run time. Compare that with the old implementation where memory for new dataset is 5.9GB (23% increase) and time went up from 40s to 60s (so a 50% increase). I'll try chucking in more data but it'd be good if one of you could get a Mem(data) and Runtime(data) plot for a standardized machine so we know we're not going logarithmically up for any of these resources

**Comment by stefsmeets** [comment]
We have also noticed that how the data are chunked has a big effect on the outcome. The good news is that it is completely configurable through dask. Maybe we have not found the sweet spot yet.

Can you let me know which span and statistic you are using for these tests?

**Comment by valeriupredoi** [comment]
yeah the art of chunking is a tough nut, I've had to deal with that recently. Yeah, stat is mean and span is overlap :+1:

**Comment by Peter9192** [comment]
Hey @valeriupredoi I think it's important to understand how the lazy computation works here. The computation is mapped onto the chunks. According to https://docs.dask.org/en/latest/scheduling.html, dask's default setting for data-arrays is to use multithreading. If I understand it correctly, the memory intake is thus approximately the `memory needed to compute a single chunk` times the `number of chunks processed in parallel`. Like @stefsmeets said, this can be completely tuned. Make the chunks very small and set `dask.config.set(scheduler='synchronous')` and you should see that the memory intake is considerably lower (at the cost of longer computation times I would expect).

**Comment by valeriupredoi** [comment]
absolutely! I know these things are tunable and tweakable - the question to answer here is what is the one size fits all setting because I don't expect - in fact, I don't want - the user to set chunk sizes in the config user file. Maybe we can work out an equation to get that and set it in the run environment depending how much data is chucked in the analysis?

**Comment by Peter9192** [comment]
> Maybe we can work out an equation to get that and set it in the run environment depending how much data is chucked in the analysis?

Like the idea, but this sounds like something that warrants a wider discussion as we are in the process of making the entire preprocessor lazy. We should also consider the interplay between our own multiprocessing and dask's multithreading.... 

What I'm interested in right now is how to get this PR merged. I think we should accept that it'll not yet use the full potential of dask. So maybe we can just turn off daks's multithreading (set it to synchronous) for now and set a reasonable block size that should work on most systems?

**Comment by stefsmeets** [comment]
Hi @valeriupredoi , we just updated the 

1. Remove the temporary coord from the resulting cube. 
2. The other attributes you mentioned were generated by iris using the `equalise_attributes` function, and thus should correspond only to those attributes that are available in all input cubes.
3. When there is no lazy function available in iris, we resort to slicing over the time axis. This should limit memory usage in a way similar to your implementation. This is a _temporary_ solution, until iris 2.3 is available in October, which will have the lazy percentile / median.
4. We also removed the dask config, because we should not be setting global variables inside this function. This is best done elsewhere in the esmvaltool global config.

Could you have a look whether the memory is back to acceptable levels for your test recipe? 😅

**Comment by valeriupredoi** [comment]
cheers guys! Just having lunch now and will test after :hamburger:

**Comment by Peter9192** [comment]
> cheers guys! Just having lunch now and will test after 🍔

Bon appetit!

**Comment by valeriupredoi** [comment]
> > cheers guys! Just having lunch now and will test after hamburger
> 
> Bon appetit!

Merci beaucoup :fr: 

OK guys - first off - delighted with the MM performance: for the more stretchy data for my last tests (150 years, 4-5 odd models, overlap, mean) I have you penned at a whooping 200M of memory compared to 6GB  for the previous implementation and wrt time 10 odd seconds compared to 50 odd seconds (did multiple runs but this doesn't mean the numbers are 100% accurate since JASMIN/sci2 is a shared machine, but the numbers are good ballparks). So - well done! :beer: Unfortunately, there is a `but` here - what we gain in performance at the MM step run, we lose quite badly at the `save` step, compared to the old implementation: this branc does the MM file save in about 45 odd seconds and takes 6.5-6.8GB memory, whereas the old implementation is doing the save in <3s and is taking up <1GB memory. Overall the runs are 10-15% slower with this new implementation - but don't worry, this could be due to a lot of stuff happening before and after MM. The save step, if not directly linked to the now super tuned MM step, is a cause of how the data to be saved is packed (or unpacked, since it's all over the threads in chunks), so maybe we can ask the iris guys what is the best strategy to save? I would really really like to approve this now since it's improved the MM computation by miles, but am afraid we need to understand the save issue first. Terrific work so far nonetheless! :beer:

**Comment by valeriupredoi** [comment]
BTW I have not touched `median` again for fear of my life :rofl:

**Comment by Peter9192** [comment]
Hey @valeriupredoi the median should now work exactly like it did before. The point you raised about saving makes total sense. If the mm-stats preprocessor is lazy, it means that when that code is executed, it will only create a task graph, or a future if you will. The actual calculation happens whenever the data is realized. In this case, that is on save. So timing the execution of individual preprocessor functions once they're lazy is not very informative. 

I hope this helps to understand what's happening. Note that we've not tried to optimize for speed here. The main feature of this PR is that it re-implements the multi-model statistics using more iris functions. This enables lazy execution and more, but the real performance gains will come once we start tuning the dask settings.

**Comment by valeriupredoi** [comment]
@Peter9192 right you are, mate! But my question is this: if the overall MM process (lazy computation + realization + save) in this branch is not offering an improvement in performance now (slightly slower and slightly more memory intensive than the old implementation, not a buzz killer but the odd 15% worse for both), when do you think the optimal tuning will be available and also, how many extra user inputs will be needed so they get the most of their specific case via a specific setup (if at all)? This is like the case of active suspension in Formula 1 in the early 90s: normal suspension was working fine, they introduced the computer-assisted active suspension and the car's performance dropped while they were trying to set that up proper, but when set up proper that made a huge difference - we are now in the set up stage, so I am a bit uneasy allowing this to replace the old mechanical MM before we complete the optimal set up of the lazy implementation. I would accept that right now if I get a promise from you guys that you will be committed to working on figuring out the setup so we use the full potential of the new iris functionality and the lazy potential, but my fear is that this, once merged, we'll all go do something else and end up with a slightly less performant MM module, that is not used to its full potential. Sorry for the rant BTW :grin:

**Comment by stefsmeets** [comment]
> 
> 
> @Peter9192 right you are, mate! But my question is this: if the overall MM process (lazy computation + realization + save) in this branch is not offering an improvement in performance now (slightly slower and slightly more memory intensive than the old implementation, not a buzz killer but the odd 15% worse for both), when do you think the optimal tuning will be available and also, how many extra user inputs will be needed so they get the most of their specific case via a specific setup (if at all)? This is like the case of active suspension in Formula 1 in the early 90s: normal suspension was working fine, they introduced the computer-assisted active suspension and the car's performance dropped while they were trying to set that up proper, but when set up proper that made a huge difference - we are now in the set up stage, so I am a bit uneasy allowing this to replace the old mechanical MM before we complete the optimal set up of the lazy implementation. I would accept that right now if I get a promise from you guys that you will be committed to working on figuring out the setup so we use the full potential of the new iris functionality and the lazy potential, but my fear is that this, once merged, we'll all go do something else and end up with a slightly less performant MM module, that is not used to its full potential. Sorry for the rant BTW 😁

Your approach is fine as long as we accept being Williams and be awesome, but engineer everything ourselves. On the other hand, we could be Aston Marting/Racing point and borrow all the performant bits (gear-box, suspension, engine) from Mercedes, while still being awesome. This means hanging in the mid-field for a while, but we save a lot of effort in doing so that we can spend on doing something more useful. 

Until we make sure all our preprocessors are lazy, we will not see massive performance gains. On the other hand, now that this PR fully relies on iris/dask, any performance improvements there we will benefit from without any effort on our side. 

In my opinion, it is a worthwhile effort to chart out which preprocessors are lazy, and which ones are not. And commit to making the whole chain lazy. The alternative is keeping the current implementation around and make the new preprocessor like a v2 or something (not ideal imo).

**Comment by Peter9192** [comment]
👍 for the racecar comparisons guys! I think I agree with both of you. Let me see if I can get tweak the performance of the PR a bit more, without going into the rabbit hole of tuning dask.

**Comment by Peter9192** [comment]
Okay so what I've done is this: there is now a much stricter separation between the lazy and the eager evaluation paths. Only statistics for which a lazy iris operator is available follow the lazy path. The eager path is equivalent to the current master, except that it operates directly on iris cubes instead of numpy arrays. IMHO, this is safer and also more readable, because we don't have to worry about indexing. 

I've done some profiling on this as well (using the recipe somewhere above). Total runtimes were quite capricious, so I'm not sure if we can conclude anything else than "more or less the same". 

Branch

| regrid | time | statistic | total run time | max memory |
| --- | --- | --- | --- | --- |
| 10*10 | 2000-2004 | mean (lazy) | 0:24 | 0.3 GB |
| 1*1 | 2000-2004 | mean (lazy) | 0:28 | 0.7 GB |
| 1*1 | 1900-2004 | mean (lazy) | 1:12 / 2:38  / 1:33 | 4.9 GB / 5.0 GB /  6.1 GB |
| 10*10 | 2000-2004 | median (not lazy) | 0:30 | 0.3 GB |
| 1*1 | 2000-2004 | median (not lazy) | 0:29 | 0.6 GB |
| 1*1 | 1900-2004 | median (not lazy) | 3:28 | 4.5 GB |

Master

| regrid | time | statistic | total run time | max memory |
| --- | --- | --- | --- | --- |
| 10*10 | 2000-2004 | mean | 0:23 | 0.3 GB |
| 1*1 | 2000-2004 | mean | 0:24 | 0.6 GB |
| 1*1 | 1900-2004 | mean | 1:07 | 4.3 GB |
| 10*10 | 2000-2004 | median | 0:25 | 0.3 GB |
| 1*1 | 2000-2004 | median | 1:01 / 2:20 | 0.6 GB / 0.5 GB |
| 1*1 | 1900-2004 | median | 4:25 / 3:01 / 3:12 | 4.5 GB / 4.6 GB / 4.4 GB |

Both implementations lead to memory issues as I'm pushing the dataset size. The current master just has a hard limit, whereas the lazy implementation should accomodate bigger datasets, but sometimes bumps into a slow situation that seems similar to what's described here: https://github.com/dask/dask/issues/2866#issuecomment-447154895. 

@stefsmeets would you be able to go over the changes I've made to check for technical issues, and maybe see if you spot anything that can still be optimized?

**Comment by stefsmeets** [comment]
Nice work @Peter9192 , looks good to me. Just fixed the tests 👍

**Comment by valeriupredoi** [comment]
haven't forgotten about this guys, sorry for slower response times, I will re-run the tests asap (by that I mean after Easter, JASMIN is being flaky8 again and I need som e vacation time over Cheesus' day of issues with the Roman authorities ;grin: )

**Comment by bouweandela** [comment]
It would be interesting to see the improvement in memory use for a computation that is actually using a lot of memory, e.g. a case where multimodel statistics as it is currently implemented in `master` uses 50GB or a 100 GB of RAM.

**Comment by valeriupredoi** [comment]
I'll have a test run :+1:

**Comment by bouweandela** [comment]
Maybe try recipe perfmetrics?

**Comment by bouweandela** [comment]
I tried this recipe for testing:
```yaml
preprocessors:
  preproc:
    regrid:
      target_grid: 0.5x0.5
      scheme: linear
    multi_model_statistics:
      span: overlap
      statistics: [mean]

diagnostics:
  tas:
    variables:
      tas:
        preprocessor: preproc
        mip: Amon
        project: CMIP5
        exp: historical
        ensemble: r1i1p1
        start_year: 1850
        end_year: 2005
    additional_datasets:
      - {dataset: ACCESS1-0}
      - {dataset: ACCESS1-3}
      - {dataset: bcc-csm1-1}
      - {dataset: bcc-csm1-1-m}
      - {dataset: BNU-ESM}
      - {dataset: CanESM2}
      - {dataset: CCSM4}
      - {dataset: CESM1-BGC}
      - {dataset: CESM1-CAM5}
      - {dataset: CESM1-FASTCHEM}
      - {dataset: CESM1-WACCM}
      - {dataset: CSIRO-Mk3-6-0}
      - {dataset: FIO-ESM}
      - {dataset: inmcm4}
      - {dataset: IPSL-CM5A-LR}
      - {dataset: IPSL-CM5A-MR}
      - {dataset: IPSL-CM5B-LR}
      - {dataset: MPI-ESM-LR}
      - {dataset: MPI-ESM-MR}
      - {dataset: MPI-ESM-P}
      - {dataset: NorESM1-M}
      - {dataset: NorESM1-ME}
    scripts: null
```
and the results were not quite what I had hoped for. With the current master branch runtime was 10 minutes and maximum 50 GB or RAM, while with this branch runtime was 27 minutes and 61 GB of RAM.

The increase in runtime is not unexpected, because I would expect the processing to be done twice, once for the multimodel statistics computation and once for the individual models, but there has been an increase in memory use as well..

Peter9192:
> The increase in runtime is not unexpected, because I would expect the processing to be done twice, once for the multimodel statistics computation and once for the individual models, but there has been an increase in memory use as well..

@bouweandela thanks for testing! Now you see why I was hesitant about this..

During my tests I had the impression that dask fills up the memory by trying to do many operations in parallel as long as there is enough memory available. I expect you do see a reduction in memory when you set `dask.config.set(scheduler='synchronous')`. Could that explain the observed behaviour?

zklaus:
So where do we go with this? I see three options:

- Get it in 2.3.0
For this we need to decide rather soon what should be the outcome and how to get there.
- Bump it to 2.4.0
This would give us some more time to figure things out. Since it doesn't look critical at the moment, this might be reasonable.
- Scrap it
If we don't manage to define a positive outcome, we might want to abandon this.

Peter9192:
If we want to get it in for 2.3 or 2.4, I see two options

1. Fix the performance issue. Not sure how though; maybe @bouweandela wants to give it a try?
2. Keep the refactoring, but not the lazy stuff. Since we've already implemented a fallback for the functions like median that are not lazy yet in iris, we could use that for all operators right now.  This would be quite straightforward, as there are now two paths through the code: `compute_eager` and `compute_lazy`. We could redirect everything to `compute_eager` and keep or remove the code related to the lazy pathway.

zklaus:
What performance problem? The memory use? Of course, dask should use all the available memory. And start as many threads as there are (hyperthread) cores. If that is not what we want, then that is a bug in our dask setup, not with the multi-model stuff, no?

bouweandela:
> What performance problem? The memory use?

The goal of this pull request was to make it possible to process datasets that currently require more RAM than there is available in the machine. Some first tests I did, did not look promising though: a slowdown and larger memory use for existing cases and a complete hang instead of a crash for larger datasets. But I didn't have time to have a detailed look. There are still 2 weeks to go until the feature freeze, right? I agree with @Peter9192 that if we cannot improve the performance in time, we probably still want to merge the cleaner code for the non-lazy version.

zklaus:
@Peter9192, would it be too much work to split this up into a cleanup PR that we can merge now and a lazy PR that we can bump to 2.4.0 if needed?

Peter9192:
I'll try to find some time soon.

bouweandela:
@Peter9192 Could you please merge `main` into this branch so the diff is up to date?

Peter9192:
Hmm that was not intended, wait a sec...

Peter9192:
> 
> 
> @Peter9192 Could you please merge `main` into this branch so the diff is up to date?

Done. All conflicts resolved, but haven't really tested it yet.

bouweandela:
Some new performance test results:

On my laptop (2 physical/4 virtual CPU cores, 16 GB RAM) I was able to run the example recipe https://github.com/ESMValGroup/ESMValCore/pull/968#issuecomment-833601231 with
- target grid 1x1: by setting the default dask array chunk-size to 32 MiB, runtime ? minutes, memory use 6 GB
- target grid 0.5x0.5: by setting the default dask array chunk-size to 8 MiB, runtime 20 minutes, memory use 8.5 GB

On a cluster node (16 physical CPU cores, 64 GB RAM):
- target grid 1x1: runtime 4:40 minute, memory use 30 GB
- target grid 0.5x0.5: runtime ? minutes, memory use ? -> Ran out of memory

With the current main branch, I am unable to run the recipe on my laptop. On the cluster node:
- target grid 1x1: runtime 9:15 minute, memory use 14 GB
- target grid 0.5x0.5: runtime 19 minutes, memory use 53 GB (two times slower than previous result, but I cannot remember if it was on the same cluster/node)

The input data size after regridding to 0.5x 0.5 degree is 22 models x 155 years x 12 months x 360 latitudes x 720 longitudes x 4 byte float = 40 GB, after regridding to 1x1 it is 10 GB. Data size before regridding is 3.8 GB.

What I do not understand is why I need to [reduce the default chunk size](https://docs.dask.org/en/latest/configuration.html) to be able to run the recipe? Without reducing it the recipe runs out of memory. If anyone has ideas?

Another question: is regridding to a (considerably) higher resolution a realistic use case?

zklaus:
Regridding to a higher resolution is realistic. This happens for example when you want to compare a global model at a local scale. It is important to think about the right interpolation method in these cases.

bouweandela:
> This happens for example when you want to compare a global model at a local scale.

Would that mean regridding the full globe, or just some local area?

zklaus:
Usually to a local area, I think. Though people are also interested to compare the same model at two different resolution (in some sense, this is what PRIMAVERA was all about).

bouweandela:
This will probably have to wait for v2.5, unless we find an easy solution. We have requested extra funding in the IS-ENES3 project to look into this and related problems with high memory use by dask.

schlunma:
> This will probably have to wait for v2.5, unless we find an easy solution. We have requested extra funding in the IS-ENES3 project to look into this and related problems with high memory use by dask.

@bouweandela any change this will be ready for v2.5?

bouweandela:
No. So far, we have not been able to find additional funding to work on this, unfortunately.

rcomer:
https://github.com/SciTools/iris/pull/3901 has now been merged, so the lazy percentile aggregation will be available at Iris v3.3, which is due out in the summer.

valeriupredoi:
many thanks for the heads up @rcomer

sloosvel:
> No. So far, we have not been able to find additional funding to work on this, unfortunately.

Should this PR be removed from the 2.6 milestone then?

zklaus:
[Check EVT PRs](https://trello.com/c/0cYTuIRY/5-check-evt-prs)

bouweandela:
Example recipe for testing:
```yaml
---
documentation:
  description: Test recipe for multimodel statistics
  authors:
    - andela_bouwe

  references:
    - gleckler08jgr

  projects:
    - c3s-magic

preprocessors:

  preproc:
    regrid:
      target_grid: 0.5x0.5
      scheme: linear
    multi_model_statistics:
      span: overlap
      statistics: [mean]
      keep_input_datasets: true

diagnostics:

  tas:
    variables:
      tas:
        preprocessor: preproc
        product: output1
        mip: Amon
        project: CMIP5
        exp: historical
        ensemble: r1i1p1
        start_year: 1850
        end_year: 2005
    additional_datasets:
      - {dataset: ACCESS1-0}
      - {dataset: ACCESS1-3}
      - {dataset: bcc-csm1-1}
      - {dataset: bcc-csm1-1-m}
      - {dataset: BNU-ESM}
      - {dataset: CanESM2}
      - {dataset: CCSM4}
      - {dataset: CESM1-BGC}
      - {dataset: CESM1-CAM5}
      - {dataset: CESM1-FASTCHEM}
      - {dataset: CESM1-WACCM}
      - {dataset: CSIRO-Mk3-6-0}
      - {dataset: FIO-ESM}
      - {dataset: inmcm4}
      - {dataset: IPSL-CM5A-LR}
      - {dataset: IPSL-CM5A-MR}
      - {dataset: IPSL-CM5B-LR}
      - {dataset: MPI-ESM-LR}
      - {dataset: MPI-ESM-MR}
      - {dataset: MPI-ESM-P}
      - {dataset: NorESM1-M}
      - {dataset: NorESM1-ME}
      - {project: CMIP6, dataset: TaiESM1, ensemble: r1i1p1f1, grid: gn}
    scripts: null
```

bouweandela:
@fnattino and I have some time to work on this now, but it seems unlikely that it will be ready in time for v2.7.

bouweandela:
Tests by @fnattino appear to show that this works quite well when used with #1714.

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -578,12 +601,12 @@ def _multicube_statistics(cubes, statistics, span, ignore_scalar_coords=False):
         )
 
     # Avoid modifying inputs
-    copied_cubes = [cube.copy() for cube in cubes]
+    cubes = [cube.copy() for cube in cubes]
```
valeriupredoi, line 604: I don't understand the purpose of this - if you are dereferencing the initial `cubes` then you lost them, and only the garbage collector knows about them now

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -609,13 +632,14 @@ def _multicube_statistics(cubes, statistics, span, ignore_scalar_coords=False):
 
     # Calculate statistics
     statistics_cubes = {}
+    lazy_input = any(cube.has_lazy_data() for cube in cubes)
```
valeriupredoi, line 635: shouldn't be `all()`? Or you accept non-lazy cubes too

esmvalcore/preprocessor/_regrid.py
```diff
@@ -650,11 +651,38 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
...
+        # because rechunking will keep the original chunk in memory.
+        return cube
+
+    data = cube.lazy_data()
+
+    # Compute a good chunk size for the target array
+    tgt_shape = data.shape[:-2] + target_grid.shape
+    tgt_chunks = data.chunks[:-2] + target_grid.shape
+    tgt_data = da.empty(tgt_shape, dtype=data.dtype, chunks=tgt_chunks)
+    tgt_data = tgt_data.rechunk({i: "auto" for i in range(cube.ndim - 2)})
```
valeriupredoi, line 677: dask rechunking is a bit slow and bulky, but I don't know actual performance parameters for it, unfortunately - are we sure this adds a benefit and doesn't just make things slower?

valeriupredoi (approved):
many thanks to all who contributed to this - it looks great and I bet it's also adding quite a bit of performance improvement (I've not tested though); I am approving, but I have a few, minor questions. Cheers :beer:

esmvalcore/preprocessor/_multimodel.py
bouweandela: The purpose of renaming `copied_cubes` to `cubes` is to make the code easier to understand. Using fewer variable names reduces the cognitive load: instead of remembering what `cubes`, `copied_cubes`, and `aligned_cubes` are, the reader now only needs to remember that `cubes` is the input data to the statistics computation.

esmvalcore/preprocessor/_multimodel.py
bouweandela: Yes, if one or a few of the cubes is not lazy (due to fixes or temporal interpolation), it's probably still more efficient not to realize all cubes. Especially if you're doing a statistics computation over e.g. all of CMIP6. If this turns out to be a problem when testing the ESMValTool recipes for the upcoming v2.9 release, we can always change it.

esmvalcore/preprocessor/_regrid.py
bouweandela: Yes, it reduces memory consumption: if you're computing multimodel statistics you will load the n_datasets x chunksize into memory of a worker, which may be much already. If you combine that with regridding to a higher resolution grid, it really blows up memory use. Maybe we should do something similar for the multimodel statistics too. If I remember correctly, we even had code to do that in this pull request at some point.

bouweandela: Note that it's only rechunking if the regridded chunks are at least twice as big as the input chunks (see also explanation on line 667).

esmvalcore/preprocessor/_multimodel.py
valeriupredoi: OK but then why copy them at all? Destroying the original `cubes` and copying them to a new address is a bit of an excercise in futility IMO but maybe am missing something - at any rate, don't let this hold the merge :+1:

esmvalcore/preprocessor/_multimodel.py
valeriupredoi: sounds good and a good call too! Maybe add a #TODO there? Not too fussy about it tho

esmvalcore/preprocessor/_regrid.py
valeriupredoi: yes, that is a very good point that I completely missed, cheers, bud! :beer:

esmvalcore/preprocessor/_multimodel.py
bouweandela: Let's create an issue when it actually turns out to be an issue ;-)

esmvalcore/preprocessor/_multimodel.py
valeriupredoi: indeed - it's just a pointer in a dict :grin:

PR #979, opened 2021-02-02
Set remove_preproc_dir to false in default config-user

valeriupredoi:
The `remove_preproc_dir` needs to be set to `false` by default because 80% users actually need the preprocessor files and may contribute to confusion amongst uninitiated users


schlunma (approved):
Yes, fully agree! It would be nice to include this in v2.2!

valeriupredoi:
cheers Manu! :beer:

Issue #972, opened 2021-02-01
Code freeze for v2.2.0


jvegasbsc:
Today we freeze the code for the 2.2.0 release. I ask for your help to review and merge as many pending issues in the milestone before today at 22:00 CES. Thanks!

@ESMValGroup/scientific-lead-development-team 
@ESMValGroup/technical-lead-development-team

valeriupredoi:
@jvegasbsc I see you are more lenient than @bouweandela and myself for last release, we decided to freeze it in the afternoon :grin: For the freeze, you know the drill - create a branch called something like `release_2.2.0` and ask @mattiarighi to set nobody-move-a-finger permissions on it ie nobody pushes to it; we should also ask the people to try run as many recipes as possible with the release branch; also, we should get on a zoom call next week when we do the release :+1:

jvegasbsc:
Branch `release_2.2.0` created. @remi-kazeroni can you please protect it?

remi-kazeroni:
> Branch `release_2.2.0` created. @remi-kazeroni can you please protect it?

I don't think I have the right to do that. I can't access the settings of the repo. My guess is that only @axel-lauer can do that at the moment.

axel-lauer:
Done!

valeriupredoi:
cheers @axel-lauer - for the sake of redundancy it'd be very good if you allowed @remi-kazeroni and at least another core member privileges for these types of operations (of your choice, could be @bouweandela , myself or @jvegasbsc ) :+1:

axel-lauer:
@valeriupredoi good point. I propose to give @remi-kazeroni as "successor" of @mattiarighi admin rights right now but to leave it up to the new steering committee to decide on how admin rights are handled in the future.

jvegasbsc:
Please, @ESMValGroup/esmvaltool-developmentteam go ahead and test the branch `release_2.2.0` with your favourite recipes so we can get proper testing in the new releases. Please, bear in min that you may need ESMValGroup/ESMValTool#1874 to make them work with Iris 3

valeriupredoi:
> @valeriupredoi good point. I propose to give @remi-kazeroni as "successor" of @mattiarighi admin rights right now but to leave it up to the new steering committee to decide on how admin rights are handled in the future.

cheers @axel-lauer - it all sounds very reasonable to me :beer:

valeriupredoi:
[...]

PR #987, opened 2021-02-05
Move fiona to be installed from conda forge

valeriupredoi:
`fiona` is again giving us headaches, the reason why I am moving it to be a `conda-forge` dependency instead of PyPi is in this [comment](https://github.com/ESMValGroup/ESMValCore/issues/972#issuecomment-774111132) as a result of @remi-kazeroni noticing fiona trying to read an old csv file from GDAL that has disappeared from GDAL>=3.1 and asking us about it in this [comment](https://github.com/ESMValGroup/ESMValCore/issues/972#issuecomment-773321942), also an issue documented here https://github.com/Toblerity/Fiona/issues/897 (towards the bottom)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,9 +32,6 @@ REQUIREMENTS = {
     'install': [
         'cf-units',
         'dask[array]',
-        # fiona: 1.8.18/py39, they seem weary to build manylinux wheels
-        # so we may have to install from conda-forge in the future
-        'fiona',
         'fire',
         'nc-time-axis',  # needed by iris.plot
         'netCDF4',
--- a/environment.yml
+++ b/environment.yml
@@ -6,6 +6,7 @@ channels:
 dependencies:
   - cftime  # iris=3.0.1 needs <=1.2.1; >=1.3.0 - years<999 get a 0 instead of empty space
   - compilers
+  - fiona  # 1.8.18/py39, they seem weary to build manylinux wheels and pypi ver built with older gdal
   - esmpy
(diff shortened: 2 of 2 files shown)


remi-kazeroni:
I have just checked and this PR fixed this issue mentioned above. Thanks for looking into this @valeriupredoi 👍

valeriupredoi:
brilliant, cheers Remi! can you pls approve so we get it merged and cherrypicked in the release? :beer:

remi-kazeroni (approved):
Thanks @valeriupredoi 👍

setup.py
```diff
@@ -32,9 +32,6 @@
     'install': [
         'cf-units',
         'dask[array]',
-        # fiona: 1.8.18/py39, they seem weary to build manylinux wheels
-        # so we may have to install from conda-forge in the future
-        'fiona',
```
bouweandela, line 37: The list in setup.py lists all dependencies, regardless of how they're installed. Could you put fiona back her @valeriupredoi? If it's installed by conda already, pip won't try to install it.

valeriupredoi: aha will do now!

valeriupredoi: https://github.com/ESMValGroup/ESMValCore/pull/990 pls approve so we merge and release the fiona baby

bouweandela: Thanks!

valeriupredoi: cheers for pointing out and reviewing, bud!

Issue #134, opened 2019-06-28
Feature request - ESMValTool html reports


ledm:
For the past two years, we've been using [BGC-val](https://www.geosci-model-dev.net/11/4215/2018/gmd-11-4215-2018.html) to evaluate the marine component of the spin up and CMIP6 runs of the UK Earth System Model (UKESM). One of the main selling points of BGC-val, was that it can summarise the results of its metrics in a simple, clear, shareable and mobile-friendly html website. Here's an older example report that we showed at [EGU2017](http://gws-access.ceda.ac.uk/public/esmeval/EGU2017/).

The report maker tool was crucial for the uptake in use of the tool and it allowed several users to share monitoring and evaluating the model. Similarly, the [ILAMB ](https://www.ilamb.org/) interface is very popular among land surface modellers for this reason, among others. I'd like to implement a similar feature for ESMValTool. I've discussed this with @valeriupredoi a few times, and he recommended starting a discussion issue on github.

This is how I envisage the interface with ESMValTool:
- A `MakeReport` Boolean will need to be added to the global config.
- The html will be added to a new directory, `report`, alongside `plots`, `preproc`, `run` and `logs`. This new directory will contain an `index.html` file, an `images` folder and an `assets` directory. The `index.html` file will contain the bulk of the report. The `report` directory would be fully portable, and would contain everything needed to show the report.
- The report maker would have to run run after the diagnostics have completed, and search recursively through the plots folder of the output directory, adding every figure to the `index.html` page. 
- Each image would be shown alongside some metadata in a table. The metadata would include information like original dataset, scenario, ensemble member, preprocessors, diagnostic scripts, etc.... In some cases, metadata can be added directly into the image file metadata. However, this might not be compatible with all supported image formats. In either case, we'd need to make a wrapper for `pyplot.savefig`/`iris.savefig` which adds metadata into the image.  
- The report maker would likely not be compatible with tools that already do this, like autoassess, but would instead function as a "native" ESMValTool version of such tools. 

I'm happy to reuse the code that we had in BGC-val. It would need a bit of tweaking, but wouldn't be a major issue. The HTML was initially taken from [html5up.net](https://html5up.net/) and was under a creative commons licence. This was a html5 ready, mobile friendly, scalable website template. However, I'm not 100% confident that this would be the wisest option. It may be more sensible to built a new report template from scratch. Do we have any html experience in ESMValTool?

[...]

Issue #133, opened 2018-06-11
Output pdf report with all plots and metadata for recipe


nielsdrost:
For the MAGIC project we would like to offer users (of our web portal) a way to download information on a namelist. For this we would like ESMValTool to be able to create a multi-page report for a namelist, with all plots, references, perhaps provenance info, etc, for the namelist run.

bouweandela:
A html report is generated since #991 was merged.

Issue #826, opened 2020-10-12
`DIAGNOSTICS_PATH` silently defaults to current directory if esmvaltool cannot be imported


stefsmeets:
**Describe the bug**
I just had a problem with my esmvaltool installation, and I noticed that `DIAGNOSTICS_PATH` simply defaults to the CWD without any warning. In my opinion, it should either crash hard or make it very clear where it is looking for recipes. The current behaviour is not transparent, and it took me a while to figure it out.

bouweandela:
Maybe a logger.info message displaying the diagnostics path would make this more obvious?

It would be nice if diagnostics_path was configurable in config user, with as default value the location of the esmvaltool installation or if that is not present, maybe the current working directory, but I'm not sure if that is a very sensible backup value.

stefsmeets:
Agreed, that could be useful. I was thinking it would also make sense to configure it as an environment variable.

What do you think about further decoupling esmvaltool and esmvalcore? At the moment, esmvalcore presents itself as esmvaltool (just look at the entry point and `esmvalcore._main.py`) and installs the command line utility. This feels a bit the wrong way around. Right now esmvalcore depends on esmvaltool to ask where its diagnostics are located. When I think about it, esmvaltool should import esmvalcore and tell it where the diagnostics are located (easy to do if it is configurable). Then, you could have esmvalcore present a simple API to run recipes. Something to consider for the API.

bouweandela:
Indeed the `esmvaltool` command is provided by the ESMValCore package. We kept the name because it's familiar, though if you have an idea for a better name, we could provide that as an alternative I guess and then move the `esmvaltool` command to the ESMValTool package at some point.

It is possible to use the ESMValCore package and the `esmvaltool` command without installing the ESMValTool package and this works just fine, but it is not well documented what files and directories are expected if you want things to be fully working. This is something that definitely could use some improvements.

stefsmeets:
Yeah, I was thinking too that the esmvaltool command line program logically belongs to esmvaltool. We could have a trimmed down version for esmvalcore that just runs recipes and handle configs (pretty much what it currently is). We could import the base command line in ESMValTool and customize it with more esmvaltool related commands if necessary. I think this would be good to have better isolation between the two packages. ESMValTool is already installed as a python package anyway, so it should not be too difficult.

Issue #28, opened 2019-05-15
We shouldn't require authors to be in config-references.yml to run a recipe


ledm:
In the documentation section of recipes, there is an obligatory field: `authors`. This requires the names listed in the `authors` section of the recipe to be in the `config-references.yml` file.

Otherwise, you will get an error like this one:
```
ValueError: Tag 'lee' does not exist in section 'authors' of /users/modellers/ledm/workspace/ESMValToolTest/ESMValTool/esmvaltool/config-references.yml

```

This should 100% be a requirement for pushing recipes back to github.  

However, this excludes new users from ownership of their own recipe if they don't have access to the source code. For instance they are running a shared installation or ESMValTool, or they are using the `user` installation instead of the `develop` installation.

Can we relax this requirement when running a recipe, but keep it for submitting recipes? 

@valeriupredoi and I spoke about it earlier today. Any thoughts @mattiarighi, @bouweandela ?

bouweandela:
I took this into account when splitting the packages. @ledm After https://github.com/ESMValGroup/ESMValTool/pull/1148 is merged, can you please test if things now work as you would like?

bouweandela:
This has now been implemented in the code, but it should still be documented.

Note that when the config-references.yml file is not available, no information from that file can be retrieved, so authors should write their full name and references etc. in the recipe and diagnostic scripts if they still want reasonable looking provenance output.

stefsmeets:
Hi @bouweandela , has this ever been resolved? As we are expanding the API, it is getting increasingly difficult for me to test recipe / metadata rendering since this depends on the ability of author names (tags) being resolved. Part of this is the forward dependency on `esmvaltool` to define author names, and the other is that there is no 'default' user to fall back to or to test with.

bouweandela:
No, this has not been completely solved, or the issue would have been closed. If you try to run a recipe with an author tag that is not available in the config-references.yml file (author `some_author` in this case), the run will stop with this error message:
```
ValueError: Tag 'some_author' does not exist in section 'authors' of /home/bandela/src/esmvalgroup/esmvaltool/esmvaltool/config-references.yml
```

The problem is that a collection of recipes, diagnostic scripts, bibtex references, and config-references.yml files needs to match each other. However, it is possible to use ESMValCore to just run a recipe and associated diagnostic script without the bibtex references and config-references.yml file, but obviously, you cannot replace tags with a value if you do not have the contents to replace them with, so this information will be missing from the output.

[...]

PR #991, opened 2021-02-08
Generate HTML report with recipe output

stefsmeets:
## Description

Hi everyone, this PR lays the foundation for generating html reports from esmvaltool output. I'm using `jinja2` as the rendering engine. 

This PR will enable rendering recipe output (images + data using relative paths) using the API in a jupyter notebook. It will also automatically create and place a html report (`index.html`) at the root of the output directory on each run through the CLI.

The best way to test it is to run:
`esmvaltool run examples/recipe_python.yml`

-   Addresses #133, #134
-   Link to documentation: https://esmvaltool--991.org.readthedocs.build/projects/ESMValCore/en/991/quickstart/output.html#output

***

## Before you get started

## Checklist

***


valeriupredoi:
just wanted to say - dis very very cool! very well done starting this @stefsmeets :beer:

valeriupredoi:
one thing I would like to suggest is the additional copying of the results to a `public_html`-type of directory if the tool runs on a HPC site (in CLI), I will set one up on CEDA-JASMIN and would encourage @remi-kazeroni do the same on DKRZ and others that can do it at other clusters :+1:

stefsmeets:
Hi @Peter9192 @nielsdrost I think this is ready to review! Let me know what you think.

For Codacy:
- The `import outside toplevel` is necessary to avoid a circular import, I thought this was the cleanest way to do it.
- `Attribute 'session_name' defined outside __init__` is in a method to convert from `config-user` to `Session`, i.e. between the API and current core. I did not want to update the init function to keep the interface clean within the API.

Both will eventually go away if we include integrate more of the API into the core.

Peter9192:
> 
> 
> one thing I would like to suggest is the additional copying of the results to a `public_html`-type of directory if the tool runs on a HPC site (in CLI), I will set one up on CEDA-JASMIN and would encourage @remi-kazeroni do the same on DKRZ and others that can do it at other clusters 👍

I think this is indeed what we had in mind, right @nielsdrost ? Similar to what we do with the output of the bot. It'd be great if you could do that for JASMIN as well!

esmvalcore/experimental/templates/RecipeInfo.j2
```diff
@@ -0,0 +1,37 @@
+<h1>{{ info.name }}</h1>
```
Peter9192, line 1: Would it make sense for these templates to be in a folder called 'output' or something like that?

esmvalcore/experimental/templates/TaskOutput.j2
```diff
@@ -0,0 +1,23 @@
+<h2>{{ task.name }}</h2>
```
Peter9192, line 1: Would it make sense to wrap these tasks in <divs> so we can easily style them? This would make it very easy for example to make the these divs collapsible. https://www.w3schools.com/howto/howto_js_collapsible.asp

tests/sample_data/experimental/test_run_recipe.py
```diff
@@ -18,6 +20,26 @@
...
+    reason='The behaviour of these tests depends on what ``DIAGNOSTICS.path``'
+    'points to. This is defined by a forward-reference to ESMValTool, which'
+    'is not installed in the CI, but likely to be available in a developer'
+    'or user installation.')
+
+TAGS = TagsManager({
+    'authors': {
+        'doe_john': {
+            'name': 'Doe, John',
+            'institute': 'Testing Institute',
```
Peter9192: Haha!

tests/unit/experimental/references/doe2021.bibtex
```diff
@@ -0,0 +1,6 @@
+@misc{doe2021,
+    doi = {0},
+    year = {2021},
+    author = {J. Doe},
+    title = {Test Free or Fail Hard},
```
Peter9192, line 5: xD

stefsmeets:
I had to do a bit of refactoring in the tests to avoid this issue: https://github.com/ESMValGroup/ESMValCore/issues/28#issuecomment-778170951
The problem is that `authors` is a required keyword, but if `esmvaltool` is not installed, the tags default to `{}` (#826). This means the recipe runner will crash when it is asked to write an html file, because it will need the author tags to be resolved. Since the tests did not encounter this situation, it was fine until now.

I think it is sorted now and tests should be working again 👍

Peter9192 (changes requested):
Hey @stefsmeets very neat work!! This is amazing. One suggestion would be to also include the original recipe file with which this output was created (just like you add the logs). 

Also I noticed that this PR breaks the nice printing of the recipe info in the notebook...

And there seems to be an ugly black border around my image, but maybe that's because I was putting it to the test with a different recipe than the example. I also tried to include a vegalite image (like we talked about this week), and that doesn't seem to work yet... But that would be for a follow up PR I suppose

stefsmeets:
> Hey @stefsmeets very neat work!! This is amazing. One suggestion would be to also include the original recipe file with which this output was created (just like you add the logs).

Yeah, good idea!

> Also I noticed that this PR breaks the nice printing of the recipe info in the notebook...

That's not meant to happen... I cannot reproduce it, can you show me what happens?

> And there seems to be an ugly black border around my image, but maybe that's because I was putting it to the test with a different recipe than the example. I also tried to include a vegalite image (like we talked about this week), and that doesn't seem to work yet... But that would be for a follow up PR I suppose

I added a black border so I can see where the image ends. You are more than welcome to make a follow-up PR to pimp the output style 😎

nielsdrost:
> > one thing I would like to suggest is the additional copying of the results to a `public_html`-type of directory if the tool runs on a HPC site (in CLI), I will set one up on CEDA-JASMIN and would encourage @remi-kazeroni do the same on DKRZ and others that can do it at other clusters +1
> 
> I think this is indeed what we had in mind, right @nielsdrost ? Similar to what we do with the output of the bot. It'd be great if you could do that for JASMIN as well!

Yep, this is only step one of our cunning plan! Next we plan to:
- Dust of the old cylc suite for running all recipes automatically once in a while
- Create a overall page that can show the output of multiple recipes (as some sort of gallery)
- Make this html have some styling
- Host the result on the esmvalbot machine

But I like your idea of serving up user's recipe output too (a bit like the bot does).

stefsmeets:
@esmvalbot please run examples/example_python.yml

esmvalcore/experimental/templates/TaskOutput.j2
stefsmeets: Yeah, that would be useful! Let's talk about styling in a follow-up PR.

esmvalcore/experimental/templates/RecipeInfo.j2
stefsmeets: I think 'templates' is a suitable name. Jinja2 works with 'environments', so once an environment path is defined, you can retrieve templates everywhere.

Peter9192: Yeah I agree templates makes sense, they're often used in web projects I think. But since ESMValTool is not really a web project I'm not sure if it is immediately obvious what kind of templates they are in this case. That's why I thought it might make sense to make this a bit more explicit.

esmvalcore/_main.py
```diff
@@ -94,14 +94,16 @@ def process_recipe(recipe_file, config_user):
             "NetCDF compression.")
 
     # copy recipe to run_dir for future reference
-    shutil.copy2(recipe_file, config_user['run_dir'])
+    shutil.copy2(recipe_file, os.path.join(config_user['run_dir'],
```
Peter9192, line 97: Not sure if renaming the recipe will set people off. I think you did this for making it available in the generated html, right? Could we instead glob for something like `recipe_*.yml`? It would also be nice if the user saw the name of the original recipe so they can easily find it if they wanna reproduce.

Peter9192 (changes requested):
Hey @stefsmeets, just one more remark. And do you know why the tests are failing now? It'll be easier to get it merged by a core team member if they pass.

esmvalcore/_main.py
stefsmeets: Yeah, that is right. I think it fits with giving everything a standard name inside the output directory, this makes it easy to find. I had to do something here so that running a recipe from the API and the CLI has the recipe in the same place. This is by far the easiest solution, because otherwise I would have to generate the name of the recipe in the session object. I can have another go at that. Globbing will also work, but it seems a bit fragile to me.

The name of the recipe in the output directory is not covered by the tests, which is why I was confident to change it. The name of the original recipe is in the output directory.

stefsmeets:
> Hey @stefsmeets, just one more remark. And do you know why the tests are failing now? It'll be easier to get it merged by a core team member if they pass.

No clue, it seemed to happen out of the blue 🤷‍♂️

Peter9192:
@stefsmeets can you try to merge master to see if the tests are fixed?

Peter9192 (approved):

setup.py
```diff
@@ -43,7 +44,6 @@
         'pyyaml',
         'requests',
         'scitools-iris>=3.0.1',
-        'requests',
```
bouweandela, line 46: Why are you removing this dependency? It is used in `esmvalcore/_citation.py`.

esmvalcore/_config/_diagnostics.py
```diff
@@ -77,6 +77,13 @@ class TagsManager(dict):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
+        # Set a default author for testing purposes
```
bouweandela, line 80: Adding code just to be able to run tests is a bit awkward. Could you create a config-references.yml file with the required information for running your tests instead?

bouweandela (commented):
Nice work @stefsmeets! I promise I won't review this pull request too, as you're probably busy enough already with the comments from @Peter9192 and @nielsdrost, but I just checked to make sure it doesn't break anything. Could you please have a look at the comments?

Also, do not forget to add `jinja2` as a requirement in package/meta.yml.

setup.py
stefsmeets: I noticed it's in the list twice (line 44) while I was adding the ninja2 dependency 😅

esmvalcore/_config/_diagnostics.py
stefsmeets: I agree it is a bit awkward, but creating a new `config-references.yml` means we are committing to creating a mini-repo for testing. The way the diagnostics and tags are initiated (with the dependency on `esmvaltool` and fall-back to nothing) make it very difficult to test this properly.

I tried the monkeypatching route, but this became complex very quickly, because the tags import needs to be overwritten in several places, and it caused the tests to fail in other places that monkeypatched in tags.

I eventually figured it out with monkeypatching, but there are two tests that test Julia / R support through a subprocess, and these bypass any variables that we can set. In absence of an `esmvaltool` import, they will initialize `config-references.yml` (TAGS) as an empty dict. Yet, authors are a required keyword. So writing the html file requires the authors to be resolvable, and will thus crash the program.

Adding a new `config-references.yml` also does not crate a clean path to fix the two tests. The issue is that there is no way to specify the location of `config-references.yml` (or diagnostics path for that matter). I will happily create an issue for it and work on it separately, but to me this seemed out of scope for this PR. 

I hope you understand why adding a default to the TAGS makes sense not only for the tests, but also to have a dummy author for new contributors that they can use while they are developing a recipe (#28).

Peter9192: Hey @stefsmeets this sounds reasonable and I think it's a good compromise. Let's see if we can make time to address #28 separately.

esmvalcore/_main.py
stefsmeets: Thanks for the suggestion, this is fixed in the latest commit 👍

bouweandela:
When I run examples/recipe_python.yml with this branch, I now get the following messages towards the end of the run:
```
2021-02-19 09:15:27,201 UTC [21648] WARNING /home/bandela/src/esmvalgroup/esmvalcore/esmvalcore/experimental/_warnings.py:18: UserWarning: 
  Thank you for trying out the new ESMValCore API.
  Note that this API is experimental and may be subject to change.
  More info: https://github.com/ESMValGroup/ESMValCore/issues/498

2021-02-19 09:15:27,205 UTC [21648] WARNING /home/bandela/src/esmvalgroup/esmvalcore/esmvalcore/experimental/config/_config_validators.py:254: ESMValToolDeprecationWarning: `write_plots` will be removed in 2.4.0.

2021-02-19 09:15:27,206 UTC [21648] WARNING /home/bandela/src/esmvalgroup/esmvalcore/esmvalcore/experimental/config/_config_validators.py:255: ESMValToolDeprecationWarning: `write_netcdf` will be removed in 2.4.0.

2021-02-19 09:15:28,025 UTC [21648] WARNING /home/bandela/src/esmvalgroup/esmvalcore/esmvalcore/experimental/config/_validated_config.py:80: MissingConfigParameter: `drs` is not defined (https://docs.esmvaltool.org/projects/ESMValCore/en/latest/quickstart/configure.html)

```
Would it be possible to not show these warnings?

esmvalcore/_config/_diagnostics.py
bouweandela: To summarize the offline discussion we had about this: running ESMValCore without a config-references.yml file is a feature that we would like to keep. The current behaviour is to not replace any tags if config-references.yml is missing. If config-references.yml is present, the tool will try to replace all tags and raise an exception if a tag that is used in the recipe or diagnostic script is not defined in config-references.yml.

bouweandela:
> Also, do not forget to add jinja2 as a requirement in package/meta.yml.

@stefsmeets Did you see this comment?

esmvalcore/_config/_diagnostics.py
stefsmeets: I made it so that it keeps the old behaviour. 👍

However, it creates a very fragile foundation to test from.

1. User defined an author tag that is in `esmvaltool` -> works both with and without `esmvaltool`
2. User defines author tag *not* in `esmvaltool` -> works only when `esmvaltool` is *not* installed, crashes when `esmvaltool` *is* installed

We should have a way to make the define of the diagnostics path i.e. `config-references.yml` so that the outcome of a test no longer depends on whether `esmvaltool` is installed or not.

stefsmeets:
Hi @bouweandela , I addressed your 'I won't review this pull request too'-comments. Can you let me know if your concerns have been addressed?

stefsmeets:
I had to update all the tests using `TAGS`, because monkey patches are persistent. More info: https://stackoverflow.com/questions/46899708/monkeypatch-persisting-across-unit-tests-python

In particular, because I wanted to make sure that the `test_diagnostic_run` passes when the `TAGS` dict is empty (as we discussed offline and in the comments above). This is currently not the case, afaict.

bouweandela:
> I had to update all the tests using TAGS, because monkey patches are persistent. More info: https://stackoverflow.com/questions/46899708/monkeypatch-persisting-across-unit-tests-python

If you do it like the person in that Stack Overflow question does it, this is indeed the case. However, if you use `pytest`'s [`monkeypatch`](https://docs.pytest.org/en/stable/monkeypatch.html) fixture as we do, all modifications will be undone once the test function has finished.

> Can you let me know if your concerns have been addressed?

Looks good now, thanks!

stefsmeets:
> If you do it like the person in that Stack Overflow question does it, this is indeed the case. However, if you use `pytest`'s [`monkeypatch`](https://docs.pytest.org/en/stable/monkeypatch.html) fixture as we do, all modifications will be undone once the test function has finished.

Thanks @bouweandela , for the clarification. I figured that is how it should work.

> > Can you let me know if your concerns have been addressed?
> 
> Looks good now, thanks!

@ESMValGroup/esmvaltool-coreteam As far as I'm concerned, this PR can then be merged!

Issue #961, opened 2021-01-25
Add a coverage reporting service to increase test coverage


bouweandela:
To make us all more aware of how well we are testing new code, it would be good to add a coverage reporting service that comments on a pull request and shows how well we are doing. An example of such a service is https://about.codecov.io/product/features/

valeriupredoi:
good idea! I'll self assign meself to this :+1: We need to make this easy to understand since we'd want to be used not only by geeks like us but rather any of our developers

PR #994, opened 2021-02-09
Report coverage for tests that run on any pull request

bouweandela:
At the moment, coverage reports are only uploaded to Codacy if there are changes to files listed in .circleci/install_triggers. For most pull request that means that there is no coverage information available on Codacy. This pull request fixes that problem by uploading the coverage report from the tests that always run.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,6 +43,9 @@ jobs:
           path: test-reports/
       - store_artifacts:
           path: test-reports/
+      - coverage-reporter/send_report:
+          coverage-reports: 'test-reports/coverage.xml'
+          project-token: $CODACY_PROJECT_TOKEN
 
   install:
     # Test installation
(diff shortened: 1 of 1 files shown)


valeriupredoi:
cheers @bouweandela but don't mind my peasantry but where can I see the coverage in a human-readable form? It'd be nice to put the instructions how to check coverage in the documentation too :+1:

bouweandela:
If you click the 'Files' tab on Codacy. However, I think we should make this much more visible by using a service as described in #961 and then we can add that to the documentation.

valeriupredoi:
yes [index](https://19616-190192145-gh.circle-artifacts.com/0/test-reports/coverage_html/index.html) is nice and gives me the summary I am after, but I needs to scroll to neverland to get to it, - I am fine with it now that I know what I should look for, that's why I'd add a line in the documentation a la "If you want to check the coverage status of the tests please go to commit -> test (if succesful) -> ARTIFACTS -> scroll to test-reports/coverage_html/index.html"

jvegasbsc (approved):

PR #999, opened 2021-02-10
Load fx variables as cube cell measures / ancillary variables

### LINKED ISSUE #436: Preprocessor chain is not being applied to cell measures
Basically, if your preprocessor changes the shape of the diagnostic data cube (for instance a spatial or temporal slice, such as the `annual_statistics` or  `extract_region` preprocesors), before using the  `area_statistics`, zonal_mean` and `volume_statistics`  preprocessors, then the resulting cube will no longer have the same shape as the FX data. 

This means that we can only use the mean operation in `area_statistics`, zonal_mean` and `volume_statistics` can only be applied on th global scale. It is not currently possible to use them on a regional scale.

Many of the preprocessors need to be applied to both the diagnostic dataset, but also the FX data, in the cases when the `mean` operator is used in `area_statistics`, zonal_mean` and `volume_statistics`.

This issue is related to the recent fixes merged by @mattiarighi, @schlunma and @valeriupredoi. The PR's #429 and #433 have closed some FX related issues recently, but I don't want this issue to get lost with the other closed issues and PR.

sloosvel:
## Description

This pull requests adds a new default preprocessor step called `add_cell_measures`. This step cmor-checks and loads areas and volumes as cell measures into a cube variable if the preprocessor steps `area_statistics` and `volume_statistics` are called in the recipe. Thus allowing the preprocessor chain to be applied in fx variables. If those steps are not included, nothing happens to the original cube.

It should also solve a couple of issues that there are with the current fx handling:

- If the volume is time dependent, all files are loaded instead of only keeping the first file.
- I found that the current implementation allows to define the `fx_variables` option as both a list and a list of dicts. Instead this pull requests proposes to call the fx_variables using YAML-file identations so that when the recipe is loaded, it is always loaded as a dict. Because allowing both lists and dicts ends up complicating the the code.

I open this as a draft because there may be some issues in terms of where to define this preprocessor step (it's added in `_other.py`, but that may not be the best option). Also because I had to modify code that I did not originally write, so even though tests are passing, maybe I changed something that should not have been changed.

-   Closes #436 
-   Link to documentation:

***

## Before you get started

## Checklist

-   [ ] PR has a descriptive title for the [changelog](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#branches-pull-requests-and-code-review)
-   [ ] Labels are assigned so they can be used in the [changelog](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#branches-pull-requests-and-code-review)
-   [ ] Code follows the [style guide](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#code-style)

***


esmvalcore/preprocessor/_other.py
```diff
@@ -38,3 +39,97 @@ def clip(cube, minimum=None, maximum=None):
...
+    for fx_files in fx_variables.values():
+        if isinstance(fx_files, str):
+            fx_files = [fx_files]
+        if not fx_files:
+            continue
+        for fx_file in fx_files:
+            loaded_cube = iris.load(fx_file)
+            short_name = loaded_cube[0].var_name
+            mip = loaded_cube[0].attributes['table_id']
+            freq = loaded_cube[0].attributes['frequency']
```
bouweandela, line 84: It might be better to pass these in as function arguments rather than relying on the input NetCDF files to have these global attributes

bouweandela (commented):
Very nice work @sloosvel, it would be great if you could make the pull request ready for review!

valeriupredoi:
nice work @sloosvel - could you please address the Codacy complaints? I will have a good look after that, still stuck in Tool-land at the moment :+1:

esmvalcore/preprocessor/_area.py
```diff
@@ -157,45 +157,6 @@ def meridional_statistics(cube, operator):
...
-                grid_areas = da.tile(grid_areas, [cube.shape[0], 1, 1])
-            else:
-                raise ValueError('Grid and dataset number of dimensions not '
-                                 'recognised: {} and {}.'
-                                 ''.format(cube.ndim, grid_areas.ndim))
-    return grid_areas
-
-
-# get the area average
 def area_statistics(cube, operator, fx_variables=None):
```
bouweandela, line 160: Can the `fx_variables` argument now be removed from this function, since they are now available in the cube?

doc/recipe/preprocessor.rst
```diff
@@ -388,7 +391,11 @@ experiment is preferred for fx data retrieval:
       landmask:
         mask_landsea:
           mask_out: sea
-          fx_variables: [{'short_name': 'sftlf', 'exp': 'piControl'}, {'short_name': 'sftof', 'exp': 'piControl'}]
+          fx_variables:
```
bouweandela, line 394: I like the new syntax, but could you make it backward compatible so the old syntax keeps working, at least for a while?

esmvalcore/preprocessor/__init__.py
```diff
@@ -41,7 +41,7 @@
     mask_outside_range,
 )
 from ._multimodel import multi_model_statistics
-from ._other import clip
+from ._other import clip, add_cell_measure
```
bouweandela, line 44: Maybe you could add a new module esmvalcore/preprocessor/_ancillary_variables.py and keep the code there? `other` is not very clear.

esmvalcore/preprocessor/_volume.py
```diff
@@ -211,26 +211,17 @@ def volume_statistics(
...
+            'Cell measure "ocean_volume" not found in cube. '
+            'Check fx_file availability.'
+        )
+        logger.info('Attempting to calculate grid cell volume...')
 
-    # Check whether the dimensions are right.
-    if cube.data.ndim == 4 and grid_volume.ndim == 3:
-        grid_volume = np.tile(grid_volume,
-                              [cube_shape[0], 1, 1, 1])
+    if not fx_variables:
```
bouweandela, line 223: Same for this function, I think the `fx_variables` argument can be removed now, right?

bouweandela (dismissed):
Thanks, I think this is a very nice new feature. Could you please have a look at the comments?

Could you also add documentation for the new preprocessor function `add_cell_measure` to  doc/recipe/preprocessor.rst ?

Are you planning to add another very similar preprocessor function for adding other ancillary variables in a follow up pull request? E.g. land-sea masks for the masking preprocessor functions?

bouweandela:
@valeriupredoi It would be great if you could also review and test this, as you have a much better understanding than me about what this new feature is supposed to do, because you've worked on it previously.

doc/recipe/preprocessor.rst
bouweandela: Maybe until version 2.4 or 2.5?

esmvalcore/preprocessor/_other.py
```diff
@@ -38,3 +39,93 @@ def clip(cube, minimum=None, maximum=None):
...
+    check_level: CheckLevels
+        Level of strictness of the checks.
+
+
+    Returns
+    -------
+    iris.cube.Cube
+        Cube with added cell measures.
+    """
+    from esmvalcore.preprocessor._io import concatenate
```
bouweandela, line 64: Could you move the imports to the top of the module please?

esmvalcore/preprocessor/_other.py
```diff
@@ -38,3 +39,93 @@ def clip(cube, minimum=None, maximum=None):
...
+    from esmvalcore.cmor.fix import fix_metadata, fix_data
+    from esmvalcore.cmor.check import cmor_check_metadata, cmor_check_data
+
+    if not fx_variables:
+        return cube
+    fx_cubes = iris.cube.CubeList()
+    for _, fx_info in fx_variables.items():
```
bouweandela, line 71:
```suggestion
    for fx_info in fx_variables.values():
```

esmvalcore/preprocessor/_other.py
```diff
@@ -38,3 +39,93 @@ def clip(cube, minimum=None, maximum=None):
...
+        if not fx_info:
+            continue
+        if isinstance(fx_info['filename'], str):
+            fx_info['filename'] = [fx_info['filename']]
+        for fx_file in fx_info['filename']:
+            loaded_cube = iris.load(fx_file)
```
bouweandela, line 77: It would probably be safer to use
https://github.com/ESMValGroup/ESMValCore/blob/9f603be2d726e99fcf069ad747ddc8a73a24e89a/esmvalcore/preprocessor/_io.py#L104
with
https://github.com/ESMValGroup/ESMValCore/blob/9f603be2d726e99fcf069ad747ddc8a73a24e89a/esmvalcore/preprocessor/_io.py#L89
here, so the behaviour is the same as in the preprocessing chain for any other variable.

esmvalcore/preprocessor/_other.py
```diff
@@ -38,3 +39,93 @@ def clip(cube, minimum=None, maximum=None):
...
+
+        fx_cube = cmor_check_metadata(fx_cube, cmor_table=project, mip=mip,
+                                      short_name=short_name, frequency=freq,
+                                      check_level=check_level)
+
+        fx_cube = fix_data(fx_cube, short_name=short_name, project=project,
+                           dataset=dataset, mip=mip, frequency=freq,
+                           check_level=check_level)
+
+        fx_cube = cmor_check_data(fx_cube, cmor_table=project, mip=mip,
```
bouweandela, line 99: For the future, I think it would be a nice idea to create a new `load` function that does all of the preprocessing steps needed to load the data:
https://github.com/ESMValGroup/ESMValCore/blob/9f603be2d726e99fcf069ad747ddc8a73a24e89a/esmvalcore/preprocessor/__init__.py#L79-L94
and then use that both here and also in the usual preprocessor chain, but maybe that's best left for another pull request.

esmvalcore/preprocessor/_area.py
bouweandela: Oh, I now see that you're already broadcasting when loading the data.

esmvalcore/preprocessor/_other.py
```diff
@@ -38,3 +39,93 @@ def clip(cube, minimum=None, maximum=None):
...
+                                  short_name=fx_cube.var_name, frequency=freq,
+                                  check_level=check_level)
+
+        measure_name = {
+            'areacella': 'area',
+            'areacello': 'area',
+            'volcello': 'volume'
+        }
+
+        if fx_cube.var_name in measure_name.keys():
```
bouweandela, line 109:
```suggestion
        if fx_cube.var_name in measure_name:
```
there is no need to use the `.keys()` method, `in` will look at the keys in a dict by default

esmvalcore/preprocessor/_other.py
```diff
@@ -38,3 +39,93 @@ def clip(cube, minimum=None, maximum=None):
...
+                ) from exc
+            measure = iris.coords.CellMeasure(
+                fx_data,
+                standard_name=fx_cube.standard_name,
+                units=fx_cube.units,
+                measure=measure_name[fx_cube.var_name],
+                var_name=fx_cube.var_name,
+                attributes=fx_cube.attributes)
+            cube.add_cell_measure(measure, range(0, measure.ndim))
+            logger.info('Added %s as cell measure in cube of %s.',
```
bouweandela, line 126:
```suggestion
            logger.debug('Added %s as cell measure in cube of %s.',
```
It's better to only use debug log messages from preprocessor functions, because the terminal will become extremely cluttered for a recipe with e.g. 50 datasets and 10 variables.

esmvalcore/preprocessor/_area.py
```diff
@@ -157,45 +157,6 @@ def meridional_statistics(cube, operator):
         raise ValueError(msg)
 
 
-def tile_grid_areas(cube, fx_files):
```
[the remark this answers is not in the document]
bouweandela: Sorry, I noticed just now that you're already broadcasting when loading, so this looks fine.

valeriupredoi:
@bouweandela good call, yes, I'll have a look first thing tomorrow :beer:

esmvalcore/preprocessor/_area.py
sloosvel: I thought of that, but then how would someone specify in the recipe the fx file that is needed?

doc/recipe/preprocessor.rst
sloosvel: I can try, but it won't look too good I think.

sloosvel:
>Are you planning to add another very similar preprocessor function for adding other ancillary variables in a follow up pull request? E.g. land-sea masks for the masking preprocessor functions?

I kind of tried that but had some issues with the iris CellMeasures because it only accepts variables with the `measure` attribute set to either `area` or `volume`, so I am not sure if these types of names are appropiate for masks variables.

bouweandela:
For those variables you will need to use [`iris.cube.Cube.add_ancillary_variable`](https://scitools-iris.readthedocs.io/en/stable/generated/api/iris/cube.html#iris.cube.Cube.add_ancillary_variable) to add them as an ancillary variable instead of as a cell measure.

sloosvel:
Sorry this became really big and with lots of changes to address all the comments. Here is a summary:

- The option to add vars as ancillary vars, instead of only measures, has been added. It's much more convenient than only adding support for measures.
- All broadcasting has been removed from specific preprocessor functions and is now done when adding the measures/ancillary vars.
- The `fx_variables` argument that is set in the recipe for each preprocessor that needs it accepts a list, a list of dicts and yaml-file identations. This argument is later removed from the options of each preprocessor since it's not really needed, but it's kept as an argument in the `add_fx_variables` step.
- I added a bunch of test but I still need to try it with a real recipe to make sure nothing breaks.

valeriupredoi:
@sloosvel awesome job - is it ready for review? Lemme know pls, and if so, pls change it from Draft to RfR :beer:

sloosvel:
It's already open as ready for review!

valeriupredoi:
cheers muchly @sloosvel I promise I'll look at it a bit later today :+1:

valeriupredoi:
later today proved to be 7 days later, sorries, almost like the movie 28 Days Later :radioactive: :grin: Am looking at it now @sloosvel - could you please address the conflict? In the meantime I'll have a good look at le code

doc/recipe/preprocessor.rst
```diff
@@ -325,8 +365,25 @@ experiment is preferred for fx data retrieval:
...
+
+    preprocessors:
+      preproc_weighting:
+        weighting_landsea_fraction:
+          area_type: land
+          exclude: ['CanESM2', 'reference_dataset']
+          fx_variables: [
+            {'short_name': 'sftlf', 'exp': 'piControl'}, 
+            {'short_name': 'sftof', 'exp': 'piControl'}
+            ]
```
valeriupredoi, line 386: last time I coded up this `fx_variables` dicts (or list of dicts) could accept the same number of data parameters as `dataset` - is this still the case? If so, it's prob good to mention it (unless I missed it)

doc/recipe/preprocessor.rst
valeriupredoi: yesh, I was asking a related question in my previous comment. I agree with @bouweandela on this one, when a new functional feature is implemented it's always nice to have a temporary overlap with its previous syntax :+1:

doc/recipe/preprocessor.rst
```diff
@@ -1295,18 +1380,36 @@ region, depth layer or time period is required, then those regions need to be
...
 
 .. code-block:: yaml
 
-    fx_variables: ['areacello', 'volcello']
+    fx_variables: 
+      areacello:
+      volcello:
 
-or as list of dictionaries, with specific variable parameters (they key-value pair may be as specific
+or calling the variables and adding specific variable parameters (they key-value pair may be as specific
```
valeriupredoi, line 1428:
```suggestion
or calling the variables and adding specific variable parameters (the key-value pair may be as specific
```

doc/recipe/preprocessor.rst
```diff
@@ -1355,18 +1458,36 @@ No depth coordinate is required as this is determined by Iris. This function
...
```
valeriupredoi, line 1506:
```suggestion
or calling the variables and adding specific variable parameters (the key-value pair may be as specific
```

esmvalcore/_recipe.py
```diff
@@ -330,77 +333,64 @@ def _add_fxvar_keys(fx_var_dict, variable):
...
+    project_tables = CMOR_TABLES[var_project].tables
 
     # force only the mip declared by user
-    if 'mip' in fx_variable:
-        fx_mips = [fx_variable['mip']]
+    found_mip = False
+    if not fx_info['mip']:
+        for mip in project_tables:
+            fx_cmor = project_tables[mip].get(fx_info['short_name'])
+            if fx_cmor:
```
valeriupredoi: can you add a debug message/exception if the var has no cmor table?

esmvalcore/_recipe.py
```diff
@@ -330,77 +333,64 @@ def _add_fxvar_keys(fx_var_dict, variable):
...
+                found_mip = True
+                fx_info['mip'] = mip
+                fx_info = _add_fxvar_keys(fx_info, variable)
+                logger.debug(
+                    "For fx variable '%s', found table '%s'",
+                    fx_info['short_name'], mip)
+                fx_files = _get_input_files(fx_info, config_user)[0]
+                if fx_files:
+                    logger.debug(
+                        "Found fx variables '%s':\n%s",
```
valeriupredoi:
```suggestion
                        "Found files for fx variables '%s':\n%s",
```

esmvalcore/_recipe.py
```diff
@@ -330,77 +333,64 @@ def _add_fxvar_keys(fx_var_dict, variable):
...
+                        fx_info['short_name'],
+                        pformat(fx_files))
+                    break
```
valeriupredoi: please add a debug message for the case `fx_files` is empty

esmvalcore/_recipe.py
```diff
@@ -330,77 +333,64 @@ def _add_fxvar_keys(fx_var_dict, variable):
...
+            fx_files = _get_input_files(fx_info, config_user)[0]
 
     # If fx variable was not found in any table, raise exception
-    if not searched_mips:
+    if not found_mip:
         raise RecipeError(
-            f"Requested fx variable '{fx_varname}' not available in "
-            f"any 'fx'-related CMOR table ({fx_mips}) for '{var_project}'")
+            f"Requested fx variable '{fx_info['short_name']}' "
+            f"not available in any CMOR table for '{var_project}'")
```
valeriupredoi, line 392: ah thanks, I was complaining about this upstream :+1:

esmvalcore/_recipe.py
```diff
@@ -330,77 +333,64 @@ def _add_fxvar_keys(fx_var_dict, variable):
...
 
     # flag a warning
     if not fx_files:
```
valeriupredoi, line 395: note that this will fail with `fx_files` variable not defined if `fx_cmor` is False, so prob move this up in the loop

esmvalcore/_recipe.py
```diff
@@ -426,14 +416,35 @@ def _update_fx_files(step_name, settings, variable, config_user, fx_vars):
     """Update settings with mask fx file list or dict."""
     if not fx_vars:
         return
+    for fx_var, fx_info in fx_vars.items():
+        if not fx_info:
```
valeriupredoi, line 431: isn't `fx_info` an empty dict `{}` to the very least at all times?

esmvalcore/preprocessor/_ancillary_vars.py
```diff
@@ -0,0 +1,165 @@
...
+        short_name = fx_info['short_name']
+        project = fx_info['project']
+        dataset = fx_info['dataset']
+        mip = fx_info['mip']
+        freq = fx_info['frequency']
+        loaded_cube = fix_metadata(loaded_cube, short_name=short_name,
+                                   project=project, dataset=dataset,
+                                   mip=mip, frequency=freq,
+                                   check_level=check_level)
+        fx_cubes.append(loaded_cube[0])
```
valeriupredoi, line 30: loaded_cube is a CubeList right? why not check and extract the cube you need from the get-going and save a bit of resources?

esmvalcore/preprocessor/_ancillary_vars.py
```diff
@@ -0,0 +1,165 @@
...
+        Iris cube with fx data.
+    measure: str
+        Name of the measure, can be 'area' or 'volume'.
+
+
+    Returns
+    -------
+    iris.cube.Cube
+        Cube with added ancillary variables
+    """
```
valeriupredoi, line 68: please add what it raises as well as `Raises\n-----`

esmvalcore/preprocessor/_ancillary_vars.py
```diff
@@ -0,0 +1,165 @@
...
+    if measure not in ['area', 'volume']:
+        raise ValueError(f"measure name must be 'area or volume, "
```
valeriupredoi, line 70:
```suggestion
        raise ValueError(f"measure name must be 'area' or 'volume', "
```

valeriupredoi (changes requested):
OK a few comments now since the diff needs refreshing

esmvalcore/preprocessor/_ancillary_vars.py
```diff
@@ -0,0 +1,165 @@
...
+        Iris cube with input data.
+    fx_cube: iris.cube.Cube
+        Iris cube with fx data.
+
+
+    Returns
+    -------
+    iris.cube.Cube
+        Cube with added ancillary variables
+    """
```
valeriupredoi, line 106: same here pls add what it raises

esmvalcore/preprocessor/_area.py
```diff
@@ -246,9 +205,17 @@ def area_statistics(cube, operator, fx_variables=None):
...
         if input data cube has different shape than grid area weights
     """
-    grid_areas = tile_grid_areas(cube, fx_variables)
+    grid_areas = None
+    try:
+        grid_areas = cube.cell_measure('cell_area').core_data()
+    except iris.exceptions.CellMeasureNotFoundError:
+        logger.info(
+            'Cell measure "cell_area" not found in cube. '
+            'Check fx_file availability.'
```
valeriupredoi, line 214: print the cube as well, better for identification during a hefty run

esmvalcore/preprocessor/_mask.py
```diff
@@ -73,37 +51,32 @@ def _get_fx_mask(fx_data, fx_option, mask_type):
...
 def _apply_fx_mask(fx_mask, var_data):
     """Apply the fx data extracted mask on the actual processed data."""
-    # Broadcast mask
-    var_mask = np.zeros_like(var_data, bool)
-    var_mask = np.broadcast_to(fx_mask, var_mask.shape).copy()
-
     # Apply mask across
     if np.ma.is_masked(var_data):
-        var_mask |= var_data.mask
+        fx_mask |= var_data.mask
```
valeriupredoi, line 56: you 100% sure they have the same dims? Is there a check that checks the mask array dims and barks at the user if they're not the same? Plan for the worst, always :grin:

esmvalcore/preprocessor/_mask.py
```diff
@@ -132,30 +105,24 @@ def mask_landsea(cube, fx_variables, mask_out, always_use_ne_mask=False):
...
+        fx_cube = None
+        try:
+            fx_cube = cube.ancillary_variable('land_area_fraction')
+        except iris.exceptions.AncillaryVariableNotFoundError:
+            try:
+                fx_cube = cube.ancillary_variable('sea_area_fraction')
+            except iris.exceptions.AncillaryVariableNotFoundError:
+                logger.debug(
+                    'Ancillary variables land/sea area fraction '
+                    'not found in cube. Check fx_file availability.')
```
valeriupredoi, line 119: so (for me to understand) this is the equivalent of first looking for sftlf then sftof right?

valeriupredoi (changes requested):
wow! really nice refactoring @sloosvel and quite a lot of code too, many thanks for this! I have not checked the tests because I trust they're good - I was more interested in the functional implementation and that looks very good to me :+1: Just a few minor questions and comments

tests/unit/preprocessor/_weighting/test_weighting_landsea_fraction.py
```diff
@@ -153,56 +85,31 @@ def test_shape_is_broadcastable(shape_1, shape_2, out):
...
-    if fx_files.get('sftof'):
-        fx_cubes.append(CUBE_SFTOF)
-    mock_iris.load_cube.side_effect = fx_cubes
-    weighted_cube = weighting.weighting_landsea_fraction(
-        cube, fx_files, area_type)
-    assert weighted_cube == cube
+    weighted_cube = weighting.weighting_landsea_fraction(cube, area_type)
+    assert np.array_equal(weighted_cube.data, cube.data)
     assert weighted_cube is cube
-    mock_iris.reset_mock()
```
valeriupredoi, line 208: OK now I've sped through the tests too :grin:

valeriupredoi:
Very nice functional addition! I am well happy I didn't have to do it since I've had it up to here with these fx pests :laughing: The question that remains is have you tested with real life recipes and will you be doing the changes in ESMValTool's recipes to accomodate the new recipe syntax? Cheers again, Saskia!

esmvalcore/_recipe.py
sloosvel: It can be `None` as well, for example when the input is given as a list such as ['areacello', 'areacella']. That is why it's inizialited as a dict.

doc/recipe/preprocessor.rst
sloosvel: Yes, it's still the case. Otherwise some tests would not be passing, but I can add more examples in the documentation.

esmvalcore/_recipe.py
sloosvel: But a False fx_cmor raises a RecipeError, so the tool would stop before getting here no?

esmvalcore/preprocessor/_mask.py
sloosvel: yes, that's the idea.

esmvalcore/preprocessor/_mask.py
sloosvel: Well the fx_var has been broadcasted to the var's shape when added, so I assumed they would be the same at this point. Is there a need for a further check if that's the case?

esmvalcore/preprocessor/_ancillary_vars.py
sloosvel: If the variables are time dependent they might be stored in different files. The fix metadata function works by file and then the cubes get concatenated to continue the checks. But I am not sure what do you mean here, sorry.

valeriupredoi (commented):
sorry Saskia, I thought I'd left these comments yesterday, it appears not for some githubby reason

esmvalcore/preprocessor/_ancillary_vars.py
valeriupredoi: yeah sorry, porr phrasing on my part - what I mean is `fx_file` can be loaded a cube with `load_cube()` since it'll never be a `CubeList` object (I believe) and then can be propagated though as a single cube. If this is not the case, and concatenation needs to be done on multiple cubes then I believe it's safer to concatenate using the `_io.concatenate()` function, so you don't risk concatenation fails due to issues with cube time overlaps or metadata

esmvalcore/preprocessor/_mask.py
valeriupredoi: no, I missed the broadcasting to shape step :)

esmvalcore/preprocessor/_mask.py
valeriupredoi: right one! cheers :+1:

esmvalcore/_recipe.py
sloosvel: Sorry what I meant is that if `fx_cmor` is `None`, then `found_mip` will stay `False` (as in line 325). And that will raise the error before getting to line 384.  Anyway I don't think I changed the order of reporting that there is already in master.

esmvalcore/preprocessor/_ancillary_vars.py
sloosvel: >yeah sorry, poor phrasing on my part - what I mean is fx_file can be loaded a cube with load_cube() since it'll never be a CubeList object (I believe) and then can be propagated though as a single cube

Ah ok! @bouweandela preferred to use the `_io.load()` function though.

>I believe it's safer to concatenate using the _io.concatenate() function, so you don't risk concatenation fails due to issues with cube time overlaps or metadata.

Yes, the proper concatenation happens in the next line!

valeriupredoi: indeed it does and I am fine with `_io.load()` too, it's more secure indeed, cheers Saskia! :+1:

esmvalcore/_recipe.py
valeriupredoi: OK cool, I trust you on that :grin:

sloosvel:
>Very nice functional addition! I am well happy I didn't have to do it since I've had it up to here with these fx pests 😆. The question that remains is have you tested with real life recipes and will you be doing the changes in ESMValTool's recipes to accomodate the new recipe syntax? Cheers again, Saskia!

Sure I'll take care of that as well! Thanks for taking a look.

tests/integration/test_recipe.py
```diff
@@ -2165,7 +2165,8 @@ def test_fx_list_mip_change_cmip6(tmp_path, patched_datafinder, config_user):
     assert '_Ofx_' in fx_variables['areacello']['filename']
     assert '_Efx_' in fx_variables['clayfrac']['filename']
     assert '_fx_' in fx_variables['sftlf']['filename']
-    assert '_fx_' in fx_variables['sftgif']['filename']
+    assert len(fx_variables['sftgif']['filename']) == 1
+    assert 'sftgif_' in fx_variables['sftgif']['filename'][0]
```
valeriupredoi, line 2169: @sloosvel am a bit concerned about this fix I had to introduce, isn't `fx_variables['sftgif']['filename']` a string no matter what fx variable we look at? Why should it be a list for `sftgif` and regular string for the other fx vars?

sloosvel: It can be a list if the fx var is time dependent, since therefore the data can be split across different files if it's chunked. And `sftgif` is present in many of the CMOR tables. I think that the difference is that the current implementation look only in the `fx` mip (which is time independent), unless specified by the user. Whereas this one looks at all the tables unless specified. I think during the integration tests it finds an sftgif that belongs to `mip: IyrAnt`, but I don't think that is a problem?

valeriupredoi: OK good to know Saskia, cheers for the clarification, makes good sense. The test is failing on Circle because of the test codebase being used from the release, and not from current development branch, see #1063 - please don't fix it, I need to run a GA test be 100% sure that issue is valid :+1:

valeriupredoi:
@sloosvel I am very happy for this PR to be merged asap, I have gone through it in detail and I reckon it's a very nice one and you've answered all my queries (cheers for being patient with me!). The only thing preventing a merge is the broken test on Circle that is not broken in reality, I'd like @bouweandela to shed light on this via #1063 (when he's got time) and then we'll get this in right away :beer:

valeriupredoi (approved):

tests/integration/test_recipe.py
sloosvel: Actually I think it's an old test, because my local branch does not look like that. In that case sftgif is only a string. But it could also be a list if the file was time dependent. I'll update the branch so this test is fixed.

valeriupredoi: cool, cheers Saskia! note that the issue is still present (not related to your PR, but the issue of older codebase tested with with up-to-date tests in Circle). Best advice is to run tests locally too, everytime one declares a PR ready for merge :+1:

bouweandela:
> The only thing preventing a merge is the broken test on Circle that is not broken in reality, I'd like @bouweandela to shed light on this via #1063 (when he's got time) and then we'll get this in right away :beer:

The test is fixed now, let's get this merged if you think it's ready! If needed, we can continue the discussion on the tests in the issue. I won't be back at work for another two weeks, let's not wait for that.

sloosvel:
Actually I have been wondering if it would be better to remove the ancillary vars/ measures after the preproc steps that need them finish.

valeriupredoi:
@bouweandela cheers! @sloosvel all fine and ready for merge by me, just let me know if you think this is ready-ready :+1: BTW I am totally fine with the Codacy issues too :beer:

bouweandela:
@sloosvel I think that would be better, maybe just before the `save` step? Because of the broadcasting, keeping them would result in rather large output files.

valeriupredoi:
good to merge @sloosvel ? I feel the need for speed(merge) :grin:

sloosvel:
Well it seemed to work for what I have tested but if anyone else wants to double check they are welcome to!

valeriupredoi:
I am very happy to merge now and fix whatevs later, I tested this and it works - if there are any corner cases we can always open new PRs, this ia great amount of work and added functionality that needs not wait no more :+1:

valeriupredoi:
thank you very very much again @sloosvel :beer:

PR #1003, opened 2021-02-12
Fix tests for multi-processing with spawn method (i.e. macOSX with Python>3.8)

### LINKED ISSUE #1002: Error message for 'spawn' multiprocessing method
ESMValTool tests fail when the Python multiprocessing method is `spawn` instead of `fork`. This makes ESMValTool not usable on Windows, and it is a tricky issue on MacOSX, where [Python 3.8 uses `spawn`, while earlier versions used `fork`](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods). However, installation of ESMValTool works fine on MacOSX with Python 3.8, so the user is only alerted to issues much later, and it is unclear to a user why problems occur.

An informative error message early in the process would be helpful. Perhaps already with installation?

The multiprocessing method in Python can be checked as follows:
```
import multiprocessing as mp
mp.get_context()
```

bvreede:
## Description
Fixed a test that caused problems with MacOSX when using multiprocessing set to spawn (standard in Python 3.8).

-   Closes #1002
-   Link to documentation:

***

## Before you get started

## Checklist

***

### DIFF
--- a/tests/integration/test_task.py
+++ b/tests/integration/test_task.py
@@ -8,28 +8,9 @@ import esmvalcore
 from esmvalcore._task import BaseTask, TaskSet
 
 
-@pytest.fixture
-def example_tasks():
-    """Example tasks for testing the task runners."""
-    tasks = TaskSet()
-    for i in range(3):
-        task = BaseTask(
-            name=f'task{i}',
-            ancestors=[
-                BaseTask(name=f'task{i}-ancestor{j}') for j in range(3)
-            ],
-        )
-        for task0 in task.flatten():
-            task0.priority = i
-        tasks.add(task)
-
-    return tasks
-
-
-@pytest.mark.parametrize('max_parallel_tasks', [1, 2, 3, 4, 16, None])
-def test_run_tasks(monkeypatch, tmp_path, max_parallel_tasks, example_tasks):
-    """Check that tasks are run correctly."""
+class MockBaseTask(BaseTask):
     def _run(self, input_files):
+        tmp_path = self._tmp_path
         output_file = tmp_path / self.name
 
         msg = ('running {} in thread {}, using input {}, generating {}'.format(
@@ -50,7 +31,29 @@ def test_run_tasks(monkeypatch, tmp_path, max_parallel_tasks, example_tasks):
 
         return [output_file]
 
-    monkeypatch.setattr(BaseTask, '_run', _run)
+
+@pytest.fixture
+def example_tasks(tmp_path):
+    """Example tasks for testing the task runners."""
+    tasks = TaskSet()
+    for i in range(3):
+        task = MockBaseTask(
+            name=f'task{i}',
+            ancestors=[
+                MockBaseTask(name=f'task{i}-ancestor{j}') for j in range(3)
+            ],
+        )
+        for task0 in task.flatten():
+            task0.priority = i
+            task0._tmp_path = tmp_path
+        tasks.add(task)
+
+    return tasks
+
+
+@pytest.mark.parametrize('max_parallel_tasks', [1, 2, 3, 4, 16, None])
+def test_run_tasks(monkeypatch, tmp_path, max_parallel_tasks, example_tasks):
+    """Check that tasks are run correctly."""
 
     example_tasks.run(max_parallel_tasks=max_parallel_tasks)
 
@@ -72,7 +75,7 @@ def test_runner_uses_priority(monkeypatch, runner, example_tasks):
         order.append(self.priority)
         return [f'{self.name}_test.nc']
 
-    monkeypatch.setattr(BaseTask, '_run', _run)
+    monkeypatch.setattr(MockBaseTask, '_run', _run)
     monkeypatch.setattr(esmvalcore._task, 'Pool', ThreadPool)
 
     runner(example_tasks)
--- a/tests/integration/test_task.py
+++ b/tests/integration/test_task.py
@@ -1,3 +1,4 @@
+import multiprocessing
 import os
 from functools import partial
 from multiprocessing.pool import ThreadPool
@@ -51,10 +52,13 @@ def example_tasks(tmp_path):
     return tasks
 
 
+@pytest.mark.parametrize('mpmethod', ["spawn", "fork"])
 @pytest.mark.parametrize('max_parallel_tasks', [1, 2, 3, 4, 16, None])
-def test_run_tasks(monkeypatch, tmp_path, max_parallel_tasks, example_tasks):
+def test_run_tasks(monkeypatch, tmp_path, max_parallel_tasks, example_tasks,
+                   mpmethod):
     """Check that tasks are run correctly."""
-
+    monkeypatch.setattr(esmvalcore._task, 'Pool',
+                        multiprocessing.get_context(mpmethod).Pool)
     example_tasks.run(max_parallel_tasks=max_parallel_tasks)
 
     for task in example_tasks:


valeriupredoi:
nice one @bvreede - but the important question is if an actual task (not a mock task) runs on OSX without setting the multiproc method at the start of the script? Since that's what was getting raised by the failing test :+1:

bvreede:
Thanks @valeriupredoi! As far as I understand, the task itself runs with either `spawn` or `fork`; the issue we zeroed in on was with the test. Fwiw running recipes on mac osx with python >3.8 works well as far as I've seen — have yet to encounter an error. Are there any issues outside of this failing test that we know of, that were fixed by setting multiprocessing to `fork`?

The method that was used in this test is a monkeypatched _run method that is still used, except now called in a different way; i.e. MockBaseTask._run.

valeriupredoi:
great! if you confirm that you are able to run a few example recipes on OSX with Python>=3.8 and `max_parallel_tasks:` set to 2 or more in the config-user then it means we good to go and in business :+1: :beer:

bvreede:
I can confirm this!

bvreede:
Is that enough? :)

valeriupredoi (approved):
excellent work :partying_face:

valeriupredoi:
> Is that enough? :)

absolutely! I feel like a blind man trying to figure all this stuff only relying on the remote GA machine. Good stuff! If I am not too much of a nuisance, would you be able to have a look at the GA tests next include OSX fully - for tests and builds etc :beer:

bvreede:
If you can explain to me what the GA tests are 😅  absolutely. 
[Apologies; new to this!]

valeriupredoi:
sorry - Github Actions, you can find them in `.github/workflows` and let me know if you have questions about them, I'd be very happy to assist!

stefsmeets (commented):
To ensure that this is working with spawn, can you add a test with a context manager:
i.e. something like this:

```
with multiprocessing.get_context('spawn').Pool() as pool:
    pool.map(function, args)
```

tests/integration/test_task.py
```diff
@@ -50,7 +31,29 @@ def _run(self, input_files):
...
+        tasks.add(task)
+
+    return tasks
+
+
+@pytest.mark.parametrize('max_parallel_tasks', [1, 2, 3, 4, 16, None])
+def test_run_tasks(monkeypatch, tmp_path, max_parallel_tasks, example_tasks):
+    """Check that tasks are run correctly."""
 
     example_tasks.run(max_parallel_tasks=max_parallel_tasks)
```
stefsmeets, line 58: Maybe add a parameterizer with 'fork' / 'spawn' that does something like this:

```suggestion
    monkeypatch.setattr(esmvalcore._task, 'Pool', mp.get_context('spawn').Pool)
    example_tasks.run(max_parallel_tasks=max_parallel_tasks)
```

bvreede:
@valeriupredoi 

The failing test in GA is the same as the one that's failing on my machine. The [latest run on master](https://github.com/ESMValGroup/ESMValCore/runs/1878594842) shows the following failure summary:
```
FAILED tests/integration/test_task.py::test_run_tasks[2] - TypeError: 'NoneTy...
FAILED tests/integration/test_task.py::test_run_tasks[3] - TypeError: 'NoneTy...
FAILED tests/integration/test_task.py::test_run_tasks[4] - TypeError: 'NoneTy...
FAILED tests/integration/test_task.py::test_run_tasks[None] - TypeError: 'Non...
FAILED tests/integration/test_task.py::test_run_tasks[16] - TypeError: 'NoneT...
FAILED tests/integration/cmor/_fixes/cmip6/test_cesm2.py::test_cl_fix_file
```

The last one is new to me, however; and it does not fail locally with python 3.9. [I am trying to reproduce this with python 3.8 but having a fight with conda so no dice rn — this can be entirely chalked up to my incompetence btw.]
Can we add 3.9 to GA tests for OSX? There's a comment that ESMValTool does not work well for 3.9, but it's not clear to me why. It would be informative to see whether this test also fails for 3.9, for one.

valeriupredoi:
sure - go for it! That's my old comment before we've done all the good stuff with Python 3.9. Pls reinstate all the tests I commented out and add whatever you think it's needed so we start understanding what's going on in OSX land (well, more like you start understanding coz I am OSX-less :grin: ). But let's first merge this and continue work in a new PR, I reckon

bvreede:
Thanks for the review @stefsmeets! I agree it's a good idea to make sure that the tests (and the software) work with both multiprocessing methods — but given that we have now explicitly removed the dependency on spawn/fork from this test, it doesn't make sense to me to add this condition. Or am I missing something?
It does make sense to test the task function itself under both methods — would this do that?
Appreciate your clarification, I'm learning tons!

stefsmeets:
> Thanks for the review @stefsmeets! I agree it's a good idea to make sure that the tests (and the software) work with both multiprocessing methods — but given that we have now explicitly removed the dependency on spawn/fork from this test, it doesn't make sense to me to add this condition. Or am I missing something?
> It does make sense to test the task function itself under both methods — would this do that?
> Appreciate your clarification, I'm learning tons!

Hi @bvreede , good question! This is to make sure that we do not silently introduce a bug that affects the multiprocessing behaviour. i.e. if we want to transition to using 'spawn' eventually / support osx, we might as well add a test right now to ensure any change in behaviour of esmvalcore will cause the test to fail for both methods.

bouweandela:
@bvreede Could you please merge the master branch into this pull request, to see if all tests now pass?

bvreede:
Not sure if I did that in the most elegant way, but all tests pass now! Merci bien.

tests/integration/test_task.py
```diff
@@ -1,3 +1,4 @@
+import multiprocessing
```
stefsmeets, line 1:
```suggestion
import multiprocessing
```
```suggestion
import multiprocessing as mp
```

tests/integration/test_task.py
```diff
@@ -50,8 +32,33 @@ def _run(self, input_files):
...
+    return tasks
+
+
+@pytest.mark.parametrize('mpmethod', ["spawn", "fork"])
+@pytest.mark.parametrize('max_parallel_tasks', [1, 2, 3, 4, 16, None])
+def test_run_tasks(monkeypatch, tmp_path, max_parallel_tasks, example_tasks,
+                   mpmethod):
+    """Check that tasks are run correctly."""
+    monkeypatch.setattr(esmvalcore._task, 'Pool',
+                        multiprocessing.get_context(mpmethod).Pool)
```
stefsmeets, line 61:
```suggestion
    monkeypatch.setattr(esmvalcore._task, 'Pool', mp.get_context(mpmethod).Pool)
```

tests/integration/test_task.py
bvreede: Haha, that was my original line, but changed with advice from @bouweandela :)
I did agree with his argument though: because we only use the multiprocessing package once, it only reduces readability and doesn't add much to shorten it.

valeriupredoi: yes, if a module is used <2 times it should be kept with its name, unless of course it's a German name that is made up of >300 letters :grin:

stefsmeets (changes requested):
Hi @bvreede , nice work! Looks fine to me. I'm just wondering _why_ this works, because it seems like it is almost the same as before. I'm not entirely happy how the second test (`test_runner_uses_priority`) now monkeypatches `MockBaseTask`, but I guess it is the way it is for now. I hope it won't cause issues for the next person debugging this in the future.

Just a small comment on the code and then it is good to go. 🚀

tests/integration/test_task.py
stefsmeets: This is the standard way to import multiprocessing 😅

stefsmeets (approved):

stefsmeets:
All good, I have not strong enough opinions whether this should be `mp` or `multiprocessing`, so this can be merged @valeriupredoi 👍

Issue #1020, opened 2021-02-24
Possibility: run tests in tests/integration/cmor/_fixes with one thread only


valeriupredoi:
I am currently conducting an experiment and am running the tests every hour on the GA machine, but with `pytest -n 1` see if my previous find that tests don't segfault when run in single-thread mode. If this proves to be true, I would argue that we should run a number of tests like this - not all of them because for one, not all tests result in segfaults and two, I have noticed only the ones in `tests/integration/cmor/_fixes` to be segfaulting (well 99.9% of times). @bouweandela or @jvegasbsc do you guys know how to run a subset of the tests in single-thread whereas keeping the others in multi-threading? BTW if you see a ton of Test jobs on the GA machine, that's me experiment :grin:

valeriupredoi:
and I totally forgot that scheduled GA jobs run only on `master` so am gonna make that a PR and decrease the frequency from hourly to 6-hourly so we don't clog the GA machine too badly and ask to merge into `master` so we can observe

stefsmeets:
I looked into this a while ago, and I could not find a way to run a subset of tests in a single thread. There is an open issue for this: https://github.com/pytest-dev/pytest-xdist/issues/385.

Could mark the segfaulting tests using `@pytest.mark.sequential`, and then do `pytest -n1 -m"sequential"`

valeriupredoi:
@stefsmeets cheers! that's a good idea - but decorating them that way means they will run only when they get invoked with the runtime flag?

stefsmeets:
Yeah, you would have to do the following I think:

```
pytest -n1 -m"sequential"
pytest -n4 -m"not sequential"
```

Maybe one or the other can be set as a default in `setup.cfg`.

valeriupredoi:
unfortunately this is not a solution to get us rid of the Segmentation Faults. I've just looked at the 3-hourly tests that have been running with `-n 1` and segfaults creep in there too, going to revert the change to the GA workflow (so we don't run that test like pogos every 3 hours for nothing) and close the issue after, the mystery still persists... :confounded:

valeriupredoi:
reverted the changes in #1025 if someone is nice to approve it, we'll forget about this. On to figuring out another cause of segfaults

PR #1022, opened 2021-02-25
Switch to running the Github Action test workflow every 3 hours in single thread mode to observe if Sementation Faults occur

valeriupredoi:
I am trying to gather enough statistic to prove that if we run the test in single thread mode we will not see the Segmentation Faults we've been seeing for nearly a year now. So to do that I want to have the GA run the test only workflow every three hours; unfortunately it's not possible to have this done on a branch, it has to be the main branch, but we can let this go for a few days, gather enough statistic then revert to the nightly run as we had it so far. If this is true then we can do what @stefsmeets suggests in this [comment](https://github.com/ESMValGroup/ESMValCore/issues/1020#issuecomment-785809295) and hopefully rid ourselves of those pesty SegFaults for good. Please see logic behind this PR in https://github.com/ESMValGroup/ESMValCore/issues/1020

### DIFF
--- a/.github/workflows/action-test.yml
+++ b/.github/workflows/action-test.yml
@@ -22,15 +22,14 @@ on:
   push:
     branches:
     - master
-    - run_test_hourly
   # run the test only if the PR is to master
   # turn it on if required
   #pull_request:
   #  branches:
   #  - master
   schedule:
-    # - cron: '0 0 * * *'
-    - cron: '0 * * * *'
+    # - cron: '0 0 * * *'  # nightly
+    - cron: '0 */3 * * *'  # every 3 hours
 
 jobs:
   linux:
--- a/.github/workflows/action-test.yml
+++ b/.github/workflows/action-test.yml
@@ -22,14 +22,15 @@ on:
   push:
     branches:
     - master
-    - github-actions2
+    - run_test_hourly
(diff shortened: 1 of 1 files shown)


.github/workflows/action-test.yml
```diff
@@ -22,14 +22,14 @@ on:
...
-    - github-actions2
   # run the test only if the PR is to master
   # turn it on if required
   #pull_request:
   #  branches:
   #  - master
   schedule:
-    - cron: '0 0 * * *'
+    # - cron: '0 0 * * *'  # nightly
+    - cron: '0 */3 * * *'  # every 3 hours
```
stefsmeets, line 32: Are you going to change this back? Otherwise I'd suggest to just replace the line:

```suggestion
    - cron: '0 */3 * * *'  # every 3 hours
```

stefsmeets (approved):
Looks fine to me! 🚀

.github/workflows/action-test.yml
valeriupredoi: yeah no, gonna change it back, we really don't want to run that test every 3h for the rest of our time :grin:

valeriupredoi:
cheers @stefsmeets :beer:

Issue #1018, opened 2021-02-23
Preprocessor chain 'climate_statistics' and 'multimodel_statistics' fails as time dimension is not present


tomaslovato:
The construction of a preprocessor that computes first `climate_statistics` and then `multimodel_statistics` fails with a missing `time` coordinate error from the latter.

This is reasonable as for individual model `climate_statistics` the time axis is replaced with the one corresponding to requested operation (either 'day_of_year', 'month_number', 'season_number'), but `multimodel_statistics` expects to receive a cube with the `time` coordinate

I made a test and it should be a big issue to replace the deletion of `time` coordinate from the cube after aggregation
https://github.com/ESMValGroup/ESMValCore/blob/7f8406bdafe898c05d37f398457023b228e79324/esmvalcore/preprocessor/_time.py#L537-L539

with a demotion of the cube `time` coordinate to auxiliary (time will have values coherent with the aggregation operation), as in the following 
```python
@@ -535,7 +535,8 @@ def climate_statistics(cube,
     clim_coord = _get_period_coord(cube, period, seasons)
     operator = get_iris_analysis_operation(operator)
     clim_cube = cube.aggregated_by(clim_coord, operator)
-    clim_cube.remove_coord('time')
+    iris.util.demote_dim_coord_to_aux_coord(clim_cube,'time')
     if clim_cube.coord(clim_coord.name()).is_monotonic():
         iris.util.promote_aux_coord_to_dim_coord(clim_cube, clim_coord.name())
     else:
```

valeriupredoi:
cheers @tomaslovato - related to https://github.com/ESMValGroup/ESMValCore/issues/890 - I'll have to find some time next week sit down with the code and get all this jazz in (note the associated issues related to `multimodel`, there are a few)

tomaslovato:
@valeriupredoi I had a look to changes of time handling in PR #1023, but yet the code for multimodel statistics requires that `time` is a coordinate of the cube
https://github.com/ESMValGroup/ESMValCore/blob/0c798617b885777a5ffa1697c39bb46974a73bab/esmvalcore/preprocessor/_multimodel.py#L208-L212

while `time` coordinate is removed when computing the `climate_statistics`
https://github.com/ESMValGroup/ESMValCore/blob/0c798617b885777a5ffa1697c39bb46974a73bab/esmvalcore/preprocessor/_time.py#L544-L546

I think that the solution proposed above in this discussion should be a viable fix.

[...]

Issue #1009, opened 2021-02-19
Preprocessor multi_model_statistics cannot handle altitude coordinate


axel-lauer:
The preprocessor calculating multi_model_statistics cannot handle data with a vertical coordinate "altitude" and crashes with "Expected to find exactly 1 air_pressure coordinate, but found none." in _multimodel.py. This error can reproduced with this recipe: [recipe_test_mm.yml.txt](https://github.com/ESMValGroup/ESMValCore/files/6009023/recipe_test_mm.yml.txt)

Log file: [main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/6009000/main_log_debug.txt)

stefsmeets:
Hi @axel-lauer , could you add a test for this bug so that we can make sure it does not happen in #968 ?

Issue #975, opened 2021-02-01
Multi model statistics: missing files for individual models 


remi-kazeroni:
**Describe the bug**
When using the `multi_model_statistics` preprocessor step, the `preproc` directory contains only the global file for multi model statistics. The files for each datasets are missing.

Note that everything seems fine with the branch `release_2.1`. We should aim at fixing that before the next release @ESMValGroup/esmvaltool-coreteam 

**Please attach**
  - The recipe that you are trying to run, you can find a copy in the `run` directory in the output directory
  - The `main_log_debug.txt` file, this can also be found in the `run` directory in the output directory
[recipe_test_mmm.yml.txt](https://github.com/ESMValGroup/ESMValCore/files/5904676/recipe_test_mmm.yml.txt)
[main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/5904677/main_log_debug.txt)

remi-kazeroni:
Mistral being super slow today, I haven't been able to create an up to date environment... Mine is about one week old...

schlunma:
I'm pretty sure this was introduced in #949:

The old version of the cube added all products to the output of `multi_model_statistics`:

https://github.com/ESMValGroup/ESMValCore/blob/7d829964e633e65031a709568502c6df505216bd/esmvalcore/preprocessor/_multimodel.py#L391

The new version does not:

https://github.com/ESMValGroup/ESMValCore/blob/0fb657d07194309e8801979d02feb4a5a9cd25a4/esmvalcore/preprocessor/_multimodel.py#L368-L379

Peter9192:
Good spot! The behaviour shouldn't have changed like this, but also I think maybe we shouldn't change it back. Let me explain.

#949 is actually working towards support for an ensemble statistics preprocessor. We first implemented all of this in #673 but then figured it had become too big and we wanted to split it up in smaller tasks. Hence I copied some of the changes to #949. At this point it was not necessary to change the behaviour, and I should have spotted that. 

However, when we continue towards the ensemble stats, we need this behaviour to change. See for example [this](example) use case. If you, at some point in the future, want to combine multimodel and ensemble statistics in a single preprocessor (e.g. first compute ens means, then multimodel between the ensemble means), you don't want to pass the original products to multimodel as well. 

So a preprocessor that contains multimodel should *only* return multimodel products. An alternative way to get the output from the original cubes as well is to split the preprocessor:

```yaml
preprocessors:
  everything_else: &everything_else
[...]

Issue #891, opened 2020-11-30
Preprocessor `multimodel_statistics` fails when data have no horizontal dimension


stefsmeets:
**Describe the bug**
Hi all, I'm developing some tests for the multimodel statistics using real data (#856), and I'm coming accross this bug:

Running multimodel statistics with a list of cubes with **no horizontal dimension**, will result in
`iris.exceptions.CoordinateNotFoundError: 'Expected to find exactly 1 depth coordinate, but found none.'`

The bug can be reproduced by:
```python
cubes = [cube[:, :, 0, 0] for cube in cubes]
multi_model_statistics(cubes, span='full', statistics=['mean'])
```

<details>
  <summary>See the full stack trace below.</summary>

```
timeseries_cubes_month = [<iris 'Cube' of air_temperature / (K) (time: 14; air_pressure: 2; latitude: 3; longitude: 2)>, <iris 'Cube' of air_te... 3; longitude: 2)>, <iris 'Cube' of air_temperature / (K) (time: 14; air_pressure: 2; latitude: 3; longitude: 2)>, ...]

    @pytest.mark.functional
    # @pytest.mark.xfail('iris.exceptions.CoordinateNotFoundError')
    def test_multimodel_no_horizontal_dimension(timeseries_cubes_month):
        """Test statistic without horizontal dimension using monthly data."""
        span = 'full'
        cubes = timeseries_cubes_month
        cubes = [cube[:, :, 0, 0] for cube in cubes]
        # Coordinate not found error
        # iris.exceptions.CoordinateNotFoundError:
        # 'Expected to find exactly 1 depth coordinate, but found none.'
>       multimodel_test(cubes, span=span, statistic='mean')

tests/functional/test_multimodel.py:207:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/functional/test_multimodel.py:125: in multimodel_test
    output = multi_model_statistics(cubes, span=span, statistics=statistics)
esmvalcore/preprocessor/_multimodel.py:439: in multi_model_statistics
    statistic_cube = _assemble_full_data(cubes, statistic)
[...]

PR #1023, opened 2021-02-25
Update tests for multimodel statistics preprocessor

stefsmeets:
## Description

Hi everyone, we are developing a new implementation of multi-model statistics preprocessor in #968. The goal of this PR is to to better chart out the requirements for the preprocessor, and to have a more complete description of the functionality through the unit tests, so that no functionality / features are lost.

The current set of tests does not cover all the functionality and is difficult to extend as it still makes use of the old `unittest` framework. This PR re-implements the functionality of the old tests using `pytest`.

-   Closes #issue_number
-   Link to documentation:

***

## Before you get started

## Checklist

***

### DIFF
--- a/tests/unit/preprocessor/_multimodel/test_multimodel.py
+++ b/tests/unit/preprocessor/_multimodel/test_multimodel.py
@@ -117,7 +117,7 @@ VALIDATION_DATA_SUCCESS = (
         'std', (5.656854249492381, 4, 2.8284271247461903),
         marks=pytest.mark.xfail(
             raises=AssertionError,
-            reason='Iris 3.0.1 uses `ddof=1` for calculation of std. dev.')),
+            reason='https://github.com/ESMValGroup/ESMValCore/issues/1024')),
     ('full', 'min', (1, 1, 1)),
     ('full', 'max', (9, 9, 5)),
     ('full', 'median', (5, 5, 3)),
@@ -129,7 +129,7 @@ VALIDATION_DATA_SUCCESS = (
         'std', (5.656854249492381, 4),
         marks=pytest.mark.xfail(
             raises=AssertionError,
-            reason='Iris 3.0.1 uses `ddof=1` for calculation of std. dev.')),
+            reason='https://github.com/ESMValGroup/ESMValCore/issues/1024')),
     ('overlap', 'min', (1, 1)),
     ('overlap', 'max', (9, 9)),
     ('overlap', 'median', (5, 5)),
--- a/tests/unit/preprocessor/_multimodel/test_multimodel.py
+++ b/tests/unit/preprocessor/_multimodel/test_multimodel.py
@@ -1,267 +1,305 @@
-"""Unit test for :func:`esmvalcore.preprocessor._multimodel`."""
+"""Unit test for :func:`esmvalcore.preprocessor._multimodel`"""
 
-import unittest
+from datetime import datetime
 
 import iris
 import numpy as np
+import pytest
 from cf_units import Unit
+from iris.cube import Cube
 
-import tests
+import esmvalcore.preprocessor._multimodel as mm
 from esmvalcore.preprocessor import multi_model_statistics
-from esmvalcore.preprocessor._multimodel import (
-    _assemble_data,
-    _compute_statistic,
-    _get_time_slice,
-    _multiproduct_statistics,
-    _plev_fix,
-    _put_in_cube,
-    _unify_time_coordinates,
+
+SPAN_OPTIONS = ('overlap', 'full')
+
+FREQUENCY_OPTIONS = ('daily', 'monthly', 'yearly')  # hourly
+
+CALENDAR_OPTIONS = ('360_day', '365_day', 'gregorian', 'proleptic_gregorian',
+                    'julian')
+
+
+def assert_array_almost_equal(this, other):
+    """Assert that array `this` almost equals array `other`."""
+    if np.ma.isMaskedArray(this) or np.ma.isMaskedArray(other):
+        np.testing.assert_array_equal(this.mask, other.mask)
+
+    np.testing.assert_allclose(this, other)
+
+
+def timecoord(frequency,
+              calendar='gregorian',
+              offset='days since 1850-01-01',
+              num=3):
+    """Return a time coordinate with the given time points and calendar."""
+
+    time_points = range(1, num + 1)
+
+    if frequency == 'hourly':
+        dates = [datetime(1850, 1, 1, i, 0, 0) for i in time_points]
+    if frequency == 'daily':
+        dates = [datetime(1850, 1, i, 0, 0, 0) for i in time_points]
+    elif frequency == 'monthly':
+        dates = [datetime(1850, i, 15, 0, 0, 0) for i in time_points]
+    elif frequency == 'yearly':
+        dates = [datetime(1850, 7, i, 0, 0, 0) for i in time_points]
+
+    unit = Unit(offset, calendar=calendar)
+    points = unit.date2num(dates)
+    return iris.coords.DimCoord(points, standard_name='time', units=unit)
+
+
+def generate_cube_from_dates(
+    dates,
+    calendar='gregorian',
+    offset='days since 1850-01-01',
+    fill_val=1,
+    len_data=3,
+    var_name=None,
+):
+    """Generate test cube from list of dates / frequency specification.
+
+    Parameters
+    ----------
+    calendar : str or list
+        Date frequency: 'hourly' / 'daily' / 'monthly' / 'yearly' or
+        list of datetimes.
+    offset : str
+        Offset to use
+    fill_val : int
+        Value to fill the data with
+    len_data : int
+        Number of data / time points
+    var_name : str
+        Name of the data variable
+
+    Returns
+    -------
+    iris.cube.Cube
+    """
+    if isinstance(dates, str):
+        time = timecoord(dates, calendar, offset=offset, num=len_data)
+    else:
+        unit = Unit(offset, calendar=calendar)
+        time = iris.coords.DimCoord(unit.date2num(dates),
+                                    standard_name='time',
+                                    units=unit)
+
+    return Cube((fill_val, ) * len_data,
+                dim_coords_and_dims=[(time, 0)],
+                var_name=var_name)
+
+
+def get_cubes_for_validation_test(frequency):
+    """Set up cubes used for testing multimodel statistics."""
+
+    # Simple 1d cube with standard time cord
+    cube1 = generate_cube_from_dates(frequency)
+
+    # Cube with masked data
+    cube2 = cube1.copy()
+    cube2.data = np.ma.array([5, 5, 5], mask=[True, False, False])
+
+    # Cube with deviating time coord
+    cube3 = generate_cube_from_dates(frequency,
+                                     calendar='360_day',
+                                     offset='days since 1950-01-01',
+                                     len_data=2,
+                                     fill_val=9)
+
+    return [cube1, cube2, cube3]
+
+
+VALIDATION_DATA_SUCCESS = (
+    ('full', 'mean', (5, 5, 3)),
+    ('full', 'std', (5.656854249492381, 4, 2.8284271247461903)),
+    # ('full', 'std_dev', (5.656854249492381, 4, 2.8284271247461903)),
+    ('full', 'min', (1, 1, 1)),
+    ('full', 'max', (9, 9, 5)),
+    ('full', 'median', (5, 5, 3)),
+    ('full', 'p50', (5, 5, 3)),
+    ('full', 'p99.5', (8.96, 8.96, 4.98)),
+    ('overlap', 'mean', (5, 5)),
+    ('overlap', 'std', (5.656854249492381, 4)),
+    # ('overlap', 'std_dev', (5.656854249492381, 4)),
+    ('overlap', 'min', (1, 1)),
+    ('overlap', 'max', (9, 9)),
+    ('overlap', 'median', (5, 5)),
+    ('overlap', 'p50', (5, 5)),
+    ('overlap', 'p99.5', (8.96, 8.96)),
+    # test multiple statistics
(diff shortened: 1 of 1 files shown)


stefsmeets:
Although `iris` uses the same underlying function to calculate the standard deviation, they use a default setting for `ddof=1`, instead of `0`, which is the default in `numpy`. Since the new implementation of the multimodel tests uses the `iris.analysis` package, means that using the `std` operator will return a different value. 

I consider this a bug, because in other parts of `esmvaltool`, `iris.analysis.STD_DEV` is also used.

The corresponding tests have been marked with `xfail`.

```python
In [1]: import numpy as np

In [2]: data = np.ma.array(
    data=[[  1,   1,   1],
          [999,   5,   5],
          [  9,   9, 999]],
    mask=[[False, False, False],
          [ True, False, False],
          [False, False,  True]])

In [3]: # old behaviour

In [4]: np.ma.std(data, axis=0)
Out[4]:
masked_array(data=[4.0, 3.265986323710904, 2.0],
             mask=[False, False, False],
       fill_value=1e+20)

In [5]: import iris.analysis

In [6]: # new behaviour

In [7]: iris.analysis.STD_DEV.aggregate(data, axis=0)  
Out[7]:
masked_array(data=[5.656854249492381, 4.0, 2.8284271247461903],
             mask=[False, False, False],
       fill_value=1e+20)

In [8]: # new behaviour can be reproduced with ddof=1 (default in iris 3.0.1)

In [9]: np.ma.std(data, axis=0, ddof=1)
Out[9]:
masked_array(data=[5.656854249492381, 4.0, 2.8284271247461903],
             mask=[False, False, False],
       fill_value=1e+20)
```

The other failing test (marked with `xfail`) is related to passing cubes with non-overlapping timespans. This returns the same cubes in the current implementation, but raises an error in the new implementation.

tests/unit/preprocessor/_multimodel/test_multimodel.py
```diff
@@ -1,267 +1,314 @@
...
-    def test_get_time_slice_part(self):
-        """Test get time slice if all cubes have data."""
-        cubes = [self.cube1, self.cube2]
-        result = _get_time_slice(cubes, time=14)
-        masked = np.ma.empty(list(cubes[0].shape[1:]))
-        masked.mask = True
-        expected = [self.cube1[0].data, masked]
-        self.assert_array_equal(expected, result)
-
-    def test_return_products(self):
```
Peter9192, line 224: This test shouldn't be discarded as it will still be relevant, also in the new implementation

tests/unit/preprocessor/_multimodel/test_multimodel.py
```diff
@@ -1,267 +1,314 @@
...
-        expected_full_mean.mask[1] = False
-        self.assert_array_equal(comp_full_mean.data, expected_full_mean)
-
-    def test_plev_fix(self):
-        """Test plev fix."""
-        fixed_data = _plev_fix(self.cube2.data, 1)
-        expected_data = np.ma.ones((3, 2, 2))
-        self.assert_array_equal(expected_data, fixed_data)
-
-    def test_unify_time_coordinates(self):
```
Peter9192, line 193: I think we might want to keep (or re-implemen) this test. Eventually I think this will be moved to the time preprocessor tests in https://github.com/ESMValGroup/ESMValCore/issues/744

tests/unit/preprocessor/_multimodel/test_multimodel.py
```diff
@@ -1,267 +1,314 @@
...
+        generate_cube_from_dates(dates2),
+    )
+
+
+@pytest.mark.xfail(reason='Multimodel statistics returns the original cubes.')
+def test_edge_case_time_no_overlap_fail():
+    """Test case when time coords do not overlap using span='overlap'.
+
+    Expected behaviour: `multi_model_statistics` should fail if time
+    points are not overlapping.
```
Peter9192, line 244: I'm not sure if the desired behaviour here is to fail. I would say so if I was designing it from scratch, but maybe this is actually considered a 'feature' by some users/recipes?? Any thoughts on this @valeriupredoi @LisaBock

Peter9192 (changes requested):
Nice job @stefsmeets! I think these modified tests make the requirements for a new implementation much more explicit. As far as I can see, the new tests cover all of the old tests except for two of them. Could you have a look at how we could add those within the new test framework?

schlunma:
Hi guys, great job you're doing here!

I have a question regarding this PR: Should it already include tests for all functionalities that we want the future `multimodel_statistics` preprocessor to have (and mark those that fail right now with an `xfail`)? If no, you can stop reading :smile: 

If yes, do we want want to include tests that ensure that the preprocessor works with cube that have arbitrary coordinates? I think that's a really desired features (#1018 #891 #890). Right now, I think all the tests include the `time` coordinate somehow (please correct me if I'm wrong). I can add some tests to this branch directly if you agree!

Peter9192:
Hi @schlunma yes I think this test-driven development makes a lot of sense! It helps to map all the requirements for a new implementation.

schlunma:
Cool, sounds great! I will add some tests later today.

stefsmeets:
Hi @schlunma , we already have tests for #890 and #891

https://github.com/ESMValGroup/ESMValCore/blob/fee8d93faaefaa7e042401592718065700f08b2d/tests/sample_data/multimodel_statistics/test_multimodel.py#L227)

https://github.com/ESMValGroup/ESMValCore/blob/fee8d93faaefaa7e042401592718065700f08b2d/tests/sample_data/multimodel_statistics/test_multimodel.py#L251

Peter9192:
What's maybe even more important than tests for new desired features, is tests for 'hidden' requirements. We recently saw an example of this in #975. Are there any other 'features' or 'requirements' you can think are currently not covered by the tests?

schlunma:
> Hi @schlunma , we already have tests for #890 and #891

Ahh, nice, so that should be covered then.

> What's maybe even more important than tests for new desired features, is tests for 'hidden' requirements. We recently saw an example of this in #975. Are there any other 'features' or 'requirements' you can think are currently not covered by the tests?

Not really. One thing that recently came up is #1009 (handling of `altitude` coordinate), but once the preprocessor can handle arbitrary coordinates this should not be an issue anymore.

schlunma:
One suggestion: Since we want to support cubes with no time coordinate, I think that the `span` argument should be optional in the future. I really doesn't make sense to specify this when there is no time dimensions.

tests/unit/preprocessor/_multimodel/test_multimodel.py
stefsmeets: I thought this was already covered elsewhere. I re-implemented it and added it back. 👍

tests/unit/preprocessor/_multimodel/test_multimodel.py
stefsmeets: Yep, I added it back. 👍

tests/unit/preprocessor/_multimodel/test_multimodel.py
Peter9192: @schlunma what do you think?

schlunma: My first intuition would also be to let it fail when there is no overlap, but I guess returning an empty cube (or no cube) could also be a reasonable option. I've never used that "feature", and I don't really know if anyone uses it. I would be fine with changing this behavior if it is not used anywhere.

Peter9192 (approved):

tests/unit/preprocessor/_multimodel/test_multimodel.py
Peter9192: Okay, let's leave it like this then 👍

Peter9192:
> 
> 
> One suggestion: Since we want to support cubes with no time coordinate, I think that the `span` argument should be optional in the future. I really doesn't make sense to specify this when there is no time dimensions.

I like the idea, but I'm not sure we'll address this in #968 already. I think it sounds more like a follow up. But we can certainly open an issue about it and maybe already start thinking about a test that belongs to it.

stefsmeets:
Hi @ESMValGroup/esmvaltool-coreteam , I think this PR is ready to be merged!

tests/unit/preprocessor/_multimodel/test_multimodel.py
valeriupredoi: remember the purpose of the test in general is to verify the integrity of the functional behavior, not to assume things on its own, so if the function fails if there is no overlap, the test should check for that :+1:

tests/unit/preprocessor/_multimodel/test_multimodel.py
```diff
@@ -1,256 +1,331 @@
...
+CALENDAR_OPTIONS = ('360_day', '365_day', 'gregorian', 'proleptic_gregorian',
+                    'julian')
+
+
+def assert_array_almost_equal(this, other):
+    """Assert that array `this` almost equals array `other`."""
+    if np.ma.isMaskedArray(this) or np.ma.isMaskedArray(other):
+        np.testing.assert_array_equal(this.mask, other.mask)
+
+    np.testing.assert_allclose(this, other)
```
valeriupredoi, line 27: any reason why you add the same function as the one from the tests' init file here? well, you are adding the allclose check to it, but why do you need it? please use the imported function and if there is a need for an almost equal check, add it in the test that needs it

valeriupredoi (changes requested):
@stefsmeets good job, tests are looking much more comprehensive now! Just a wee change related to the array equality checker func and a comment related to the overlap check from me :beer:

valeriupredoi:
oh and another comment: could you please tell me what's the coverage percentage now compared to the previous module?

valeriupredoi:
> One suggestion: Since we want to support cubes with no time coordinate, I think that the `span` argument should be optional in the future. I really doesn't make sense to specify this when there is no time dimensions.

 we should account for a `nospan` or `notime` case, but making the argument fully optional may lead to user errors when they forget to set it and we'd have to make the function use a default span - there are major differences between, overlap and full, and even if you throw a warning that complains the span parameter was not set, defaulting to blah, users don't read warnings

Peter9192:
> 
> 
> > One suggestion: Since we want to support cubes with no time coordinate, I think that the `span` argument should be optional in the future. I really doesn't make sense to specify this when there is no time dimensions.
> 
> we should account for a `nospan` or `notime` case, but making the argument fully optional may lead to user errors when they forget to set it and we'd have to make the function use a default span - there are major differences between, overlap and full, and even if you throw a warning that complains the span parameter was not set, defaulting to blah, users don't read warnings

One way to do it is to add an additional (not the default) option that says `span=None` if you don't want it. But again, let's discuss that in another issue.

valeriupredoi:
> > > One suggestion: Since we want to support cubes with no time coordinate, I think that the `span` argument should be optional in the future. I really doesn't make sense to specify this when there is no time dimensions.
> > 
> > 
> > we should account for a `nospan` or `notime` case, but making the argument fully optional may lead to user errors when they forget to set it and we'd have to make the function use a default span - there are major differences between, overlap and full, and even if you throw a warning that complains the span parameter was not set, defaulting to blah, users don't read warnings
> 
> One way to do it is to add an additional (not the default) option that says `span=None` if you don't want it. But again, let's discuss that in another issue.

yip, agreed on both counts!

tests/unit/preprocessor/_multimodel/test_multimodel.py
stefsmeets: Hi @valeriupredoi , I'm not entirely sure how to import this file without messing with `sys.path`. The default mechanics for defining imports is in conftest.py, which we do not use atm.

valeriupredoi: no need to import per se it's already imported at module run level, just use it from there without duplicating it here

tests/unit/preprocessor/_multimodel/test_multimodel.py
stefsmeets: Good point @valeriupredoi , the point of this PR is to establish the definition so that we have a better description of the functionality that we want. Currently there is no test that explicitly checks for this scenario, so the behaviour is not defined. 

Therefore we want to clarify the desired behaviour with this PR. Since this return statement is inconsistent with the normal output (a list is returned rather than a dict with statistics), I marked this with XFAIL to indicate a known problem (which will be fixed with #968 😁). 

If this is behaviour that we need with this function, that is also OK for me, then I will adjust the test to ensure that a list of cubes is given back if there is no overlap and update #968 accordingly.

stefsmeets:
> oh and another comment: could you please tell me what's the coverage percentage now compared to the previous module?

For `_multimodel.py`, 
this PR: `statements: 196, missing: 13, coverage: 93%`,
master: `statements: 196, missing: 17, coverage: 91%`

Note that the coverage says very little about the quality of the tests.

valeriupredoi:
shweet! yes, it's a number that measures brute coverage, just wanted to check there are no omissions in statement checks that were done before, this PR introduces a better quality test module :+1:

tests/unit/preprocessor/_multimodel/test_multimodel.py
valeriupredoi: sure, good call! I am perfectly fine with this as it is here, discussing further in the PR is cool, the only bit that I think needs addressing is that array equality func below, other than that - good work here :beer:

tests/unit/preprocessor/_multimodel/test_multimodel.py
stefsmeets: Like you said, the functions are different (`array_allclose` vs `array_equal`), so the other one cannot be used directly. Instead, I changed the function name to better reflect its function. Does that work for you?

valeriupredoi: OK we should move this one upstream in the init at some point and just use it from there

valeriupredoi (approved):

valeriupredoi:
cheers for the good work @stefsmeets :beer:

PR #1032, opened 2021-03-02
Increased flexibility of CMOR checks for datasets with generic alevel coordinates

### LINKED ISSUE #1029: Increase flexibility of CMOR checks for datasets with generic alevel coordinates
Background
---

Some 4D atmospheric variables are defined with a `alevel` instead of a `plev` Z dimension. Example: [`cl`](https://github.com/ESMValGroup/ESMValCore/blob/2966f5448a7aa9e3cc53058b6e2bbe69256c9296/esmvalcore/cmor/tables/cmip6/Tables/CMIP6_Amon.json#L206).

In these datasets, the Z coordinate is time, lat and lon-dependent and can be calcualted from 1D and 3D arrays (see [here](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#_atmosphere_hybrid_sigma_pressure_coordinate)). Saving these 1D and 3D arrays takes much less memory than saving a 4D array. When loading these datasets, `iris` automatically transforms these generic level coordinates into `air_pressure` or `altitude` (depends on the exact type of level coordinate that is used) coordinates via [`iris.aux_factories`](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/aux_factory.html). Thus, something like

```py
cube = iris.load_cube('path/to/data_with_alevel_coordinate.nc')
air_pressure_coord = cube.coord('air_pressure')
```

even thought the original netCDF does **NOT** have an `air_pressure` coordinate is totally viable. Cubes like these behave exactly like cubes with an `air_pressure` coordinate defined. Moreover, diagnostic scripts can assume that these datasets have these derived coordinate, which is also vertical coordinate that is most relevant from a scientific point of view.

For some observational dataset, e.g. ERA5, it is be the case that these 4D variables are defined with a default (1D) `air_pressure` coordinate ( `plev`). This causes datasets like this to fail during our CMOR checks, since the tool expects a generic level coordinate. While this makes sense from a technical point of view, it does not make sense from a scientific point of view. After loading both cubes will have `air_pressure` coordinates, with the only exception that in the first case this coordinate is 4D and in the second case it is 1D. One obvious solution would be to create a generic level coordinate in the CMORizer, but this is not reasonable and unnecessary work.

Feature request
---

Would it be possible to let datasets with "non-generic" level coordinates pass our CMOR checks if the original variable is defined with a generic level coordinate? As I explained above, the generic level coordinates are actually only an intermediate step in the creation of a "non-generic" coordinate. This would only affect few predefined combinations, e.g. for the generic level coordinate `alevel`, the coordinates `air_pressure` (`plev`) or `altitude` (`alt`) would be allowed alternatives. This probably also applies to ocean coordinates, but I don't know the details here.

This feature would simplify the handling of these datasets a lot!

schlunma:
This PR proposes a possible solution for #1029.

-   Closes #1029
-   Link to documentation:

***

## Before you get started

## Checklist

***


valeriupredoi:
@schlunma good pointer to this during the ESMValTool telecon today, I am very much for a more flexible cmor check strategy (the idea about it as it is implemented now was mine after all, always been pro-flexible towards cmor checks) - I think we need to also include a table in the documentation [here](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/develop/fixing_data.html#customizing-checker-strictness) where we list (not exhaustively though) the cases where errors or warning etc get returned :+1:

sloosvel:
@schlunma maybe this can be marked as ready for review now, I tested a recipe with multiple datasets and things seemed to work as expected. Let me know how it works for you.

schlunma:
@sloosvel Awesome, thanks for testing! :tada:

Let me add some unit tests for it (but probably only in two weeks; I don't know if I manage today and next week I'm on vacation), then I will mark it as ready for review!

esmvalcore/cmor/check.py
```diff
@@ -464,17 +464,50 @@ def _check_generic_level_dim_names(self, key, coordinate):
...
+                        f'{out_name} has wrong standard_name or is not set.')
                 else:
-                    self.report_critical(self._does_msg, key, 'exist')
+                    self._check_alternative_dim_names(key)
+
+    def _check_alternative_dim_names(self, key):
+        """Check for viable alternatives to generic level coordinates."""
+        alternative_coord = None
+        allowed_alternatives = {
+            'alevel': ['alt40', 'plev19', 'plevs'],
```
zklaus, line 476: I think it would be helpful to indicate (probably in a comment) where this information is coming from. It seems to somehow come from CMIP6 tables, but then, there is no mention of `plevs` in the tables, and on the other hand one might wonder why not mention `alt16` or the other `plevXX`.

schlunma: Yes, very good point! I will add an explanation of that. `plevs` comes from the CMIP5 table, btw.

schlunma:
This PR is ready for review now!

schlunma:
Is there a chance that this can go into 2.4? I can only work on this today and tomorrow. Afterwards I will be on vacation and will not be back before the core feature freeze.

sloosvel:
I can check tomorrow then, it look good to me back then so I guess it should be fine now. But I'll look with more detail tomorrow

esmvalcore/cmor/check.py
```diff
@@ -464,17 +464,87 @@ def _check_generic_level_dim_names(self, key, coordinate):
...
+        level coordinates. Right now, only alternative level coordinates for
+        the atmosphere ('alevel' or 'zlevel') are supported.
+
+        Note that only the "simplest" CMOR table entry per coordinate is
+        specified (e.g., only 'plev3' for the pressure level coordinate and
+        'alt16' for the altitude coordinate). These different versions (e.g.,
+        'plev3', 'plev19', 'plev39', etc.) only differ in the requested values.
+        We are mainly interested in the metadata of the coordinates (names,
+        units), which is equal for all coordinate versions. In the DEFAULT
+        strictness or lower, differing requested values only produce a warning.
```
sloosvel, line 502: So this could print  "fake" warnings? I get that 90% of people don't read them, but I still think it would be a bit confusing if you get warnings because you are comparing right or wrong requested values of plev19 to plev3. There are a few plev coords, but it's not like you have thousands of options either. Maybe you could use the `len` of the depth coord to choose which plevX could be checked against to.

schlunma: Yes, there could be "fake" warnings. The problem here is that most datasets that need this fix are observational datasets, and they don't really care about `plev19`, `plev3`, etc. at all, so even checking the `len` of the coordinate would be useful. Would it be an alternative to add an argument `check_requested` to `_check_coord` and set it to `False` in this specific case?

sloosvel: I guess that would the quick solution. Or print a message that requested values will not be checked.

A more elaborate option could be to try to work with the `cmor_strict` parameter in the config developer. If `True` like in CMIP projects, check the requested values. If `False` like in OBS project, skip the check. But it's quite the work, so maybe it's something that could be considered for the future.

schlunma: I will add a message and a note to the code that this might be expanded in the future

schlunma:
Any further comments on this?

sloosvel:
Not from my side

sloosvel (approved):

PR #1034, opened 2021-03-05
Add option for regional regridding to regrid preprocessor

### LINKED ISSUE #493: Regional regridding
The current `regrid` preprocessor function only allows global grids, e.g. "1x1" degrees. It would be nice to have a more flexible preprocessor function that allows you to define a regional grid.

Maybe `regrid_region` with arguments `latitude` and `longitude`, which take 3 numbers as input, start, stop and step (or number of steps)?

Alternatively, we could combine the regrid and extract_region functions, so we first create the grid with the "1x1" syntax and then feed that to `extract_region` preprocessor before regridding? This could be implemented by just supplying the arguments of `extract_region` to `regrid`.

Opinions anyone? @mattiarighi @zklaus ?

stefsmeets:
## Description

Hi everyone, this PR implements new arguments to the regrid preprocessor so that a regional target grid can be specified directly. 

To use:

```yaml
    regrid:
      target_grid: 
         lon_start: 40
         lon_end: 60
         lon_step: 2
         lat_start: -10
         lat_end: 30
         lat_step: 2
      scheme: linear
```

-   Closes #493 
-   Link to api reference: https://esmvaltool--1034.org.readthedocs.build/projects/ESMValCore/en/1034/api/esmvalcore.preprocessor.html#esmvalcore.preprocessor.regrid 
- Link to documentation: https://esmvaltool--1034.org.readthedocs.build/projects/ESMValCore/en/1034/recipe/preprocessor.html#regridding-on-regional-target-grid-specification

***

## Before you get started

## Checklist

***


stefsmeets:
Hi @Peter9192 @SarahAlidoost , would you be willing to review this PR?

esmvalcore/preprocessor/_regrid.py
```diff
@@ -104,9 +104,51 @@ def parse_cell_spec(spec):
     return dlon, dlat
 
 
-def _stock_cube(spec, lat_offset=True, lon_offset=True):
+def _stock_cube(latdata, londata, circular: bool = False):
```
valeriupredoi, line 107: probably better to rename to eg `_regional_stock_cube` and below to `_global_stock_cube`

esmvalcore/preprocessor/_regrid.py
```diff
@@ -151,25 +192,95 @@ def _stock_cube(spec, lat_offset=True, lon_offset=True):
...
+    lat_start : float
+        Latitude value of the first grid cell center. The grid includes this
+        value.
+    lat_end : float
+        Maximum latitude value of the last grid cell. The grid includes this
+        value only if it falls on a grid point. Otherwise, it cuts off at the
+        first previous value.
+    lat_step : float
+        Latitude spacing between grid points, i.e. the distance between two
+        adjecent values.
```
valeriupredoi, line 221: is this the distance between the centers of two neighboring cells? if so, probably better to say it that way

esmvalcore/preprocessor/_regrid.py
```diff
@@ -151,25 +192,95 @@ def _stock_cube(spec, lat_offset=True, lon_offset=True):
...
+
+    Returns
+    -------
+    xvals : np.array
+        List of longitudes
+    yvals : np.array
+        List of latitudes
+    """
+    if lat_step == 0:
+        raise ValueError('Latitude step must be larger than 0, '
```
valeriupredoi, line 241: is there a reason as to why adding a default step might help here? Can we add a default step maybe based on the original grid spec?

esmvalcore/preprocessor/_regrid.py
```diff
@@ -151,25 +192,95 @@ def _stock_cube(spec, lat_offset=True, lon_offset=True):
...
+    if (lat_start < _LAT_MIN) or (lat_end > _LAT_MAX):
+        raise ValueError(
+            f'Latitude values must lie between {_LAT_MIN}:{_LAT_MAX}, '
+            f'got lat_start={lat_start}:lat_end={lat_end}.')
+
+    def get_points(start, stop, step):
+        """Calculate grid points."""
+        num = int((stop - start) // step)
+        stop = start + num * step
+        return np.linspace(start, stop, num + 1)
```
valeriupredoi, line 257: don't like functions inside functions, can we take this outside?

esmvalcore/preprocessor/_regrid.py
```diff
@@ -151,25 +192,95 @@ def _stock_cube(spec, lat_offset=True, lon_offset=True):
...
+
+    latitudes = get_points(lat_start, lat_end, lat_step)
+    longitudes = get_points(lon_start, lon_end, lon_step)
+
+    return latitudes, longitudes
+
+
+def _stock_regional_cube(spec: dict):
```
valeriupredoi, line 265: ok this is getting confusing - so we have stock_cube, stock_global_cube and now stock_regional_cube...

esmvalcore/preprocessor/_regrid.py
```diff
@@ -260,42 +367,61 @@ def extract_point(cube, latitude, longitude, scheme):
...
+    """Perform horizontal regridding.
+
+    Note that the target grid can be a cube (:py:class:`~iris.cube.Cube`),
+    path to a cube (``str``), a grid spec (``str``) in the form
+    of `MxN`, or a ``dict`` specifying the target grid.
+
+    For the latter, the ``target_grid`` should be a ``dict`` with the
+    following keys:
+
+    - ``lon_start`` is the longitude of the first grid cell.
```
valeriupredoi, line 379: I'd go with "longitude value at the center of the first grid cell" so you can explain what the step is as the "distance between grid cell centers" - I find it more intuitive this way

esmvalcore/preprocessor/_regrid.py
```diff
@@ -151,25 +192,95 @@ def _stock_cube(spec, lat_offset=True, lon_offset=True):
...
+
+    Parameters
+    ----------
+    lat_start : float
+        Latitude value of the first grid cell center. The grid includes this
+        value.
+    lat_end : float
+        Maximum latitude value of the last grid cell. The grid includes this
+        value only if it falls on a grid point. Otherwise, it cuts off at the
+        first previous value.
```
valeriupredoi, line 218: please tell the user which one is the start and which one is the end one ie (-20, -90) or is it (-90, -20) ie again is it abs(max) or max

valeriupredoi (changes requested):
nice work @stefsmeets :+1: :beer: A few minor notes on mostlydocstringing around, but the main thing I think this PR still needs is actual documentation - will you please add that?

valeriupredoi:
apologies for crashing the review party btw :laughing:

stefsmeets:
Hi @valeriupredoi , no worries about crashing the party 😅 Thanks for the feedback!

I want to clarify what you mean by 'adding actual documentation'. I have completed the docstrings, which will appear in readthedocs. Is this not enough?

valeriupredoi:
> Hi @valeriupredoi , no worries about crashing the party sweat_smile Thanks for the feedback!
> 
> I want to clarify what you mean by 'adding actual documentation'. I have completed the docstrings, which will appear in readthedocs. Is this not enough?

that's cool but we should have a decent explanation of this functionality in the actual sphinx documentation, in [here](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/preprocessor.html#horizontal-regridding) where we introduce the user to all manners of horizontal regridding :+1:

esmvalcore/preprocessor/_regrid.py
stefsmeets: Thanks for pointing this out. The function names were a bit of a poor choice from my side. I have updated them to make more sense 👍

esmvalcore/preprocessor/_regrid.py
stefsmeets: See above

esmvalcore/preprocessor/_regrid.py
stefsmeets: Yep, this could be worded better. I have updated the text 👍

esmvalcore/preprocessor/_regrid.py
stefsmeets: Will do 👍

esmvalcore/preprocessor/_regrid.py
valeriupredoi: cheers :beer:

esmvalcore/preprocessor/_regrid.py
valeriupredoi: all in good time, we can't both code and be Jane Austens at the same time :laughing:

esmvalcore/preprocessor/_regrid.py
stefsmeets: Normally, I would agree with you, but this function is very small and only relevant in this context that I'd prefer to keep it here for clarity. I did it this way to prevent repeating the same code twice for latitude/longitude.

esmvalcore/preprocessor/_regrid.py
stefsmeets: I check for `lat_step==0` to guard against a `ZeroDivisionError` down the line

stefsmeets:
Hi @valeriupredoi , I addressed your comments, could you let me know if you are happy with the changes? 🥂

SarahAlidoost (changes requested):
@stefsmeets to review this PR, I ran the [recipe_lisflood.yml](https://github.com/ESMValGroup/ESMValTool/blob/master/esmvaltool/recipes/hydrology/recipe_lisflood.yml) by changing the preprocessors section as:

```yml
preprocessors:
  general: &general
    custom_order: true
    regrid:
      target_grid: 
         lon_start: 0
         lon_end: 9
         lon_step: 0.1
         lat_start: 45
         lat_end: 54
         lat_step: 0.1
      lon_offset: true
      lat_offset: true
      scheme: linear
```

It results in an error as:

```
  File "miniconda3/envs/esmvaltool/lib/python3.8/site-packages/esmvalcore/preprocessor/_regrid.py", line 85, in parse_cell_spec
    cell_match = _CELL_SPEC.match(spec)
TypeError: expected string or bytes-like object
```

Another issue is related to the naming of the keys i.e. 
       
```yml
lon_start: 0
lon_end: 9
lon_step: 0.1
lat_start: 45
lat_end: 54
lat_step: 0.1
```

Can we use the names that are used in the [extract_region](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/preprocessor.html#extract-region) processor i.e. 

```yml
start_longitude: 
end_longitude: 
start_latitude: 
end_latitude: 
```

These are my comments for now. I will re-run the recipe when the error is fixed. Thank you.

valeriupredoi (approved):
great, cheers for the changes @stefsmeets - looks tiptop to me now :beer:

stefsmeets:
Hi @SarahAlidoost , thanks for the review!

> ```yaml
>       lon_offset: true
>       lat_offset: true
> ```

You are getting this error, because `lon_offset` and `lat_offset` expect a float or int. This is unrelated to this PR, because these variables are not used.

```yaml
start_longitude: 0
end_longitude: 9
step_longitude: 0.1
start_latitude: 45
end_latitude: 54
step_latitude: 0.1
```

Would you prefer this? Personally, I think these variable names are too long and difficult to parse and type, which is why I opted for the shorter versions.

SarahAlidoost:
> Hi @SarahAlidoost , thanks for the review!
> 
> I will have a look at the error, because it should not be triggered. EDIT: did you checkout the correct branch? Because I cannot reproduce the error.

my bad, I noticed that the development installation didn't include the changes in this PR. Now, the recipe run is successful including `lon_offset` and `lat_offset`, no need to remove these keys. I compared the output of the recipe using this PR with those using the master branch, as an example please see this figure
lisflood_diff 
I couldn't find a reason for the differences in the pixels that are extracted by the polygon. It seems that the differences are more related to `extract_shape` rather than `regrid`.

> 
> ```yaml
> start_longitude: 0
> end_longitude: 9
> step_longitude: 0.1
> start_latitude: 45
> end_latitude: 54
> step_latitude: 0.1
> ```
> 
> Would you prefer this? Personally, I think these variable names are too long and difficult to parse and type, which is why I opted for the shorter versions. That said, I don't have a strong opinion on what it should be (as I'm not going to be the one who has to type it a million times ).

Yes, many thanks. My reason is being consistent with what is already there i.e. names in `extract_region` processor. If some developers use other names for `start_longitude` in the future for example start_long, long_start, start_lon, or longitude_start, eventually, we will have an esmvaltool that uses several different names for one thing. I was trying to avoid this here. :-)

stefsmeets:
Thanks for the review @SarahAlidoost , and good to see that the output looks the same for you! The implementation in this PR calls `cube.regrid` directly (which includes some degree of interpolation), whereas `extract_region` performs an intersect or a mask depending on the input data. This may give small differences.

Would you mind approving the PR, sot that it can be merged?

SarahAlidoost:
> Thanks for the review @SarahAlidoost , and good to see that the output looks the same for you! The implementation in this PR calls `cube.regrid` directly (which includes some degree of interpolation), whereas `extract_region` performs an intersect or a mask depending on the input data. This may give small differences.
> 
> Would you mind approving the PR, sot that it can be merged?

Thank you for your explanations. I run a simple recipe without `extract_shape` processor.  The recipe run using this pull request is much faster and uses less memory than the master branch. However, looking at the results, it seems that the end longitude and latitude points are missing. This is due to the function `get_points`, as an example, I run this function with these values:

```yml
start_longitude: 0
end_longitude: 9
start_latitude: 45
end_latitude: 54
```

```python
get_points(start_latitude, end_latitude, step_latitude):
array([45. , 45.1, 45.2, 45.3, 45.4, 45.5, 45.6, 45.7, 45.8, 45.9, 46. ,
       46.1, 46.2, 46.3, 46.4, 46.5, 46.6, 46.7, 46.8, 46.9, 47. , 47.1,
       47.2, 47.3, 47.4, 47.5, 47.6, 47.7, 47.8, 47.9, 48. , 48.1, 48.2,
       48.3, 48.4, 48.5, 48.6, 48.7, 48.8, 48.9, 49. , 49.1, 49.2, 49.3,
       49.4, 49.5, 49.6, 49.7, 49.8, 49.9, 50. , 50.1, 50.2, 50.3, 50.4,
       50.5, 50.6, 50.7, 50.8, 50.9, 51. , 51.1, 51.2, 51.3, 51.4, 51.5,
       51.6, 51.7, 51.8, 51.9, 52. , 52.1, 52.2, 52.3, 52.4, 52.5, 52.6,
       52.7, 52.8, 52.9, 53. , 53.1, 53.2, 53.3, 53.4, 53.5, 53.6, 53.7,
       53.8, 53.9])

```
The result of  `get_points` does not include the `end_latitude: 54`. It is the same for longitude values. Let's have a look at the output together. ;-)

stefsmeets:
> The result of `get_points` does not include the `end_latitude: 54`. It is the same for longitude values. Let's have a look at the output together. ;-)

Good find, that is not supposed to happen. It is supposed to include this grid point if it falls on the 'end_latitude'.

stefsmeets:
Ok, so the problem is with the target grid step of `0.1`, which is represented as  `Decimal('0.1000000000000000055511151231257827021181583404541015625')`, which caused the number of steps to round down to 89 instead of 90. It is fixed in the latest commit 👍

SarahAlidoost:
> Ok, so the problem is with the target grid step of `0.1`, which is represented as `Decimal('0.1000000000000000055511151231257827021181583404541015625')`, which caused the number of steps to round down to 89 instead of 90. It is fixed (and tested for) in the latest commit

Thanks, I re-ran the recipe and the problem is now fixed.

SarahAlidoost (approved):

stefsmeets:
Hi @ESMValGroup/esmvaltool-coreteam , this PR is ready to be merged! 🚀

valeriupredoi:
great! @stefsmeets could you please do the following before I merge: run the tests locally (on your machine) - we've found an issue with flake8 not running on the CI and that may lead to fails that get seen only after the merge and fix the Codacy issues (or as much of them as you can, certainly two of them can be fixed with zero effort). Cheers :beer:

stefsmeets:
> great! @stefsmeets could you please do the following before I merge: run the tests locally (on your machine) - we've found an issue with flake8 not running on the CI and that may lead to fails that get seen only after the merge and fix the Codacy issues (or as much of them as you can, certainly two of them can be fixed with zero effort). Cheers 🍺

I'm not getting these on my local machine, must be some Codacy caching thing going on (notice that the line numbers do not even make sense).

nielsdrost:
Thanks @stefsmeets for merging in master so the Flake8 checks run.

Note that Codacy misreports 3 existing errors as new.

Peter9192:
I tried using this feature with decimal start/stop latitude/longitude values, but I get 
```
  File "/home/peter/ESMValCore/esmvalcore/preprocessor/_regrid.py", line 257, in get_points
    num = int((stop - start) // Decimal(str(step)))
TypeError: unsupported operand type(s) for //: 'float' and 'decimal.Decimal'
```

stefsmeets:
> 
> 
> I tried using this feature with decimal start/stop latitude/longitude values, but I get
> 
> ```
>   File "/home/peter/ESMValCore/esmvalcore/preprocessor/_regrid.py", line 257, in get_points
>     num = int((stop - start) // Decimal(str(step)))
> TypeError: unsupported operand type(s) for //: 'float' and 'decimal.Decimal'
> ```

That's odd, works on my pc, python 3.8:

```python
In [3]: 6 // decimal.Decimal(str('2'))
Out[3]: Decimal('3')
```

Which version are you on? Can you open an issue?

Peter9192:
```
In [3]: 6.5 // decimal.Decimal(str('2'))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-9-a29fc17ef4fc> in <module>
----> 1 6.5 // decimal.Decimal(str('2'))

TypeError: unsupported operand type(s) for //: 'float' and 'decimal.Decimal'

```

stefsmeets:
Missed the float part, also fails here 👍

Peter9192:
I think the closing bracket for `int` should come before the `//`, could it be that this was misplaced accidentally?

stefsmeets:
No, `num` must be an integer to calculate the correct endpoint.

Peter9192:
so you need 2 ints.... 😕

stefsmeets:
Or a smarter way to calculate the grid spacing 😅

PR #1037, opened 2021-03-08
Move _get_time_bounds from preprocessor._time to cmor.check to avoid circular import with cmor module

### LINKED ISSUE #1036: Conda build broken
The conda build is broken because the [tests fail](https://app.circleci.com/pipelines/github/ESMValGroup/ESMValCore/4401/workflows/e663b52a-28a5-44f0-bf04-5b7495d28ef6/jobs/20571):
```
  File "/opt/conda/conda-bld/esmvalcore_1615163441366/test_tmp/run_test.py", line 8, in <module>
    import esmvalcore.cmor.check
  File "/opt/conda/conda-bld/esmvalcore_1615163441366/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/python3.9/site-packages/esmvalcore/cmor/check.py", line 12, in <module>
    from esmvalcore.preprocessor._time import _get_time_bounds
  File "/opt/conda/conda-bld/esmvalcore_1615163441366/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/python3.9/site-packages/esmvalcore/preprocessor/__init__.py", line 11, in <module>
    from ..cmor.check import cmor_check_data, cmor_check_metadata
ImportError: cannot import name 'cmor_check_data' from partially initialized module 'esmvalcore.cmor.check' (most likely due to a circular import) (/opt/conda/conda-bld/esmvalcore_1615163441366/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/python3.9/site-packages/esmvalcore/cmor/check.py)
Tests failed for esmvalcore-2.2.0-py_0.tar.bz2 - moving package to /opt/conda/conda-bld/broken
WARNING:conda_build.build:Tests failed for esmvalcore-2.2.0-py_0.tar.bz2 - moving package to /opt/conda/conda-bld/broken
WARNING conda_build.build:tests_failed(2955): Tests failed for esmvalcore-2.2.0-py_0.tar.bz2 - moving package to /opt/conda/conda-bld/broken
TESTS FAILED: esmvalcore-2.2.0-py_0.tar.bz2
```

The build started failing after #849 was merged.

valeriupredoi:
Closes #1036 

as a thumb rule, we should really not use variable names like `time` as well as we should make a proper `__init__.py` file for the `cmor` module - this fix here is not elegant at all but until we de-convolute `cmor` module from `preprocessor` modue via a good `__init__` file I can't see another solution

### DIFF
--- a/esmvalcore/cmor/check.py
+++ b/esmvalcore/cmor/check.py
@@ -1,4 +1,5 @@
 """Module for checking iris cubes against their CMOR definitions."""
+import datetime
 import logging
 from enum import IntEnum
 
@@ -22,6 +23,47 @@ CheckLevels = IntEnum('CheckLevels', 'DEBUG STRICT DEFAULT RELAXED IGNORE')
    - RELAXED: Fail if cubes present severe discrepancies with CMOR standards.
    - IGNORE: Do not fail for any discrepancy with CMOR standards.
 """
+
+
+def _get_next_month(month, year):
+    if month != 12:
+        return month + 1, year
+    return 1, year + 1
+
+
+def _get_time_bounds(time, freq):
+    bounds = []
+    for step, point in enumerate(time.points):
+        month = time.cell(step).point.month
+        year = time.cell(step).point.year
+        if freq in ['mon', 'mo']:
+            next_month, next_year = _get_next_month(month, year)
+            min_bound = time.units.date2num(
+                datetime.datetime(year, month, 1, 0, 0))
+            max_bound = time.units.date2num(
+                datetime.datetime(next_year, next_month, 1, 0, 0))
+        elif freq == 'yr':
+            min_bound = time.units.date2num(
+                datetime.datetime(year, 1, 1, 0, 0))
+            max_bound = time.units.date2num(
+                datetime.datetime(year + 1, 1, 1, 0, 0))
+        elif freq == 'dec':
+            min_bound = time.units.date2num(
+                datetime.datetime(year, 1, 1, 0, 0))
+            max_bound = time.units.date2num(
+                datetime.datetime(year + 10, 1, 1, 0, 0))
+        else:
+            delta = {
+                'day': 12 / 24,
+                '6hr': 3 / 24,
+                '3hr': 1.5 / 24,
+                '1hr': 0.5 / 24,
+            }
+            min_bound = point - delta[freq]
+            max_bound = point + delta[freq]
+        bounds.append([min_bound, max_bound])
+
+    return np.array(bounds)
 
 
 class CMORCheckError(Exception):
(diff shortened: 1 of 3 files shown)


esmvalcore/cmor/check.py
```diff
@@ -509,6 +508,7 @@ def _check_coord_bounds(self, cmor, coord, var_name):
                     coord.var_name, var_name)
 
     def _check_time_bounds(self, freq, time):
+        from esmvalcore.preprocessor._time import _get_time_bounds
```
bouweandela, line 511: Would it be possible to just solve the problem instead of this hacky workaround? I think if you just move the function _get_time_bounds to the the file esmvalcore/cmor/check.py and import it in the preprocessor, things should work just fine. The preprocessor depends on the cmor check module, there should be no need to have the reverse dependency.

Could you also add a unit tests so we do not find out that stuff broke only when building the conda package next time?

bouweandela (changes requested):

bouweandela:
Thanks for looking into this @valeriupredoi!

esmvalcore/cmor/check.py
valeriupredoi: okay. I thought about that but that func is not really a check function but rather a preprocessor func - the test is there, the test for check.py but for some reason is not run in the test suite? Anyway will have a look now

valeriupredoi: ok I moved around the `_get_time_bounds()` func, even though I don't like it this way either, but heyho. How do you write a test to test for a circular import? Got no clue, the previous case was a function imported in preprocessor from check that was in turn importing the preprocessor, as long as we stay away from `from esmvalcore... import ...` inside any of the cmor modules we're grand

bouweandela: > How do you write a test to test for a circular import? 

Just import the module and see if it works.

`import esmvalcore.cmor.check`

raised the circular import exception, so if you add a test that does that, I think we should be good to go. Maybe create a new file and call it `tests/unit/test_cmor_api.py` and have a test there that does the imports of modules documented in the public API? https://docs.esmvaltool.org/projects/esmvalcore/en/latest/api/esmvalcore.cmor.html

valeriupredoi: that is a good idea! I added that and added top-level imports for all the cmor modules but fine-grained imports only for check module - you reckon that'll do for now?

bouweandela (approved):

PR #1043, opened 2021-03-12
Test flake8 in test and develop CI tests to catch flake8 fails before reaching master

valeriupredoi:
Analogous to @bouweandela 's fix in https://github.com/ESMValGroup/ESMValTool/pull/2070


valeriupredoi:
OK this is useless - just seen that the flake8 issue gets nicely picked up by the `test` test without any further massage. Am closing it so one less for @bouweandela to revew :grin:

PR #1051, opened 2021-03-17
Fix 'positive' attribute for rsns/rlns variables

lukasbrunner:
## Description

As per the discussion in #1050 the custom CMOR file for the variables rsns and rlns (Surface Net **downward** Shortwave/Longwave Radiation) contain a wrong value in the field `positive: up`. This pull request corrects them to `down.` 

-   Closes  #1050
-   Link to documentation:

***

## Before you get started

-   [x] [☝ Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do

## Checklist

-   [x] PR has a descriptive title for the [changelog](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#branches-pull-requests-and-code-review)
-   [x] Labels are assigned so they can be used in the [changelog](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#branches-pull-requests-and-code-review)
-   [x] Code follows the [style guide](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#code-style)
-   [x] [Circle/CI tests pass](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#branches-pull-requests-and-code-review)
-   [x] [Codacy code quality checks pass](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#branches-pull-requests-and-code-review)
-   [x] [Documentation builds successfully](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#branches-pull-requests-and-code-review) on readthedocs
-   [x] [Unit tests](https://docs.esmvaltool.org/projects/esmvalcore/projects/esmvalcore/en/latest/contributing.html#contributing-to-the-esmvalcore-package) are available

***

PR #1052, opened 2021-03-17
Add fixes for the net & up radiation variables from ERA5

### LINKED ISSUE #1050: Using derived ERA5 data
Question for the community:

I'm currently developing a recipe which is comparing the CMIP6 variable rsus (long_name: Surface Upwelling Shortwave Radiation) with data from ERA5. This variable does not exist in ERA5 so in a first step I did:
`cdo setpartabn,parameter_table -sub surface_solar_radiation_downwards/era5_surface_solar_radiation_downwards_1979_monthly.nc surface_net_solar_radiation/era5_surface_net_solar_radiation_1979_monthly.nc surface_solar_radiation_upwards/era5_surface_solar_radiation_upwards_1979_monthly.nc`
(where setpartabn renames the variable)
This did not work because the preprocessor could not make units fit (ERA5 is in J m-2 and CMIP6 is in W m-2). NOTE: for the native ERA5 radiation files (like surface_solar_radiation_downwards and the CMIP6 equivalent rsds) this conversion works so I'm assuming this could be added somewhere. 

**Question 1: Is it possible to make the preprocessor aware of new ERA5 variables like in this case?**

For now I have solved this problem by converting the unit offline like:
`cdo -b F32 setpartabn,parameter_table -divc,86400 -sub surface_solar_radiation_downwards/era5_surface_solar_radiation_downwards_1979_monthly.nc surface_net_solar_radiation/era5_surface_net_solar_radiation_1979_monthly.nc surface_solar_radiation_upwards/era5_surface_solar_radiation_upwards_1979_monthly.nc`
(where setpartabn renames the variable and now also the unit)
This works but I wonder if this is the way it should be done...

**Question 2: Is there a protocol to add new observations to ESMValTool and how does the code check (ESMValBot) work in this case? Is there need/way to upload the data as well?**

lukasbrunner:
## Description

As discussed in #1050 we need some additional radiation data from ERA5 (namely rsus, rlus, rsns, rlns). This pull request adds fixes to these variables equivalent to the existing fixes for rsds and rlds. 

Additional information: radiation upward variables are not available for ERA5. We discussed several ideas but have not really come up with a solution how to fix this on the fly (e.g. https://github.com/ESMValGroup/ESMValCore/issues/1050#issuecomment-800271692). Fixing it in the diagnostic (as I suggested here: https://github.com/ESMValGroup/ESMValCore/issues/1050#issuecomment-800226289) does not work because some of the calculations are not linear (e.g., there is a difference if we do std(rsus) or std(rsds) - std(rsns)). 
Therefore I now calculate the upwards variables offline as a first fix. 

-   Closes #1050  
-   Link to documentation:

***

## Before you get started

## Checklist

***

### DIFF
--- a/esmvalcore/cmor/_fixes/native6/era5.py
+++ b/esmvalcore/cmor/_fixes/native6/era5.py
@@ -181,6 +181,30 @@ class Rlds(Fix):
         return cubes
 
 
+class Rlns(Fix):
+    """Fixes for Rlns."""
+    def fix_metadata(self, cubes):
+        """Fix metadata."""
+        for cube in cubes:
+            fix_hourly_time_coordinate(cube)
+            fix_accumulated_units(cube)
+            cube.attributes['positive'] = 'down'
+
+        return cubes
+
+
+class Rlus(Fix):
+    """Fixes for Rlds."""
+    def fix_metadata(self, cubes):
+        """Fix metadata."""
+        for cube in cubes:
+            fix_hourly_time_coordinate(cube)
+            fix_accumulated_units(cube)
+            cube.attributes['positive'] = 'up'
+
+        return cubes
+
+
 class Rls(Fix):
     """Fixes for Rls."""
     def fix_metadata(self, cubes):
@@ -200,6 +224,30 @@ class Rsds(Fix):
             fix_hourly_time_coordinate(cube)
             fix_accumulated_units(cube)
             cube.attributes['positive'] = 'down'
+
+        return cubes
+
+
+class Rsns(Fix):
+    """Fixes for Rsns."""
+    def fix_metadata(self, cubes):
+        """Fix metadata."""
+        for cube in cubes:
+            fix_hourly_time_coordinate(cube)
+            fix_accumulated_units(cube)
+            cube.attributes['positive'] = 'down'
+
+        return cubes
+
+
+class Rsus(Fix):
+    """Fixes for Rsus."""
+    def fix_metadata(self, cubes):
+        """Fix metadata."""
+        for cube in cubes:
+            fix_hourly_time_coordinate(cube)
+            fix_accumulated_units(cube)
+            cube.attributes['positive'] = 'up'
 
         return cubes
 
--- a/esmvalcore/cmor/_fixes/native6/era5.py
+++ b/esmvalcore/cmor/_fixes/native6/era5.py
@@ -194,7 +194,7 @@ class Rlns(Fix):
 
 
 class Rlus(Fix):
-    """Fixes for Rlds."""
+    """Fixes for Rlus."""
     def fix_metadata(self, cubes):
         """Fix metadata."""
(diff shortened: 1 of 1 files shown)


esmvalcore/cmor/_fixes/native6/era5.py
```diff
@@ -204,6 +228,30 @@ def fix_metadata(self, cubes):
...
+
+
+class Rsus(Fix):
+    """Fixes for Rsus."""
+    def fix_metadata(self, cubes):
+        """Fix metadata."""
+        for cube in cubes:
+            fix_hourly_time_coordinate(cube)
+            fix_accumulated_units(cube)
+            cube.attributes['positive'] = 'up'
```
valeriupredoi, line 250: I think it's best we change these (the positive attribute) directly in the custom CMOR tables, could you do that please? Cheers for cleaning up my old mistakes :beer:

valeriupredoi (changes requested):

esmvalcore/cmor/_fixes/native6/era5.py
lukasbrunner: I'm not sure what you mean by that (sorry I'm a bit out of my depth here - first time working with the core). Setting the attribute here is not necessary, is that what you are saying? I just copied the exact syntax from the other fixes. So do you want me to remove all of the lines where attributes are set?

bouweandela:
> Fixing it in the diagnostic (as I suggested here: #1050 (comment)) does not work because some of the calculations are not linear (e.g., there is a difference if we do std(rsus) or std(rsds) - std(rsns)).

If this doesn't work, the only remaining solution would be to implement it as a derived variable. Go for it!

esmvalcore/cmor/_fixes/native6/era5.py
bouweandela: That's done in #1051

valeriupredoi: @lukasbrunner sorry for being a bit too technical, mate - yeah, the changes I was suggesting were to be done straight in the custom CMOR tables (we have CMOR tables that are standard for the standard variables, and those are the PCMDI CMOR tables, and we have custom ones that we construct to include all these types of non-standard variables, and the ones for the rsus etc variables were written by me when I was young and innocent and made those attribute mistakes) :+1:

valeriupredoi (approved):
cool! cheers @lukasbrunner :beer:

valeriupredoi:
@lukasbrunner could you pls address the codacy complaints (add those pesty blank lines) :beer:

esmvalcore/cmor/_fixes/native6/era5.py
lukasbrunner: Right thanks for explaining @valeriupredoi  :) I may have another one for you: #1054

PR #1053, opened 2021-03-18
Add derived variable rsus

### LINKED ISSUE #1050: Using derived ERA5 data

lukasbrunner:
## Description

Add the option to derive the variable rsus (long_name: Surface Upwelling Shortwave Radiation) for ERA5 where it is not present as discussed in #1050. This is exactly equivalent to existing derivations for, e.g., rsns. 

-   Closes #1050
-   Link to documentation:

***

## Before you get started

## Checklist

***

### DIFF
--- a/esmvalcore/preprocessor/_derive/rsus.py
+++ b/esmvalcore/preprocessor/_derive/rsus.py
@@ -1,0 +1,38 @@
+"""Derivation of variable `rsus`.
+
+authors:
+    - lukas_brunner
+
+"""
+from iris import Constraint
+
+from ._baseclass import DerivedVariableBase
+
+
+class DerivedVariable(DerivedVariableBase):
+    """Derivation of variable `rsus`."""
+
+    @staticmethod
+    def required(project):
+        """Declare the variables needed for derivation."""
+        required = [
+            {
+                'short_name': 'rsds'
+            },
+            {
+                'short_name': 'rsns'
+            },
+        ]
+        return required
+
+    @staticmethod
+    def calculate(cubes):
+        """Compute upwelling shortwave flux from downwelling and net."""
+        rsds_cube = cubes.extract(
+            Constraint(name='surface_downwelling_shortwave_flux_in_air'))
+        rsns_cube = cubes.extract(
+            Constraint(name='surface_net_downward_shortwave_flux'))
+
+        rsus_cube = rsds_cube - rsns_cube
+
+        return rsus_cube


valeriupredoi:
OK @lukasbrunner I fixed the lon and lat var names as per our discussion in #1054 and fixed to `extract_cube` rather than `extract` so those are cubes and CubeLists, and I added a test as well, please have a look and see if your problems in #1054 are corrected for, mate :+1: :beer:

esmvalcore/preprocessor/_derive/rsus.py
```diff
@@ -0,0 +1,41 @@
...
+
+    @staticmethod
+    def calculate(cubes):
+        """Compute upwelling shortwave flux from downwelling and net."""
+        rsds_cube = cubes.extract_cube(
+            Constraint(name='surface_downwelling_shortwave_flux_in_air'))
+        rsns_cube = cubes.extract_cube(
+            Constraint(name='surface_net_downward_shortwave_flux'))
+        # fix latitude and longitude var_name
+        rsns_cube.coord("longitude").var_name = "Longitude"
```
bouweandela, line 36: Rather than making a choice, it would  probably be safer to just make sure they are the same for both variables, e.g.:
```suggestion
        rsns_cube.coord(axis='X').var_name = rsds_cube.coord(axis='X').var_name
```

esmvalcore/preprocessor/_derive/rsus.py
```diff
@@ -33,7 +33,7 @@ def calculate(cubes):
         rsns_cube = cubes.extract_cube(
             Constraint(name='surface_net_downward_shortwave_flux'))
         # fix latitude and longitude var_name
-        rsns_cube.coord("longitude").var_name = "Longitude"
+        rsns_cube.coord(axis='X').var_name = rsds_cube.coord(axis='X').var_name
         rsns_cube.coord("latitude").var_name = "Latitude"
```
valeriupredoi, line 37:
```suggestion
        rsns_cube.coord(axis='Y').var_name = rsds_cube.coord(axis='Y').var_name
```

valeriupredoi (commented):
@lukasbrunner it's prob best to change the Y-axis one too :+1:

esmvalcore/preprocessor/_derive/rsus.py
valeriupredoi: yeh that's not a bad idea at all - more general, and if the X-azis is called `cows` that'll get picked up by the checker (or not, since it's var_name anyway) :+1:

lukasbrunner:
Okay so I had to add another two lines of fixes since the long_name was also causing problems. In addition I have to fix the `positive` attribute after the calculation. There should be a fix for this in cmor/_fixes/native6/era5.py so I can only assume this does not get called after derivation?

Can you maybe have a look if this makes sense since the test fails now.

valeriupredoi:
I fixed the test, but my question is - is `rsus` `positive` attribute `up`? - it's derived from two `down` variables, so I'd naively expect it to be `down` as well. You will have to add a custom CMOR table in `esmvalcore/cmor/tables/custom/` so that the derived variable be checked and fixed by CMOR :+1:

lukasbrunner:
Good question @valeriupredoi! I am not 100% sure but this is what makes most sense to me and what is consistent with the calculation as it is now. I started my considerations from this

`rsns (down) = rsds (down) - rsus (up)`

which transforms into

`rsus (up) = rsds (down) - rsns (down)`

which is what we do now. And this definition is also consistent with CMIP6 
https://github.com/ESMValGroup/ESMValCore/blob/b58bc5d7dad645cf4517f4cb9cc64db45ce95ff5/esmvalcore/cmor/tables/cmip6/Tables/CMIP6_Amon.json#L989-L1001

It is also intuitive to me that upward radiation would be `positive: up` but again I'm not 100% sure and I think it is most important to be consistent here.

> You will have to add a custom CMOR table in esmvalcore/cmor/tables/custom/ so that the derived variable be checked and fixed by CMOR

Okay that was not clear to me: even though rsus is in the CMOR tables already I sill need to add it again?

valeriupredoi:
ah you are absolutely right! It is indeed in the CMOR tables (I should have checked) and it's `upwelling`-something so `up` makes total sense. So no need to write a custom CMOR table and the native6/era checker should use the CMIP6 table (I guess, is that right @jvegasbsc @sloosvel ?)

valeriupredoi:
note that the rsus fix from the native6/era5 fix code is only applied now, after merging `master` branch in here (it wasn't in here before), so you should be ready to go :+1:

lukasbrunner:
Great I had not thought of that! What I would do now (if we agree that everything is fine) is to do the exact same thing we did for rsus (shortwave radiation) also for rlus (longwave radiation). Add a derivation script and a test? I don't need it at the moment but since it is exactly the same thing I guess it makes sense to add that as well whilst we are on it?

valeriupredoi:
yeah just add it in here, two derived vars in one PR is fine, if we go to more than two it becomes a but stuffy. Yes, pplease, add a test as well, when you done just change the PR status from Draft to ready for review and I'll approve it - cheers :beer:

lukasbrunner:
Hm I'm not sure about this test fail. Is it in an unrelated test? Could you have a look at it @valeriupredoi

valeriupredoi:
> Hm I'm not sure about this test fail. Is it in an unrelated test? Could you have a look at it @valeriupredoi

you were importing the `rsus` derivation module :grin:

lukasbrunner:
my search and replace failed me! thanks :beers:

valeriupredoi:
no worries, that's failed me on many occasions :grin: Good stuff, if you done changing can make the PR ready for review so I can go ahead and review&approve :+1:

valeriupredoi (approved):
good stuff @lukasbrunner thanks very much :beer:

lukasbrunner:
> good stuff @lukasbrunner thanks very much :beer:

Thanks for your help!

PR #1064, opened 2021-03-30
Run problematic cmor tests sequentially to avoid segmentation faults on CircleCI

valeriupredoi:
Attempting to minimize the effects of random segmentation faults described in #644 and based on finding in this [comment](https://github.com/ESMValGroup/ESMValCore/issues/644#issuecomment-810276519) and using implementation suggestion by @stefsmeets (cheers, mate!)

EDIT: huzzah! This thing really works!
Closes #644

PR #1065, opened 2021-03-31
Run conda build parallel AND sequential tests

valeriupredoi:
Continuation from #1064 to include the same test run strategy for conda build tests ~and try bump the number of threads we use for running the parallel tests on Circle.~ Bumping number of threads with `-n 4` definitely leads to memory issues for the conda build tests and might lead to the same issue for the regular CI tests, so let's keep them at 2

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,7 @@ jobs:
             . /opt/conda/etc/profile.d/conda.sh
             conda activate esmvaltool
             pip install .[test]
-            pytest -n 2 -m "not installation and not sequential"
+            pytest -n 4 -m "not installation and not sequential"
             pytest -n 0 -m "sequential"
       - save_cache:
           key: test-{{ .Branch }}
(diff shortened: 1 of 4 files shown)


stefsmeets:
`FAILED tests/integration/cmor/_fixes/cmip6/test_cesm2.py::test_cl_fix_file`

Oh boy... 🍿

valeriupredoi:
> `FAILED tests/integration/cmor/_fixes/cmip6/test_cesm2.py::test_cl_fix_file`
> 
> Oh boy... popcorn

yeah another one that we might have to run sequentially

valeriupredoi:
OK conda build test fails because flake8 is eager enough to check conda's run_test.py that we don't want to be checked

stefsmeets (approved):
Nice work 👍

valeriupredoi:
many thanks @stefsmeets Happy Easter :rabbit:

PR #1068, opened 2021-04-07
Fix: Flip big endian arrays before concatenation

markelg:
## Description

This fix is needed to support the CCLM CORDEX model data as it is available in ESGF and in the CDS. Endianness between files is not consistent, and this raises an error when concatenating. The best way to solve this we saw is to look for the ">" that flags big endian datatypes and, if found,  call numpy methods byteswap and newbyteorder to reverse the endianness of the underlying arrays.

Link to documentation:

***

***

### DIFF
--- a/esmvalcore/preprocessor/_io.py
+++ b/esmvalcore/preprocessor/_io.py
@@ -171,12 +171,22 @@ def _get_concatenation_error(cubes):
     raise ValueError(f'Can not concatenate cubes: {msg}')
 
 
+def _fix_cube_endianess(cubes):
+    """Transform cubes in big endian to little"""
+    for cube in cubes:
+        if cube.dtype.byteorder == ">":
+            logger.warning("Changing cube endianess to little. This may be "
+                           "memory intensive.")
+            cube.data = cube.data.byteswap().newbyteorder()
+
+
 def concatenate(cubes):
     """Concatenate all cubes after fixing metadata."""
     if len(cubes) == 1:
         return cubes[0]
 
     _fix_cube_attributes(cubes)
+    _fix_cube_endianess(cubes)
 
     if len(cubes) > 1:
         # order cubes by first time point


esmvalcore/preprocessor/_io.py
```diff
@@ -171,12 +171,22 @@ def _get_concatenation_error(cubes):
     raise ValueError(f'Can not concatenate cubes: {msg}')
 
 
+def _fix_cube_endianess(cubes):
+    """Transform cubes in big endian to little"""
```
valeriupredoi, line 175:
```suggestion
    """Transform cubes in big endian to little."""
```

esmvalcore/preprocessor/_io.py
```diff
@@ -171,12 +171,22 @@ def _get_concatenation_error(cubes):
...
+    for cube in cubes:
+        if cube.dtype.byteorder == ">":
+            logger.warning("Changing cube endianess to little. This may be "
+                           "memory intensive.")
+            cube.data = cube.data.byteswap().newbyteorder()
```
valeriupredoi, line 180: this is a flat-out data realization in memory, could you try exploring the use of `core_data()` instead of `data` please? That way we keep things lazy

valeriupredoi (changes requested):
cheers for the code changes! Please see a couple comments from me, also could I ask you to please write a unit test for the new function? Cheers :beer:

valeriupredoi:
Oh an one more thing - could you please open a PR directly into the ESMValCore repo rather than using a fork to the repo? Testing is better that way :+1:

markelg:
Thank you @valeriupredoi Shall I write the test in tests/integration/._io? I don't see tests for this module in the tests/unit

> Oh an one more thing - could you please open a PR directly into the ESMValCore repo rather than using a fork to the repo? Testing is better that way +1

zklaus:
Does the endianness extend to the coordinates as well? What about scalar coordinates or cell measures?

valeriupredoi:
> Shall I write the test in tests/integration/._io? I don't see tests for this module in the tests/unit

Maybe you can put this utility function in `esmvalcore/preprocessor/_other.py` and add the test to `tests/unit/preprocessor/_other/test_other.py`? I am not 100% sure it warrants itself inside the IO module since it's not something that gets used very often? Up to you -

markelg:
> Does the endianness extend to the coordinates as well? What about scalar coordinates or cell measures?

Yes, it does extend. I did not take that into account since it is the variable the one that is raising the error. _io_concatenate seems to handle well the coordinates. Now, maybe the function should flip all the arrays in cube.coords() too, this would be more consistent, but I don't now how to do that, since I don't see a setter for the cube coords. I have it now solved for the data, also for the lazy case.

zklaus:
Nice that you addressed the lazy thing. Could you push the related commit and we can continue the discussion from there?

markelg:
I think it is almost ready. Please have a look. I managed to fix the coordinates and bounds too. Apart from the unit test, I tested in with the CORDEX data that were causing the error and it worked fine.

zklaus:
Nice and good dig! Depending on your ambition, you might want to consider adding this functionality to dask itself by addressing the issue you linked. Then we could do away with the type detection workaround that you had to put in place here.

Do we need to take cell_measures into account as well? This would typically play a role for `areacella` or similar and is quite important particularly for regional grids since the cell size calculations for projections are not trivial in general.

bouweandela:
> Endianness between files is not consistent, and this raises an error when concatenating.

I think it would be good to open an issue about this in [iris](https://github.com/scitools/iris) and see if can be solved there.

PR #1076, opened 2021-04-21
Remove references to ESMValGroup/esmvalgroup conda channel

valeriupredoi:
I have started to remove references to the esmvalgroup channel from within esmvalcore codebase and documentation. We still have a couple things to sort out here:

- pointer to the ESMValGroup [packages](https://anaconda.org/esmvalgroup/repo?type=conda&label=main) from within the README's install with conda badge
- uploading the package at release point from Github Actions - this is something we still want to do just for security reasons and then and only then render and build the package on conda-forge? Or just drop that completely and build and upload on conda-forge onlt? I am not sure what are the steps to automate the conda-forge package build+upload so maybe @zklaus could edit `doc/contributing.rst` with the new steps and update `.github/workflows/action-conda-publish.yml` if it needs any editing :beer:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,7 +188,7 @@ jobs:
             conda create -y --name esmvaltool
             set +x; conda activate esmvaltool; set -x
             # Install
-            conda install -y esmvalcore -c esmvalgroup -c conda-forge
+            conda install -y esmvalcore -c conda-forge
             # Log versions
             conda env export > /logs/environment.yml
             # Test installation
--- a/.github/workflows/action-publish-pypi-conda-forge.yml
+++ b/.github/workflows/action-publish-pypi-conda-forge.yml
@@ -1,0 +1,154 @@
+name: PyPi and conda-forge Build and Deploy 🐍📦
+
+on:
+  release:
+    types: [published]
+  # use this for testing
+  push:
+    branches:
+      - master
+
+jobs:
+  pypi-build-n-publish:
+    name: Build and publish ESMValCore on PyPi
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.9
+      - name: Install pep517
+        run: >-
+          python -m
+          pip install
+          pep517
+          --user
+      - name: Build a binary wheel and a source tarball
+        run: >-
+          python -m
+          pep517.build
+          --source
+          --binary
+          --out-dir dist/
+          .
+      #- name: Publish distribution 📦 to Test PyPI
+      #  uses: pypa/gh-action-pypi-publish@master
+      #  with:
+      #    password: ${{ secrets.test_pypi_password }}
+      #    repository_url: https://test.pypi.org/legacy/
+      - name: Publish distribution 📦 to PyPI
+        if: startsWith(github.ref, 'refs/tags')
+        uses: pypa/gh-action-pypi-publish@master
+        with:
+          password: ${{ secrets.pypi_password }}
+
+  pypi-package-verify:
+    name: verify / python-${{ matrix.python-version }} / ${{ matrix.os }}
+    if: startsWith(github.ref, 'refs/tags')
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ['ubuntu-latest', 'macos-latest']
+        python-version: ["3.6", "3.7", "3.8", "3.9"]
+    runs-on: ${{ matrix.os }}
+    needs: pypi-build-n-publish
+    steps:
+      - uses: actions/checkout@v2
+        with:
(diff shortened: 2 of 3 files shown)


zklaus:
The conda-forge package will be build from the pypi sdist tarball. So pypi publishing has to happen first.

valeriupredoi:
> The conda-forge package will be build from the pypi sdist tarball. So pypi publishing has to happen first.

OK so we should include the workflow in the pypi publish workflow, do you think this can be done automatically (the conda-forge upload, not the workflow inclusion, this I'll do now and ask you to manage it). Also, do we still want to upload to the esmvalgroup channel too? I don't think that'd be needed anymore

valeriupredoi:
@zklaus I have just added a composite GA workflow file https://github.com/ESMValGroup/ESMValCore/pull/1076/commits/f20a99dc7a4f7574118f18af62470c5b5dfe64e2 that:

- builds and deploys the pypi package
- verifies it by installing from pip
- builds and deploys to conda, provided the pip verification was succesful
- verifies the conda package

Note that step 3 (building and uploading to conda) needs your attention and changes for the new conda-forge deployment strategy, which I don't know much about, could you have a look and change it so it will work please, when you haz some spare time :beer:

valeriupredoi:
after that we can safely remove `action-pypi-build-and-deploy.yml` and `action-conda-publish.yml` :+1:

zklaus:
The way conda-forge works normally is like this: When a new version is uploaded to pypi, conda-forge notices this automatically and opens a PR in the respective feedstock. This notifies the maintainers of the feedstock that will check the PR and if all is well merge it. The rest happens automatically.

valeriupredoi:
aha! I reckon we should probably just build it anyway without publishing just to track any issues that the automated conda-forge may run into, what do you think?

zklaus:
Sorry, I don't understand what you mean. Could you clarify? Specifically, what do you mean here with "build it anyway" and "publishing"?

valeriupredoi:
> Sorry, I don't understand what you mean. Could you clarify? Specifically, what do you mean here with "build it anyway" and "publishing"?

run `conda-build` nightly since that's how the conda-forge package gets built too right? We need to replicate in a nightly test every step that leads to publishing of the package onto conda-forge, bar, of course, the publishing itself. Me not being familiar with the conda-forge package build and deployment I am loading all my guns in https://github.com/ESMValGroup/ESMValCore/commit/f20a99dc7a4f7574118f18af62470c5b5dfe64e2 - it's you who decides which bit should be kept and which changed :beer:

zklaus:
I don't think we need to build this to (pre-?)check for conda-forge. I think it might be rather hard to replicate the build environment anyway and I don't see what we would gain. I wouldn't build a conda package at all, to be honest.

valeriupredoi:
excellent, so if `conda-build` is not used at all in the conda-forge package building then we can just retire the tests that run it?

zklaus:
Oh I think it is being used. We just don't need to care about it. When we publish on Pypi conda-forge should notice, open a PR and this will send us a notification. Then we check conda-forges CI checks, which include builds on all platforms, and if that works we merge, if it doesn't we investigate.

valeriupredoi:
aha, good plan! But how about our internal nightly tests that run `conda-build` - should we keep them?

zklaus:
That depends. Are they here to guarantee that conda-build works? Then I would say we don't need to bother. Are they actually used for something, for example to have a nightly-builds channel for esmvalcore that allows development with esmvaltool against the latest features that are not released yet? Then we should keep them.

I personally would lean towards scrapping them and saying: If you want to develop against a development version of esmvalcore, you need to make a development installation of esmvalcore.

valeriupredoi:
they are there just to check if the package builds well, there is no actual use for the package after it has been built, unless, of course, in the case of the (old) upload to `esmvalgroup`

zklaus:
In that case I would say we don't need it. @nielsdrost, @bouweandela, what say you?

valeriupredoi:
@zklaus what you reckon about this PR?

zklaus:
I am a bit unsure about the github action workflow. Putting questions about the workflow to one side for a moment, it doesn't seem to be on topic in this PR.

bouweandela:
I would be fine with removing all GitHub actions related to the conda build, but I do have a slight preference to keep at least one nightly conda build (i.e. the one on CircleCI) to gain some confidence that the conda build on conda-forge will actually work after making a release, even if it's not exactly the same build.

The documentation on building and uploading the conda package would also need to be updated: https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#create-and-upload-the-conda-package

valeriupredoi:
we should have ideally finished this before the 2.3.0 release :cry:

valeriupredoi:
we have done this, gradually, from different PRs, currently it's only the docker bit that's referencing `esmvalgroup` :+1:

PR #1079, opened 2021-04-22
Implemented ICON project to read native ICON model output

### LINKED ISSUE #1073: Include ICON in ESMValCore
ICON Data is currently not readable by the ESMValTool

I'm going to start with a small prototype to regrid ICON to a lat-lon grid.

### LINKED ISSUE #1355: `extra_facets` are broken
#1259 introduced a bug in the `extra_facets` functionality, there is a `[mip_]` missing in the access of the nested `dict` here:

https://github.com/ESMValGroup/ESMValCore/blob/090962bdc9ba1d55fdfed8858762a17c7594f123/esmvalcore/_config/_config.py#L80

bsolino:
## Description

Closes #1073
Closes #1355

Link to documentation:

- https://esmvaltool--1079.org.readthedocs.build/projects/ESMValCore/en/1079/develop/fixing_data.html#icon
- https://esmvaltool--1079.org.readthedocs.build/projects/ESMValCore/en/1079/quickstart/configure.html#developer-configuration-file

***

***


zklaus:
Medium-term, it would be nice to align this with [the extra_facets facility](https://esmvaltool--1153.org.readthedocs.build/projects/ESMValCore/en/1153/quickstart/configure.html#extra-facets).

schlunma:
This PR is ready for review now! :+1:

zklaus:
At first glance, this looks good! Please complete the checklist to make it ready for review.

schlunma:
Done!

schlunma:
Is there a chance that this can go into 2.4? I can only work on this today and tomorrow. Afterwards I will be on vacation and will not be back before the core feature freeze.

zklaus:
I'll start the review now. I see that this PR has no labels yet despite the checkbox in the list. The labels are important because they are used to generate the changelog for the release. Please have another look.

esmvalcore/cmor/_fixes/icon/icon.py
```diff
@@ -0,0 +1,163 @@
...
+            cube.standard_name = None
+        else:
+            cube.standard_name = self.vardef.standard_name
+        cube.var_name = self.vardef.short_name
+        cube.long_name = self.vardef.long_name
+        if cube.units != self.vardef.units:
+            cube.convert_units(self.vardef.units)
+
+    @staticmethod
+    def _fix_height(cube, plev_points_cube, plev_bounds_cube):
```
zklaus, line 68: I have to look at some example data. Will add `_fix_height` review later.

esmvalcore/_config/extra_facets/icon-mapping.yml
```diff
@@ -0,0 +1,23 @@
...
+  CFday:
+    clwvi: {raw_name: cllvi}
+  CFsubhr:
+    clwvi: {raw_name: cllvi}
+  E3hr:
+    clwvi: {raw_name: cllvi}
+  fx:
+    areacella:
+      latitude: grid_latitude
+      longitude: grid_longitude
```
zklaus, line 18: `grid_latitude` and `grid_longitude` are strictly reserved for rotated grids (see [here](cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html), particularly the note under example 5.6
> A CF compliant application can determine that rlon and rlat are longitude and latitude values in the rotated grid by recognizing the standard names grid_longitude and grid_latitude.

), it might be a good idea to communicate this upstream since that is most likely not what is meant here.

esmvalcore/cmor/_fixes/icon/icon.py
```diff
@@ -0,0 +1,163 @@
...
+
+        # Fix metadata
+        lat.var_name = "lat"
+        lon.var_name = "lon"
+        lat.standard_name = "latitude"
+        lon.standard_name = "longitude"
+        lat.long_name = "latitude"
+        lon.long_name = "longitude"
+        lat.convert_units('degrees')
+        lon.convert_units('degrees')
```
zklaus, line 111: What are the units before the conversion?

esmvalcore/cmor/_fixes/icon/icon.py
```diff
@@ -0,0 +1,163 @@
...
+        lat.units = "degrees_north"
+        lon.units = "degrees_east"
+
```
zklaus, line 114: I wonder about bounds. It's probably better to leave this to the future, but perhaps put a comment or otherwise a pin in it? Btw the mesh support is in [Iris (main) now](https://github.com/SciTools/iris/blob/main/docs/src/whatsnew/latest.rst). So maybe we can revisit that soonish.

esmvalcore/cmor/_fixes/icon/icon.py
```diff
@@ -0,0 +1,163 @@
...
+        t_coord.points = new_dt_points
+        t_coord.units = new_t_unit
+
+
+class Siconca(Fix):
+    """Fixes for ``siconca``."""
+
+    def fix_metadata(self, cubes):
+        """Fix metadata."""
+        # Note: wrong var_name gets fixed in AllVars
```
zklaus, line 160: Maybe make it clear that AllVars gets called later and that that also fixes the units (which should be "%" according to https://github.com/PCMDI/cmip6-cmor-tables/blob/master/Tables/CMIP6_SImon.json).

esmvalcore/cmor/check.py
```diff
@@ -647,7 +645,8 @@ def _check_coord_points(self, coord_info, coord, var_name):
                                          valid_max)
 
         if l_fix_coord_value:
-            if coord.ndim == 1:
+            # cube.intersection only works for cells with 0 or 2 bounds
+            if coord.ndim == 1 and coord.nbounds in (0, 2):
```
zklaus, line 649: What's `nbounds` on longitudes for icon? If it's neither 0 nor 2, is there an easy way to understand it?

esmvalcore/cmor/check.py
```diff
@@ -696,6 +695,11 @@ def _check_requested_values(self, coord, coord_info, var_name):
                 cmor_points = [float(val) for val in coord_info.requested]
             except ValueError:
                 cmor_points = coord_info.requested
+            if coord.points.ndim != 1:
+                self.report_debug_message(
+                    "Cannot check requested values of {}D coordinate {} since "
+                    "it is not 1D", coord.points.ndim, var_name)
+                return
```
zklaus, line 702: I presume this was added because some variable is requested on specific pressure levels in CMIP6, but not delivered on pressure levels in ICON?

esmvalcore/preprocessor/_regrid.py
```diff
@@ -470,11 +473,19 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
...
             cube = esmpy_regrid(cube, target_grid, scheme)
         else:
             cube = cube.regrid(target_grid, HORIZONTAL_SCHEMES[scheme])
+        try:
+            cube.data = cube.core_data().astype(original_dtype,
+                                                casting='same_kind')
+        except TypeError as exc:
+            logger.warning(
+                "dtype of data changed during regridding from '%s' to '%s': "
+                "%s", original_dtype, cube.core_data().dtype, str(exc))
```
zklaus, line 488: I think the potential impact on other data is too large to include it here. Can we keep it separate? Why does it affect ICON particularly?

zklaus (changes requested):
Overall nice, clean code. Good job! I made some comments and will turn to _fix_height and the tests only now. I'll look at the documentation last.

esmvalcore/_config/extra_facets/icon-mapping.yml
```diff
@@ -0,0 +1,23 @@
...
+      raw_name: cell_area
+  SIday:
+    siconca: {raw_name: sic}
+  SImon:
+    siconca: {raw_name: sic}
```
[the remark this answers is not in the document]
schlunma: This is definitely not complete, but covers many variables (a list of the variables that we analyzed is given in the recipes here https://github.com/ESMValGroup/ESMValTool/tree/icon_eval/esmvaltool/recipes/icon). Most of the variables actually have the correct `short_name`.

esmvalcore/cmor/_fixes/icon/icon.py
```diff
@@ -0,0 +1,163 @@
...
+        if cube.coords("time"):
+            self._fix_time(cube)
+        if cube.coords("height"):
+            plev_points_cube = cubes.extract_cube(var_name_constraint('pfull'))
+            plev_bounds_cube = cubes.extract_cube(var_name_constraint('phalf'))
+            cube = self._fix_height(cube, plev_points_cube, plev_bounds_cube)
+        lat_name = self.extra_facets.get('latitude', 'latitude')
+        lon_name = self.extra_facets.get('longitude', 'longitude')
+        if cube.coords(lat_name) and cube.coords(lon_name):
+            self._fix_lat_lon(cube, lat_name, lon_name)
```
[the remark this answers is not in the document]
schlunma: I remember that @bouweandela once told me (> 2 years ago) to use `self` instead of the class name :smile: Is there any particular reason to avoid `self`?

esmvalcore/cmor/_fixes/icon/icon.py
```diff
@@ -0,0 +1,163 @@
...
+    @staticmethod
+    def _fix_time(cube):
+        """Fix time coordinate of cube."""
+        t_coord = cube.coord("time")
+        t_coord.var_name = 'time'
+        t_coord.standard_name = 'time'
+        t_coord.long_name = 'time'
+
+        # Convert units to CF format
+        t_unit = t_coord.attributes.pop("invalid_units")
```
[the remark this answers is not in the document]
schlunma: Good idea, will do!

esmvalcore/cmor/_fixes/icon/icon.py
```diff
@@ -0,0 +1,163 @@
...
+        timestep, _, t_fmt_str = t_unit.split(" ")
+        new_t_unit_str = f"{timestep} since 1850-01-01"
```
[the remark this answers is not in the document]
schlunma: In the ICON files that I have `{timestep}` is `day`, but I think we can leave it like this to support different units. Is there any reason to not use 1850-01-01 as the baseline?

esmvalcore/cmor/_fixes/icon/icon.py
```diff
@@ -0,0 +1,163 @@
...
+        new_t_unit = cf_units.Unit(new_t_unit_str, calendar="standard")
```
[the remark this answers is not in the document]
schlunma: It's actually `proleptic_gregorian`, so we probably should use this. There is no way of reading this from the file, since iris removes the information and just saves the `invalid_units` without the calender, e.g.

```
DimCoord(array([20110201.]), standard_name='time', units=Unit('unknown'), var_name='time', attributes={'invalid_units': 'day as %Y%m%d.%f'})
```

esmvalcore/cmor/_fixes/icon/icon.py
```diff
@@ -0,0 +1,163 @@
...
+
+        new_datetimes = [datetime.strptime(str(dt), t_fmt_str) for dt in
+                         t_coord.points]
+        new_dt_points = [new_t_unit.date2num(new_dt) for new_dt in
+                         new_datetimes]
```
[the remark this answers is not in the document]
schlunma: Yes, good point, will do!

esmvalcore/cmor/_fixes/icon/icon.py
schlunma: Yes!

esmvalcore/_config/extra_facets/icon-mapping.yml
schlunma: Yes, but we are changing the names here. The "problem" here is that `areacella` is read from the "grid information" file which has not been CMORized. All other variables correctly use `latitude` and `longitude`

esmvalcore/cmor/_fixes/icon/icon.py
schlunma: `radians`

esmvalcore/cmor/_fixes/icon/icon.py
schlunma: The bounds are set correctly already by iris, there is nothing that we need to change.

esmvalcore/cmor/check.py
schlunma: `nbounds=3` for ICON, which correspond to the vertices of triangles, see

grafik

esmvalcore/cmor/check.py
schlunma: Yes, ICON only provides 4D pressure levels. We could of course use interpolation in the CMORizer to provide the data on the requested 1D levels, but I don't think that's a good idea since it will add unnecessary errors to the data.

esmvalcore/preprocessor/_regrid.py
schlunma: For an unknown (to me) reason the `unstructured_nearest` scheme always converts the data to `float64`, this is why it only affects ICON now. Since I'm using the `try...except` here and `casting='same_kind'` I think we are pretty safe here: The data type is only altered if it doesn't change the data values dramatically (e.g., no `1.5 (float) -> 1 (int)`) or changes from one float type to another.

esmvalcore/cmor/_fixes/icon/icon.py
```diff
@@ -0,0 +1,163 @@
...
+            standard_name='air_pressure',
+            long_name='pressure',
+            units=plev_points_cube.units,
+            attributes={'positive': 'down'},
+        )
+        cube.add_aux_coord(air_pressure_coord, np.arange(cube.ndim))
+
+        # Add 'positive' attribute to height so iris can determine Z axis
+        # correctly
+        cube.coord('height').attributes['positive'] = 'down'
```
zklaus, line 91: The standard name "height" has canonical units "m", whereas this height coordinate doesn't measure distance but is only an index (with a slightly odd choice of bounds). Not completely sure how to address this. I think the coordinate should not be height. For `long_name` it uses `generalized_height`, perhaps that could be appropriate, or something with index in its name.

esmvalcore/cmor/_fixes/icon/icon.py
```diff
@@ -0,0 +1,163 @@
...
+
+        # Reverse height axis so that index 0 is surface level
```
zklaus, line 93: Perhaps clearer to write:
```suggestion
        # Reverse entire cube along height axis so that index 0 is surface level
```

esmvalcore/cmor/_fixes/icon/icon.py
zklaus: Actually, looking at some of the exampels in the CF conventions, `model_levels` seems a clear candidate.

zklaus (changes requested):

esmvalcore/_config/extra_facets/icon-mapping.yml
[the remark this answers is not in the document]
zklaus: Nice. Perhaps add a short comment to that effect at the top of the file.

esmvalcore/cmor/_fixes/icon/icon.py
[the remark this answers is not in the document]
zklaus: :laughing: Well, for me it makes it clearer that this is not a method that can interact with self. But a bit of searching didn't give me any great reference. @bouweandela, do you have an opinion? I think if Bouwe doesn't way in, we shouldn't let this stop us.

esmvalcore/cmor/_fixes/icon/icon.py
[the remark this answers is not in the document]
zklaus: Hm. Looking at the example I think I might be even stricter here and match this against the fixed string `day as %Y%m%d.%f`. I mean, what could you envision as an alternative? `hour as ???`

Also, looking a bit closer at the documentation of [`datetime.strptime`](https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes), I worry a wee bit about the `%f`. Do you perhaps have some sub-daily data (6hr, 3hr, ...) to check that everything works out there?

esmvalcore/cmor/_fixes/icon/icon.py
schlunma: Ok, I will change this to `model_levels`. Any preference for the `long_name` and `standard_name`?

esmvalcore/cmor/_fixes/icon/icon.py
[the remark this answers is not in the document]
zklaus: On the basetime, well, CMIP6 makes a bunch of provisions (see the note linked above), but it's not critical, I think.

esmvalcore/cmor/_fixes/icon/icon.py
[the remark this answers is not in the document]
zklaus: Ah yes, I'd consider this an iris bug. Perhaps raise it over there when you come back from vacation.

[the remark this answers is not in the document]
zklaus: Oh but definitely put the proleptic_gregorian. That really could have an impact for spinup runs.

esmvalcore/_config/extra_facets/icon-mapping.yml
zklaus: Yeah, I get it. I just thought you could tell the modelers so that they can fix their wrong use of these standard names and we can put a comment that makes us reevaluate the situation in a couple of years.

esmvalcore/cmor/_fixes/icon/icon.py
zklaus: Ok, that makes sense. Does a direct conversion to 'degrees_north' (and '_east') work?

esmvalcore/cmor/_fixes/icon/icon.py
zklaus: Ah yes, got it.

esmvalcore/preprocessor/_regrid.py
zklaus: See https://github.com/SciTools/iris/blob/3ff02699b7b6ea341895c4b893febe29670f0435/lib/iris/analysis/trajectory.py#L421 :facepalm: . I'd suggest to report this as an issue to iris and here to do the dtype conversion based on the `scheme`.

esmvalcore/cmor/_fixes/icon/icon.py
schlunma: It works! Just changed it.

esmvalcore/cmor/_fixes/icon/icon.py
zklaus: I was inspired by [CF Conventions 1.9, Example 6.2](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.9/cf-conventions.html#alternative-coordinates), so perhaps you could follow that? Notably, `standard_name` is `None`.

esmvalcore/cmor/_fixes/icon/icon.py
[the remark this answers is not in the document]
schlunma: Unfortunately, I don't have sub-daily data to test this right now. I will fix the invalid units and add a check. If it doesn't work out of the box, support for sub-daily data can be implemented at a later stage

schlunma:
Any further comments on this?

esmvalcore/cmor/_fixes/icon/icon.py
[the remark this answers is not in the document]
zklaus: I still prefer a clearer distinction of static methods, but let's not consider it critical.

esmvalcore/cmor/_fixes/icon/icon.py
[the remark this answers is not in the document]
zklaus:
```suggestion
        new_t_unit_str = f"days since 1850-01-01"
```
 CMIP6 requires the plural form "days", which also makes more sense to me. If you want to keep it tied to the `timestep`, I would introduce a translation dictionary like `dict(day="days")`, but honestly, I think keeping it to literally "days" here is the simplest and only if and when we have an example that doesn't come as "day" we'll know how to do better.

esmvalcore/cmor/check.py
zklaus: Oh yeah, of course. In that case, I suggest adding a brief comment to the effect of:
> `nbounds==0` means there are no bounds given, `nbounds==2` implies a regular grid with bounds in the grid direction, `nbounds>2` implies an irregular grid with bounds given as vertices of the cell polygon.

esmvalcore/_config/extra_facets/icon-mapping.yml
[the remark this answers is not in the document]
zklaus: I think `component` and `var_type` should appear in this file as well. This would simplify the recipes and it should be clear for every variable, right? You might want to experiment even with something like
```
  Amon:
    *:
      component: atm
      var_type: atm_2d_ml
```
though I suspect that ultimately this approach still means that more variables need to appear here.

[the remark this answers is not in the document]
zklaus: Perhaps that should even cover `version`.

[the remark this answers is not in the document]
schlunma: No, these the values for `component` and `var_type` are completely arbitrary, so it doesn't make sense to fix them here. Also, version could change as well, so we definitely want this in the recipe.

tests/integration/cmor/_fixes/icon/test_icon.py
```diff
@@ -0,0 +1,410 @@
...
+
+    assert cube.coords('latitude', dim_coords=False, dimensions=(0, 1))
+    assert cube.coords('longitude', dim_coords=False, dimensions=(0, 1))
+    assert cube.coords('height', dim_coords=False, dimensions=())
+
+
+# Test fix with empty standard_name
+
+
+def test_empty_standard_name_fix(cubes_2d):
```
zklaus, line 385: I don't understand the point of this test, less since obviously there is a correct standard name for tas. Could you explain?

tests/integration/preprocessor/_regrid/test_regrid.py
```diff
@@ -35,9 +35,69 @@ def test_regrid__linear(self):
                                     coord_system=self.cs)
         coords_spec = [(lats, 0), (lons, 1)]
         grid = iris.cube.Cube(data, dim_coords_and_dims=coords_spec)
-        result = regrid(self.cube, grid, 'linear')
+        self.grid_for_linear = grid
+
+        # Setup unstructed cube and grid
```
zklaus, line 40:
```suggestion
        # Setup unstructured cube and grid
```

tests/integration/preprocessor/_regrid/test_regrid.py
```diff
@@ -35,9 +35,69 @@ def test_regrid__linear(self):
...
+        data = np.empty((1, 1))
```
zklaus, line 41:
```suggestion
        data = np.zeros((1, 1))
```
for reproducability.

tests/integration/preprocessor/_regrid/test_regrid.py
```diff
@@ -35,9 +35,69 @@ def test_regrid__linear(self):
...
+
+        lats = iris.coords.AuxCoord(
+            x,
+            standard_name=lats.metadata.standard_name,
+            long_name=lats.metadata.long_name,
+            var_name=lats.metadata.var_name,
+            units=lats.metadata.units,
+            attributes=lats.metadata.attributes,
+            coord_system=lats.metadata.coord_system,
+            climatological=lats.metadata.climatological)
```
zklaus, line 69: `**lats.metadata` ?

tests/integration/preprocessor/_regrid/test_regrid.py
```diff
@@ -35,9 +35,69 @@ def test_regrid__linear(self):
...
+        self.grid_for_unstructured_nearest = iris.cube.Cube(
+            data, dim_coords_and_dims=coords_spec)
+
+        # Replace 1d spatial coords with 2d spatial coords.
+        lons = self.cube.coord('longitude')
+        lats = self.cube.coord('latitude')
+        x, y = np.meshgrid(lons.points, lats.points)
+
+        lats = iris.coords.AuxCoord(
+            x,
```
zklaus, line 62: lats should be y and lons x, no?

zklaus (changes requested):

esmvalcore/_config/extra_facets/icon-mapping.yml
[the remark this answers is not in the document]
zklaus: Version could definitely change, that was really only a question.

But `component` and `var_type` are really completely arbitrary? `(Amon, tas)` will not always come from the atmosphere? That sounds surprising.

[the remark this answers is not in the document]
schlunma: The values are set by the person that runs the model, so we need them to stay here.

tests/integration/preprocessor/_regrid/test_regrid.py
schlunma: That was copy-pasted from existing code, and I really do not want to mess with other tests

tests/integration/preprocessor/_regrid/test_regrid.py
schlunma: That was copy-pasted from existing code, and I really do not want to mess with other tests

tests/integration/preprocessor/_regrid/test_regrid.py
schlunma: That was copy-pasted from existing code, and I really do not want to mess with other tests

tests/integration/cmor/_fixes/icon/test_icon.py
schlunma: To test these lines:

https://github.com/ESMValGroup/ESMValCore/blob/df4458f4079c6b031254d8d30c589bd009a820b2/esmvalcore/cmor/_fixes/icon/icon.py#L58-L61

I know that there is a standard name for tas, this is mainly done to please codecov so that every line is tested.

tests/integration/cmor/_fixes/icon/test_icon.py
```diff
@@ -0,0 +1,410 @@
...
+    """Check latitude and longitude coordinates of cube."""
+    assert cube.coords('latitude', dim_coords=False)
+    lat = cube.coord('latitude', dim_coords=False)
+    assert lat.var_name == 'lat'
+    assert lat.standard_name == 'latitude'
+    assert lat.long_name == 'latitude'
+    assert lat.units == 'degrees_north'
+    np.testing.assert_allclose(lat.points, [-90.0, -45.0, 0.0,  45.0, 90.0])
+    np.testing.assert_allclose(lat.bounds,
+                               [[-100.0, -90.0, -80.0],
```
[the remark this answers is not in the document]
schlunma: I added this to make sure that no unintended change of lat and lon values takes place (this is done by the CMOR checker later, e.g. transform lons from -180...180 to 0...360). I don't think we need a special grid here, all things that needs to get tested also work with this grid. If we need a more complex one in the future we can implement it later.

tests/integration/cmor/_fixes/icon/test_icon.py
```diff
@@ -0,0 +1,410 @@
...
+
+@pytest.fixture
+def cubes_2d_lat_lon_grid():
+    """Cube with 2D latitude and longitude."""
+    lat_coord = iris.coords.AuxCoord([[0.0, 1.0], [2.0, 3.0]], var_name='lat',
+                                     standard_name='latitude',
+                                     long_name='latitude', units='degrees')
+    lon_coord = iris.coords.AuxCoord([[-1.0, 1.0], [2.0, 3.0]], var_name='lon',
+                                     standard_name='longitude',
+                                     long_name='longitude', units='degrees')
```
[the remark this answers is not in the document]
schlunma: Yes, but they really do not matter here. This tests merely checks that the CMORizer also works for data that is already on a regular grid.

esmvalcore/_config/extra_facets/icon-mapping.yml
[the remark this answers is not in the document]
zklaus: Ok. We need a list with a short explanation of all supported extra facets. That should definitely be in the documentation part (will add a comment there), but might be repeated in the comment at the top of this file, at least stating which facets are expected here with a link to the full documentation.

tests/integration/cmor/_fixes/icon/test_icon.py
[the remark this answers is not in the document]
zklaus: I agree that we don't need a special or particular grid here, but this "grid" is a sequence of unconnected, degenerate triangles with area 0. As such, the cell areas are wrong and the regridding is questionable (after all, the support of the grid is empty). While of course, the nearest scheme works, this is just asking for trouble in the future. The points south of the south pole and north of the north pole are not even well-defined.

If this really is all supposed to test what happens with an invalid grid, that should be stated clearly at least in a comment.

I think a nice simple reasonable grid would be a better choice.

tests/integration/cmor/_fixes/icon/test_icon.py
zklaus: Is there an actual example of a variable with no standard name? That might be a better test. Otherwise, I suggest adding a comment.

tests/integration/preprocessor/_regrid/test_regrid.py
zklaus: It shows up as your code, hence you are responsible.

tests/integration/preprocessor/_regrid/test_regrid.py
zklaus: Still your responsibility, but in this case, I am ok with this code, just wanted to offer a possible simplification, so here your call.

tests/integration/preprocessor/_regrid/test_regrid.py
zklaus: Your code, your responsibility. The issue is that we have a few lines before this
```
x, y = np.meshgrid(lons.points, lats.points)
```
clearly saying that x is longitude and y is latitudes (which makes sense). If the switch is purposeful, it should be explained in a comment, otherwise, it should be corrected.

doc/develop/fixing_data.rst
```diff
@@ -372,6 +372,27 @@ Native models
...
+       exp: amip, grid: R2B5, ensemble: r1v1i1p1l1f1, mip: Amon,
+       short_name: tas, var_type: atm_2d_ml, start_year: 2000, end_year: 2014}
+    - {project: ICON, dataset: ICON, component: atm, version: 2.6.1,
+       exp: amip, grid: R2B5, ensemble: r1v1i1p1l1f1, mip: Amon,
+       short_name: ta, var_type: atm_3d_ml, start_year: 2000, end_year: 2014}
+
+Please note the duplication of the name ``ICON`` in ``project`` and
+``dataset``, which is necessary to comply with ESMValTool's data finding and
+CMORizing functionalities.
+
```
zklaus, line 395: We should add a (tabular? list?) overview of all supported extra facets with a brief explanation.

zklaus (changes requested):

esmvalcore/preprocessor/_regrid.py
```diff
@@ -469,13 +472,25 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
...
         else:
             cube = cube.regrid(target_grid, HORIZONTAL_SCHEMES[scheme])
 
+        # Preserve dtype for 'unstructured_nearest' scheme
+        if scheme == 'unstructured_nearest':
+            try:
+                cube.data = cube.core_data().astype(original_dtype,
+                                                    casting='same_kind')
+            except TypeError as exc:
+                logger.warning(
```
bouweandela, line 489:
```suggestion
                logger.debug(
```
Does this happen often? Is this really important info for the user?

tests/integration/cmor/_fixes/icon/test_icon.py
schlunma: I am not aware of an actual use case, but in the checks we treat `None` and `''` equally (see e.g.,

https://github.com/ESMValGroup/ESMValCore/blob/d587d0c245ac52cf32326d9ca5a46bc53b982b90/esmvalcore/cmor/check.py#L400-L401

so I figure it's best if we do the same here. `''` is not a valid standard name and `iris` will crash if we try to set that.

Where do you want me to put the comment? In the actual code or the test?

zklaus: I was thinking of a comment in the test to the effect of "We know that tas has a standard name, but this being native model output there may be variables with no standard name. The code is designed to handle this gracefully and here we test it with an artificial, but realistic case."

tests/integration/cmor/_fixes/icon/test_icon.py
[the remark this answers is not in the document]
schlunma: I still don't fully agree with you (we do not test the regridding here, so imho the grid doesn't matter), but I can change that. Let me try to use an octahedron instead.

tests/integration/cmor/_fixes/icon/test_icon.py
```diff
@@ -0,0 +1,445 @@
...
+
+@lru_cache
+@pytest.fixture
+def cubes_3d(test_data_path):
+    """3D sample cubes."""
+    nc_path = test_data_path / 'icon_3d.nc'
+    return iris.load(str(nc_path))
+
+
+@lru_cache
```
bouweandela, line 31: Please don't use `lru_cache` for caching fixtures, because there is a suspicion that this causes segmenation faults in the tests. You can set scope="module" if you want to re-use the fixture for the entire module, see https://docs.pytest.org/en/6.2.x/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session.

esmvalcore/preprocessor/_regrid.py
zklaus: Quite right, see above: https://github.com/ESMValGroup/ESMValCore/pull/1079#discussion_r710075266.
But that's beyond us here. Let's keep the warning for now and try to fix upstream in the long term.

PR #1082, opened 2021-04-26
Allow wildcard searches when specifying fx variables in preprocessor

### LINKED ISSUE #1081: Implement multiple ensemble style syntax used for datasets in fx variable descriptions for pre processor
I'm writing recipes using a large number of the CORDEX EUR-11 models, and need to do some land sea masking. Unfortunately, for some (not all) of the CORDEX models there are not separate sftlf fx files for each ensemble member, and instead, the fx fields exist under ensemble r0i0p0 whereas the rest of the variables are under their own ensemble numbers r1i1p1 etc. See for example: https://esgf-index1.ceda.ac.uk/esg-search/search/?offset=0&limit=10&type=Dataset&replica=false&latest=true&domain=EUR-11&rcm_name=RCA4&time_frequency=fx%2Cmon&variable=sftlf%2Ctas&project=CORDEX&facets=rcm_name%2Cproject%2Cproduct%2Cdomain%2Cinstitute%2Cdriving_model%2Cexperiment%2Cexperiment_family%2Censemble%2Crcm_version%2Ctime_frequency%2Cvariable%2Cvariable_long_name%2Ccf_standard_name%2Cdata_node&format=application%2Fsolr%2Bjson
This causes the preprocessor to break because it is searching for the fx files under the same ensemble number as the rest of the variables (and they don't exist).
What would be great, would be if the user could use the syntax developed for specifying multiple ensembles in datasets at https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/overview.html#recipe-section-datasets e.g. by specifying something like:
```
mask_landsea:
      mask_out: sea
      fx_variables:
        - { short_name: sftlf, ensemble: "r(0:9)i(0:9)p(0:9)" }
 ```
In the preprocessor. However, the preprocessor interprets the string literally and fails to find the files.
It's not possible to use the Natural Earth dataset with this data, since the data is on a rotated pole and thus has multidimensional "latitude" and "longitude" co-ordinates which is not yet supported for use with shapefiles. (That's another issue I need to look into soon). 

Would you have the time and skills to implement the solution yourself?
I can have a look, but I'll need some pointers on where to get started etc.

### PR DESCRIPTION
**Pull request by thomascrocker**
## Description

Closes #1081
Finding data for fx variables is a bit of a pain depending on MIP project.
For CMIP5, fx variables are usually stored under a new ensemble (r0i0p0) although there are a small number of exceptions, see for example, 
https://esgf-index1.ceda.ac.uk/esg-search/search/?offset=0&limit=10&type=Dataset&replica=false&latest=true&project=CMIP5&cmor_table=fx&ensemble=r0i0p1%2Cr0i0p141%2Cr0i0p142%2Cr0i0p150%2Cr0i0p151%2Cr0i0p2%2Cr0i0p3%2Cr0i0p4%2Cr0i0p5%2Cr0i0p6%2Cr1i1p1&facets=project%2Cproduct%2Cinstitute%2Cmodel%2Cexperiment%2Cexperiment_family%2Ctime_frequency%2Crealm%2Ccmor_table%2Censemble%2Cvariable%2Cvariable_long_name%2Ccf_standard_name%2Cdata_node&format=application%2Fsolr%2Bjson
For CORDEX: It appears to vary arbitrarily depending on modelling centre. "Invariant fields (frequency=fx) may have the value r0i0p0 or that of the corresponding GCMEnsembleMember attribute" http://is-enes-data.github.io/cordex_archive_specifications.pdf
For CMIP6: I can't find any specifications, but it seems fx files are mostly under individual ensemble memebers, but not sure if this is the case for all models and fx fields.

~~This fix essentially detects if finding fx files has failed. If it has, it will automatically search again under r0i0p0. This way, the user can still specify explicitly the fx ensemble to use (and ESMValTool will prefer the files stored under the same ensemble as the variable if the user doesn't specify anything), and CMIP6 will work correctly. For CORDEX and CMIP5 the software will happily find the correct files under r0i0p0 and writes a line to the debug logger to record as such.~~
This fix solves the issue by allowing the user to specify FX variables in the recipe using glob style wildcards, this can then be used to allow the pre-processor to find FX data that is stored in different locations.
***

- [x] [☝ Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do (#1081)

- [X] [🛠][1] This pull request has a [descriptive title and labels](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-title-and-label)
- [X] [🛠][1] Changes are [backward compatible](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#backward-compatibility)

***

### CODE REVIEW

**Comment by schlunma** [comment]
Hi @thomascrocker, thanks for your contribution!

I had a brief look at this and found that our tests for `data_finder.py` are not covering `fx` files at the moment. I will add thorough tests for that tomorrow. I think it makes sense to include the additional tests first before reviewing/merging this PR. I hope that is okay for you!

Sorry for any inconvenience!

File: esmvalcore/_data_finder.py [hunk 1]
```diff
@@ -241,11 +241,21 @@ def _find_input_files(variable, rootpath, drs):
...
     """Return the full path to input files."""
-    # change ensemble to fixed r0i0p0 for fx variables
-    # this is needed and is not a duplicate effort
-    if variable['project'] == 'CMIP5' and variable['frequency'] == 'fx':
-        variable['ensemble'] = 'r0i0p0'
     (files, dirnames, filenames) = _find_input_files(variable, rootpath, drs)
+    if (
+        variable['frequency'] == 'fx'
+        and variable['ensemble'] != 'r0i0p0'
+        and dirnames == []
```
valeriupredoi, line 248: you are strangling the code here ie you are looking for a very specific combination of parameters that may result in the combined condition to be missed a lot of the times. Note that for CMIP5 the ensemble is always `r0i0p0` for fx variables, and the ones you found out that don't stick to this rule should still have the fx data files in `/r0i0p0` dir - if not, that's a problem of the data release. For CMIP6 they will never be in `r0i0p0` but dotted around in various ensembles `rXiXpXfX`, as you well note, but not always in *all* ensemble dirs (they may be in `r1i1p1f1` but not in `r2i1p1f1`) so a `*` search on the ensemble dir will do the job, or a quasi-explicit `r*` for ensemble, the fx data files should all be the same no matter the ensemble. So CMIP6 should be covered as it is now, the only problem I see is the CMIP5 ones that are not `r0i0p0`, which kind of alarms me and am thinking what sort of files are those

thomascrocker: Thanks for your comments. I'm not sure how a user would write a recipe to handle data where the ensemble containing the fx info is stored under a different ensemble to that specified in the variable (tas, pr or whatever), which as you say can happen under CMIP6.
Similarly for CORDEX, the data specification I linked to above states that the fx info can live either under the r0i0p0 ensemble or the ensemble of the regular variables in question.

My issue is I'm trying to develop a recipe that performs the same analysis on large multi model ensembles from different projects (~30 models each from CMIP5/6, CORDEX and UKCP) I want to include a pre processor that does the land sea masking of the models..
i.e.
```
    mask_landsea:
      mask_out: sea
```
In the current version of the code the data finder looks in `r0i0p0` for CMIP5, and the same ensemble as the variable for everything else. So at the moment this will not find the fx files for a large proportion of the CMIP6 and CORDEX data. I don't see any way in a recipe, short of writing a separate pre-processor for each dataset, to deal with looking in different ensemble folders depending on the dataset.
One other issue I've also just discovered, is for some of the CORDEX datasets, sometimes the data exists under the historical experiment, but not the rcp85 for example. :-/ As you say, the fx files should be the same across ensembles and experiments, so ideally it would be good to search across.

Are you suggesting a modification of some sort to the https://github.com/ESMValGroup/ESMValCore/blob/574f9f8054a99aedfc4b4606682f367ad82b1e6d/esmvalcore/_data_finder.py#L201 function to replace the ensemble part of the dir name with a wildcard?

thomascrocker:
> Hi @thomascrocker, thanks for your contribution!
> 
> I had a brief look at this and found that our tests for `data_finder.py` are not covering `fx` files at the moment. I will add thorough tests for that tomorrow. I think it makes sense to include the additional tests first before reviewing/merging this PR. I hope that is okay for you!
> 
> Sorry for any inconvenience!

Thanks, I saw that some of the tests had failed under the commit test, but didn't understand enough about them to be honest. Happy for you to add whatever you feel is necessary.

esmvalcore/_data_finder.py
valeriupredoi: cheers for the detailed explanation :beer: - yes, probably a wildcard would be better, or better off, a the use of `glob.glob()` with a wildcard. Do you think that'd be beyond the scope of this PR?

esmvalcore/_data_finder.py
```diff
@@ -156,16 +160,26 @@ def _resolve_latestversion(dirname_template):
...
-    if '{latestversion}' not in dirname_template:
+    if "{latestversion}" not in dirname_template:
         return dirname_template
 
     # Find latest version
-    part1, part2 = dirname_template.split('{latestversion}')
+    part1, part2 = dirname_template.split("{latestversion}")
+    # resolve any wildcards entered for fx variables
+    if "/fx/" in part1:
+        dirs = glob.glob(part1)
```
valeriupredoi, line 170: ah very nice, I just wrote my comment to the review post without seeing the new changes :+1:

valeriupredoi (approved):

valeriupredoi:
nice work @thomascrocker :+1:

esmvalcore/_data_finder.py
valeriupredoi: yeah I just fixed those tests, @schlunma would you mind having a look too, these are tests that have been freshly added from `master` after a previous merge of your data finder test improving branch :+1: :beer:

valeriupredoi: cheers for willing to do dox @thomascrocker :beer:

valeriupredoi:
frenly ping @schlunma for a final review (tests changed) and @thomascrocker fix the conflict pls :beer:

bouweandela:
Is this still needed, given that an improved syntax for specifying fx variables was introduced in #999?

@zklaus As our expert in conventions, do you have an opinion on these changes?

valeriupredoi:
> Is this still needed, given that an improved syntax for specifying fx variables was introduced in #999?
> 
> @zklaus As our expert on conventions, do you have an opinion on these changes?

yes it's needed to find the actual files, this doesn't impact the loading and use of the fx variables

esmvalcore/_data_finder.py
schlunma: I'm not really convinced of this solution:

1. Sometimes fx variables for CMIP6 are stored in other tables, e.g. `Omon` (if the files are time dependent) or `Ofx`. Both cases are not considered here.
2. Why would we ignore subsequent directories if more than one is found?

esmvalcore/_data_finder.py
```diff
@@ -231,63 +245,60 @@ def _find_input_dirs(variable, rootpath, drs):
...
+    variable["short_name"] = short_name
     return (files, input_dirs, filenames_glob)
 
 
 def get_input_filelist(variable, rootpath, drs):
     """Return the full path to input files."""
-    # change ensemble to fixed r0i0p0 for fx variables
-    # this is needed and is not a duplicate effort
-    if variable['project'] == 'CMIP5' and variable['frequency'] == 'fx':
-        variable['ensemble'] = 'r0i0p0'
```
schlunma, line 254: I really do not like this. Changes like this are not backwards compatible, i.e. we have to change every recipe that uses `fx` variables for CMIP5 data.

tests/integration/data_finder.yml
```diff
@@ -422,14 +422,14 @@ get_input_filelist:
...
+      - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r1i1p1/v20110330/sftlf/sftlf_fx_HadGEM2-ES_historical_r1i1p1.nc
     dirs:
-      - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r0i0p0/v20110330/sftlf
+      - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r1i1p1/v20110330/sftlf
     file_patterns:
-      - sftlf_fx_HadGEM2-ES_historical_r0i0p0*.nc
+      - sftlf_fx_HadGEM2-ES_historical_r1i1p1*.nc
     found_files:
-      - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r0i0p0/v20110330/sftlf/sftlf_fx_HadGEM2-ES_historical_r0i0p0.nc
+      - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r1i1p1/v20110330/sftlf/sftlf_fx_HadGEM2-ES_historical_r1i1p1.nc
```
schlunma, line 432: Same as comment above: This is bad, if we want to ensure backwards compatibility, we want to ensure that the `r0i0p0` ensemble is selected for `fx` variables is found when no further information is given in the recipe.

tests/integration/data_finder.yml
```diff
@@ -448,11 +448,11 @@ get_input_filelist:
...
       preprocessor: test_preproc
     available_files:
-      - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r1i1p1/v20110330/sftlf/sftlf_fx_HadGEM2-ES_historical_r0i0p0.nc
-      - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r1i1p1/v20110330/areacella/areacella_fx_HadGEM2-ES_historical_r0i0p0.nc
+      - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r1i1p1/v20110330/sftlf/sftlf_fx_HadGEM2-ES_historical_r1i1p1.nc
+      - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r1i1p1/v20110330/areacella/areacella_fx_HadGEM2-ES_historical_r1i1p1.nc
     dirs: []
     file_patterns:
-      - orog_fx_HadGEM2-ES_historical_r0i0p0*.nc
+      - orog_fx_HadGEM2-ES_historical_r1i1p1*.nc
```
schlunma, line 455: See comment in l. 430

schlunma (commented):
Sorry for the slow response - was on vacation last week.

I really like the idea of specifying a wildcard for `ensemble`. However, I have some issues with the current implementation:
- Right now, it only works for `fx`, not for `Ofx`, etc.
- Cases where the wildcard returns more than one results are ignored.
- There are no tests for the new functionality.

Moreover, I really do not like the removal of the automatic setting of `r0i0p0` for the `ensemble` key for CMIP5 fx variables. We would need to change all recipes that use this feature and test the output thoroughly. I don't think that's really desirable.

As an alternative, you could soften the check by performing this automatic change only if `variable['ensemble'] != '*'`, this ensures backwards compatibility with the current recipes but also allows us to use the new feature if we want. How does that sound?

valeriupredoi:
@schlunma very good points, I was a bit uneasy myself that's why I kept pinging you :grin: - the thing about `r0i0p0` is that if that's the only one that has fx files in it then it should be fine, it'll find those and get over with it, problem is that if there are fx files in eg `r1i1p1` then it'll prefer those, but again should be fine, since they *should* be the same files since are produced only once per model. About the eg `Omon` stuff for CMIP6 - yeah, I totally overlooked that case

esmvalcore/_data_finder.py
valeriupredoi: I guess we can generalize without the conditional on `/fx` which I totally overlooked and it's indeed extremely restrictive, but we can't do a `.*` coz that'll take forever and a half, maybe just a `*` on ensembles? Even that will take quite a bit with models with 30-40 ensembles in CMIP6

thomascrocker:
Ok, I've added some more commits to address these issues.

However, longer term I'm a bit uncomfortable with the way this all works...

Firstly, the way FX files are found is not consistent across projects, for CMIP5 the code automatically looks in `r0i0p0`, even if the user specifies a different ensemble member to be chosen in the fx_variables section of the preprocessor. For all other projects though, the software looks in the same ensemble member and experiment as the dataset in question, (unless the user overrides this behaviour manually). The intention of ESMValTool (I think) is that, where they exist, the software should automatically find fx files, but if a user specifies explicitly an ensemble and experiment (or in fact any other information) by explicitly defining an fx variable, then only that information should be searched for and returned.

The problem we have while for CMIP5 99% of the fx files are stored under `r0i0p0` the location of fx files in other projects is more complex. For CORDEX the fx files are mixed between being stored under `r0i0p0` or the ensemble of the rest of the data, and for CMIP6 it looks like the files can also be in different places depending on the model. For example, on JASMIN, fx files for HadGEM3-GC31-LL are stored under picontrol ensemble r1i1p1 `/badc/cmip6/data/CMIP6/CMIP/MOHC/HadGEM3-GC31-LL/piControl/r1i1p1f1/fx` and are not present under the historical experiment, but for MRI-ESM2-0 they are stored under the both the historical and piControl experiment.

However, for a user that wants to do multimodel research, they really shouldn't need to manually find the location of each fx file they need to use and then explicitly state it individually in their recipe with a separate preprocessor for each model.

I wonder if a better longer term approach might be to generate some specific rules for these projects, but make additions to the logic where fx variables are built in the recipes, e.g. somewhere around this code:
https://github.com/ESMValGroup/ESMValCore/blob/347d19d28c0b91891180a58a211ad7781c2d6999/esmvalcore/_recipe.py#L334
However, I'm not familiar with that bit of the software, so would be interested to see what people thought. If someone could dig up some documentation from CMIP6 regarding the location of fx files in the DRS, then perhaps some sensible rules could be setup that would cover CMIP5, CORDEX and CMIP6.

esmvalcore/_data_finder.py
schlunma: I'm still not convinced by this solution. Checking paths directly just doesn't feel right (e.g. I guess if we ever support Windows, this wouldn't work for paths that use `\` instead of `/`). Moreover, I find it weird to put that code into the function `_resolve_latestversion`.

As an alternative, how about you write a dedicated function `_resolve_fx_wildcards` (or a similar name) and put it right before the [call to `_resolve_latestversion`](https://github.com/ESMValGroup/ESMValCore/blob/87b5f489c6aa10d168e8924af521379fb10fb260/esmvalcore/_data_finder.py#L235)? You can pass the argument `variable` to this function, which can then be checked for `variable['frequency'] == 'fx'`. I think this is way safer than the current check of the path.

thomascrocker: Thanks for the suggestion, I have refactored my approach using a new function as you suggested. I've also added a couple of tests for CMIP6 and CORDEX using the wildcard functionality. The original FX tests remain in place and the code passes them also. Let me know what you think.

doc/recipe/preprocessor.rst
```diff
@@ -374,6 +374,15 @@ or alternatively:
             {'short_name': 'sftof', 'exp': 'piControl'}
             ]
 
+Additionally, it is possible to search across all ensembles and experiments when specifying
```
schlunma, line 377: In theory, the wildcards also work for all other keys (not only `ensemble` and experiment`). This could be mentioned as a side note here.

esmvalcore/_data_finder.py
```diff
@@ -253,41 +290,42 @@ def get_input_filelist(variable, rootpath, drs):
     if variable['project'] == 'CMIP5' and variable['frequency'] == 'fx':
```
schlunma, line 290:
```suggestion
    if all([variable['project'] == 'CMIP5',
            variable['frequency'] == 'fx',
            variable.get('ensemble') == '*']):
```

esmvalcore/_recipe.py
```diff
@@ -443,9 +443,9 @@ def _update_fx_files(step_name, settings, variable, config_user, fx_vars):
                 fx_var: fx_info
             })
 
-    logger.info('Using fx_files: %s for variable %s during step %s',
-                pformat(settings['add_fx_variables']['fx_variables']),
-                variable['short_name'], step_name)
+    logger.debug('Using fx_files: %s for variable %s during step %s',
+                 pformat(settings['add_fx_variables']['fx_variables']),
+                 variable['short_name'], step_name)
```
```suggestion
    logger.info('Using fx_files: %s for variable %s during step %s',
                 [n for n in settings['add_fx_variables']['fx_variables']],
                 variable['short_name'], step_name)
    logger.debug('Using fx_files: %s for variable %s during step %s',
                 pformat(settings['add_fx_variables']['fx_variables']),
                 variable['short_name'], step_name)
```

tests/integration/data_finder.yml
```diff
@@ -422,15 +422,75 @@ get_input_filelist:
...
+      - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r1i1p1/v20110330/sftlf/sftlf_fx_HadGEM2-ES_historical_r1i1p1.nc
     dirs:
       - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r0i0p0/v20110330/sftlf
     file_patterns:
       - sftlf_fx_HadGEM2-ES_historical_r0i0p0*.nc
     found_files:
       - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r0i0p0/v20110330/sftlf/sftlf_fx_HadGEM2-ES_historical_r0i0p0.nc
 
+  # CMIP6 wildcard specification for fx variable
+
```
schlunma, line 435:
```suggestion

  - drs: DKRZ
    variable:
      variable_group: test
      short_name: sftlf
      original_short_name: sftlf
      dataset: HadGEM2-ES
      project: CMIP5
      cmor_table: CMIP5
      institute: [INPE, MOHC]
      frequency: fx
      modeling_realm: [atmos]
      mip: fx
      exp: *
      ensemble: r1i1p1
      diagnostic: test_diag
      preprocessor: test_preproc
    available_files:
      - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r1i1p1/v20110330/sftlf/sftlf_fx_HadGEM2-ES_historical_r1i1p1.nc
      - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r2i1p1/v20110330/sftlf/sftlf_fx_HadGEM2-ES_historical_r2i1p1.nc
      - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r0i0p0/v20110330/sftlf/sftlf_fx_HadGEM2-ES_historical_r0i0p0.nc
    dirs:
      - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r0i0p0/v20110330/sftlf
    file_patterns:
      - sftlf_fx_HadGEM2-ES_historical_**.nc
    found_files:
      - MOHC/HadGEM2-ES/historical/fx/atmos/fx/r0i0p0/v20110330/sftlf/sftlf_fx_HadGEM2-ES_historical_r0i0p0.nc
```

To test prioritization of `r0i0p0`.

schlunma (commented):
Nice! Thanks for addressing my comments.

I have some minor comments and suggestions for two further tests.

Apart from that I think this PR is ready to go :tada:

esmvalcore/_data_finder.py
thomascrocker: I'm not sure of the intention here. Is it to not override the ensemble number for CMIP5 when the user specifies ``*`` for the ensemble? In that case is there a typo here and you mean ``!=`` in the last comparison?

schlunma: Yes, I mean `!=`. Sorry!!

tests/integration/data_finder.yml
thomascrocker: I added the two tests you had suggested. However, I think you meant
```
exp: historical
ensemble: '*'
```
rather than:
```
exp: *
ensemble: r1i1p1
```
So swapped them around.

thomascrocker:
> Nice! Thanks for addressing my comments.
> 
> I have some minor comments and suggestions for two further tests.
> 
> Apart from that I think this PR is ready to go tada

Great! Thanks very much for taking the time to review.

schlunma:
Just tested with real data and found two issues:

1. The wildcards still appear in the output files:

```
preproc
└── test
    └── orog_2
        ├── CMIP5_CanESM2_fx_historical_*_orog.nc
        └── metadata.yml
```
This does not look nice, but shouldn't be a big issue since we only allow one fx file, not multiple of them. To solve this we could use the contents of [`config-developer.yml`](https://github.com/ESMValGroup/ESMValCore/blob/master/esmvalcore/config-developer.yml) to reverse-engineer the `tags` from the path.

2. A much more serious problem: When checking for the existence of a directory, only the first part of the path (before `latestversion`) is checked: https://github.com/ESMValGroup/ESMValCore/blob/3961cfc0e55059c3677d07b94428615eced21cf5/esmvalcore/_data_finder.py#L162-L170
If you have a file structure that looks like this
```
            ├── fx
            │   ├── atmos
            │   │   └── fx
            │   │       ├── r0i0p0
            │   │       │   └── v20120410
            │   │       │       ├── orog
            │   │       │       │   └── orog_fx_CanESM2_historical_r0i0p0.nc -> /mnt/lustre02/work/bd0854/DATA/ESMValTool2/CMIP5_DKRZ/CCCma/CanESM2/historical/fx/atmos/fx/r0i0p0/v20120410/orog/orog_fx_CanESM2_historical_r0i0p0.nc
            │   │       │       └── sftlf
            │   │       │           └── sftlf_fx_CanESM2_historical_r0i0p0.nc -> /mnt/lustre02/work/bd0854/DATA/ESMValTool2/CMIP5_DKRZ/CCCma/CanESM2/historical/fx/atmos/fx/r0i0p0/v20120410/sftlf/sftlf_fx_CanESM2_historical_r0i0p0.nc
            │   │       ├── r1i1p1
            │   │       │   └── v20120410
            │   │       │       ├── areacella
            │   │       │       │   └── areacella_fx_CanESM2_historical_r1i1p1.nc -> /mnt/lustre02/work/bd0854/DATA/ESMValTool2/CMIP5_DKRZ/CCCma/CanESM2/historical/fx/atmos/fx/r0i0p0/v20120410/areacella/areacella_fx_CanESM2_historical_r0i0p0.nc
            │   │       │       └── sftlf
            │   │       │           └── sftlf_fx_CanESM2_historical_r1i1p1.nc -> /mnt/lustre02/work/bd0854/DATA/ESMValTool2/CMIP5_DKRZ/CCCma/CanESM2/historical/fx/atmos/fx/r0i0p0/v20120410/sftlf/sftlf_fx_CanESM2_historical_r0i0p0.nc
```
and ask for `areacella` with `ensemble: '*'`, it is never found, since `resolve_fx_wildcards` returns only the `r0i0p0` directory which does not contain `areacella`. I'm currently not sure how we could solve this, since we cannot look up the entire path before resolving `latestversion`, but `latestversion` can also only be setup with the wildcards resolved. Classic chicken-and-egg situation :unamused:

thomascrocker:
Well done for spotting those issues. I don't think I've ever worked with a folder structure that doesn't have the version folder after the experiment and ensemble information.

One possible solution might be to resolve fx wildcards and the version all at the same time. I.e. a function: `resolve_wildcards_and _version()` that replaces the `{latest_version}` tag with `*` (if found) and then globs the path. Then choose any folder containing `latest` if found, otherwise, regex search for `v\d{8}`, sort and pick the highest.

Would that work or have I missed something? I'll give it a try.

schlunma:
Yes, that sounds like an excellent solution! Couple things I recommend here:

- Make sure that you don't change the behavior of `_find_input_dirs`for non-fx variables, e.g. use something like:
```py
if variable["frequency"] == "fx":
    dirname = _resolve_fx_wildcards_and_version(dirname)
else:
    dirname = _resolve_latestversion(dirname)
```
so we do not need to worry that this PR changes other variables and do not need to test that.
- Use the function `_resolve_latestversion` in your new function `_resolve_fx_wildcards_and_version` for the actual resolving of the version. That should save you a lot of work and ensures that the behavior for fx and non-fx variables is similar.
- Add a unit test for the new function `_resolve_fx_wildcards_and_version`, so we can be sure that it does what we intend it to do.
- Add the test case I described [here](https://github.com/ESMValGroup/ESMValCore/pull/1082#issuecomment-845286058) to [`data_finder.yml`](https://github.com/ESMValGroup/ESMValCore/blob/3961cfc0e55059c3677d07b94428615eced21cf5/tests/integration/data_finder.yml) so we know that it actually works.

Thanks @thomascrocker for all the effort you put into this! Let me know if you need anything, happy to help!

thomascrocker:
@schlunma 
Ok, I have refactored again with an approach that I think works. 

The `_resolve_latestversion` function is unchanged. We now also have a `_resolve_globs` function and a `_resolve_wildcards_and_version` function, which uses both of those functions.

I added a test to cover the example with `areacella` mentioed above. I wasn't sure how to write a unit test for my new functions, since they are dependent on the file systems, so if you are able to do that, that would be great.

Cheers

zklaus:
How is this PR affected by https://github.com/ESMValGroup/ESMValCore/issues/1138#issuecomment-845202897 ?

thomascrocker:
> How is this PR affected by [#1138 (comment)](https://github.com/ESMValGroup/ESMValCore/issues/1138#issuecomment-845202897) ?

I think the suggestion made by @bouweandela is a good one (i.e. allowing specification of `cell_measures` and `ancillary_variables` in a dataset definition) this would ensure we can continue to have explicitly defined recipes, and should probably form it's own PR.
However even if that was implemented, the functionality added by this PR is still very useful from a convenience perspective, in that the user can very easily use fx variables for multiple models without needing to trawl the file system themselves when writing the recipe to determine exactly under which ensemble and experiment each fx file is stored for each dataset.

schlunma:
Hi @thomascrocker, unfortunately it still didn't work.

I implemented an alternative approach, extended the tests and extensively tested it with real data (e.g. I ran all example recipes). All worked well!

Can you please give it a try and see if it still works for your purposes?

@valeriupredoi it would be nice if you could have a look at the code again, there were quite some changes since you last reviewed it.

If all works out, I think this is ready to go :tada:

thomascrocker:
Thanks very much for fixing. I woke up this morning and realised I'd forgotten that the `_resolve_globs` function I wrote was still only returning one folder which may not be the parent of the actual full path we need to find further down the line.

I just tested with the recipe I'm developing and it works fine. Well done. 🎉

doc/recipe/preprocessor.rst
```diff
@@ -374,6 +374,17 @@ or alternatively:
             {'short_name': 'sftof', 'exp': 'piControl'}
             ]
 
+Additionally, it is possible to search across all ensembles and experiments (or any other keys)
+when specifying the fx variable, by using the ``*`` character, which is useful for some projects
+where the location of the fx files is not consistent.
+This makes it possible to search for fx files under multiple ensemble members or experiments.
+For example: ``ensemble: '*'``. Note that the ``*`` character must be quoted since ``*`` is a
+special charcter in YAML. This functionality is only supported for time invariant fx variables
+(i.e. frequency ``fx``). Note also that if multiple folders of matching fx files are found,
```
valeriupredoi, line 383:
```suggestion
(i.e. frequency ``fx`` or ``Ofx`` or ``Efx``). Note also that if multiple folders of matching fx files are found,
```

doc/recipe/preprocessor.rst
```diff
@@ -507,6 +529,16 @@ or alternatively:
...
+ESMValTool will default to ensemble r0i0p0 if it exists and then first folder found only 
+if it does not.
+
```
valeriupredoi, line 541: I think you can just write this paragraph once (at the top, or wherever you think it's best suited) and then just reference it via a Markdown reference, rather tha writing it three times. Up to you

esmvalcore/_data_finder.py
```diff
@@ -163,60 +162,114 @@ def _resolve_latestversion(dirname_template):
...
     return dirname_template
 
 
+def _resolve_wildcards_and_version(dirname, basepath, project, drs):
+    """Resolve wildcards and latestversion tag."""
+    if "{latestversion}" in dirname:
+        dirname_version_wildcard = dirname.replace("{latestversion}", "*")
+
+        # Find all directories that match the template
+        all_dirs = sorted(glob.glob(dirname_version_wildcard))
```
valeriupredoi, line 188: I think you want `reverse=True` here since `latest` should be first in line

esmvalcore/_data_finder.py
```diff
@@ -163,60 +162,114 @@ def _resolve_latestversion(dirname_template):
...
+
+        # Sort directories by version
+        all_dirs_dict = {}
+        for directory in all_dirs:
+            version = dir_to_var(
+                directory, basepath, project, drs)['latestversion']
```
valeriupredoi, line 194: if you find `latest` exit the loop and return quickly to save time (hence `latest` to be first in the sorted list of dirs

esmvalcore/_data_finder.py
```diff
@@ -163,60 +162,114 @@ def _resolve_latestversion(dirname_template):
...
     root = get_rootpath(rootpath, project)
-    path_template = _select_drs('input_dir', drs, project)
+    path_template = _select_drs("input_dir", drs, project)
 
     dirnames = []
     for dirname_template in _replace_tags(path_template, variable):
         for base_path in root:
             dirname = os.path.join(base_path, dirname_template)
-            dirname = _resolve_latestversion(dirname)
+            if variable["frequency"] == "fx" and "*" in dirname:
```
valeriupredoi, line 264: do you also want `Ofx`, `Efx`? I think those are all time-independent too

esmvalcore/_data_finder.py
```diff
@@ -231,65 +284,104 @@ def _find_input_dirs(variable, rootpath, drs):
...
 
 def get_input_filelist(variable, rootpath, drs):
     """Return the full path to input files."""
     # change ensemble to fixed r0i0p0 for fx variables
     # this is needed and is not a duplicate effort
-    if variable['project'] == 'CMIP5' and variable['frequency'] == 'fx':
+    if all([
+            variable['project'] == 'CMIP5', variable['frequency'] == 'fx',
+            variable.get('ensemble') != '*'
+    ]):
```
valeriupredoi, line 309: same here - how about `Ofx`, `Efx`?

esmvalcore/_data_finder.py
```diff
@@ -231,65 +284,104 @@ def _find_input_dirs(variable, rootpath, drs):
...
+        "{variable_group}",
+        "{dataset}_{mip}_{short_name}_{start_year}-{end_year}.nc",
     )
 
     outfile = template.format(**variable)
 
     return outfile
+
+
+def dir_to_var(dirname, basepath, project, drs):
```
valeriupredoi, line 355: this is a public function, so you either add numpy-style API documentation for it, or you make it private (even if private I'd reckon it needs a bit more description - lots of operations it performs and not well documented)

esmvalcore/_data_finder.py
```diff
@@ -231,65 +284,104 @@ def _find_input_dirs(variable, rootpath, drs):
...
+        regex = rf"{full_key.replace(key, '(.*)')}"
+        regex = regex.replace('{', '').replace('}', '')
+        matches = re.findall(regex, vals[idx])
+        while '' in matches:
+            matches.remove('')
+        if len(matches) != 1:
+            raise ValueError(
+                f"Regex pattern '{regex}' for '{full_key}' cannot be "
+                f"(uniquely) matched to element '{vals[idx]}' in directory "
+                f"'{dirname}'")
```
valeriupredoi, line 385: this is a brilliantly opaque error message, can you pls make it more user-friendly? Imagine you're a first time user and this pops up :laughing:

valeriupredoi (changes requested):
cheers muchly guys, sorry I left this behind in the jungle, I took another look - looks good to me, apart from a couple minor comments :+1:

bouweandela:
>>    How is this PR affected by #1138 (comment) ?
>
> I think the suggestion made by @bouweandela is a good one (i.e. allowing specification of cell_measures and ancillary_variables in a dataset definition) this would ensure we can continue to have explicitly defined recipes, and should probably form it's own PR.
> However even if that was implemented, the functionality added by this PR is still very useful from a convenience perspective, in that the user can very easily use fx variables for multiple models without needing to trawl the file system themselves when writing the recipe to determine exactly under which ensemble and experiment each fx file is stored for each dataset. (This also keeps recipes shorter and neater)

I'm a bit concerned about the order in which things are added here. I think it would be better to first implement support for per dataset ancillary/cell measure variables without wildcards. On top of that, a convenience method for recipe authors that automatically expands the wildcards and saves the expanded recipe in the output directory could be added. Does this pull request save the expanded recipe in the output directory?

If we follow the order proposed above, we are sure that any recipes that are developed can be added to the ESMValTool repository. On the other hand, if we just implement the wildcard feature here, the explicit way of specifying the data may never get implemented and that would be a major departure from the philosophy we've followed for recipes so far.

In order to start supporting wildcards, we would also need a unit test in the ESMValTool repository that checks that no recipes containing wildcards are added there.

schlunma:
> I'm a bit concerned about the order in which things are added here. I think it would be better to first implement support for per dataset ancillary/cell measure variables without wildcards. On top of that, a convenience method for recipe authors that automatically expands the wildcards and saves the expanded recipe in the output directory could be added. Does this pull request save the expanded recipe in the output directory?

No, it does not save the expanded recipe since there is no way to expand it yet. We do not support the syntax yet to define individual fx files for individual models (for usage in the preprocessor).

> If we follow the order proposed above, we are sure that any recipes that are developed can be added to the ESMValTool repository. On the other hand, if we just implement the wildcard feature here, the explicit way of specifying the data may never get implemented and that would be a major departure from the philosophy we've followed for recipes so far.

It would have been nice to know that before we put hours of coding into this PR...Adding the possibility to explicitly specify the fx files and an automatic expansion will require substantial changes in this PR, so a lot of work has been useless.

I cannot put any more work into this PR or the explicit expansion. Who will implement this?

In my opinion this feature here is really useful.

bouweandela:
> It would have been nice to know that before we put hours of coding into this PR...

I'm sorry to hear that. That's why I always recommend discussing plans in an issue and with the core development team before starting with the implementation. However, I think the work has not been useless: it will be much faster to get this right even with adding support for specifying fx variables per dataset first.

thomascrocker:
Is it going to be possible to merge this pull request imminently? Personally the functionality is very useful (/ essential to my work), and I'd rather not have to wait for implementation of per dataset ancillary variables, since that would probably take a while, plus it doesn't actually solve the problem that this PR is addressing.

esmvalcore/_data_finder.py
thomascrocker: see above

ledm:
This functionality is a life saver! Thanks everyone!

bouweandela:
>  Adding the possibility to explicitly specify the fx files and an automatic expansion will require substantial changes in this PR ..
> 
> I cannot put any more work into this PR or the explicit expansion. Who will implement this?

It looks like I will need functionality that is very similar to this for getting #1128 to work, so I can make a start with this over the next few months.

valeriupredoi:
@ledm keeps needing this functionality (and it's a very good functionality to have regardless) so I merged `main` and solved the conflicts. What are our plans wrt this PR, folks? I am thinking we should get it in at some point soon - there are still unresolved review comments and tests are still needed I believe, but I think this branch needs a revival in attention :grin:

schlunma:
I totally agree that this is a very good functionality and we should implement it very soon, but from the review comments above I think that there are different opinions on this on how to proceed. I'd like to quite my comment from earlier:

> > @bouweandela: I'm a bit concerned about the order in which things are added here. I think it would be better to first implement support for per dataset ancillary/cell measure variables without wildcards. On top of that, a convenience method for recipe authors that automatically expands the wildcards and saves the expanded recipe in the output directory could be added. Does this pull request save the expanded recipe in the output directory?
> 
> @schlunma: No, it does not save the expanded recipe since there is no way to expand it yet. We do not support the syntax yet to define individual fx files for individual models (for usage in the preprocessor).
> 
> > @bouweandela: If we follow the order proposed above, we are sure that any recipes that are developed can be added to the ESMValTool repository. On the other hand, if we just implement the wildcard feature here, the explicit way of specifying the data may never get implemented and that would be a major departure from the philosophy we've followed for recipes so far.
> 
> @schlunma: It would have been nice to know that before we put hours of coding into this PR...Adding the possibility to explicitly specify the fx files and an automatic expansion will require substantial changes in this PR, so a lot of work has been useless. I cannot put any more work into this PR or the explicit expansion. Who will implement this? In my opinion this feature here is really useful.

valeriupredoi:
right on Manu! That's why I am reopening the discussion floor here - we may have to wait for @bouweandela 's opinion since he's a bit away these days, but what I can add to this is that we want to wait until Bouwe lays down a plan for #1128 - in fact, I think we should all discuss that (the Core geeks, that is). Let's keep it as a detached prototype for now, until we decide that approach; it's gonna get stale very soon but I can periodically refresh it with `main`, this way we keep @ledm marginally happy too :beer:

ledm:
Running this, I'm currently getting the following error:

```
2021-11-17 14:12:03,282 UTC [11009] ERROR   Program terminated abnormally, see stack trace below for more information:
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_main.py", line 479, in run
    fire.Fire(ESMValTool())
  File "/home/users/ldemora/miniconda3/envs/esmvaltool/lib/python3.9/site-packages/fire/core.py", line 141, in Fire
  File "/home/users/ldemora/miniconda3/envs/esmvaltool/lib/python3.9/site-packages/fire/core.py", line 466, in _Fire
  File "/home/users/ldemora/miniconda3/envs/esmvaltool/lib/python3.9/site-packages/fire/core.py", line 681, in _CallAndUpdateTrace
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_main.py", line 423, in run
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_main.py", line 121, in process_recipe
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_recipe.py", line 65, in read_recipe_file
    return Recipe(raw_recipe,
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_recipe.py", line 1048, in __init__
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_recipe.py", line 1523, in initialize_tasks
    tasks = self._create_tasks()
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_recipe.py", line 1484, in _create_tasks
    new_tasks, failed = self._create_preprocessor_tasks(
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_recipe.py", line 1459, in _create_preprocessor_tasks
    task = _get_preprocessor_task(
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_recipe.py", line 1008, in _get_preprocessor_task
    task = _get_single_preprocessor_task(
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_recipe.py", line 864, in _get_single_preprocessor_task
    products = _get_preprocessor_products(
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_recipe.py", line 797, in _get_preprocessor_products
    ancestors = _get_ancestors(variable, config_user)
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_recipe.py", line 597, in _get_ancestors
    filenames) = _get_input_files(variable, config_user)
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_recipe.py", line 563, in _get_input_files
    filenames) = get_input_filelist(variable=variable,
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_data_finder.py", line 351, in get_input_filelist
    (files, dirnames, filenames) = _find_input_files(variable, rootpath, drs)
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_data_finder.py", line 335, in _find_input_files
    input_dirs = _find_input_dirs(variable, rootpath, drs)
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_data_finder.py", line 314, in _find_input_dirs
    matches = glob.glob(dirname)
  File "/home/users/ldemora/miniconda3/envs/esmvaltool/lib/python3.9/glob.py", line 22, in glob
    return list(iglob(pathname, recursive=recursive))
  File "/home/users/ldemora/miniconda3/envs/esmvaltool/lib/python3.9/glob.py", line 43, in _iglob
    dirname, basename = os.path.split(pathname)
  File "/home/users/ldemora/miniconda3/envs/esmvaltool/lib/python3.9/posixpath.py", line 103, in split
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```


As you can see, it looks like `dirname`  in `_find_input_dirs` is sometimes somehow set to `None`, which breaks glob. Here's the code where it happens:

```
def _find_input_dirs(variable, rootpath, drs):
    """Return a the full paths to input directories."""
    project = variable["project"]

    root = get_rootpath(rootpath, project)
    path_template = _select_drs("input_dir", drs, project)

    dirnames = []
    for dirname_template in _replace_tags(path_template, variable):
        for base_path in root:
            dirname = os.path.join(base_path, dirname_template)
            if variable["frequency"] == "fx" and "*" in dirname:
                dirname = _resolve_wildcards_and_version(dirname, base_path,
                                                         project, drs)
                var_from_dir = dir_to_var(dirname, base_path, project, drs)
                for (key, val) in variable.items():
                    if val == '*':
                        variable[key] = var_from_dir.get(key, '*')
            else:
                dirname = _resolve_latestversion(dirname)
            print('dirname', dirname)
            matches = glob.glob(dirname)
            matches = [match for match in matches if os.path.isdir(match)]
            if matches:
                for match in matches:
                    dirnames.append(match)
            else:
                logger.debug("Skipping non-existent %s", dirname)

    return dirnames

```

valeriupredoi:
yeh that's why the tests are failing too, lemme investigado :+1:

esmvalcore/_data_finder.py
```diff
@@ -247,18 +292,24 @@ def get_rootpath(rootpath, project):
...
+            if variable["frequency"] == "fx" and "*" in dirname:
+                dirname = _resolve_wildcards_and_version(dirname, base_path,
+                                                         project, drs)
+                var_from_dir = dir_to_var(dirname, base_path, project, drs)
+                for (key, val) in variable.items():
+                    if val == '*':
+                        variable[key] = var_from_dir.get(key, '*')
+            else:
+                dirname = _resolve_latestversion(dirname)
             matches = glob.glob(dirname)
```
ledm, line 313: This line tries to glob a `None`, which breaks glob. Try to skip it, if a None is returned
```suggestion
            if not dirname:
                continue
            matches = glob.glob(dirname)
```

valeriupredoi: that's the one, hang on am gonna commit that from local repo, am running the tests now

valeriupredoi: done

bouweandela:
> I think it would be better to first implement support for per dataset ancillary/cell measure variables without wildcards. 

As soon as I'm done with the IS-ENES3 reporting stuff, hopefully next week, I will have time to start working on improving the dataset definition in the tool and that could then be used for better defining ancillary files as well as for finding out where the data lives, which we need for the distributed stuff.

ledm:
How does this PR handle wildcards or absencses in the `activity` field? I've been running it without specifying an activity. This is so that I can use the same preprocessor for both `historical` and `ssp???` experiments. 

ie:
```
    volume_statistics:
      operator: mean
      fx_variables:
      - {grid: gn, short_name: volcello, exp: '*', ensemble: '*', }
```

However, it looks like the data_finder is unable to find `volcello`. It's looking for the ssp volcello in places like this:
```
/badc/cmip6/data/CMIP6/CMIP/MOHC/UKESM1-0-LL/*/*/Omon/volcello/gn/{latestversion}
```
but it doesn't try the correct activity path:
```
/badc/cmip6/data/CMIP6/ScenarioMIP/MOHC/UKESM1-0-LL/*/*/Omon/volcello/gn/{latestversion}
```

My work around is to create a different preprocessor for historical activity and for scenarioMIP activity, but it would be better if the activity wildcard were addressed, somehow. Cheers!

ledm:
Just adding another comment, that I think that the problem here is because our previous edit [33cd03a](https://github.com/ESMValGroup/ESMValCore/pull/1082/commits/33cd03a94e4bb3515132e834125f679fd07a72da) didn't solve this problem, but hid it.

ledm:
Continuing to work on this today (instead of paying attending to the  workshop - sorry!), I think one of the minor problems with this PR is that whenever it checks whether  `mip == 'fx'`. There are lots of places beyond `fx` where fx files can live. For instance, `Ofx` for Ocean fx files. Similarly, some fx data in some models can be time dependent, like volcello in `Omon` or `Oyr` or even `Oday`.

ledm:
While this PR is very powerful and I love it, it's power has actually hid some missing data from me. It failed to find a volcello file because that file did not exist on jasmin. However, the output error message was:
```
2021-11-24 22:33:46,779 UTC [43615] ERROR   esmvalcore._main:491 Program terminated abnormally, see stack trace below for more information:
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/preprocessor/_volume.py", line 206, in volume_statistics
    grid_volume = cube.cell_measure('ocean_volume').core_data()
  File "/home/users/ldemora/miniconda3/envs/esmvaltool/lib/python3.9/site-packages/iris/cube.py", line 1976, in cell_measure
    raise iris.exceptions.CellMeasureNotFoundError(msg)
iris.exceptions.CellMeasureNotFoundError: 'Expected to find exactly 1 ocean_volume cell_measure, but found none.'

  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_main.py", line 479, in run
    fire.Fire(ESMValTool())
  File "/home/users/ldemora/miniconda3/envs/esmvaltool/lib/python3.9/site-packages/fire/core.py", line 141, in Fire
  File "/home/users/ldemora/miniconda3/envs/esmvaltool/lib/python3.9/site-packages/fire/core.py", line 466, in _Fire
  File "/home/users/ldemora/miniconda3/envs/esmvaltool/lib/python3.9/site-packages/fire/core.py", line 681, in _CallAndUpdateTrace
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_main.py", line 423, in run
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_main.py", line 125, in process_recipe
    recipe.run()
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_recipe.py", line 1563, in run
    self.tasks.run(max_parallel_tasks=self._cfg['max_parallel_tasks'])
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_task.py", line 707, in run
    self._run_sequential()
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_task.py", line 718, in _run_sequential
    task.run()
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_task.py", line 249, in run
    input_files.extend(task.run())
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/_task.py", line 253, in run
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/preprocessor/__init__.py", line 489, in _run
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/preprocessor/__init__.py", line 356, in apply
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/preprocessor/__init__.py", line 302, in preprocess
    result.append(_run_preproc_function(function, item, settings))
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/preprocessor/__init__.py", line 285, in _run_preproc_function
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/preprocessor/_volume.py", line 211, in volume_statistics
    grid_volume = calculate_volume(cube)
  File "/home/users/ldemora/workspace/ESMValTool/ESMValCore/esmvalcore/preprocessor/_volume.py", line 165, in calculate_volume
    area = iris.analysis.cartography.area_weights(cube)
  File "/home/users/ldemora/miniconda3/envs/esmvaltool/lib/python3.9/site-packages/iris/analysis/cartography.py", line 407, in area_weights
    raise iris.exceptions.CoordinateMultiDimError(lat)
iris.exceptions.CoordinateMultiDimError: Multi-dimensional coordinate not supported: 'latitude'
2021-11-24 22:33:46,799 UTC [43615] INFO    esmvalcore._main:495
If you have a question or need help, please start a new discussion on https://github.com/ESMValGroup/ESMValTool/discussions
If you suspect this is a bug, please open an issue on https://github.com/ESMValGroup/ESMValTool/issues
To make it easier to find out what the problem is, please consider attaching the files run/recipe_*.yml and run/main_log_debug.txt from the output directory.
```

So the fatal error was actually occured because we're trying to calculate the volume without fx files. The absence of fx files should have caused the fatal error, but the preprocessing was allowed to continue. 

We should:

1. Add a error message in volume_statistics saying that the FX was not found, and turn on a fatal.
2. Add a fatal message earlier in the process if the fx file is not found and not calculable.

thomascrocker:
> Continuing to work on this today (instead of paying attending to the workshop - sorry!), I think one of the minor problems with this PR is that whenever it checks whether `mip == 'fx'`. There are lots of places beyond `fx` where fx files can live. For instance, `Ofx` for Ocean fx files. Similarly, some fx data in some models can be time dependent, like volcello in `Omon` or `Oyr` or even `Oday`.

I don't think this is an issue. (Unless the code has changed significantly since I last looked at it). As far as I'm aware the code doesn't check for `mip==fx` it checks for `frequency==fx`. This means it should pick up any fx files that are time invariant, so it would still find `Ofx` files. I admit though I don't think time dependent fx data would be captured (as frequency would not be `fx`), unfortunately I've never worked with this type of FX data before so I'm not sure the best way to approach it, or what the convention is (if any) around where they are stored.

valeriupredoi:
how about we strive to get this in 2.5 - it looks a treat and I reckon it adds very good value and functionality - at least @ledm will be happy and I'll have quieter Fridays :laughing: @schlunma what do you reckon?

schlunma:
To quote myself a second time: :smile:

I think this is totally useful, but before continuing with this we (i.e., the @ESMValGroup/esmvaltool-coreteam) need to decide if we want to have this feature implemented like this or not. I don't think we've reached a consensus yet. 

> @schlunma: I totally agree that this is a very good functionality and we should implement it very soon, but from the review comments above I think that there are different opinions on this on how to proceed. I'd like to quite my comment from earlier:
> 
> > > @bouweandela: I'm a bit concerned about the order in which things are added here. I think it would be better to first implement support for per dataset ancillary/cell measure variables without wildcards. On top of that, a convenience method for recipe authors that automatically expands the wildcards and saves the expanded recipe in the output directory could be added. Does this pull request save the expanded recipe in the output directory?
> > 
> > 
> > @schlunma: No, it does not save the expanded recipe since there is no way to expand it yet. We do not support the syntax yet to define individual fx files for individual models (for usage in the preprocessor).
> > > @bouweandela: If we follow the order proposed above, we are sure that any recipes that are developed can be added to the ESMValTool repository. On the other hand, if we just implement the wildcard feature here, the explicit way of specifying the data may never get implemented and that would be a major departure from the philosophy we've followed for recipes so far.
> > 
> > 
> > @schlunma: It would have been nice to know that before we put hours of coding into this PR...Adding the possibility to explicitly specify the fx files and an automatic expansion will require substantial changes in this PR, so a lot of work has been useless. I cannot put any more work into this PR or the explicit expansion. Who will implement this? In my opinion this feature here is really useful.

valeriupredoi:
that "@schlunma" guy sure looks like he know his stuff in that convo :grin: Yeah, am totally for getting this in, we can tweak later, but at least we'd keep some scientist users happy for a while, and I can have lunches before 3pm :hamburger:

ledm:
In a perfect world, we wouldn't need this as every model would have it's fx data in the correct place and we wouldn't need wildcards. More pragmatically, I've been using this code for at least 6 months and it has been massively labour saving. I have no doubt that others would find it just as helpful as I have in the short term. 

In the longer term, while I don't fully understand what @zklaus and @bouweandela are asking for instead of this, it sounds like it may require a significant re-write of good parts of this code. (And may be superseded by up-coming changes in iris?) I think that so long as it's passing tests, and works as described, merging this now might be helpful for a lot of people. 

It's a fantastic feature that is very powerful and very user friendly.

bouweandela:
The problem with this is that **recipes that use this feature are not reproducible and you will never notice**. Let me try to get the fx file specification in the recipe right for version v2.6 and if I don't manage by then you can merge this. Does that sound like a compromise?

ledm:
As I'm still in the process of using this code, (it's really the only tool for the job), can you please help with the conflicts in _data_finder at some point? There's something weird going on there that's preventing me from running.

schlunma:
I can have a look at the conflicts, but probably won't have the time to test everything.

schlunma:
I solved the conflicts now, let me know if the branch works again @ledm

ledm:
You're a hero. Thanks @schlunma

ledm:
I used this branch without the most recent patch (5c455ed) and with several additional fixes but no significant changes, to produce an analysis over the weekend on jasmin. The preprocessor ran to completion, and everything appears fine in the output text, however all netcdfs in preproc are fully masked and contain no data. It's really weird and I don't get it. Any ideas?

zklaus:
@bouweandela, I think @ledm's comment refers to your earlier comment

> The problem with this is that **recipes that use this feature are not reproducible and you will never notice**. Let me try to get the fx file specification in the recipe right for version v2.6 and if I don't manage by then you can merge this. Does that sound like a compromise?

So... what do you think? Did we make progress? Is there an issue or perhaps even already a PR to track here?

bouweandela:
Only [a branch](https://github.com/ESMValGroup/ESMValCore/compare/main...split-recipe) so far, the deadline for the release seemed so far away ;-) I'll try to make some progress today.

bouweandela:
Just fixed the merge conflict and turned the branch into a draft pull request: #1609

bouweandela:
It seems unlikely that I'll get #1609 finished in time for v2.6, it turns out to be quite a bit of work, but I'm off to a good start so can I ask that you please wait with this for one more release?

ledm:
So we haven't merged this yet? 

I still need to use wildcards for FX files every time that I run ESMValTool, and probably will need to keep doing so for the foreseeable future. What is the suggested best alternative practice instead of using this method?  #1609 is still a draft and this #1082 conflicts with main now.

bouweandela:
Hi @ledm, #1609 is almost ready to be used, maybe I can show you how to use it and you can try it out at the upcoming workshop?

ledm:
Yes please, I'm incredibly frustrated with this. FX files are crucial, and yet they have never worked properly or had a flexible enough interface. It's been four years that I've been raising this as an issue and it's still broken and undocumented. 

The solutions that I wrote for this work in 2021 and earlier no longer work, so now I'm about to submit two different papers on the basis of this PR, and neither of them can be reproduced right now (or re-run!) This is the exact reason why you initially declined this PR - reproducibility of recipes in publications. 

However, for obvious reasons, I'm extremely keen to not have to re-write everything at the last minute right before submission.

ledm:
Also, there's no changes to documentation in #1609 yet, so if you want me to use it, you can expect a lot of emails/whatsapp messages! (Just ask @valeriupredoi about that, lol!)

bouweandela:
I added an example recipe to the description of the pull request. I'm still working on it so expect things to break, especially commits called WIP (Work In Progress), but the example runs if you check out commit https://github.com/ESMValGroup/ESMValCore/pull/1609/commits/98f052f55629667b9d51c34067327f3e1f855542 and set `offline: false` in config-user.yml.

bouweandela:
> It's been four years that I've been raising this as an issue and it's still broken and undocumented.

Raising an issue is a good start, but in the end, someone needs to do the actual work.

Note that the current functionality is actually [documented](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/preprocessor.html#fx-variables-as-cell-measures-or-ancillary-variables), it just doesn't work very well because it assumes a certain organization of files that just isn't the reality on ESGF.

schlunma:
Closing this but keeping the branch alive; this has been properly implemented in #1609.

Issue #890, opened 2020-11-30
Preprocessor `multimodel_statistics` fails when data have no time dimension


stefsmeets:
**Describe the bug**
Hi all, I'm developing some tests for the multimodel statistics using real data (#856), and I'm coming accross this bug:

Running multimodel statistics with a list of cubes with **no time dimension**, will result in
`ValueError: Cannot guess bounds for a coordinate of length 1.`

The bug can be reproduced by:
```python
cubes = [cube[0] for cube in cubes]
multi_model_statistics(cubes, span='full', statistics=['mean'])
```

<details>
  <summary>See the full stack trace below.</summary>

```
timeseries_cubes_month = [<iris 'Cube' of air_temperature / (K) (time: 14; air_pressure: 2; latitude: 3; longitude: 2)>, <iris 'Cube' of air_te... 3; longitude: 2)>, <iris 'Cube' of air_temperature / (K) (time: 14; air_pressure: 2; latitude: 3; longitude: 2)>, ...]

    @pytest.mark.functional
    # @pytest.mark.xfail('ValueError')
    def test_multimodel_no_time_dimension(timeseries_cubes_month):
        """Test statistic without time dimension using monthly data."""
        span = 'full'
        cubes = timeseries_cubes_month
        cubes = [cube[0] for cube in cubes]
        # ValueError: Cannot guess bounds for a coordinate of length 1.
>       multimodel_test(cubes, span=span, statistic='mean')

tests/functional/test_multimodel.py:227:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/functional/test_multimodel.py:125: in multimodel_test
    output = multi_model_statistics(cubes, span=span, statistics=statistics)
esmvalcore/preprocessor/_multimodel.py:439: in multi_model_statistics
    statistic_cube = _assemble_full_data(cubes, statistic)
esmvalcore/preprocessor/_multimodel.py:323: in _assemble_full_data
    time_axis = [float(fl) for fl in _monthly_t(cubes)]
esmvalcore/preprocessor/_multimodel.py:277: in _monthly_t
    days = {day for cube in cubes for day in _datetime_to_int_days(cube)}
[...]

PR #1097, opened 2021-05-05
demote time coordinate in climate_statistics

### LINKED ISSUE #1018: Preprocessor chain 'climate_statistics' and 'multimodel_statistics' fails as time dimension is not present

tomaslovato:
## Description

The construction of a preprocessor that computes, e.g., first `climate_statistics` and then `multimodel_statistics` fails with a missing time coordinate error from the latter because the time coordinate is removed from the cube in 
https://github.com/ESMValGroup/ESMValCore/blob/7f8406bdafe898c05d37f398457023b228e79324/esmvalcore/preprocessor/_time.py#L537-L539

The proposed changes aims to demote the time coordinate instead of removing it (see details in #1018)

Closes #1018

Link to documentation:

***

***


tomaslovato:
@valeriupredoi "you're my only hope" 🚀

schlunma:
Unfortunately I don't know why the original `time` coordinate is removed here. If the original author had a good reason for this then I really don't want to break anything here. Maybe @valeriupredoi knows more about this?

I personally like your change (i.e., keeping the `time` coordinate), but if we do this, we need to do this for all time-related preprocessors, need to add some unit tests and also need to perform some tests if this change breaks anything (I don't think it would, but you'll never know 😁)

However, I think once #968 is merged this change shouldn't be necessary anymore, since #968 generalized `multi_model_statistics` in such a way that it can deal with input data with arbitrary dimensions. Would it be okay for you to wait for that? I hope that this PR gets merged soon!

tomaslovato:
> Unfortunately I don't know why the original `time` coordinate is removed here. If the original author had a good reason for this then I really don't want to break anything here. Maybe @valeriupredoi knows more about this?
> 
> I personally like your change (i.e., keeping the `time` coordinate), but if we do this, we need to do this for all time-related preprocessors, need to add some unit tests and also need to perform some tests if this change breaks anything (I don't think it would, but you'll never know 😁)
> 
> However, I think once #968 is merged this change shouldn't be necessary anymore, since #968 generalized `multi_model_statistics` in such a way that it can deal with input data with arbitrary dimensions. Would it be okay for you to wait for that? I hope that this PR gets merged soon!

Thanks @schlunma for having a look a this! Certainly some testing will be a healthy step!
After digging a bit in the development, I found that time coord was removed in bfe1456e by @jvegasbsc but I did not found any specific comment for that choice in the old PR #87.

So far, only `climate_statistics` remove `time` in favour of either `month_number`, `day_of_month`, `day_of_year` (maybe also season but not checked yet) while other functions like `annual_statistics` or `decadal statistics` preserve time.

About #968 it is my understanding that in `multi_model_statistics` that time is always required, at least I got this by looking at the code and the function preamble https://github.com/ESMValGroup/ESMValCore/blob/99159ac7fda6fff442725502c988951d1bbc78b5/esmvalcore/preprocessor/_multimodel.py#L393-L395

schlunma:
> About #968 it is my understanding that in `multi_model_statistics` time is always required, at least I got this by looking at the code and the function preamble
> 
> https://github.com/ESMValGroup/ESMValCore/blob/99159ac7fda6fff442725502c988951d1bbc78b5/esmvalcore/preprocessor/_multimodel.py#L393-L395

I certainly hope that's not the case, we already have lots of issues that describe the problem (e.g., #890, #891) and I think they are considered in #968. Maybe @Peter9192 or @stefsmeets can briefly clarify that for us?

stefsmeets:
> > About #968 it is my understanding that in `multi_model_statistics` time is always required, at least I got this by looking at the code and the function preamble
> > https://github.com/ESMValGroup/ESMValCore/blob/99159ac7fda6fff442725502c988951d1bbc78b5/esmvalcore/preprocessor/_multimodel.py#L393-L395
> 
> I certainly hope that's not the case, we already have lots of issues that describe the problem (e.g., #890, #891) and I think they are considered in #968. Maybe @Peter9192 or @stefsmeets can briefly clarify that for us?

#890 are not resolved with the #968. It's not the goal of the PR, but it may be tackled in a follow-up PR, since #968 already simplifies the time implementation. At the moment multimodel statistics function fails because iris cannot guess the bounds of a cube with no time coordinate.

schlunma:
All right, then I misunderstood this, sorry. As I said, I'm in favor for merging this PR (once it has been tested and some unit tests), but I think we need some more discussion about this, especially about the original purpose of the removal of the `time` coordinate.

Peter9192:
To add to Stef's comment: making mmstats independent of time is a bit of a challenge since we have the span argument that is explicitly about time atm. So to do that you'd have to reinterpret this argument.

zklaus:
Is the time coordinate that comes out of `climate_statistics` still adequate for the data? It is possible to construct an appropriate time coordinate following [CF conventions 1.8, Section 7.4 "Climatological Statistics"](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#climatological-statistics), but I am not sure that merely aggregating the cube gives you the correct information.

@tomaslovato, if you look at the demoted time coordinate in the output, does the information make sense?

zklaus:
Hm. Iris has gained the ability to deal with climatological bounds relatively recently. I think it would be worthwhile to checkout how we might leverage that, but it seems to me the problem here really is #890, i.e. a bug in multimodel statistics, that really should not depend on the time axis.

So my suggestion is to not change climate_statistics for now, but rather fix #890 and open a feature request to include proper treatment of climatologies in the future. That might be as simple as saying `cube.coord('time').climatological = True` in the preprocessor, but further study of the documentation of iris and of the cf conventions is needed to confirm.

tomaslovato:
@zklaus As I see it by now the time (and bounds) are revised within multi-model stats by `_unify_time_coordinates` so I won't be much worried about it. Moreover, the cube generated from `climate_statistics` still retain its original main axis based on the monthly, yearly or daily counter ...
I personally think that it is useful to allow users creating a processing chain to compute climatologies and multimodel stats.

zklaus:
> I personally think that it is useful to allow users creating a processing chain to compute climatologies and multimodel stats.

Oh definitely. It just shouldn't depend on a time axis. That is simply a bug in multimodel statistics.

Peter9192:
> by now the time (and bounds) are revised within multi-model stats by `_unify_time_coordinates` so I won't be much worried about it

See https://github.com/ESMValGroup/ESMValCore/issues/744

tomaslovato:
@zklaus Given that the issue with time dependency in `multimodel` statistics won't be fixed in the short period, should we add this fix to enable the correct concatenation of these preprocessors?

bouweandela:
Now that #1150 has been merged, I think this is no longer needed. Could you please check @tomaslovato?

tomaslovato:
@bouweandela I made a test with the recipe detailed here below and it fails with the following 
```iris.exceptions.CoordinateNotFoundError: 'Expected to find exactly 1 time coordinate, but found none.'``` (see also
[main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/6654228/main_log_debug.txt)

It is my guess that the code is crashing in the check for aligned cubes here
https://github.com/ESMValGroup/ESMValCore/blob/f9481e5504c8860205b3e64f4668bb94a83dfe25/esmvalcore/preprocessor/_multimodel.py#L313
that goes directly into the check for `time` coords 
https://github.com/ESMValGroup/ESMValCore/blob/f9481e5504c8860205b3e64f4668bb94a83dfe25/esmvalcore/preprocessor/_multimodel.py#L69-L72

I frankly don't see why it is so terrible in demoting the `time` to an auxiliary in this operation and have a fix for the time being ... otherwise it would be necessary to instruct the `_multimodel.py` to handle similarly to time also `year`, `month`, `day` coords derived in `climate_statistics`. The latter seems to me very unlikely as it will generate a significant dependence form changes in `climate_statistics`.

<details>
  
  ```py
documentation:
  description: test
  authors:
    - lovato_tomas

datasets:
  - {dataset: CMCC-ESM2,     grid: gn, project: CMIP6,  exp: historical,  ensemble: r1i1p1f1,  start_year: 2000,  end_year: 2014}
  - {dataset: UKESM1-0-LL, grid: gn, project: CMIP6,  exp: historical,  ensemble: r1i1p1f2,  start_year: 2001,  end_year: 2004}


preprocessors:
  prep_test:
    custom_order: true
    extract_time:
      { start_year: 2000, start_month: 1, start_day: 1, end_year: 2001, end_month: 12, end_day: 31 }
    regrid:
      target_grid: 2x2
      scheme: linear
    climate_statistics:
      operator: mean
      period: month
    multi_model_statistics:
      span: overlap
      statistics: [mean]
      exclude: ['WOA']

diagnostics:
  test:
    variables:
      tos:
         preprocessor: prep_test
         mip: Omon
    scripts:  ```
</details>

**Comment by tomaslovato** [comment]
@bouweandela there is one thing that I forgot to add in my previous comment, 
I made a test with `annual_statistics` operator and after the time agregation the resulting cube has `time` as Dimcoord and `year` as an auxiliary coord, which looks the opposite of what is done in `climate_statistics` ...

```
 2021-06-15 09:16:28,860 UTC [296784] DEBUG   esmvalcore.preprocessor:279 Running area_statistics(sea_surface_temperature                                                    (time: 2; first spatial index for variables stored on an unstructured grid: 292; second spatial index for variables stored on an unstructured grid: 362)
     Dimension coordinates:
          time                                                                  x                                                                    -                                                                       -
          first spatial index for variables stored on an unstructured grid      -                                                                    x                                                                       -
          second spatial index for variables stored on an unstructured grid     -                                                                    -                                                                       x
     Auxiliary coordinates:
          year                                                                  x                                                                    -                                                                       -
          latitude                                                              -                                                                    x                                                                       x
          longitude                                                             -                                                                    x                                                                       x
     Attributes:
```

tomaslovato:
> What is needed, is a bit of code in the multimodel statistics preprocessor function to align the time coordinate from all the input models in case that coordinate has only 1 value
Now that I know this I feel kind of lucky since I always tried with monthly climatological stats !!

> I agree that there should be a single consistent behaviour. Probably it would be best to just keep the time coordinate as it is after the statistics operation, because that is how the iris authors thought it made most sense.

I fully support the consistent behaviour toward iris philosophy, so if we agree on that I can at least give a try in modifying  `climate_statistics` by retaining the time resulting after the computations and demoting the coordinate derived from climate aggregation ...

tomaslovato:
@bouweandela I commited a modification to the code of `climate_statistics` to produce a cube that retains `time` as Coord and the `climate coordinate` as an auxiliary one to align with the general behaviour of iris. 
In this way the problem with time handling is also fixed.

[main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/6673103/main_log_debug.txt)

bouweandela:
I suspect that the time coordinate on the resulting multmodel statistic cube is completely wrong if you try to fix it like this. Because we're actively working on fixing the multimodel statistics preprocessor, I think it would be best to complete that work before starting to modify others just so they work with the omissions in the current implementation of the multimodel statistics.

If you need a workaround for now, you could also compute the multimodel statistics before the climate statistics using a [custom preprocessor order](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/overview.html#recipe-section-preprocessors).

tomaslovato:
> I suspect that the time coordinate on the resulting multimodel statistic cube is completely wrong if you try to fix it like this.

Actually it should not, as there is the alignement of cubes `time` coord before computing the multimodel statistics https://github.com/ESMValGroup/ESMValCore/blob/f9481e5504c8860205b3e64f4668bb94a83dfe25/esmvalcore/preprocessor/_multimodel.py#L77-L91

which is indeed at the origin of this issue (see my comments above) since there is the missing `time` coord in the cube  that  is explicitly requested by this align procedure, namely in function `_get_consistent_time_unit`.

> 
> If you need a workaround for now, you could also compute the multimodel statistics before the climate statistics using a [custom preprocessor order](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/overview.html#recipe-section-preprocessors).

I tested the option you suggest, but it works at the cost of a larger computational burden...

Issue #876, opened 2020-11-23
cftime>=1.3.0 completes years<1000 with zeros instead of spaces


valeriupredoi:
as the title says, now we have zeros, see below:

run this script:

```python
import cftime
from cf_units import Unit, date2num, num2date
d = cftime.DatetimeGregorian(40, 1, 16, 12, 0, 0, 0)
print([d.strftime('%Y%m%d%H%M')])
```

with `cftime<1.3.0` the result is `['  4001161200']` whereas with `cftime=1.3.0` the result is `['004001161200']`

As a consequence, we have two failed tests, see [here](https://github.com/ESMValGroup/ESMValCore/runs/1431151389?check_suite_focus=true)

I propose we either place a check on version in the test or we pin the package, either or - what do you think Javi and Bouwe?

bouweandela:
Pin cftime to the most recent version and fix the tests

valeriupredoi:
aye, just aboot to do that, capt'n :ship:

PR #1124, opened 2021-05-12
Add basic support for variable mappings

### LINKED ISSUE #1157: Support extra information for additional datasets
Sometimes it is useful to provide extra information for the loading of data, particularly in the case of native model data, or observational or other data, that generally follows the established standards, but is not part of the big supported projects like CMIP, CORDEX, obs4MIPs.

To support this, it would be nice to have a facility that allows the passing of such extra information to both the data finder and possibly needed fixes.

zklaus:
## Description

Closes #1157 

Link to documentation: https://esmvaltool--1124.org.readthedocs.build/projects/ESMValCore/en/1124/develop/extra_facets.html#

***

***


esmvalcore/_recipe.py
```diff
@@ -1061,6 +1062,12 @@ def expand_ensemble(variable):
 
         return expanded
 
+    def _add_project_variable_mappings(self, variable):
+        mappings = get_variable_mappings(variable["project"],
+                                         variable["dataset"])
+        mapping = mappings[variable["mip"]][variable["short_name"]]
```
senesis, line 1068: "mip" shouldn't be a key in mapping : the actual need for mapping is to describe various alternate variable names, and maybe also scaling/offset attributes; these attributes depend on the base variable name, but not depend  on the CMIP table name, which depend on data frequency; and there is no ambiguity of base variable name across tables (except maybe for one single variable in CMIP6)

zklaus: Hm. I am not sure about this. For example in era5 data, we might want to derive daily means from hourly data and monthly means from, well, monthly data. Those live in different files and thus we need a different mapping. For models, the needs may be different, but first, the overhead is relatively low because the same variable appears only in a handful of tables, and second, it may still sometimes be useful. For example, occasionally there is both an instantaneous version of, say `pr`, and a time integrated one. One should  be used for tables like `3hrPt`, the other for tables like `3hr`.

esmvalcore/cmor/_fixes/fix.py
```diff
@@ -2,15 +2,30 @@
...
 
 from ..table import CMOR_TABLES
 
 
+@lru_cache
+def get_variable_mappings(project, dataset):
+    DEFAULT_PATH = (Path(__file__).parents[0] / project /
+                    f"{dataset}-mappings.yml")
+    mapping_path = CFG.get(project, {}).get("mapping_path", DEFAULT_PATH)
+    with open(mapping_path, "r") as mapping_file:
```
senesis, line 20: Must handle the cases where there is no mapping file, neither in default location, nor specified in the project definition

esmvalcore/_recipe.py
senesis, line 1068: A mapping dict is not supposed to have keys for all mips, and when it has, not all variables have necessary an entry

esmvalcore/cmor/_fixes/fix.py
```diff
@@ -2,15 +2,30 @@
...
+import yaml
+
+from esmvalcore._config._config import CFG
 
 from ..table import CMOR_TABLES
 
 
+@lru_cache
+def get_variable_mappings(project, dataset):
+    DEFAULT_PATH = (Path(__file__).parents[0] / project /
```
senesis, line 17: project names and dataset names are usually used in lower case when looking in source directories

esmvalcore/_config/_config.py
```diff
@@ -14,14 +14,24 @@
...
+    from importlib.resources import files as importlib_files
+except ImportError:
+    from importlib_resources import files as importlib_files
+
 
 @lru_cache
 def _get_project_mappings(project, dataset):
-    DEFAULT_PATH = (Path(__file__).parents[0] / project
-                    / f"{dataset}-mappings.yml")
-    mapping_path = CFG.get(project, {}).get("mapping_path", DEFAULT_PATH)
```
senesis: Deleting the possibility to have a mapping path defined in the config-developer file means keeping users unable to re-define the mapping when they change the way their model's outputs
(but maybe I misunderstand which is the actual value of SEARCH_PATHS

zklaus:
I think this is functionally ready. It is still missing documentation and tests, plus the checks fail.

On the failing checks, I think all of them are spurious. In detail:

Codacy complains about two things, namely that `_get_variable_details_for_project` is too long a name, and that I disable one error locally. The local disabling is the right thing to do because it is complaining about a missing import precisely where I am testing for the right python version to allow the import.

The commit checks fail as follows:
- conda-build exceeds time limit
- documentation: success
- test: The tests succeed, but the uploading of the coverage fails due to missing CODACY_PROJECT_TOKEN, presumably because this branch lives in my fork.
- install and develop: these fail due to what looks like a typo in the test for a cmip5 fix:
```
>      self.assertEqual(dates[0].strftime('%Y%m%d%H%M'), ' 30001161200')
E       AssertionError: '30001161200' != ' 30001161200'
```

Questions:
@bouweandela, are you happy with the new naming?
@senesis, I think I addressed all your comments. Sorry for not responding directly to them. Do you think this is ok now?

bouweandela:
> install and develop: these fail due to what looks like a typo in the test for a cmip5 fix:

This looks like #876. I think the test could be improved so it uses the actual date instead of a string formatted version.

esmvalcore/_config/_config.py
```diff
@@ -13,6 +16,46 @@
...
+    for key, value in update.items():
+        if isinstance(value, collections.abc.Mapping):
+            dictionary[key] = _deep_update(dictionary.get(key, {}), value)
+        else:
+            dictionary[key] = value
+    return dictionary
+
+
+@lru_cache
+def _get_variable_details_for_project(project):
```
bouweandela, line 36: Maybe you could rename the function to `_get_variable_details` because `project` is already the name of the argument, so it's obvious that it's `for_project`. Or maybe a name like `_load_extra_facets`? Or `_load_project_facets`?

esmvalcore/_config/_config.py
```diff
@@ -13,6 +16,46 @@
...
+    config = {}
+    config_paths = [
+        importlib_files("esmvalcore._config"),
+        Path.home() / ".esmvaltool",
+    ]
+    for config_path in config_paths:
+        search_path = config_path / "variable_details"
+        config_file_paths = search_path.glob(f"{project.lower()}-*.yml")
+        for config_file_path in sorted(config_file_paths):
+            with config_file_path.open() as config_file:
```
bouweandela, line 46: Could you please log the filename for easier debugging?

esmvalcore/_config/_config.py
```diff
@@ -13,6 +16,46 @@
...
+                config_piece = yaml.safe_load(config_file)
+            if config_piece:
+                _deep_update(config, config_piece)
+    return config
+
+
+@lru_cache
+def get_variable_details(project, dataset, mip, short_name):
```
bouweandela, line 54: Maybe name this `get_extra_facets`?

esmvalcore/cmor/_fixes/fix.py
```diff
@@ -150,6 +154,7 @@ def get_fixes(project, dataset, mip, short_name):
         """
         cmor_table = CMOR_TABLES[project]
         vardef = cmor_table.get_variable(mip, short_name)
+        var_mapping = get_variable_details(project, dataset, mip, short_name)
```
bouweandela, line 157: As discussed earlier, do you think it would make sense to pass these values in from the recipe instead of loading them here? That would have the advantage that the facets from the configuration file could be overwritten from the recipe.

They could be passed into the preprocessor functions `fix_file`, `fix_metadata`, and `fix_data` from the recipe here:
https://github.com/ESMValGroup/ESMValCore/blob/347d19d28c0b91891180a58a211ad7781c2d6999/esmvalcore/_recipe.py#L270-L278
Do check if that also works for fx files though, i.e. in the file [esmvalcore/preprocessor/_ancillary_vars.py](https://github.com/ESMValGroup/ESMValCore/blob/master/esmvalcore/preprocessor/_ancillary_vars.py) these functions should also receive the correct arguments.

bouweandela (commented):
Looks good @zklaus!

senesis:
> @senesis, I think I addressed all your comments. Sorry for not responding directly to them. Do you think this is ok now?

I inserted one comment as a review (see above).

I also suggested (maybe in your private repository's branch) to make a change in cmor/_fixes/fix.py, to write 
```
    def get_cube_from_list(self, cubes, short_name=None, mapping_key=None):
    ....
        if short_name is None:
            short_name = self.vardef.short_name
        short_name=self.var_mapping.get(mapping_key,short_name)
```
in order to add a capacity to this function (with backward compatibility), which allows fix method to easily access cubes using project's specific variable name (by providing project's specific key for this variable name)

senesis (commented):
That's look fine, except for the review comment below and maybe an extended version of get_cube_from_list (see other comment)

I was able to run a basic test with a recipe using simultaneously IPSL data in its two formats (mono-var and multi-var files) by defining two projects for that. For the time being, I cannot use project native6 for such a simultaneous use of two data formats (as long as the DRS cannot be chosen on a dataset instance basis)

esmvalcore/_config/_config.py
```diff
@@ -37,6 +37,7 @@ def _get_project_mappings(project):
     config = {}
     config_paths = [
         importlib_files("esmvalcore._config"),
+        Path.home() / ".esmvaltool",
```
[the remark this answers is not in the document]
zklaus: As far as I understand it, it is quite uncommon and unintended that users change the `config-developer.yml` file. They do change the `config-user.yml` file that lives in `~/.esmvaltool` and it might be an idea to add a config option there. But then again, if you are willing to change something in `~/.esmvaltool`, why can this not be done in the sub-directory?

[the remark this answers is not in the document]
senesis: The config-user file is an argument of the esmvaltool command (and defaults to one in ~/.esmvaltool), and the config-dev file is specified in the config-user file 

If we agree that a model developer user should be allowed to change what the model does output, and so that he should be able to change the mapping dicts, then he has to be allowed to run two instances of ESMValTool at the same time, one with a legacy mapping, and another one with a changed mapping. So, yes, he should at least have a config mapping option in the config-user file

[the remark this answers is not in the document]
bouweandela: @zklaus The idea was indeed to make it possible to specify the path to the new configuration files in config-developer.yml. It is possible to specify the path to config-developer.yml in config-user.yml. I know it is fairly unusual for users to change config-developer.yml, but it is a feature we support. Changing this kind of mapping file is probably similarly unsual, so I think it should be fine to add it there. Most users will not need this as they are not running models anyway.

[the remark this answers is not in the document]
zklaus: I did have a custom config-developer.yml in the past, but got bitten by updates in the core one too often, so I moved away from that. Which is probably an indication that it would be nice to have a better way to overwrite individual settings. But that's a topic for another day.

So, on topic, I understand that we had that idea to name a mappings or extra facets file in the configuration. But since, we have moved to the idea of supporting multiple mapping files. Should a possible configuration option then point to a directory? Or maybe accept a list of files? Also, should we still support a generic directory in the users configuration directory, or do away with this completely and require all configuration over and above the one shipped in the package to be listed explicitly?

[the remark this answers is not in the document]
senesis: > So, on topic, I understand that we had that idea to name a mappings or extra facets file in the configuration

This idea still stands (if `in the configuration` is understood as `in config-dev file`)

> Should a possible configuration option then point to a directory?

I think so

> Or maybe accept a list of files?

I think it doesn't matter 

> Also, should we still support a generic directory in the users configuration directory, or do away with this completely and require all configuration over and above the one shipped in the package to be listed explicitly?

I agree with the latter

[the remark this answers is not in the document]
zklaus: After giving it some more thought, here is what I did: I added a new configuration option `extra_facets_dir`, but to `config_user.yml`, not `config_developer.yml`. This can take a single path or a list of paths that is appended to the default paths, which are `esmvaltool._config/extra_facets` and `~/.esmvaltool/extra_facets`. Since later entries effectively overwrite earlier ones, like this the user has the option to choose at runtime which replacements should take place, while still giving us all the flexibility to deliver more static versions.

What do you think about it?

zklaus:
@stefsmeets, to make the tests work, I had to add validation for a new option in `config_user.yml` to the experimental config stuff. Apart from the simple addition, which permits essentially a single path or a list of paths, I needed to move from list to tuple in order to make the result hashable, which I need since it is used in an `lru_cache`d function as an argument. So I added a `_tuplify_validator` alongside the `_listify_validator`. Could you please have a look at that and let me know what you think? Perhaps it's generally a good idea to move to tuples for these options?

zklaus:
@senesis, I have added your suggestion for a mapping_key to the get_cube_from_list method. I have also implemented a config option for the extra_facets_dir, see discussion above.

Do you have any other comments?

senesis:
> Do you have any other comments?

@zklaus No , that's fine, thanks. I would like to update PR #1143 for this changes, so let me know when you're fully done, and also if you know the best way for updating my PR : locally rebasing my branch on yours and then pushing ?

esmvalcore/experimental/config/_config_validators.py
```diff
@@ -111,6 +111,59 @@ def func(inp):
...
+    try:
+        func.__name__ = "{}tuple".format(scalar_validator.__name__)
+    except AttributeError:  # class instance.
+        func.__name__ = "{}Tuple".format(type(scalar_validator).__name__)
+    func.__qualname__ = func.__qualname__.rsplit(".",
+                                                 1)[0] + "." + func.__name__
+    if docstring is not None:
+        docstring = scalar_validator.__doc__
+    func.__doc__ = docstring
+    return func
```
stefsmeets, line 164: I don't think it is necessary to duplicate the entire `_listify_validator` code here, how about this:

```suggestion
def _tuplify_validator(scalar_validator, 
                       allow_stringlist=False,
                       *,
                       n_items=None,
                       docstring=None):
    """Apply validator to a list and return a tuple."""
     _list_validator = _listify_validator(scalar_validator, allow_stringlist, n_items=n_items, docstring=docstring)
     def func(inp):
         lst = _list_validator(inp)
         return tuple(lst)
     return func
```

stefsmeets: I had another approach using an `as_tuple` toggle on the `_listify_validator`, and then adding a `if as_tuple: return tuple(inp)` statement at the end. I think I prefer the version above.

stefsmeets:
> So I added a `_tuplify_validator` alongside the `_listify_validator`. Could you please have a look at that and let me know what you think? 

Perhaps it is not necessary to copy the entire function, see my comment for a possible different approach! Can you also add some tests for the new function?

> Perhaps it's generally a good idea to move to tuples for these options?

That depends on what is needed from the config. `yaml` arrays are loaded as lists in python. If they are needed as tuples in the code, then that could be changed in this part of the code.

esmvalcore/experimental/config/_config_validators.py
zklaus: Hm, I see your point, but would in principle rather like to avoid constructing unneeded objects. How about an option as you propose, but treating it as a selector for the iterable type used in the generator expression? I.e.
```
iterablizer = tuple if as_tuple else list
...
inp = iterablizer(...)
```

zklaus: Though I also recognize that it will hardly make any real difference in the config context and that this issue is rather tangential here and must not delay this PR.

stefsmeets: That also works for me 👍 The thing I want to avoid is that we duplicate the entire `_listify_validator`.

jvegasbsc:
Just a concept question. ¿Why you decided to keep the `extra_facets` separated instead of updating the `var_info` with them if not present? That's the way we do with things like institutes and activities for CMIP and has the advantage keeping most of the functions completely agnostic of the change

jvegasbsc:
> Just a concept question. ¿Why you decided to keep the extra_facets separated instead of updating the var_info with them if not present? That's the way we do with things like institutes and activities for CMIP and has the advantage keeping most of the functions completely agnostic of the change

I just see that this is indeed what you are doing, but then why have the extra parameter for the fix function?

zklaus:
@jvegasbsc, thanks for looking at this. I'm sorry, but I am not sure I understand. I see no `var_info` around in the code that we are talking about, but let me ask it this way around: If we don't pass `extra_facets`, how does a fix get access to this information in, say, `fix_metadata`?

esmvalcore/_recipe.py
```diff
@@ -267,6 +280,10 @@ def _get_default_settings(variable, config_user, derive=False):
         'short_name': variable['short_name'],
         'mip': variable['mip'],
     }
+    extra_facets = get_extra_facets(variable['project'], variable['dataset'],
+                                    variable['mip'], variable['short_name'],
+                                    config_user['extra_facets_dir'])
+    fix.update(extra_facets)
```
bouweandela, line 286: What @jvegasbsc  means is that instead of doing this, it would be better to just start with
```python
fix = copy.deepcopy(variable)
```
That way you have all extra facets available (because they are added to `variable` on line 1120).

esmvalcore/cmor/_fixes/fix.py
```diff
@@ -7,23 +7,20 @@
 
 
 class Fix:
-    """
-    Base class for dataset fixes.
-    """
-    def __init__(self, vardef):
+    """Base class for dataset fixes."""
+    def __init__(self, vardef, **extra_facets):
```
jvegasbsc, line 11: This modification it's also not needed if you do what is suggested by @bouweandela

esmvalcore/_recipe.py
zklaus: I am happy to do that. I did not do that because somebody before me had decided that it was unwise to start from the whole variable dict and specifically selected those four entries. Since I have no idea under what circumstances what other entries might land in there, I did not want to change that.

But if you two guys, @bouweandela and @jvegasbsc, think that is ok, I am happy to go along with it.

esmvalcore/cmor/_fixes/fix.py
zklaus: Sorry, I still don't see that. The change suggested by Bouwe simplifies lines 276­-286 somewhat but has no impact before or after that, right?

jvegasbsc: Sorry, I got mixed: vardef is straight from cmor, so you must left it untouched. 

What I am suggesting here is that Instead of passing the extra_facets, just pass the full variable that you have in the recipe and that now you will be passing to the fix, so we don't only have what is coming from the dict but also potential changes made at recipe level

esmvalcore/_recipe.py
jvegasbsc: That was because we were using only a limited amount of input to get and work the fixes, so I thought it was better to be explicit about the required keys.

This is not the first time it has to grow, so it's just better to pass everything instead of adding more and more arguments

esmvalcore/_config/_config.py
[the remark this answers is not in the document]
bouweandela: I'm not entirely convinced yet: to be able to use the extra facets for finding data, the facets from these files have to match the facets used in the DRS which is still only in config-developer.yml. Therefore it seems more logical to keep this together. Of course, it's a bit hard to judge for me, because I don't know how people will use the new feature (yet). Also, it's much harder to change or remove a setting in config-user.yml once it's in, because many more people are affected by it.

esmvalcore/_recipe.py
zklaus: Ok, let's try that. It leads to two questions:

- This trips up `esmvalcore.preprocessor.check_preprocessor_settings` because we have an introspection driven whitelist of valid arguments to the preprocessor function here. Should we remove that check for these specific steps (fix_*), or maybe detect the use of `**kwargs` (simple enough: `args.varkw is None`)?
- What should we do for `add_fx_variables`? Also add the whole variable or regress to explicit `extra_facets` here?

esmvalcore/cmor/_fixes/fix.py
zklaus: If I understand you correctly, then this amounts to do the change we are discussing above and possibly changing the name of the kwargs here, correct?

esmvalcore/_config/_config.py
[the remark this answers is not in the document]
zklaus: - The *contents* of the extra_facets files will correspond with the DRS, but their location is quite independent of it. Indeed, it is certainly possible, perhaps even sometimes useful, to have extra_facets that don't appear in the DRS at all and are only used in fixes.
- I don't think many people will be affected by this setting in `config-user.yml`. We could even leave it out of the shipped example entirely and only document it in the documentation since at the moment it is really targeted at a rather specific subset of users.
- In any case, the counterargument is that if you place it in the `config-developer.yml` you increase the number of custom `config-developer.yml` files that the user must maintain from 0 to 2, while the number of custom `config-user.yml` files remains at 2. This creates a certain maintenance problem and in my personal experience this is rather annoying, since then I am forced to update my config-developer files when something is changed in the released version of the file, but it is not trivial to notice when such a change happens.

esmvalcore/_recipe.py
zklaus: I have adopted the approach to pass the whole variable and have addressed the two points I raised by
- Changing the `check_preprocessor_settings` function to forego the check for exact argument matching if the preprocessor in question uses `*args` or `**kwargs`.
- Passing the entire `variable` dict also to `add_fx_variables`.

Please open a new conversation if you disagree about one of these points. Also note the changes that were necessary to the existing recipe integration tests.

esmvalcore/_config/_config.py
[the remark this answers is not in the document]
bouweandela: > I don't think many people will be affected by this setting in config-user.yml. We could even leave it out of the shipped example entirely and only document it in the documentation since at the moment it is really targeted at a rather specific subset of users.

Let's do that as a compromise then, to not burden starting users with too many settings.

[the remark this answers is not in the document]
zklaus: There is one little hiccup: The experimental new config system uses this as the defaults, unlike our production config system which has the defaults hard-coded, and there is a test for the experimental system that trips if a setting is used that doesn't have a default.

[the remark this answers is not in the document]
zklaus: I removed the setting and fixed the test.

zklaus:
With the addition of the documentation, I think this is complete now. Some test coverage is provided by the necessary modifications to existing tests, more will come with the application to concrete use cases.

@bouweandela, are you happy with this? Would you like to see a particular test case added or do you have any other comments?

esmvalcore/cmor/_fixes/fix.py
zklaus: The full variable is now being passed. Please comment again if this is still not sufficient in your eyes.

doc/develop/extra_facets.rst
```diff
@@ -0,0 +1,111 @@
...
+table, and finally the `short_name`. The leaf dictionary placed here gives the
+extra facets that will be made available to data finder and the fix
+infrastructure. The following example illustrates the concept.
+
+.. _extra-facets-example-1:
+
+.. code-block:: yaml
+   :caption: Extra facet example file `native6-era5.yml`
+
+   era5:
```
bouweandela, line 34:
```suggestion
   ERA5:
```

doc/develop/extra_facets.rst
```diff
@@ -0,0 +1,111 @@
+.. _extra_facets:
```
bouweandela, line 1: How would you feel about adding this section to the chapter we have on configuration instead? https://esmvaltool--1124.org.readthedocs.build/projects/ESMValCore/en/1124/quickstart/configure.html

doc/develop/extra_facets.rst
```diff
@@ -0,0 +1,111 @@
...
+default facets, for example to deal with local particularities in the data
+handling.
+
+Use of extra facets
+===================
+For extra facets to be useful, the information that they provide must be
+applied. There are fundamentally two places where this comes into play. One is
+the datafinder, the other are fixes.
+
+Use of extra facets in the datafinder
```
bouweandela, line 71: This information would be better placed here: https://esmvaltool--1124.org.readthedocs.build/projects/ESMValCore/en/1124/quickstart/find_data.html

doc/develop/extra_facets.rst
```diff
@@ -0,0 +1,111 @@
...
+   :caption: Example drs use in `config-developer.yml`
+
+   native6:
+     input_file:
+       default: '{name_in_filename}*.nc'
+
+The same replacement mechanism can be employed everywhere where tags can be
+used, particularly in `input_dir` and `input_file`.
+
+Use of extra facets in fixes
```
bouweandela, line 99: And this information could then be added here: https://esmvaltool--1124.org.readthedocs.build/projects/ESMValCore/en/1124/develop/fixing_data.html

bouweandela (changes requested):
Looks good, I was able to use this to implement some basic safety checks and data finding for ERA5 with very minor additions. 

Could you please have a look at the following?
- I think the way the information is passed on to the preprocessor function `add_fx_variables` needs to be changed (see inline comments). 
- Could you please check that you have updated the numpy style docstrings for those functions that have them with documentation on any new arguments?
- Could you please add unit tests for where coverage is missing? In particular the new functions ` _deep_update` and `_load_extra_facets` in esmvalcore/_config/_config.py are not trivial and not covered by existing tests either.

esmvalcore/cmor/_fixes/fix.py
```diff
@@ -7,23 +7,23 @@
 
 
 class Fix:
-    """
-    Base class for dataset fixes.
-    """
-    def __init__(self, vardef):
+    """Base class for dataset fixes."""
+    def __init__(self, vardef, extra_facets={}):
```
bouweandela, line 11: This will give you a Codacy error that a `dict` is an unsafe default, because the `dict` that is used as the default value here, is created when the function is loaded from the module, not when the function is called. That means that if you would modify `extra_facets` inside the function, any subsequent calls of the function will use the modified `extra_facets`, potentially leading to weird bugs.

doc/develop/fixing_data.rst
```diff
@@ -353,3 +353,19 @@ For example for monthly data, place the files in the ``/Tier3/MSWEP/latestversio
   For monthly data (V220), the data must be postfixed with the date, i.e. rename ``global_monthly_050deg.nc`` to ``global_monthly_050deg_197901-201710.nc``
 
 For more info: http://www.gloh2o.org/
+
+.. _extra-facets-fixes:
+
+Use of extra facets in fixes
+============================
+In fixes, extra facets can be used to mold data into the form required by the
```
bouweandela, line 361: Could you add a link here to the extra facets section in the configuration chapter?

doc/quickstart/find_data.rst
```diff
@@ -303,3 +303,33 @@ flexible concatenation between two cubes, depending on the particular setup:
 Note that two cube concatenation is the base operation of an iterative process of reducing multiple cubes
 from multiple data segments via cube concatenation ie if there is no time-overlapping data, the
 cubes concatenation is performed in one step.
+
+.. _extra-facets-data-finder:
+
+Use of extra facets in the datafinder
+=====================================
+Extra facets can be used to locate data files within the datafinder
```
bouweandela, line 311: Could you add a link here to the extra facets section in the configuration chapter?

bouweandela (commented):
Quickly scanned the PR now, just two more comments on the documentation. I noticed that several of the tests are failing, but that looks unrelated to this pull requests. Trying to fix that in the [fix-tests](https://app.circleci.com/pipelines/github/ESMValGroup/ESMValCore?branch=fix-tests) branch now..

bouweandela (approved):

bouweandela:
Ideally all tests would need to be green before merging this.

zklaus:
Assuming the fixes coming in from main took care of the circleci stuff (as yet to be seen), this leaves us with a couple of codacy issues. They are:

- locally disabling no-name-in-module. This should be ignored because this disabling is inside a guard that distinguished between the built-in module that's available from Python 3.9 onwards and the otherwise installed version.
- the suggestion to re-raise using from. I suggest to ignore this as well since `raise ... from` is a new Python feature and this is in a part of the code in which this PR only improves the formatting.
- A small number of formatting complaints in docstrings, that come about because of conflicting formatting standards in local and ci configuration (I think). Can also be ignored imho.

PR #1133, opened 2021-05-17
Allow to load all files, first X years or last X years in an experiment

### LINKED ISSUE #1120: Add option to load selected or all years available in an experiment
In #771 we tried to add the functionality to load all DCPP data without having to specify the `start_year` and the `end_year`, as the current time range handling is not ideal (#345). To do so, a new tag ` to load all the data available was introduced. @zklaus pointed out that it could work for all experiments. 

Yes

sloosvel:
## Description
This  pull requests introduces the tag `timerange` with possible start and end values given as a combination of dates, datetimes and duration periods following the ISO 8601 standard or wildcards. The start and end values must be separated by a `/`. Thus, in general: `timerange: '{start}/{end}'`. The `timerange` tag can be specified with a time resolution up to seconds.

The `clip_start_end_year` step has been substituted by `clip_timerange`, that works with the  `timerange`  tag in order to extract a given range of dates, datetimes or duration periods.

This improvements allow for the proper loading of data belonging to the DCPP activity in CMIP6.

Closes #1120 

Link to documentation:

***

***


zklaus:
@sloosvel, thanks for this. I will unassign myself for now to dampen the email flood. Please do assign me again when this is ready for review.

sloosvel:
I think this is kind of ready now, pending to fix a couple of codacy issues.

This pull request uses `isodate` to parse dates and durations for the new `timerange` tag. Tags `start_year` and `end_year` are still fully usable, to avoid breaking everything.  

It includes a bit of work done to replace wildcard values in recipes, but I am not too convinced this is the way to go. Or that it even belongs in this pull request.

Changes in the `clip_start_end_year` function will be added in another pull request.

doc/recipe/overview.rst
```diff
@@ -146,6 +154,23 @@ The same simplified syntax can be used to add multiple sub-experiment ids:
...
+    * ``timerange: '19800504T100000/19800504T110000'``. Spans from 04/05/1980 at 10h to 11h.
+
+  - A start point or end point, and a relative period with a resolution up to second (P[n]Y[n]M[n]DT[n]H[n]M[n]S).
+    * ``timerange: '1980/P5Y'``. Starting from 01/01/1980, spans 5 years.
+    * ``timerange: 'P2Y5M/198202``. Ending at 01/02/1982, spans 2 years and 5 months.
+  - A wildcard to load all available years, the first available start point or the last available end point.
+    * ``timerange: '*'``. Finds all available years.
+    * ``timerange: '*/1982``. Finds first available point, spans to 31/12/1982.
+    * ``timerange: '*/P6Y``. Finds first available point, spans 6 years from it.
+    * ``timerange: '198003/*``. Starting from 01/03/1980, spans until the last available point.
```
jvegasbsc, line 172:
```suggestion
    * ``timerange: '198003/*``. Starting from 01/03/1980, spans until the last available point.
    * ``timerange: 'P5M/*``. Finds latavailable point, spans 5 months backwards from it.
```

esmvalcore/_data_finder.py
```diff
@@ -95,16 +103,29 @@ def get_start_end_year(filename):
                 time = cube.coord('time')
             except iris.exceptions.CoordinateNotFoundError:
                 continue
-            start_year = time.cell(0).point.year
-            end_year = time.cell(-1).point.year
+            start_point = time.cell(0).point
+            start_datetime = datetime.datetime(
```
jvegasbsc, line 107: Why this and not directly use the point?   It is not working with the `isodate` module? The class that iris is using for this is usually close enough to the one in datetime for this replacement to work

esmvalcore/_data_finder.py
```diff
@@ -95,16 +103,29 @@ def get_start_end_year(filename):
...
+            end_year = end_point.year
             break
 
     if start_year is None or end_year is None:
         raise ValueError(f'File {filename} dates do not match a recognized'
                          'pattern and time can not be read from the file')
 
     logger.debug("Found start_year %s and end_year %s", start_year, end_year)
-    return int(start_year), int(end_year)
+    return int(start_year), int(end_year), start_date, end_date
```
jvegasbsc, line 128: Why not just return the full date and then just access the year outside?

jvegasbsc (changes requested):
Nice work.

esmvalcore/_data_finder.py
sloosvel: I think I started using another parser that was not working with the cftime Datetime object. But this one seems to work.

esmvalcore/_data_finder.py
sloosvel: Sure, I'll change that

sloosvel: Actually returning both comes in quite handy

jvegasbsc: Yes, but it is really ugly

sloosvel: start_date and end_date are strings, accessing the year outside after every time this function is called in not that direct.

jvegasbsc: We are keeping it as a string to avoid having to parse the dates

jvegasbsc:
The parts with missing tests are probably out of the scope of this PR. I will open an issue to remind us to improve the coverage of the `_main` module

jvegasbsc (approved):

jvegasbsc:
Please, fix the merge conflict and the codacy issues so we can merge

sloosvel:
Thanks for the comments @jvegasbsc. @ESMValGroup/esmvaltool-coreteam does anyone have the time to give a second opinion? Thanks in advance.

tests/integration/test_recipe.py
```diff
@@ -929,6 +929,109 @@ def test_simple_cordex_recipe(tmp_path, patched_datafinder, config_user):
...
+                timerange:
+            variables:
+              pr:
+                mip: 3hr
+              areacella:
+                mip: fx
+            scripts: null
+        """)
+
+    # Add invalid timerange
```
bouweandela, line 978: What do you mean with this comment?

esmvalcore/_data_finder.py
```diff
@@ -35,15 +36,17 @@ def find_files(dirnames, filenames):
 def get_start_end_year(filename):
     """Get the start and end year from a file name.
 
+    Additionally, returns the start and end dates as strings.
```
bouweandela, line 39: Instead of adding this new feature to this already complicated function, would it be possible to create a new function `get_start_end_date` that returns this information? If there is code that is shared with the `get_start_end_year` function, that could be moved to another function that is called by both?

esmvalcore/_recipe.py
```diff
@@ -694,6 +703,70 @@ def _update_extract_shape(settings, config_user):
...
+        'start_year': start_year,
+        'end_year': end_year
+    })
+
+    settings[step]['timerange'] = timerange
+    settings[step]['start_year'] = start_year
+    settings[step]['end_year'] = end_year
+
+    timerange = timerange.replace('/', '-')
+    filename = variable['filename'].replace('.nc', f'_{timerange}.nc')
```
bouweandela, line 766: This looks dodgy, I think you would need to make sure that the right value is passed into the function `get_output_file` from _data_finder.py in the first place, instead of changing it later on

esmvalcore/_recipe.py
```diff
@@ -1374,6 +1454,32 @@ def _resolve_diagnostic_ancestors(self, tasks):
...
+            if timerange != raw_timerange:
+                if not self._updated_recipe:
+                    self._updated_recipe = deepcopy(self._raw_recipe)
+                if datasets:
+                    (self._updated_recipe['datasets'][index]['timerange']
+                     ) = timerange
+                else:
+                    (self._updated_recipe['diagnostics'][diagnostic]
+                     ['additional_datasets'][index]['timerange']) = timerange
+
```
bouweandela, line 1482: What about the case where `additional_datasets` is specified under a variable?

esmvalcore/_recipe.py
```diff
@@ -694,6 +703,70 @@ def _update_extract_shape(settings, config_user):
...
+def _update_timerange(variable, settings, config_user):
+    if 'timerange' not in variable:
+        return
+
+    timerange = variable.get('timerange')
+    check.valid_time_selection(timerange)
+
+    step = 'clip_start_end_year'
+    if '*' in timerange:
+        (files, _, _) = _find_input_files(variable, config_user['rootpath'],
```
bouweandela, line 735: Could you add a test case to tests/integration/data_finder.yml to demonstrate that this works?

esmvalcore/_recipe.py
```diff
@@ -293,6 +296,13 @@ def _get_default_settings(variable, config_user, derive=False):
             'end_year': variable['end_year'],
         }
 
+    if 'timerange' in variable and variable['frequency'] != 'fx':
+        settings['clip_start_end_year'] = {
+            'start_year': None,
```
bouweandela, line 301: This is becoming a mess, changing the arguments of clip_start_end_year in 3 different places. Would it be possible to call the function `_update_timerange` just before `_get_default_settings` on line 830 and then set the values of the arguments just once to the final settings here?

esmvalcore/_recipe.py
```diff
@@ -503,6 +511,7 @@ def _fx_list_to_dict(fx_vars):
 
 def _update_fx_settings(settings, variable, config_user):
```
bouweandela, line 512: Something for another pull request, but what about time dependent fx variables? It looks like `clip_start_end_year` is not applied to those, this will lead to problems when trying to attach them as ancillary variables.

esmvalcore/_main.py
```diff
@@ -99,7 +99,7 @@ def process_recipe(recipe_file, config_user):
     # parse recipe
     recipe = read_recipe_file(recipe_file, config_user)
     logger.debug("Recipe summary:\n%s", recipe)
-
+    recipe.write_filled_recipe()
```
bouweandela, line 102: Maybe it would be better to include this in `recipe.run` (same goes for `recipe.write_html_summary`), because having to make all these separate calls to run a recipe makes it more complicated to use.

esmvalcore/_recipe.py
```diff
@@ -1485,6 +1593,17 @@ def get_output(self) -> dict:
 
         return output
 
+    def write_filled_recipe(self):
```
bouweandela, line 1596: Could you please add a unit test for this function?

bouweandela (dismissed):
Nice work! Here are a few ideas to make it even better

doc/recipe/overview.rst
```diff
@@ -84,10 +84,15 @@ data specifications:
 - ensemble member (key ``ensemble``, value e.g. ``r1i1p1``, ``r1i1p1f1``)
 - sub-experiment id (key `sub_experiment`, value e.g. `s2000`, `s(2000:2002)`, 
   for DCPP data only)
-- time range (e.g. key-value ``start_year: 1982``, ``end_year: 1990``. Please
-  note that `yaml`_ interprets numbers with a leading ``0`` as octal numbers,
+- time range (e.g. key-value ``start_year: 1982``, ``end_year: 1990``. 
+  Please note that `yaml`_ interprets numbers with a leading ``0`` as octal numbers,
   so we recommend to avoid them. For example, use ``128`` to specify the year
   128 instead of ``0128``.)
+  Alternatively, the time range can be specified in ISO8601 format, for both dates
```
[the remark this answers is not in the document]
sloosvel: Would the link to the wikipedia page work for you? The information in [the official iso page](https://www.iso.org/iso-8601-date-and-time-format.html) is not as clear.

esmvalcore/_recipe.py
sloosvel: Sorry about the mess, but all those changes are not really needed for the functionality on it's own, they are just to avoid having backward incompatible changes. I'll try to get it to work more clearly.

doc/recipe/overview.rst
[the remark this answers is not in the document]
bouweandela: Sure!

esmvalcore/_recipe.py
sloosvel: Forgot about it. In any case, is there a better way to access nested dicts? This function does its job, but I am not too conviced it's the best way to go about it.

bouweandela: You could write a function to do it and use that, that would be nicer.

Looking at this code in more detail, I think it doesn't work because the 'recipe_dataset_index' is computed by adding the 3 lists of datasets (at recipe, diagnostic, and variable level), so this will give the wrong position in the list if you do not take the offset into account. Would it be possible to implement this function without using the recipe_dataset_index? I only added that because some NCL diagnostics apparently depend on the ordering of the datasets in the recipe, so it is used to order the datasets before writing them to the NCL interface. I think it's not good practice to use numerical indices when there are easier to understand alternatives. It is very easy to make a mistake like this with numerical indices.

sloosvel: I have been looking into it without much success. Since datasets can get expanded, it's quite difficult to map the preprocessor output to the datasets in the recipe.

bouweandela: It looks like expansion and `recipe_dataset_index` do not work together very well, currently, all expanded datasets get the same value for `recipe_dataset_index`, so ordering them afterwards is not possible. I think this index should only be computed after expansion.

To get this to work with the expansion, I believe that you'll first need to expand, then find the dateranges for every single dataset, and then (optionally) unexpand again if this is still possible, because the files of the different datasets may not share the same daterange and/or start_year,end_year. This probably works best if you move the call to `_update_timerange` right after `variables = self._expand_tag(variables, 'sub_experiment')` on line 1242, because then you don't need to bother with the preprocessor output.

sloosvel: Thanks for the tip, let's see if I can make it work with this.

sloosvel:
I kind of addressed all the suggestions but I am having some issues with the tests. First, because of FLAKE in:

```
/test/tests/integration/test_main.py:20:1: F401 '.test_recipe.patched_datafinder' imported but unused
/test/tests/integration/test_main.py:96:34: F811 redefinition of unused 'patched_datafinder' from line 20
```
What is the proper way to call the patched datafinder?

zklaus:
You are trying to use a fixture here. Read more about them [here](https://docs.pytest.org/en/latest/explanation/fixtures.html#about-fixtures). Bottom line: you probably want to add a `conftest.py` file as described [here](https://docs.pytest.org/en/latest/reference/fixtures.html?highlight=conftest%20py#conftest-py-sharing-fixtures-across-multiple-files). Also, see https://stackoverflow.com/a/34520971 for an accessible explanation of `conftest.py`.

sloosvel:
The other test that is failing is due to the test finding a file that is not listed as an available file in the configuration of the test. Can it be added there?

zklaus:
If I read the test log correctly, the file in question is `ta_Amon_HadGEM2-ES_historical_r1i1p1_193412-195911.nc`, which weighs in at just under 70MB. That's not exactly humungous, but certainly larger than what we want in the repo.

Is the test intended to use real data?

sloosvel:
It's just to check if `_find_input_files` works, as requested by @bouweandela . But I think this is already checked when using and testing `get_input_filelist`, the only difference between the two functions is that `get_input_filelist` does a time-gating after using `_find_input_files`

jvegasbsc:
The only errors remaining are  in the `tests/unit/preprocessor/_weighting/test_weighting_landsea_fraction.py`, so they are unrelated to this PR

@sloosvel, can you address the remaining comments from @bouweandela so we can finally merge?

sloosvel:
No, no merge yet. There is another pull request that goes with this one.

zklaus:
@sloosvel, could you clarify: Does #1214 need to go in before this one? Or the other way around?

sloosvel:
This one should go first, but there should not be much of a delay between merges and the other one is not reviewed. So maybe it's better to have them both ready and then go first with this one.

valeriupredoi:
double-stack them, Saskia, like they do in Formula 1 when both cars come to the same pit :+1:

esmvalcore/_recipe.py
```diff
@@ -7,7 +7,9 @@
 from copy import deepcopy
 from pprint import pformat
 
+import isodate
```
bouweandela, line 10: This package is not yet listed in the dependencies, could you add it please?

tests/integration/test_data_finder.py
```diff
@@ -65,7 +69,7 @@ def create_tree(path, filenames=None, symlinks=None):
 def test_get_output_file(cfg):
     """Test getting output name for preprocessed files."""
     output_file = get_output_file(cfg['variable'], cfg['preproc_dir'])
-    assert output_file == cfg['output_file']
+    assert [output_file] == cfg['output_file']
```
bouweandela, line 72: Why this change? get_output_file should return a single filename, right?

tests/integration/data_finder.yml
```diff
@@ -42,6 +44,10 @@ get_input_filelist:
       - ta_Amon_HadGEM2-ES_historical_r1i1p1*.nc
     found_files:
       - ta_Amon_HadGEM2-ES_historical_r1i1p1_195912-198411.nc
+    found_timerange_files:
```
bouweandela, line 47: `get_input_filelist` should find the same files, regardless of whether a timerange or the same information with a start_year and end_year is provided. Therefore I feel that this new organization of the tests is confusing. From the associated test, it looks like this section is intended to test that using a '*' for the timerange works, right? I think it would be better to just add some new testcases to this file for that purpose, without modifying the test itself. For example
```yaml
get_input_filelist:
  - drs: default
    variable:
      variable_group: test
      short_name: ta
      original_short_name: ta
      dataset: HadGEM2-ES
      project: CMIP5
      cmor_table: CMIP5
      institute: [INPE, MOHC]
      frequency: mon
      modeling_realm: [atmos]
      mip: Amon
      exp: historical
      ensemble: r1i1p1
      timerange: '*'
      diagnostic: test_diag
      preprocessor: test_preproc
    available_files:
      - ta_Amon_HadGEM2-ES_historical_r1i1p1_193412-195911.nc
      - ta_Amon_HadGEM2-ES_historical_r1i1p1_195912-198411.nc
      - ta_Amon_HadGEM2-ES_historical_r1i1p1_198412-200511.nc
    dirs:
      - ''
    file_patterns:
      - ta_Amon_HadGEM2-ES_historical_r1i1p1*.nc
    found_files:
      - ta_Amon_HadGEM2-ES_historical_r1i1p1_193412-195911.nc
      - ta_Amon_HadGEM2-ES_historical_r1i1p1_195912-198411.nc
      - ta_Amon_HadGEM2-ES_historical_r1i1p1_198412-200511.nc
 ```

File: tests/integration/test_main.py [hunk 17]
```diff
@@ -82,9 +83,69 @@ def test_empty_run(tmp_path):
...
+    recipe_file.write_text(content)
+    Config.get_config_user(path=tmp_path)
+    excs = (ValueError, iris.exceptions.CoordinateNotFoundError)
+    with pytest.raises(excs) as exc:
+        ESMValTool.run(
+            recipe_file,
+            config_file=f"{tmp_path}/config-user.yml",
+            check_level='ignore')
+    assert str(exc.value)
+    log_dir = './esmvaltool_output'
```
bouweandela, line 143: Please don't write to the current working directory when running tests, but update the configuration so it writes to tmp_path before running the recipe

tests/integration/test_main.py
```diff
@@ -82,9 +83,69 @@ def test_empty_run(tmp_path):
...
+    filled_recipe = os.path.exists(
+        log_dir + '/' + os.listdir(log_dir)[0] + '/run/recipe_filled.yml')
+    os.system("rm -r esmvaltool_output")
```
bouweandela, line 146: If someone has the directory `esmvaltool_output` with all their recipe runs in their current working directory and spent hours computing those results, they will be very disappointed that it gets deleted by running this test. Could you please change this to something safe?

If you have to, please use `shutil.rmtree` for removing a directory instead of calling a command line tool.

tests/integration/test_main.py
```diff
@@ -82,9 +83,69 @@ def test_empty_run(tmp_path):
...
+                ensemble: r1i1p1
+                timerange: '*'
+                additional_datasets:
+                  - dataset: MPI-ESM-LR
+                    timerange: '*/P1Y'
+            scripts: null
+        """)
+    recipe_file.write_text(content)
+    Config.get_config_user(path=tmp_path)
+    excs = (ValueError, iris.exceptions.CoordinateNotFoundError)
```
bouweandela, line 136: This seems a slightly odd choice of exceptions. I guess the `iris.exceptions.CoordinateNotFoundError` comes from the extract_levels preprocessor check that tries to see if there are pressure levels in the data. Might as well not use that preprocessor in the recipe. What causes the ValueError?

esmvalcore/_recipe.py
```diff
@@ -1161,6 +1227,9 @@ def _initialize_variables(self, raw_variable, raw_datasets):
                 activity = get_activity(variable)
                 if activity:
                     variable['activity'] = activity
+            if 'timerange' in variable:
```
bouweandela, line 1230: Would it be possible to convert `start_year` and `end_year` to a timerange here and then remove `start_year` and `end_year` from the variable? That way there would be one consistent place for time information, instead of having to deal with two possible forms of the same information throughout the code.

esmvalcore/_recipe.py
sloosvel: I did not have to install it as a dependency. I think it must come with the standard Python.

tests/integration/test_data_finder.py
sloosvel: There was a complain during the test about a line being too long in the `data_finder.yml`, so I moved the filename from this:
    output_file: this/is/a/path/test_diag/test/CMIP5_HadGEM2-ES_Amon_historical_r1i1p1_ta_1960-1980.nc

to this:
    output_file:
      - this/is/a/path/test_diag/test/CMIP5_HadGEM2-ES_Amon_historical_r1i1p1_ta_1960-1980.nc
So get_output_file still returns a file, but what's in the config is a list.

I don't think anything about the functionality changed. But I have to re-write the test anyway.

tests/integration/test_main.py
sloosvel: Ok, but I copied that from `test_empty_run`. Should it be deleted from there as well?

tests/integration/test_data_finder.py
valeriupredoi: codacy/flake should never complain a line is too long in a yaml file, Saskia, you sure it wasn't in any of the :snake: codes?

sloosvel: yamllint does

tests/integration/test_main.py
sloosvel: I think it was something in the cmor checker

esmvalcore/_recipe.py
bouweandela: It is probably a dependency of one of our dependencies. However, because other packages might change their dependencies at any point without notifying us, it is important that we declare the complete list of our own dependencies. You can find the modules that are part of the Python standard library [here](https://docs.python.org/3/library/). The isodate package is an independently developed package that can be installed from PyPI, see [here](https://pypi.org/project/isodate/).

tests/integration/test_data_finder.py
bouweandela: I think this change to the yaml file is confusing, because the function returns a single string, not a list containing a string. If you want to get rid of the yamllint warning, you could [spread the string over two lines](https://stackoverflow.com/questions/3790454/how-do-i-break-a-string-in-yaml-over-multiple-lines) instead.

tests/integration/test_main.py
bouweandela: Yes, that would be great!

tests/integration/test_main.py
bouweandela: OK, I guess it will fail to process the fake data at some point. You could have a look at the sample_data tests, these use real CMIP6 data from the [ESMValTool_sample_data](https://pypi.org/project/ESMValTool-sample-data/) package so they should be able to run the recipe without error, but this is fine too. I would simplify the test by removing the extract_levels preprocessor though, just the default preprocessor tests this feature well enough.

tests/integration/test_main.py
sloosvel: At what time is the feature freeze supposed to be? Because I'm sorry but I am making a mess over this and the comment below and I don't want to miss the deadline because of it. It was copied from `test_empty_run` and no-one has complained about having their real outputs erased so far anyway.

zklaus: @sloosvel, the feature freeze is supposed to be today, but I will allow a little bit of leeway in cases like this, where work has progressed very far and we "only" need to finish the review process. So don't panic ;). Just try to finish things with the due diligence.

tests/integration/test_main.py
bouweandela: Can iris.exceptions.CoordinateNotFoundError be removed now?

bouweandela: Note that you can just pass `output_dir=str(tmp_path / 'esmvaltool_output')` as an argument to `ESMValTool.run`, but updating the config file like it is done now is fine too.

esmvalcore/_recipe.py
bouweandela: What I meant was: would it be possible to remove `start_year` and `end_year` everywhere from the code, i.e. do here something like this for backwards compatibility:
```python
if 'start_year' in variable and 'end_year' in variable:
    variable['timerange'] = f'{start_year}/{end_year}'
elif 'start_year' in variable:
    variable['timerange'] = f'{start_year}/{start_year}'
elif 'end_year' in variable:
    variable['timerange'] = f'{end_year}/{end_year}'
variable.pop('start_year', None)
variable.pop('end_year', None)
```
(not sure if I got the timerange syntax right) and then clean up the rest of the code so it no longer uses `variable['start_year']` and `variable['end_year']` anywhere anymore?

sloosvel: Wouldn't that break `extract_time`?

bouweandela: No, if users want to use `extract_time` they can configure it like any other preprocessor function from the recipe: https://esmvaltool--1133.org.readthedocs.build/projects/ESMValCore/en/1133/recipe/preprocessor.html#extract-time

sloosvel: I don't see how storing the three tags is that bad though. Having to split the timerange value across the code everytime you need to work with just the years is more work than having all three tags in `variable`.

bouweandela: Because it makes the time handling code needlessly complicated. I don't see why we would want to work with the years at all if we have a timerange. As far as I'm aware, `start_year` and `end_year` were used to do the following things:
- checking that the input files cover the requested timerange (in _recipe_checks.py) -> this can be done better with timerange
- filtering the list of preprocessor input files (in _data_finder.py and esgf/_search.py) -> this can be done better with timerange
- if the files contain more data than requested, clipping the data to the requested period using the `clip_start_end_year` preprocessor function -> you already implemented this with `clip_timerange`
- creating the preprocessor output filename -> you already implemented that this can be done with timerange

Or did I miss something?

sloosvel: Ok I will try to update it accordingly, but when timeranges are given as periods ('*/P2Y') it's useful to have a start and end year.

bouweandela: Thanks, that would be great! You already expand those wildcard style timeranges for the filled recipe, right? You can probably use the same code to convert that to a start and end datetime. The big advantage (in addition to simpler code) of using timerange for e.g. selecting the right input files is that we can just as easily support times that are stored with smaller time intervals per file than years.

sloosvel: > Thanks, that would be great! You already expand those wildcard style timeranges for the filled recipe, right? You can probably use the same code to convert that to a start and end datetime. The big advantage (in addition to simpler code) of using timerange for e.g. selecting the right input files is that we can just as easily support times that are stored with smaller time intervals per file than years.

That is quickly doable when your timerange value looks something like this `19600101/19621231'. But I am having some issues when the timerange includes periods such as `19600101/P3Y2M1D`.  Adapting the file filtering to work with periods is not that small of a change and I'm not sure I can have it ready for the release.

sloosvel: I am a bit confused about the esgf part. Does this part of the code happen before or after the download :

```
if 'start_year' in variable and 'end_year' in variable:
    variable['timerange'] = f'{start_year}/{end_year}'
elif 'start_year' in variable:
    variable['timerange'] = f'{start_year}/{start_year}'
elif 'end_year' in variable:
    variable['timerange'] = f'{end_year}/{end_year}'
variable.pop('start_year', None)
variable.pop('end_year', None)

```

sloosvel:
Not sure why `test_installation_from_source_develop_mode` is failing now.

zklaus:
> Not sure why `test_installation_from_source_develop_mode` is failing now.

I'm almost sure this is due to #1359. I kicked the test again and we'll see if that resolves things.

sloosvel:
Ok so the only line that is missing from the coverage, actually gets covered in some of the runs. It depends on the order in which the files are found and enter the loop. 

Other than that I think everything else is done. Basically in these lasts commits, I addressed and added support for file selection when there is a time period, which was something I got stuck with in https://github.com/ESMValGroup/ESMValCore/pull/1133#discussion_r722328942 . But it should work now and the corresponding tests are passing.

esmvalcore/_data_finder.py
```diff
@@ -317,18 +462,21 @@ def get_output_file(variable, preproc_dir):
...
+        timerange = variable['timerange'].replace('/', '-')
+        outfile += f'_{timerange}'
+
     outfile += '.nc'
     return outfile
 
 
 def get_statistic_output_file(variable, preproc_dir):
     """Get multi model statistic filename depending on settings."""
+    variable['timerange'] = variable['timerange'].replace('/', '-')
```
bouweandela, line 474: This results in a wrong `timerange` attribute for the multimodel statistics output file. Could you do it without modifying the `variable` dict?

bouweandela (commented):
Looks good @sloosvel!

I tried to test this with recipe_python.yml and ran into a backwards incompatibility issue: the diagnostic script depends on the availability of `start_year` and `end_year` in the metadata. Would it be possible to add those to the generated metadata.yml file here:
https://github.com/ESMValGroup/ESMValCore/blob/e4d61adfb14e3a3995abe2eb869d5f1781053416/esmvalcore/preprocessor/_io.py#L340-L345
for backwards compatibility? Or have the periods not been changed to datetimes there?

For further testing, it might be nice to merge #1214 into this pull request, because using the period selection option does not work at the moment because it crashes on the clip_timerange preprocessor. Or do you have a different idea for a testing strategy @zklaus? I know you are keen on keeping pull requests in small chunks (so am I, but it is difficult to see how this can be reduced in size).

esmvalcore/_data_finder.py
```diff
@@ -108,23 +187,89 @@ def get_start_end_year(filename):
...
     """
     selection = []
+    start_date, end_date = _parse_period(timerange)
+
+    if start_date is None and end_date is None:
+        start_date = timerange.split('/')[0]
+        end_date = timerange.split('/')[1]
+    else:
+        start_date = str(start_date)
+        end_date = str(end_date)
```
bouweandela, line 255: Why are these lines not part of the _parse_period function? Now this code is duplicated in several places where the _parse_period function is used.

esmvalcore/_data_finder.py
```diff
@@ -108,23 +187,89 @@ def get_start_end_year(filename):
...
+        end_date = isodate.datetime_isoformat(
+            end_date, format=datetime_format)
+    elif time_format == isodate.DATE_BAS_COMPLETE:
+        start_date = isodate.date_isoformat(start_date, format=time_format)
+        end_date = isodate.date_isoformat(end_date, format=time_format)
+
+    return start_date, end_date
+
+
+def _compare_dates(date, file_date):
```
bouweandela, line 227: The name of this function is confusing, because it doesn't actually compare the dates, but rather truncates the strings representing them so they are the same length, if I understand the code correctly

esmvalcore/_recipe.py
```diff
@@ -712,6 +728,31 @@ def _update_extract_shape(settings, config_user):
...
+    timerange = variable.get('timerange')
+    check.valid_time_selection(timerange)
+
+    if '*' in timerange:
+        (files, _, _) = _find_input_files(variable, config_user['rootpath'],
+                                          config_user['drs'])
+        intervals = [get_start_end_date(name) for name in files]
+
+        min_date = min(intervals)[0]
+        max_date = max(intervals)[1]
```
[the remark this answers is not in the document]
sloosvel: I have pending to address this because I am surprised it does not make the tests fail.

>Or have the periods not been changed to datetimes there?

Periods are not changed to datetimes until the clipping, because I think that it's more reliable to work with the time coordinate in the cube, than the date that appears in the filename. However, since there is the check for missing years, I added the start and end year information in that step so that it appears in the `metadata.yml` file.

sloosvel:
As suggested by @bouweandela, I am merging the clipping of #1214  here. I expect one test to break, the `test_filled_recipe` because it will try to extract the timerange in an empty cube. I don't know how to circumvent this though.

esmvalcore/esgf/_search.py
```diff
@@ -97,18 +102,29 @@ def esgf_search_files(facets):
...
-def select_by_time(files, start_year, end_year):
-    """Select files containing data between start_year and end_year."""
+def select_by_time(files, timerange):
+    """Select files containing data between a timerange."""
     selection = []
+    start_date, end_date = _parse_period(timerange)
+
+    if start_date is None and end_date is None:
+        start_date = timerange.split('/')[0]
+        end_date = timerange.split('/')[1]
```
bouweandela, line 112: Could you please move this code inside the `_parse_period` function? It appears to be duplicated everywhere the `_parse_period` function is used.

esmvalcore/_recipe_checks.py
```diff
@@ -156,7 +157,16 @@ def data_availability(input_files, var, dirnames, filenames, log=True):
...
+    start_year, end_year = _parse_period(var['timerange'])
+    if start_year is None and end_year is None:
+        start_year = int(var['timerange'].split('/')[0][0:4])
+        end_year = int(var['timerange'].split('/')[1][0:4])
+    if isinstance(start_year, str) and isinstance(end_year, str):
+        start_year = int(start_year[0:4])
+        end_year = int(end_year[0:4])
+    required_years = set(range(start_year, end_year + 1))
+    var['start_year'] = start_year
+    var['end_year'] = end_year
```
bouweandela, line 169: Changing the contents of the `var` dictionary is a really unexpected side effect of a function that is supposed to check data availability. Could you think of a better place to add this information?

bouweandela:
> As suggested by @bouweandela, I am merging the clipping of #1214 here.

Thanks @sloosvel! I ran various variations on recipe_python.yml and as far as I can see everything works fine

bouweandela:
> I expect one test to break, the test_filled_recipe because it will try to extract the timerange in an empty cube. I don't know how to circumvent this though.

I think you can avoid that problem testing smaller units of code, i.e. removing the failing test and writing a unit test that only tests  the function https://github.com/ESMValGroup/ESMValCore/blob/2f052ef82a54620b3e114a36f383ff52724c94b3/esmvalcore/_recipe.py#L1671-L1679 instead.

esmvalcore/_recipe_checks.py
sloosvel: A bunch of tests start breaking since I think they work directly with the check. Is it too bad to keep it here? Working with the start and end yeats is something that it's going to ideally be deprecated at some point anyway.

esmvalcore/esgf/_search.py
```diff
@@ -97,18 +102,25 @@ def esgf_search_files(facets):
...
 
-def select_by_time(files, start_year, end_year):
-    """Select files containing data between start_year and end_year."""
+def select_by_time(files, timerange):
+    """Select files containing data between a timerange."""
     selection = []
+    start_date, end_date = _parse_period(timerange)
+
+    start_date = str(start_date)
+    end_date = str(end_date)
```
bouweandela, line 111:
```suggestion
```
It looks like `_parse_period` always returns a tuple of str, so I think this can be deleted?

esmvalcore/_data_finder.py
```diff
@@ -108,23 +188,87 @@ def get_start_end_year(filename):
     return int(start_year), int(end_year)
 
 
-def select_files(filenames, start_year, end_year):
-    """Select files containing data between start_year and end_year.
-
-    This works for filenames matching *_YYYY*-YYYY*.* or *_YYYY*.*
+def _parse_period(timerange):
+    start_date = None
```
bouweandela, line 192: Could you add a docstring for this function and also for the other new functions that do not have one yet? A single line might be enough in many cases, but nothing at all seems a bit meagre. From our [contribution guidelines](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#what-should-be-documented):
> There is no need to write complete numpy style documentation for functions that are not visible in the [ESMValCore API Reference](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/api/esmvalcore.html#api) chapter on readthedocs. However, adding a docstring describing what a function does is always a good idea. For short functions, a one-line docstring is usually sufficient, but more complex functions might require slightly more extensive documentation.

bouweandela:
@ESMValGroup/esmvaltool-coreteam I think this is (as good as) ready to be merged. Would someone be willing to try out the new (and old) functionality? I tested it myself with a few modifications of `examples/recipe_python.yml`, but it would be good to get some more human testing done before merging because these are fairly extensive changes.

schlunma:
> @ESMValGroup/esmvaltool-coreteam I think this is (as good as) ready to be merged. Would someone be willing to try out the new (and old) functionality? I tested it myself with a few modifications of `examples/recipe_python.yml`, but it would be good to get some more human testing done before merging because these are fairly extensive changes.

Running some tests right now°

schlunma:
- [ ] `recipe_ecs.yml` fails with `Invalid value encountered for ``timerange``. Valid value must follow ISO 8601 standard for dates and duration periods, or be set to '*' to load available years. Got ['1', '150'] instead.`: [main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/7703587/main_log_debug.txt)

schlunma (changes requested):
- [ ] `recipe_ecs.yml` fails with `Invalid value encountered for ``timerange``. Valid value must follow ISO 8601 standard for dates and duration periods, or be set to '*' to load available years. Got ['1', '150'] instead.`: [main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/7703587/main_log_debug.txt)
- [ ] `recipe_tcr.yml` fails with `Invalid value encountered for ``timerange``. Valid value must follow ISO 8601 standard for dates and duration periods, or be set to '*' to load available years. Got ['1', '140'] instead.`: [main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/7703600/main_log_debug.txt)
- [ ] Discovered by accident: If I use `timerange: 19/P2Y3M` on something that provides data from 1850 to 2005, I do not get an error, but instead a preprocessed file that starts in 1901 and (correctly) constains 27 months. Am I missing something here? I would have expected an error due to missing the year "19" (recipe: see below): [main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/7703867/main_log_debug.txt)

Extended test recipe for new `timerange` argument:

```yml
# ESMValTool
# recipe_test_timerange.yml
---
documentation:
  description: test

  title: test

  authors:
    - schlund_manuel

  maintainer:
    - schlund_manuel

  references:
    - acknow_project

  projects:
    - esmval


datasets:
  - {dataset: CanESM2, project: CMIP5, exp: historical, ensemble: r1i1p1,
     start_year: 1850, end_year: 1850}
  - {dataset: CanESM2, project: CMIP5, exp: historical, ensemble: r1i1p1,
     timerange: 1860/1861}
  - {dataset: CanESM2, project: CMIP5, exp: historical, ensemble: r1i1p1,
     timerange: 187002/187005}
  - {dataset: CanESM2, project: CMIP5, exp: historical, ensemble: r1i1p1,
     timerange: 18800228/18800520}
  - {dataset: CanESM2, project: CMIP5, exp: historical, ensemble: r1i1p1,
     timerange: 18900504T100000/189006204T110000}
  - {dataset: CanESM2, project: CMIP5, exp: historical, ensemble: r1i1p1,
     timerange: 19/P2Y3M}
  # - {dataset: CanESM2, project: CMIP5, exp: historical, ensemble: r1i1p1,
  #    timerange: 1980/1982}
  # - {dataset: CanESM2, project: CMIP5, exp: historical, ensemble: r1i1p1,
  #    timerange: 1980/1982}


diagnostics:

  map:
    variables:
      tas:
        mip: Amon
    scripts:
      null
```

sloosvel:
Thanks everyone for the reviews!

I think the issues reported by @schlunma are coming from the fact that ISO8601 expects the years to have four digits. Otherwise the parser either throws an error or tries to complete the dates by adding 1s:
  - 1 can't get parsed
  - 950 gets parsed to 9501
  - 19 gets parsed to 1901

I can add a check to reformat to YYYY by adding zeroes if len(date) <4:
  - 1 would be 0001
  - 19 would be 0019 etc. 

So that existing recipes don't fail. But in general, and for future recipes, it would be better if years are given as 4-digit numbers.

schlunma:
> I think the issues reported by @schlunma are coming from the fact that ISO8601 expects the years to have four digits. Otherwise the parser either throws an error or tries to complete the dates by adding 1s:
> 
>     * 1 can't get parsed
> 
>     * 950 gets parsed to 9501
> 
>     * 19 gets parsed to 1901

Ah, yes, that makes sense!

> So that existing recipes don't fail. But in general, and for future recipes, it would be better if years are given as 4-digit numbers.

I don't think that's possible, since `yaml` interprets numbers with leading 0's as octal numbers (see https://yaml.org/spec/1.0/, sect. 2.4).

So I think adding a check that ensures that the year is in the format YYYY is the way to go here (as you suggested)!

valeriupredoi:
> I don't think that's possible, since yaml interprets numbers with leading 0's as octal numbers (see https://yaml.org/spec/1.0/, sect. 2.4).

not if they're strings :thread:

schlunma:
Not at the moment :grimacing: 

`datasets.0.start_year: '0001' is not a int.`

valeriupredoi:
> Doesn't work at the moment grimacing
> 
> `yamale.yamale_error.YamaleError: Error validating data '/home/b/b309141/recipes/recipe_test_timerange.yml' with schema '/mnt/lustre01/pf/b/b309141/ESMValCore_2/esmvalcore/recipe_schema.yml': datasets.0.start_year: '0001' is not a int.`

yeah since the recipe loader expects `int`s from the years; we could do a dual check if it's string or not, but I'm not sure if it's a good idea, if users start inputting `cow` for years, numpy will be udderly confused :cow2:

schlunma:
> yeah since the recipe loader expects `int`s from the years; we could do a dual check if it's string or not, but I'm not sure if it's a good idea, if users start inputting `cow` for years, numpy will be udderly confused cow2

:+1: 

That's why I think we should just make sure that the input is in the format YYYY and pad with 0s if necessary.

valeriupredoi:
do we have years before baby Jesus?

schlunma:
I don't think so and I don't hope so :smile:

zklaus:
> I don't think so and I don't hope so

As part of CMIP6, I give you [PMIP](http://clipc-services.ceda.ac.uk/dreq/u/PMIP.html).

schlunma:
But do these datasets use negative years? I think that's the important question. Sorry, can't find it on the website you posted.

zklaus:
> But do these datasets use negative years? I think that's the important question. Sorry, can't find it on the website you posted.

Good question. The page I linked to lists the experiments they defined. I am not sure if there are rules in place that are as clear as for historical and scenarios, but looking at actual data, MPI has published a "past2k" experiment, i.e. 2000 years before 1850. The time coordinate in the actual files seems to be about 7000-9000, but whether that is true for all, I don't know. It certainly seems plausible to me to have negative years, but if it's much hassle maybe not something we have to worry about right now.

sloosvel:
>but if it's much hassle maybe not something we have to worry about right now.

I agree. In any case I think with the latest commits the issue with the format of the years should be fixed now.

sloosvel:
Great! Anyone with the time to merge?

schlunma:
Before I merge: @bouweandela are you happy with these changes? I saw that you reviewed this PR but haven't approved it yet. Also, are the missing diff of 0.32% of codecov a problem?

valeriupredoi:
> Also, are the missing diff of 0.32% of codecov a problem?

yes, ideally we should accept only PRs that don't lower the coverage; having said that, there may be extenuating circumstances (like the removal of code that was inflating the coverage artifically before removal) - @sloosvel could pls explain the coverage delta? Very good work BTW, I will refrain from approving/merging myself since I didn't do a close review :beer:

schlunma:
> > Also, are the missing diff of 0.32% of codecov a problem?
> 
> yes, ideally we should accept only PRs that don't lower the coverage; having said that, there may be extenuating circumstances (like the removal of code that was inflating the coverage artifically before removal) - @sloosvel could pls explain the coverage delta? Very good work BTW, I will refrain from approving/merging myself since I didn't do a close review beer

As far as I understand the codecov output the overall coverage goes up, only the diff is not 100%, that means not all new lines are tested.

valeriupredoi:
yep that's right right here, Manu! You can toggle with "a" in the Files Changed to see which lines are not covered

sloosvel:
>  @sloosvel could pls explain the coverage delta?

There is a test that (sometimes) covers these lines: `/tests/integration/test_recipe/test_multi_model_filename`. The thing is that the lines compare the dates in filenames, and sometimes the date that is taken as a reference is already the first or the last, so the condition in the `if` is not always triggered. Everytime this test is run, it takes the reference file randomly. In some commits codecov reported that the coverage was 100% and in some others it misses this 0.32% I can try to write another test if it's bothering you

valeriupredoi:
@sloosvel cheers for the explanation! A test should never cover randomly, a test that takes random parameters as input is good to have in some cases, but its coverage should always be 100% ie if a certain `if` clause is not covered then the test is not working well, not the input parameters are at fault. It'd be great if you could change that :smiley:

schlunma:
Hi @sloosvel, @bouweandela just messaged me that he approves the changes in this PR :tada: 

Regarding the one line that is not always tested: I think it would make sense to write a quick unit test for the function `_get_statistic_attributes` with mocked input products that are a `list` and not a `set`. That way we ensure that the order of these products is machine-independent.

I think something like this can be used for the mocked products

https://github.com/ESMValGroup/ESMValCore/blob/b070a4b80315f21c19c1cbca3981bff94228f60e/tests/unit/preprocessor/_bias/test_bias.py#L15-L25

which can then be feed to `_get_statistic_attributes` as

```py
product_1 = PreprocessorFile(...)
...
product_n = PreprocessorFile(...)
products = [product_1, ..., product_n]

attributes = _get_statistic_attributes(products)
```

sloosvel:
>Hi @sloosvel, @bouweandela just messaged me that he approves the changes in this PR 🎉
Great!

>Regarding the one line that is not always tested: I think it would make sense to write a quick unit test for the function >_get_statistic_attributes with mocked input products that are a list and not a set. That way we ensure that the order of these >products is machine-independent.

Thanks for the tip, I'll try that. Better than whatever I was trying to do.

sloosvel:
I added a more clear test but ultimately what chooses the reference product is `next(iter(product))` and that changes with every run. I tried to increase the chances of choosing a reference product that triggers the conditions for the uncovered lined. But I'm not sure I can do much more about it.

tests/unit/test_recipe.py
```diff
@@ -327,3 +329,39 @@ def test_add_fxvar_keys_extra_facets():
         'raw_name': 'cell_area',
     }
     assert fx_var == expected_fx_var
+
+
+def test_multi_model_filename():
+    """Test timerange in multi-model filename is correct."""
+    cube = iris.cube.Cube(np.array([1]))
+    products = {
```
schlunma, line 337:
```suggestion
    products = [
```

tests/unit/test_recipe.py
```diff
@@ -327,3 +329,39 @@ def test_add_fxvar_keys_extra_facets():
...
+        PreprocessorFile(cube, 'R', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'S', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'T', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'U', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'V', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'W', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'X', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'Y', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'Z', {'timerange': '1990/1991'}),
+    }
```
schlunma, line 364:
```suggestion
    ]
```

schlunma (commented):
I think if you use a `list` instead of a `set` you should be fine. As far as I can tell the function doesn't use any set methods and `next(iter(list))` always returns the first element of the list.

tests/unit/test_recipe.py
```diff
@@ -327,3 +329,39 @@ def test_add_fxvar_keys_extra_facets():
...
+        PreprocessorFile(cube, 'Q', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'R', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'S', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'T', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'U', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'V', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'W', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'X', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'Y', {'timerange': '1990/1991'}),
+        PreprocessorFile(cube, 'Z', {'timerange': '1990/1991'}),
```
schlunma, line 363:
```suggestion
```

I think these can be removed now

sloosvel: Sure!

schlunma:
Great! @Saskia If you are happy with this PR I will merge it :tada:

sloosvel:
Sure, go ahead, many thanks!

PR #1147, opened 2021-05-25
Review fx handling

### LINKED ISSUE #1096: Issues with preprocessors that use fx variables
I found some small issues with the new implementation of the fx preprocessors in #999 (great job btw there @sloosvel!).

Issues
---------

1. Not directly related to #999, but do we have a list of all preprocessors for which `fx_variables` can be specified? Maybe I'm blind, but I didn't find it in the documentation. Basically the list given here: https://github.com/ESMValGroup/ESMValCore/blob/fee5d541dc871efe72ab972a6a52195c452d16eb/esmvalcore/_recipe.py#L482-L485

2. For `mask_landsea` we had checks for the shape of the fx variables in place. If the shapes did not match, Natural Earth masks were used. Right now, if the shapes do not match, an error is printed as reported by @LisaBock [here](https://github.com/ESMValGroup/ESMValTool/issues/2155#issuecomment-832060910): `ValueError: Dimensions of tas and sftof cubes do not match. Cannot broadcast cubes.` This is in particular problematic since for `mask_landsea`, the tool automatically adds the fx variables `sftlf` and `sftof` if none is specified. Since `sftlf` is always checked first, the tool is most likely to fail if an ocean variable is specified since the shapes of `sfltf` and ocean variables differ for most models. For example, the first recipe fails with the error `ValueError: Dimensions of tos and sftlf cubes do not match. Cannot broadcast cubes.`, and the second one runs fine:

<details>
  <summary>Recipe 1 (fails)</summary>
  
  ```yml
preprocessors:
  test:
    mask_landsea:
      mask_out: land

diagnostics:

  test:
    variables:
      tos:
        preprocessor: test
        project: CMIP5
        mip: Omon
        exp: historical
        start_year: 2000
        end_year: 2000
        ensemble: r1i1p1
        additional_datasets:
          - {dataset: CanESM2}
    scripts: null
  ```
</details>

<details>
  <summary>Recipe 2 (works)</summary>
  
  ```yml
preprocessors:
  test:
    mask_landsea:
      mask_out: land
      fx_variables:
        sftof:

diagnostics:

  test:
    variables:
      tos:
        preprocessor: test
        project: CMIP5
        mip: Omon
        exp: historical
        start_year: 2000
        end_year: 2000
        ensemble: r1i1p1
        additional_datasets:
          - {dataset: CanESM2}
    scripts: null
  ```
</details>

3. For `weighting_landsea_fraction`, we had similar checks. The tool tried `sftlf` and `sftof`, and only if both had different shapes, an error was printed. Right now, due to the prioritization of `sftlf`, an error is e.g. printed for every ocean variables that is not on an atmospheric grid (thus does not match the shape of `sftlf`), even though `sftof` would work just fine.

4. As far as I can tell from the code (didn't test though!) the remaining preprocessors should behave similarly to the old version (the error messages might differ, though).

5. The following message is now printed to the log. I think we should demote this to a debug message and maybe only print the `short_name` of the fx variables as info message.

<details>
  <summary>Log</summary>

  ```
2021-05-05 08:28:29,557 UTC [14161] INFO    Using fx_files: {'sftlf': {'alias': 'CanESM2',
           'dataset': 'CanESM2',
           'diagnostic': 'test',
           'end_year': 2000,
           'ensemble': 'r0i0p0',
           'exp': 'historical',
           'filename': '/pf/b/b309141/work/CMIP5_DKRZ/CCCma/CanESM2/historical/fx/atmos/fx/r0i0p0/v20120410/sftlf/sftlf_fx_CanESM2_historical_r0i0p0.nc',
           'frequency': 'fx',
           'institute': ['CCCma'],
           'long_name': 'Land Area Fraction',
           'mip': 'fx',
           'modeling_realm': ['atmos'],
           'original_short_name': 'sftlf',
           'preprocessor': 'test',
           'project': 'CMIP5',
           'recipe_dataset_index': 0,
           'short_name': 'sftlf',
           'standard_name': 'land_area_fraction',
           'start_year': 2000,
           'units': '%',
           'variable_group': 'sftlf'},
 'sftof': {'alias': 'CanESM2',
           'dataset': 'CanESM2',
           'diagnostic': 'test',
           'end_year': 2000,
           'ensemble': 'r0i0p0',
           'exp': 'historical',
           'filename': '/pf/b/b309141/work/CMIP5_DKRZ/CCCma/CanESM2/historical/fx/ocean/fx/r0i0p0/v20130119/sftof/sftof_fx_CanESM2_historical_r0i0p0.nc',
           'frequency': 'fx',
           'institute': ['CCCma'],
           'long_name': 'Sea Area Fraction',
           'mip': 'fx',
           'modeling_realm': ['ocean'],
           'original_short_name': 'sftof',
           'preprocessor': 'test',
           'project': 'CMIP5',
           'recipe_dataset_index': 0,
           'short_name': 'sftof',
           'standard_name': 'sea_area_fraction',
           'start_year': 2000,
           'units': '%',
           'variable_group': 'sftof'}} for variable tos during step mask_landsea
  ```
</details>

Possible solution
----------------------

The main issues here are 2. and 3. I think both can be fixed if we don't raise an exception [here](https://github.com/ESMValGroup/ESMValCore/blob/fee5d541dc871efe72ab972a6a52195c452d16eb/esmvalcore/preprocessor/_ancillary_vars.py#L80-L83) and [here](https://github.com/ESMValGroup/ESMValCore/blob/fee5d541dc871efe72ab972a6a52195c452d16eb/esmvalcore/preprocessor/_ancillary_vars.py#L120-L123) if the shapes of the ancilliary files do not match the cube. If a preprocessor cannot handle the case that no fx files are found/the shapes differ, the errors should be raised in the corresponding preprocessing function.

It would be also nice to add some tests for these cases, similiar to the [ones we had before](https://github.com/ESMValGroup/ESMValCore/pull/999/files?file-filters%5B%5D=.py#diff-e57a19c9bf65c498803a124f8d455e21ea4d03ed5d1be8f51d6fa4570bb25623L56-L92).

sloosvel:
## Description
This pull request should address the issues reported in #1096. 

Closes #1096 

Link to documentation:

***

***

### DIFF
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -357,7 +357,7 @@ def _search_fx_mip(tables, found_mip, variable, fx_info, config_user):
                 logger.debug(
                     "Found fx variables '%s':\n%s",
                     fx_info['short_name'], pformat(fx_files))
-                return found_mip, fx_info, fx_files
+    return found_mip, fx_info, fx_files
 
 
 def _get_fx_files(variable, fx_info, config_user):
@@ -441,11 +441,10 @@ def _update_fx_files(step_name, settings, variable, config_user, fx_vars):
             fx_info['filename'] = fx_files
             settings['add_fx_variables']['fx_variables'].update({
                 fx_var: fx_info
-            })
-        
+            })        
             logger.debug('Using fx_files: %s for variable %s during step %s',
-                          pformat(fx_files),
-                          variable['short_name'], step_name)
+                pformat(fx_files),
+                variable['short_name'], step_name)
 
 
 def _fx_list_to_dict(fx_vars):
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -441,8 +441,9 @@ def _update_fx_files(step_name, settings, variable, config_user, fx_vars):
             fx_info['filename'] = fx_files
             settings['add_fx_variables']['fx_variables'].update({
                 fx_var: fx_info
-            })        
-            logger.debug('Using fx_files: %s for variable %s during step %s',
+            })
+            logger.debug(
+                'Using fx_files: %s for variable %s during step %s',
                 pformat(fx_files),
                 variable['short_name'], step_name)
 
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -442,9 +442,9 @@ def _update_fx_files(step_name, settings, variable, config_user, fx_vars):
             settings['add_fx_variables']['fx_variables'].update({
                 fx_var: fx_info
             })
-            logger.debug(
-                'Using fx_files: %s for variable %s during step %s',
-                pformat(fx_files),
+            logger.info(
+                'Using fx files for variable %s during step %s: %s',
+                variable['short_name'], step_name, pformat(fx_files))
                 variable['short_name'], step_name)
 
 
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -445,7 +445,6 @@ def _update_fx_files(step_name, settings, variable, config_user, fx_vars):
             logger.info(
                 'Using fx files for variable %s during step %s: %s',
                 variable['short_name'], step_name, pformat(fx_files))
-                variable['short_name'], step_name)
 
 
 def _fx_list_to_dict(fx_vars):
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -357,7 +357,7 @@ def _search_fx_mip(tables, found_mip, variable, fx_info, config_user):
                 logger.debug(
                     "Found fx variables '%s':\n%s",
                     fx_info['short_name'], pformat(fx_files))
-    return found_mip, fx_info, fx_files
+                return found_mip, fx_info, fx_files
 
 
 def _get_fx_files(variable, fx_info, config_user):
@@ -442,10 +442,10 @@ def _update_fx_files(step_name, settings, variable, config_user, fx_vars):
             settings['add_fx_variables']['fx_variables'].update({
                 fx_var: fx_info
             })
-
-    logger.info('Using fx_files: %s for variable %s during step %s',
-                pformat(settings['add_fx_variables']['fx_variables']),
-                variable['short_name'], step_name)
+        
+            logger.debug('Using fx_files: %s for variable %s during step %s',
+                          pformat(fx_files),
+                          variable['short_name'], step_name)
 
 
 def _fx_list_to_dict(fx_vars):
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -469,15 +469,29 @@ def _update_fx_settings(settings, variable, config_user):
             user_fx_vars = _fx_list_to_dict(user_fx_vars)
             step_settings['fx_variables'] = user_fx_vars
         if not user_fx_vars:
-            if step_name in ('mask_landsea', 'weighting_landsea_fraction'):
-                user_fx_vars = {'sftlf': None}
-                if variable['project'] != 'obs4mips':
-                    user_fx_vars.update({'sftof': None})
-            elif step_name == 'mask_landseaice':
-                user_fx_vars = {'sftgif': None}
-            elif step_name in ('area_statistics', 'volume_statistics'):
-                user_fx_vars = {}
-            step_settings['fx_variables'] = user_fx_vars
+            default_fx = {
+                'area_statistics': {
+                    'areacella': None,
+                    'areacello': None,
+                },
+                'mask_landsea': {
+                    'sftlf': None,
+                },
+                'mask_landseaice': {
+                    'sftgif': None,
+                },
+                'volume_statistics': {
+                    'volcello': None,
+                },
+                'weighting_landsea_fraction': {
+                    'sftlf': None,
+                },
+            }
+            if variable['project'] != 'obs4mips':
+                default_fx['mask_landsea'].update({'sftof': None})
+                default_fx['weighting_landsea_fraction'].update(
+                    {'sftof': None})
+            step_settings['fx_variables'] = default_fx[step_name]
 
     fx_steps = [
         'mask_landsea', 'mask_landseaice', 'weighting_landsea_fraction',
--- a/esmvalcore/preprocessor/_ancillary_vars.py
+++ b/esmvalcore/preprocessor/_ancillary_vars.py
@@ -48,6 +48,7 @@ def _load_fx(var_cube, fx_info, check_level):
 
     return fx_cube
 
+
 def _is_fx_broadcastable(fx_cube, cube):
     try:
         da.broadcast_to(fx_cube.core_data(), cube.shape)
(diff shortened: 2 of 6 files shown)


esmvalcore/preprocessor/_ancillary_vars.py
```diff
@@ -46,6 +49,17 @@ def _load_fx(fx_info, check_level):
     return fx_cube
 
 
+def _is_fx_broadcastable(fx_cube, cube):
+    try:
+        da.broadcast_to(fx_cube.core_data(), cube.shape)
```
valeriupredoi, line 54: am a bit confused as to why the broadcasting test, since all we need is to check if the fx data is time dependent or not: if it's not then a simple (X, Y) shapes check is enough, if it's time-dependent then we have to be very careful time points match and broadcasting might work but time points might differ so things get messy

valeriupredoi (commented):
cheers @sloosvel - looks good to me apart from a question and a minor remark :beer:

esmvalcore/preprocessor/_ancillary_vars.py
sloosvel: I think it's more straightforward to do that. It could be that the shapes do not match because the fx var and the var are on differents grids. So this is a way to take into account all these cases with just one test.

schlunma (commented):
Nice! Tested with real data, works as expected :tada: 

Couple of comments regarding the logging. Due to the size of some messages this can get really crowded.

- https://github.com/ESMValGroup/ESMValCore/blob/933b37b225a89aeed84029d8c83105407ed7ae5f/esmvalcore/_recipe.py#L432-L448

This prints all the fx variables used for the variable. If one uses more than one preprocessor that uses fx variables, the output is wrong here. Maybe something like this is better:
```diff
+   new_fx_files = {}
    for fx_var, fx_info in fx_vars.items():
        if not fx_info:
            fx_info = {}
        if 'mip' not in fx_info:
            fx_info.update({'mip': None})
        if 'short_name' not in fx_info:
            fx_info.update({'short_name': fx_var})
        fx_files, fx_info = _get_fx_files(variable, fx_info, config_user)
        if fx_files:
            fx_info['filename'] = fx_files
+           new_fx_files[fx_var] = fx_info
-           settings['add_fx_variables']['fx_variables'].update({
-               fx_var: fx_info
-           })
+   settings['add_fx_variables']['fx_variables'].update(new_fx_files)


-   logger.info('Using fx_files: %s for variable %s during step %s',
-               pformat(settings['add_fx_variables']['fx_variables']),
-               variable['short_name'], step_name)
+   logger.info('Using fx_files: %s for variable %s during step %s',
+               list(new_fx_files), variable['short_name'], step_name)
+   logger.debug('Using fx_files: %s for variable %s during step %s',
+                pformat(new_fx_files), variable['short_name'], step_name)
```

- https://github.com/ESMValGroup/ESMValCore/blob/933b37b225a89aeed84029d8c83105407ed7ae5f/esmvalcore/preprocessor/_area.py#L212-L215
Replace `cube` with `cube.summary(shorten=True)`.

esmvalcore/preprocessor/_ancillary_vars.py
```diff
@@ -46,6 +49,17 @@ def _load_fx(fx_info, check_level):
...
+    except ValueError as exc:
+        logger.debug("Dimensions of %s and %s cubes do not match. "
+                     "Discarding use of fx_variable: %s",
+                     cube.var_name, fx_cube.var_name, exc)
```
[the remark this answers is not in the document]
sloosvel: This would print the exception message from dask:

`2021-06-01 10:15:51,152 UTC [17894] DEBUG   Dimensions of o2 and volcello cubes do not match. Discarding use of fx_variable: cannot broadcast shape (12, 75, 292, 362) to shape (1, 75, 292, 362)`

sloosvel:
> This prints all the fx variables used for the variable. If one uses more than one preprocessor that uses fx variables, the output is wrong here.

Sorry, I don't see what you mean here. Would it work for you if the debug message was printed inside the `if fx_files` statement, at the very end? That way it get's printed per fx_variable.

esmvalcore/_recipe.py
```diff
@@ -442,9 +442,9 @@ def _update_fx_files(step_name, settings, variable, config_user, fx_vars):
             settings['add_fx_variables']['fx_variables'].update({
                 fx_var: fx_info
             })
-
-    logger.info('Using fx_files: %s for variable %s during step %s',
-                pformat(settings['add_fx_variables']['fx_variables']),
+            logger.debug(
+                'Using fx_files: %s for variable %s during step %s',
+                pformat(fx_files),
```
schlunma, line 447:
```suggestion
            logger.info(
                'Using fx files for variable %s during step %s: %s',
                variable['short_name'], step_name, pformat(fx_files))
```

I think it's good to write this to `info` after all. Apart from that I'm happy with the changes! :+1:

zklaus:
This feels like we might be able to just about squeeze into 2.3.0.

@sloosvel, could you please go through the checklist and tick all the right boxes?

@schlunma, do I understand you correctly that you would approve with this last logging change in place?

@valeriupredoi, care to do a final sweep?

schlunma:
@zklaus Without this fix many of our recipes in the ESMValTool repository fail, so we have to include this into 2.3.0. Yes, I will approve after the logging change.

esmvalcore/_recipe.py
```diff
@@ -442,9 +442,9 @@ def _update_fx_files(step_name, settings, variable, config_user, fx_vars):
             settings['add_fx_variables']['fx_variables'].update({
                 fx_var: fx_info
             })
-
-    logger.info('Using fx_files: %s for variable %s during step %s',
-                pformat(settings['add_fx_variables']['fx_variables']),
+            logger.info(
+                'Using fx files for variable %s during step %s: %s',
+                variable['short_name'], step_name, pformat(fx_files))
                 variable['short_name'], step_name)
```
```suggestion
```

Sorry, I messes up the suggestion!

schlunma (approved):

valeriupredoi:
very good stuff, guys, let me just give it a one last looksee meself

esmvalcore/preprocessor/_ancillary_vars.py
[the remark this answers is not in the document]
valeriupredoi: nice, thanks Saskia! it does relay good information indeed.

esmvalcore/preprocessor/_ancillary_vars.py
valeriupredoi: OK but I am still concerned with two cases: 1. eg the fx variable is 2d (lat-lon) and the cube data is 3d (time-lat-lon) - do we not use the fx variable in that case? 2. the fx variable is on a finer or coarser time axis than the cube data - in this case we should do a time regridding operation and still use the fx data. OK these cases are outside the scope of this PR so I'll just keep this in mind for a future feature request :+1:

valeriupredoi (approved):
cheers Saskia! The only comment that I'm still thinking we should be addressing (but not in this PR) is the one about the hard cut-off due to fx data not being broadcastable onto cube data. But let's get this one in! Good work :beer:

zklaus:
Thanks @valeriupredoi, could you please open an issue about it so that we don't forget?

PR #1150, opened 2021-05-28
Use native iris functions in multi-model statistics

Peter9192:
## Description

Split off from #968 to separate the lazy and eager evaluation pathways. 

This PR re-implements the multi-model statistics preprocessor. This implementation delegates more of the work to native iris functions. This enables, among others, lazy evaluation for some common operators. The actual implementation of that lazy pathway is implemented in #968. 

Related to #781
***

- [ ] [🧪][2] and [🛠][1] ~~[Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available~~
- [ ] [🛠][1] ~Any changed [dependencies have been added or removed](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#dependencies) correctly~

***


Peter9192:
@zklaus as promised I've split off the refactoring from the lazy multi-model stats. 
@valeriupredoi perhaps you can have a quick look at this, since you've already seen all these changes in #968?

There's 2 regression tests failing now with some annoying rounding errors due to numpy changing the dtype on masked data. If you agree I'll re-generate the reference data for the regression tests so we get green lights from CircleCI.

bouweandela:
> There's 2 regression tests failing now with some annoying rounding errors due to numpy changing the dtype on masked data. If you agree I'll re-generate the reference data for the regression tests so we get green lights from CircleCI.

Masks are stored as fill values in netcdf files, so it seems unlikely that regenerating the reference data will change the data type of the resulting masks.

Peter9192:
> 
> 
> > There's 2 regression tests failing now with some annoying rounding errors due to numpy changing the dtype on masked data. If you agree I'll re-generate the reference data for the regression tests so we get green lights from CircleCI.
> 
> Masks are stored as fill values in netcdf files, so it seems unlikely that regenerating the reference data will change the data type of the resulting masks.

The point is that whenever the input data is a masked array, (even if mask is set to false everywhere) `iris.analysis.MEAN` will call `np.ma.mean`, and this one changes the dtype on the data (so also where it's not masked). This is what causes the precision error.

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -3,184 +3,65 @@
...
-
-def _compute_statistic(data, statistic_name):
-    """Compute multimodel statistic."""
-    data = np.ma.array(data)
-    statistic = data[0]
-
-    if statistic_name == 'median':
-        statistic_function = np.ma.median
-    elif statistic_name == 'mean':
-        statistic_function = np.ma.mean
```
bouweandela, line 73: The previous implementation also use `np.ma.mean`, so I still don't get why it's different now?

Peter9192: I'm also not sure, but I think maybe it's because there, we specifically allocate the output array as dtype `float32` before filling it with the result of `np.ma.mean`.

https://github.com/ESMValGroup/ESMValCore/blob/3078908caee50d4abc6a1ed4ffbbd994d594f17a/esmvalcore/preprocessor/_multimodel.py#L276-L278

zklaus: Ah yes, that would definitely do that. It is effectively casting to the dtype of the target array, thus would be roughly equivalent to sticking a `.astype(np.float32)` at the array. Should we do that instead?

valeriupredoi:
cheers @Peter9192 - thanks to @bouweandela nudging me, I am now looking at this :beer:

esmvalcore/preprocessor/_multimodel.py
valeriupredoi: well we need to make sure the data sticks to the rule to be float32 so yeah, I'd argue that's needed

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -248,115 +128,179 @@ def _unify_time_coordinates(cubes):
...
-            subset.mask = True
-        time_slice.append(subset)
-    return time_slice
+def _time_coords_are_aligned(cubes):
+    """Return `True` if time coords are aligned."""
+    first_time_array = cubes[0].coord('time').points
 
+    for cube in cubes[1:]:
+        other_time_array = cube.coord('time').points
+        if not np.array_equal(first_time_array, other_time_array):
```
valeriupredoi, line 137: they may differ at very low orders, since those are points in decimal format - maybe use np.allclose()?

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -248,115 +128,179 @@ def _unify_time_coordinates(cubes):
...
-    stats_data = np.ma.zeros(new_shape, dtype=np.dtype('float32'))
 
-    # Realize all cubes at once instead of separately for each time slice
-    _ = [cube.data for cube in cubes]
+def _subset(cube, time_points):
+    """Subset cube to given time range."""
+    begin = cube.coord('time').units.num2date(time_points[0])
+    end = cube.coord('time').units.num2date(time_points[-1])
+    constraint = iris.Constraint(time=lambda cell: begin <= cell.point <= end)
+    return cube.extract(constraint)
```
valeriupredoi, line 148: I think a try/except is needed here with except raising a nicely explained ValueError just in case

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -248,115 +128,179 @@ def _unify_time_coordinates(cubes):
...
+    cube_list.append(cube)
+
+    for time_point in pad_end:
+        new_slice = template_cube.copy()
+        new_slice.coord('time').points = float(time_point)
+        cube_list.append(new_slice)
+
+    cube_list = iris.cube.CubeList(cube_list)
+
+    new_cube = cube_list.concatenate_cube()
```
valeriupredoi, line 192: and here as well, a try/except would be better to have, exception raising a nicely explained concatenation error (iris.ConcatenationError)

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -248,115 +128,179 @@ def _unify_time_coordinates(cubes):
...
+        collapsed_slice = combined_slice.collapsed(CONCAT_DIM, operator,
+                                                   **kwargs)
+
+        # some iris aggregators modify dtype, see e.g.
+        # https://numpy.org/doc/stable/reference/generated/numpy.ma.average.html
+        collapsed_slice.data = collapsed_slice.data.astype(np.float32)
+
+        result_slices.append(collapsed_slice)
+
+    result_cube = iris.cube.CubeList(result_slices).merge_cube()
```
valeriupredoi, line 271: can this fail now? or we've already processed those cubes to a point where it will never fail

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -399,23 +343,39 @@ def multi_model_statistics(products,
...
+    decimal part optional).
+
+    Notes
+    -----
+    Some of the operators in :py:mod:`iris.analysis` require additional
+    arguments. Except for percentiles, these operators are currently not
+    supported.
+
+    Lazy operation is supported for all statistics, except
+    ``median``, ``percentile``, ``gmean`` and ``hmean``.
```
valeriupredoi, line 369: nice! we need to make sure this list gets updated when things change

valeriupredoi (commented):
very nice implementation @Peter9192 - I left a few comments here and there, not anything serious. I'll have a look at the CI tests now then approve :+1:

valeriupredoi:
ah gotcha - so the tests fail coz of the same issue I pointed out in the code, time points are not quite equal, tests can be fixed by upping the `rtol` and I'd do the same in the code too, use `allclose` with some higher `rtol` but maybe not 1 coz that's be they really are different :grin:

valeriupredoi (approved):
pre-emptive approval, but please fix the 2 tests and have a look at me comments too. Apart from that - really nice work! :+1:

esmvalcore/preprocessor/_multimodel.py
zklaus: The CubeList won't concatenate/merge if the points differ even a little bit. I guess that is the purpose of this check?

Peter9192:
Thanks for the review V, I'm on it

esmvalcore/preprocessor/_multimodel.py
valeriupredoi: right! I was out of context when I made the comment above, but yes, that'll go in `_align` then into `_compute_eager` so perfect alignment is indeed needed for merging. But if you plop an `allclose()` with a decent `rtol` here and if that's True then you can switch off all the other alignment jazz and just use eg first cube's time axis as reference?

valeriupredoi: maybe not worth the effort since those are cheap computations I reckon

esmvalcore/preprocessor/_multimodel.py
Peter9192: In the PR we already have this: https://github.com/ESMValGroup/ESMValCore/blob/223acd7841e8c3e2c2e8c3af0de57c50e80ec02b/esmvalcore/preprocessor/_multimodel.py#L267

esmvalcore/preprocessor/_multimodel.py
Peter9192: If that `allclose()` is true, you'd still need to somehow overwrite the other cubes with the reference cube, which means more conditionals, more messing with the cubes, more possibilities for errors.. 

Accomodating input data/coords with slightly (in the order of numerical precision) deviating values and treat them as equal sound to me like we should have a separate discussion about that, and not just on MM-stats.

esmvalcore/preprocessor/_multimodel.py
Peter9192: Done 👍

esmvalcore/preprocessor/_multimodel.py
Peter9192: Done 👍

esmvalcore/preprocessor/_multimodel.py
Peter9192: I would say it shouldn't, but apparently it can fail indeed. E.g. if you calculate the mean of integer arrays `[1, 2]` and `[3, 3]` one index at a time, you'll end up concatenating `int(2)` and `float(2.5)`, which fails. Normally all input should be dtype `float32` and if dtype is preserved in the calculations, that shoudn't happen. 

Just in case, I'll add a try/except clause here as well, with a nice message.

esmvalcore/preprocessor/_multimodel.py
Peter9192: Good spot! Actually I forgot to remove this one from #968 , that's were all the lazy stuff is implemented. This PR just makes the code ready for that.

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -3,184 +3,65 @@
...
+def _resolve_operator(statistic: str):
+    """Find the operator corresponding to the statistic."""
+    statistic = statistic.lower()
+    kwargs = {}
+
+    # special cases
+    if statistic == 'std':
+        logger.warning(
+            "Multicube statistics is aligning its behaviour with iris.analysis"
+            ". Please consider replacing 'std' with 'std_dev' in your code.")
```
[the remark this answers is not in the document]
Peter9192: Done 👍

Peter9192:
I think I covered all your comments @valeriupredoi. It's just those 2 regression tests that keep failing with precision errors. As I said before, I think we should just regenerate the sample data and merge. I don't think an unsurprising 0.00001% difference in a test result is worth our time and effort.

Peter9192:
@zklaus @valeriupredoi Note that I just regenerated the sample data for the failing tests. They both reported a 
`Max relative difference: 2.4267516e-07` which is almost certainly due to a back and forth conversion to dtype `float64` that's introduced because we switched from explicitly allocating the output array to having iris generate it for us.

If you agree that's okay, this PR can be merged. Otherwise, feel free to suggest/commit a different solution.

zklaus:
Thanks, I am happy with this. Could you just make a quick pass through the checklist, tick the things that are present, and strike out those you deem inapplicable? I started with this, but wanted explicit confirmation from you on backwards compatibility and potential changes in dependencies.

valeriupredoi:
excellent work, Peter! Yes, let's get this in - I'd like to carry on the discussion about (almost) equal time (and not only) coords - but defo not here, and in a different issue :beer:

Peter9192:
@zklaus done, there's no changes to the user interface, and while scipy is removed as a direct dependency for mm-stats, it's still used in many other places in the code, so I think we're good there :+1:

zklaus (approved):

axel-lauer:
We already discussed the problem introduced with too strict iris checks when averaging near-surface air temperature 'tas' as it is reported by some models at 1.5 m, by other models at 2.0 m. From the scientific point of view, regarding this as an error makes no sense as @ruthlorenz nicely explained at our last monthly meeting. Here is another example why these iris checks can be too strict and yet not useful from a scientific point of view:

Variables on hybrid vertical levels (e.g. clw) have to be converted to pressure or height levels to be processed by most diagnostics. For this conversion, some models provide an auxiliary coordinate 'p0'. Calculating the multi-model mean over such datasets converted to pressure or height levels fails because "p0" is kept after converting the vertical levels but (as expected) not identical in all models. p0 contains no useful information any more after vertical coordinate transformation has been done. Yet, this results in a failure to calculate the multi-model mean over such 3-dim variables (e.g. 3-dim cloud variables).

Peter9192:
@axel-lauer Is this the same issue as described in #1213 and addressed in #1220?

zklaus:
The `tas` issue is discussed at https://github.com/ESMValGroup/ESMValTool/discussions/2218#discussioncomment-956403 and #1204. Turns out there never was a problem; it was just a misdiagnosis.

I think the vertical levels are a different issue. @axel-lauer, could you please open a new issue for that, including the failing recipe, and perhaps also add it to ESMValGroup/ESMValTool#2218?

Issue #1143, opened 2021-05-22
Handle IPSLCM native formats  


senesis:
This **Handle IPSLCM native formats** feature is related to IS-ENES3 WP9.5 "_Enhancing the use of the ESMValTool for model development_" and more precisely its subtask "_Incorporation of ESMValTool into modelling or stand-alone workflow_" , and to IS-ENES3 deliverable D9.2 "_ESMValTool version enabling model development usage_" 

The intent is to add this feature for [release 2.3.0](https://github.com/ESMValGroup/ESMValCore/milestone/6), for which PR's are due for June, 7 (so, yet in two weeks time)

Handling IPSLCM requires :

1. to find datafiles by translating variable names to native variable names
2. to handle filenames where experiment name may include a date (independently of the CMIP sub-experiment concept), and this experiment name can be a neighbour of the dates range in the filename
3. to develop specific fixes for that model's data
4. to get an acceptable performance when using the IPSLCM data format which include multiple variables in a single NetCDF file (called 'Output')

The current status is : 
1. Item 1 is covered by #1124, but that PR is not related to [milestone 2.3.0](https://github.com/ESMValGroup/ESMValCore/milestone/6)
2. Item 2 is explained in #1127, which proposes a bugfix. Because it is related to #1078 which proposes a more generic solution, it has been closed. However, #1078 is not related to  [milestone 2.3.0](https://github.com/ESMValGroup/ESMValCore)
3. Item 3 is OK. However, until #494 is developed and merged, handling the two flavors of IPSLCM data format implies to create two projects, and thus two fixes directories (which share a single fix module)
4. Item 4 is addressed in an [ESMValCore issue](https://github.com/ESMValGroup/ESMValTool/issues/2141), and solved by a [feature of the developed fixes](https://github.com/ESMValGroup/ESMValTool/issues/2141#issuecomment-845927522) 

So, the open questions are: 
- Is it the intent that PR #1124 be merged for [milestone 2.3.0](https://github.com/ESMValGroup/ESMValCore/milestone/6)? Note that I already tested IPSLCM fixes using corresponding @zklaus repository's branch 
- Should issue #1127 be re-opened (beyond @bouweandela comment) and turned into a PR marked for [milestone 2.3.0](https://github.com/ESMValGroup/ESMValCore/milestone/6) ? (I would elaborate it)
[...]

PR #1153, opened 2021-05-31
Handle IPSL-CM6 

senesis:
## Description

Related to issue #1143. Allow to handle IPSL-CM6 data formats and file naming conventions. 

Composed of  :
- the addition of two files : a fix module and a variable mapping file (extra_facets, see #1124 )
- the update of config-developer.yml and config-user.yml
- update of the documentation 

The feature won't actually work without #1124 , but the proposed PR is compatible with the main branch before #1124 merge, in order to allow for immediate merge, and so to avoid bottlenecks. #1124 allows for finding the right file and the right variable in file when using ISPL-CM6 data.

Some documentation on the handling of native models has been introduced.

Performance in loading variables from IPSL's multi-variable datafiles with Iris  is poor, even with the Iris 3.0.2 fix on that subject. This is why pre-selecting using CDO in a fix_data method is implemented. However, because CDO licence (GPL v2) is not compatible with ESMValTool licence policy, this feature is not activated by default, and no explicit dependency is introduced. The feature can be activated through a variable mappings file (see #1124), at those sites which accept to be dependent on CDO.

***

- [x] [🧪][2] The new functionality is [relevant and scientifically sound](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#scientific-relevance) : **scientific relevance is explained in IS-ENES3 WP9.5 2nd subtask : Enhancing the use of the ESMValTool for model development**
- [x] [🧪][2] and [🛠][1] [Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available : **a slight change of doc structure was introduced (renaming quickstart section 'Finding data' to 'Input data') , with relevant changes in quickstart/configure and develop/fixing_data**
- [x] [🛠][1] Any changed [dependencies have been added or removed](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#dependencies) correctly : **see above about CDO**
- [ ] [🛠][1] All [checks below this pull request](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-checks) were successful : **Codacy check failures will be removed by merge of #1124 , except for the use of subprocess. Commit checks are no more activated, for an unknown reason**


esmvalcore/cmor/_fixes/native6/ipsl_cm6.py
```diff
@@ -0,0 +1,119 @@
...
+    """Fixes for ISPLCM 2m temperature."""
+    def fix_metadata(self, cubes):
+        """Add height2m."""
+        varname = self.var_mapping.get(KEY_FOR_VARNAME, self.vardef.short_name)
+        cube = self.get_cube_from_list(cubes, varname)
+        add_scalar_height_coord(cube)
+        return cubes
+
+
+class Huss(Fix):
```
jvegasbsc, line 112: You can do 

```python
class Huss(Tas):
    """Fixes for ISPLCM 2m specific humidity."""
    pass
```

and avoid code duplication, as both classes are identical

esmvalcore/cmor/_fixes/native6/ipsl_cm6.py
```diff
@@ -0,0 +1,119 @@
...
+        diagnosed by searching in the input filepathame for the mapping
+        value for key 'group'.
+
+        In such cases, it is worth to use an external tool for
+        filtering, at least until Iris loads fast (which is not the case
+        up to, and including, V3.0.2)
+
+        However, we take care of ESMValTool policy re. dependencies licence
+        """
+        if "_" + self.var_mapping.get("group",
```
jvegasbsc, line 31: I see several uses of `self.var_mapping` but I can not find the assigmment

jvegasbsc (commented):
One question: how can I access to a data sample so I can properly test it?

esmvalcore/cmor/_fixes/native6/ipsl_cm6.py
senesis: This PR anticipates on a merge of #1124, which assigns var_mapping through reading a mapping file (this means that IPSL-CM data cannot be processed right now, as this would raise a bug on this code line) 
I tried to explain that, and why, in the [introductory text for the PR](https://github.com/ESMValGroup/ESMValCore/pull/1153#issue-658181932). 
I tested the whole on a local branch which included #1124. I can push such a branch to the repository if useful.

senesis:
> One question: how can I access to a data sample so I can properly test it?
An example of multi-variable file for IPSL-CM6 is available [here (on my Google Drive)](https://drive.google.com/file/d/1_M7Hno-FFnkfU88jVMNCHvzyPTZWsbQA/view?usp=sharing)
Data is not enough : this PR anticipates on a merge of #1124, I tested the whole on a local branch which included #1124. I can push such a branch to the repository if you would like to run a test

esmvalcore/cmor/_fixes/native6/ipsl_cm6.py
senesis: Thanks. Fixed in second commit

senesis:
@alistairsellar : you may wish to know that this PR exists

zklaus:
Quick side note on the slow loading: I am just talking to some iris developers and it seems they are hitting the exact same problem, also while dealing with an XIOS generated multi-variable output file for the next model generation. There doesn't seem to be an issue about it yet, but I forwarded your example file and will try to add the issue here when it becomes available.

senesis:
> Quick side note on the slow loading: I am just talking to some iris developers and it seems they are hitting the exact same problem, also while dealing with an XIOS generated multi-variable output file for the next model generation. There doesn't seem to be an issue about it yet, but I forwarded your example file and will try to add the issue here when it becomes available.

I reported the problem quite early [as an Iris Issue](https://github.com/SciTools/iris/issues/4134#issue-891711169), and provided the file plus a notebook showing the problem by profiling an Iris run, and [provided a follow up in my original issue](https://github.com/ESMValGroup/ESMValTool/issues/2141#issuecomment-841086693) on this subject, which basically says that even with a recent fix in Iris V3.0.2, the load is still slow (figures availabe on demand). Do you mean that the Iris developers have some new idea here ?

pp-mo:
> @senesis even with a recent fix in Iris V3.0.2, the load is still slow

Following discussion with @abooton @zklaus , we've been looking into another possible Iris improvement : to avoid translating data-variables which are not wanted in a load.
A crude demonstration [shows a very considerable speedup for loading just one selected cube](https://github.com/SciTools/iris/pull/4176).

I rather struggled to understand exactly what you need to load here, but I'm assuming may be just a single data-variable,
based on [this comment](https://github.com/ESMValGroup/ESMValTool/issues/2141#issuecomment-845927522) and the CDO use proposed in [this PR](https://github.com/ESMValGroup/ESMValCore/pull/1153).
If it's actually more complicated than that, we can probably still adopt the general approach but it will need to be more sophisticated.

senesis:
Commit [#896764b](https://github.com/ESMValGroup/ESMValCore/pull/1153/commits/896764b877605a35931b8dc4ac1001f43714c38c) bring this PR in line with PR #1124 at [commit 50afc8b](https://github.com/ESMValGroup/ESMValCore/pull/1124/commits/50afc8ba3c1ccb699ea7d64e94a7af751af706da).  I may complement this PR at short notice if needed, including switching to the option of supporting IPSL-CM6 through a dedicated project. I will fix codacy tests diags after the merge of #1124 in main, because a number of them are due to the anticipation on fix.extra_facets existence

bouweandela:
> including switching to the option of supporting IPSL-CM6 through a dedicated project

That would be good, otherwise you will not be able to use the IPSL model data at the same time as any of the other data supported by the native6 project.

doc/quickstart/find_data.rst
```diff
@@ -15,10 +15,13 @@ the right data. We will detail below the data finding and retrieval process and
...
 
+Data types
+==========
+
 .. _CMOR-DRS:
 
-CMIP data - CMOR Data Reference Syntax (DRS) and the ESGF
-=========================================================
+CMIP data
+---------------------------------------------------------
```
bouweandela, line 24: The usual practice is to make the underline the same length as the title

esmvalcore/_config/extra_facets/native6-ipsl-cm6-mappings.yml
```diff
@@ -0,0 +1,256 @@
...
+# not compliant with ESMValtool licence policy, the default
+# configuration is to avoid using it. You may use customized settings
+# by installing a modified version of this file as
+# ~/.esmvatlool/variable_details/native6-ipsl-cm6-*.yml
+#
+# Keys scale and offset can be used for handling any unit issue that
+# is not dealt wih automatically by the udunits layer in ESMValTool
+# (and which makes use of the units metadaat in input files). In that
+# case, scaling and offseting are applied automatically by method
+# fix_data of class AllVars in cmor/_fixes/*/ipsl.py
```
bouweandela, line 24: I would recommend including the correct units of the variable in this file instead of using a scale factor and offset. That way the automatic unit conversion will be able to do its work. This is more reliable and will avoid problems if the unit in the CMOR table would change.

esmvalcore/cmor/_fixes/native6/ipsl_cm6.py
```diff
@@ -0,0 +1,114 @@
...
+        filtering, at least until Iris loads fast (which is not the case
+        up to, and including, V3.0.2)
+
+        However, we take care of ESMValTool policy re. dependencies licence
+
+        """
+        if "_" + self.extra_facets.get("group",
+                                       "non-sense") + ".nc" not in filepath:
+            # No need to filter the file
+            logger.debug("In ipsl-cm6.py : not filtering for %s", filepath)
```
bouweandela, line 36:
```suggestion
            logger.debug("Not filtering for %s", filepath)
```
there is no need to mention the filename in log messages, this is automatically recorded by the logging system in main_log_debug.txt

esmvalcore/cmor/_fixes/native6/ipsl_cm6.py
```diff
@@ -0,0 +1,114 @@
...
+            return filepath
+
+        if not self.extra_facets.get("use_cdo", False):
+            # The configuration developer doesn't provide CDO, while ESMValTool
```
bouweandela, line 40: CDO is a dependency of ESMValTool (some diagnostics use it, particularly those in R and NCL), but not of ESMValCore (and we have no intention of adding it, because we would like to keep the ESMValCore package light on dependencies). You're probably right about the licensing problems too.

esmvalcore/config-developer.yml
```diff
@@ -203,7 +208,11 @@ obs4mips:
   input_dir:
     default: 'Tier{tier}/{dataset}'
     RCAST: '/'
-  input_file: '{short_name}_{dataset}_{level}_{version}_*.nc'
+    IPSL: '{realm}/{short_name}/{freq}/{grid}/{institute}/${dataset}/{latest_version}'
```
bouweandela, line 211:
```suggestion
    IPSL: '{realm}/{short_name}/{freq}/{grid}/{institute}/{dataset}/{latest_version}'
```
is the `$` in the filename intended?

esmvalcore/config-developer.yml
```diff
@@ -203,7 +208,11 @@ obs4mips:
...
+  input_file:
+    default: '{short_name}_{dataset}_{level}_{version}_*.nc'
+    RCAST: '{short_name}_{dataset}_{level}_{version}_*.nc'
```
bouweandela, line 214: this looks the same as the `default`, does it need to be added?

esmvalcore/preprocessor/_io.py
```diff
@@ -117,8 +117,8 @@ def load(file, callback=None):
             category=UserWarning,
             module='iris',
         )
-
         raw_cubes = iris.load_raw(file, callback=callback)
+    logger.debug("Done with loading")
```
bouweandela, line 121:
```suggestion
    logger.debug("Done with loading %s", file)
```
Would be more clear, especially when the log messages get mixed up because you're running with `max_parallel_tasks` > 1. Note that the preprocessor steps are already emitting log messages though, so as soon as you get the log message from the step coming after `load` (I believe it's `fix_metadata`) coming from the same process, you know that loading is done.

doc/develop/fixing_data.rst
```diff
@@ -1,33 +1,40 @@
...
-from being processed correctly by the ESMValCore. The errors can be in
-the metadata describing the dataset and/or in the actual data.
-Typical examples of such errors are missing or wrong attributes (e.g.
-attribute ''units'' says 1e-9 but data are actually in 1e-6), missing or
-mislabeled coordinates (e.g. ''lev'' instead of ''plev'' or missing
+******************************************
+Adapting to data sources
+******************************************
+
+The baseline case for ESMValTool input data is CMOR fully compliant
```
bouweandela, line 7:
```suggestion
The baseline case for ESMValCore input data is CMOR fully compliant
```
please use the word ESMValCore everywhere in the ESMValCore documentation.

doc/develop/fixing_data.rst
```diff
@@ -1,33 +1,40 @@
...
+data that is read using Iris load fuction. ESMValTool also allows for
```
bouweandela, line 8:
```suggestion
data that is read using Iris load function. ESMValCore also allows for
```
and could you make the iris load function a link to the iris documentation for added clarity?

doc/develop/fixing_data.rst
```diff
@@ -326,30 +333,66 @@ strictness to the highest:
...
 
-For example for monthly data, place the files in the ``/Tier3/MSWEP/latestversion/mon/pr`` subdirectory of your ``native6`` project location.
+    .. code-block:: yaml
 
-.. note::
-  For monthly data (V220), the data must be postfixed with the date, i.e. rename ``global_monthly_050deg.nc`` to ``global_monthly_050deg_197901-201710.nc``
+      datasets:
+        - {simulation: CM61-LR-hist-03.1950, exp: piControl, freq: Analyse/TS_MO,
+           account: p86caub,  status: PROD, dataset: IPSL-CM6, project: native6,
+	   root: /thredds/tgcc/store}
```
bouweandela, line 357: Is `root` describing some input path? Is that still needed with the addition of the extra facets file?

doc/develop/fixing_data.rst
```diff
@@ -326,30 +333,66 @@ strictness to the highest:
...
+           account: p86caub,  status: PROD, dataset: IPSL-CM6, project: native6,
+	   root: /thredds/tgcc/store} 
+        - {simulation: CM61-LR-hist-03.1950, exp: historical, freq: Output/MO,
+           account: p86caub,  status: PROD, dataset: IPSL-CM6, project: native6,
+	   root: /thredds/tgcc/store} 
+
+    The ``Output`` format is an example of a case where variables are
+    grouped in multi-variable files, which name cannot be computed
+    directly from datasets attributes alone but requires an
+    ref:`Extra_facets file <extra_facets>`. These multi-variable
```
bouweandela, line 365: Link is broken

doc/develop/fixing_data.rst
```diff
@@ -326,30 +333,66 @@ strictness to the highest:
...
+    files must also undergo some data selection, which may involve an
+    external process for performance purpose.
+
+
+
+ERA5 and MSWEP datasets
```
bouweandela, line 371: Is the text on ERA5 and MSEP changed? If not, would it be possible to keep the original formatting so GitHub understands that it is the same?

doc/quickstart/configure.rst
```diff
@@ -176,8 +176,10 @@ It will be installed along with ESMValCore and can also be viewed on GitHub:
 `esmvalcore/config-developer.yml
 <https://github.com/ESMValGroup/ESMValCore/blob/main/esmvalcore/config-developer.yml>`_.
 This configuration file describes the file system structure and CMOR tables for several
-key projects (CMIP6, CMIP5, obs4mips, OBS6, OBS) on several key machines (e.g. BADC, CP4CDS, DKRZ,
-ETHZ, SMHI, BSC). CMIP data is stored as part of the Earth System Grid
+key projects (CMIP6, CMIP5, obs4mips, OBS6, OBS) on several key machines (e.g.
+BADC, CP4CDS, DKRZ, ETHZ, SMHI, BSC), and for native output data for some
+models (IPSL, ... see :ref:`configure_native_models`) .
```
bouweandela, line 181:
```suggestion
models (IPSL, ... see :ref:`configure_native_models`).
```

doc/quickstart/configure.rst
```diff
@@ -289,6 +292,48 @@ related to CMOR table settings available:
   to get the name of the file containing the ``mip`` table.
   Defaults to the value provided in ``cmor_type``.
 
+.. _configure_native_models:
+  
+Configuring native models and observation data sets
+----------------------------------------------------
+
+ESMValTool can take full advantage of the ability to configure
```
bouweandela, line 300:
```suggestion
ESMValCore can take full advantage of the ability to configure
```

doc/quickstart/configure.rst
```diff
@@ -289,6 +292,48 @@ related to CMOR table settings available:
...
+to host this new data source either under a dedicated project or under
+project ``native6``; when choosing the latter, such a configuration
+involves the following steps :
+
+  - allowing for ESMValTool to locate the data files :
+
+    - entry ``native6`` of ``config-developer.yml`` should be
+      complemented with sub-entries for ``input_dir`` and ``input_file``
+      that goes under a new key representing the
+      data organization (such as ``NEW_MODEL``), and these sub-entries can
```
bouweandela, line 312:
```suggestion
      data organization (such as ``INSTITUTE``), and these sub-entries can
```
these names are usually institute names, not model names, because they refer to the data storage organization used at a particular institute.

doc/quickstart/configure.rst
```diff
@@ -289,6 +292,48 @@ related to CMOR table settings available:
...
+            NEW_MODEL: '{simulation}_*.nc'
+          ...
+
+    - if necessary, provide a so-called ``extra facets file`` which
+      allows to cope e.g. with variable naming issues for finding
+      files. See `Extra_Facets`_  and here :download:`an example of
+      such a file for IPSL-CM6
+      <../../esmvalcore/_config/extra_facets/native6-ipsl-cm6-mappings.yml>`.
+
+  - ensuring that ESMValTool get the right metadata and data out of
```
bouweandela, line 333:
```suggestion
  - ensuring that ESMValCore gets the right metadata and data out of
```

doc/quickstart/configure.rst
```diff
@@ -289,6 +292,48 @@ related to CMOR table settings available:
...
+    your data files : this is described at :ref:`fixing_data`
```
bouweandela, line 334:
```suggestion
    your data files: this is described in :ref:`fixing_data`
```

doc/quickstart/find_data.rst
```diff
@@ -45,6 +48,39 @@ From the ESMValTool user perspective the number of data input parameters is
 optimized to allow for ease of use. We detail this procedure in the next
 section.
 
+Native model data
+---------------------------------------------------------
+Support for native model data is quite easy using basic
+:ref:`ESMValCore fix procedure <fixing_data>` and is yet implemented
```
bouweandela, line 54:
```suggestion
:ref:`ESMValCore fix procedure <fixing_data>` and has been implemented
```

doc/quickstart/find_data.rst
```diff
@@ -45,6 +48,39 @@ From the ESMValTool user perspective the number of data input parameters is
 optimized to allow for ease of use. We detail this procedure in the next
 section.
 
+Native model data
+---------------------------------------------------------
+Support for native model data is quite easy using basic
```
bouweandela, line 53:
```suggestion
Support for native model data that is not formatted according to a CMIP data request is quite easy using basic
```
would be a bit more specific. Some models just write their output in a way that already follow the CMIP data request, so no reformatting is needed at all.

doc/quickstart/find_data.rst
```diff
@@ -45,6 +48,39 @@ From the ESMValTool user perspective the number of data input parameters is
...
+
+  .. code-block:: yaml
+
+    OBS: /gws/nopw/j04/esmeval/obsdata-v2
+
+and the dataset:
+
+  .. code-block:: yaml
+
+    - {dataset: ERA-Interim,  project: OBS,  type: reanaly,  version: 1,  start_year: 2014,  end_year: 2015,  tier: 3}
```
bouweandela, line 70:
```suggestion
    - {dataset: ERA-Interim, project: OBS, type: reanaly, version: 1, start_year: 2014, end_year: 2015, tier: 3}
```
a single space between the items in a map is the preferred style in YAML files

doc/develop/fixing_data.rst
```diff
@@ -1,33 +1,40 @@
...
-***********
-
-Some (model) datasets contain (known) errors that would normally prevent them
-from being processed correctly by the ESMValCore. The errors can be in
-the metadata describing the dataset and/or in the actual data.
-Typical examples of such errors are missing or wrong attributes (e.g.
-attribute ''units'' says 1e-9 but data are actually in 1e-6), missing or
-mislabeled coordinates (e.g. ''lev'' instead of ''plev'' or missing
+******************************************
+Adapting to data sources
```
bouweandela, line 4: I agree that the title could be improved, but I would prefer to keep the words 'fix' and `dataset' in the title, because those are the words that everyone uses when talking about this topic.

bouweandela (changes requested):
Great that you're contributing this feature @senesis!

Could you please add unit tests for the code? They are a requirement. There is no need to add tests with sample data. You can just create some tiny iris cubes in code, as described [here](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/develop/preprocessor_function.html#unit-tests).

esmvalcore/_config/extra_facets/native6-ipsl-cm6-mappings.yml
bouweandela: As discussed on zoom, for flipping the sign I would recommend just providing the `postive: up` or `postive: down` attribute (if it's not available in the netcdf file already) and then writing some code that flips the sign and the attribute if it's the opposite from what is in the CMOR table.

senesis:
Hi, @ESMValGroup/tech-reviewers

I am trying to setup a unit test for the IPSL-CM6 fix module. When running the full test suite locally using pytest, I got that error :

> esmvalcore/cmor/_fixes/ipslcm/ipsl_cm6.py:6: in <module>
>     from ..fix import Fix
> E   ImportError: attempted relative import with no known parent package

which I do not understand, because e.g. `esmvalcore/cmor/_fixes/cmip6/gfdl_cm4.py` :
- as the same import statement, 
- is also tested using `tests/integration/cmor/_fixes/cmip6/test_gfdl_cm4.py` 
- and does not encounter such an error during testing.

What am I missing here ?

BTW : I would be glad to embark such a test for V2.3, which cutting date is today ...

zklaus:
@senesis, a bit of a shot in the dark since I believe the test code is not yet pushed into the branch: Did you remember to add a `__init__.py` file to the `esmvalcore/cmor/_fixes/ipslcm` directory to mark it as a [package](https://docs.python.org/3/tutorial/modules.html#packages)?

senesis:
> a bit of a shot in the dark since I believe the test code is not yet pushed into the branch: Did you remember to add a `__init__.py` file

Nice shot actually! I also came to the conclusion that this was the only difference with the working case. And this allowed to have succesful tests. Thanks

doc/quickstart/find_data.rst
senesis: OK; Done

esmvalcore/cmor/_fixes/native6/ipsl_cm6.py
senesis: OK. Done

esmvalcore/cmor/_fixes/native6/ipsl_cm6.py
senesis: I understood that no change is needed here.

esmvalcore/config-developer.yml
senesis: NO. Fixed

doc/quickstart/configure.rst
senesis: Fixed

doc/quickstart/find_data.rst
senesis: Changed

doc/quickstart/find_data.rst
senesis: Changed

doc/develop/fixing_data.rst
senesis: OK

doc/quickstart/find_data.rst
senesis: I fixed that, albeit I didn't wrote that ....

senesis:
> Could you please add unit tests for the code?

Done in last commit

esmvalcore/_config/extra_facets/native6-ipsl-cm6-mappings.yml
senesis: Done

esmvalcore/config-developer.yml
senesis: No, you're right. Done

esmvalcore/preprocessor/_io.py
senesis: Done

doc/develop/fixing_data.rst
senesis: Done.

doc/develop/fixing_data.rst
senesis: Done

doc/develop/fixing_data.rst
senesis: Yes, that is still needed because , for model development use case, modelers have some choice on where to put the data, which may change from one experiment to the other

doc/develop/fixing_data.rst
senesis: Because #1124 is not yet merged. I removed the link. To be polished for V2.4

doc/develop/fixing_data.rst
senesis: I tried

doc/quickstart/configure.rst
senesis: Done

doc/quickstart/configure.rst
senesis: Done

doc/quickstart/configure.rst
senesis: Done

senesis:
@zklaus 
I think I'm done now with this PR

doc/develop/fixing_data.rst
senesis: Because that text is both moved and differently indented , GitHub doesn't understand thta it is the same

doc/develop/fixing_data.rst
senesis: Yet done

senesis:
@zklaus @bouweandela 

Apart from a warning about using a subprocess, the remaining codacy diags are all more or less equal to :

> Instance of 'AllVars' has no 'extra_facets' member

which I do not understand, because AllVars inherits from Fix which now has an _init_ method which sets member extra_facets ....

Apart from that , I think everything is OK now

esmvalcore/_config/extra_facets/ipslcm-mappings.yml
```diff
@@ -0,0 +1,239 @@
...
+# Specifying 'igcm_dir' here allows to avoid having to secifiy it in
+# datasets definitions
+#
+# Key 'use_cdo' allows to choose whether CDO will be invoked for
+# selecting a variable in a multi-variable file. This generally allows
+# for smaller overal load time. But because CDO has a licence which is
+# not compliant with ESMValtool licence policy, the default
+# configuration is to avoid using it. You may use customized settings
+# by installing a modified version of this file as
+# ~/.esmvatlool/variable_details/native6-ipsl-cm6-*.yml
```
zklaus, line 18: Correct the path here (-> `~/.esmvaltool/extra_facets/ipslcm-*.yml`), perhaps add a link to the documentation, e.g. https://docs.esmvaltool.org/projects/esmvalcore/en/latest/quickstart/configure.html#location-of-the-extra-facets-files

esmvalcore/_config/extra_facets/ipslcm-mappings.yml
```diff
@@ -0,0 +1,239 @@
...
+#
+# Key 'positive' allows to tell ESMValTool when the sign convention
+# for a variable is reversed between IPSL-CM6 and CMIP6. In that case,
+# just e.g. put 'positive: down' if the CMIP6 convention is 'positive: up'
+#
+# The main key below, 'IPSL-CM6', is the one to use as a value for
+# attribute 'dataset' in the 'datasets' entry of recipes; it matches
+# the module name 'ipsl.py' in 'cmor_fixes/ipsl*/ipsl.py'
```
zklaus, line 26: This is a bit confusing. Cleary `IPSL-CM6` *doesn't* match `ipsl.py`, so what's the deal?

esmvalcore/_config/extra_facets/ipslcm-mappings.yml
```diff
@@ -0,0 +1,239 @@
...
+    hur: {ipsl_varname: hur, <<: *atm3dvars}
+
+    # ATM general variables
+    pr: {ipsl_varname: precip, <<: *atmvars}
+    psl: {ipsl_varname: slp, <<: *atmvars}
+    tas: {ipsl_varname: t2m, <<: *atmvars}
+    ts: {ipsl_varname: tsol, <<: *atmvars}
+    huss: {ipsl_varname: q2m, <<: *atmvars}
+    uas: {ipsl_varname: u10m, <<: *atmvars}
+    "vas": {ipsl_varname: v10m, <<: *atmvars}
```
zklaus, line 65: Why are the quotation marks around "vas" necessary? If it is about the pre-commit hooks, I suggest to remove them and make a commit using `git commit -n -m "A meaningful message"`

esmvalcore/_config/extra_facets/ipslcm-mappings.yml
```diff
@@ -0,0 +1,239 @@
...
+    wind10m: {ipsl_varname: wind10m, <<: *atmvars}
+
+    #  -> Turbulent fluxes
+    taux: {ipsl_varname: taux, <<: *atmvars}
+    tauy: {ipsl_varname: tauy, <<: *atmvars}
+
+    #  -> Radiative down at TOA
+    SWdnTOA: {ipsl_varname: SWdnTOA, <<: *atmvars}
+
+    #  -> Radiative down at TOA
```
zklaus, line 86: Same category? Probably the second should be `... up at TOA`?

esmvalcore/_config/extra_facets/ipslcm-mappings.yml
```diff
@@ -0,0 +1,239 @@
...
+    slp: {ipsl_varname: slp, <<: *atmvars}
+    t2m: {ipsl_varname: t2m, <<: *atmvars}
+    q2m: {ipsl_varname: q2m, <<: *atmvars}
+    u10m: {ipsl_varname: u10m, <<: *atmvars}
+    v10m: {ipsl_varname: v10m, <<: *atmvars}
+    wind10m: {ipsl_varname: wind10m, <<: *atmvars}
+
+    #  -> Turbulent fluxes
+    taux: {ipsl_varname: taux, <<: *atmvars}
+    tauy: {ipsl_varname: tauy, <<: *atmvars}
```
zklaus, line 81: What's the difference between these turbulent fluxes and the ones below (~ lines 104ff)?

esmvalcore/_config/extra_facets/ipslcm-mappings.yml
```diff
@@ -0,0 +1,239 @@
...
+    tauv: {ipsl_varname: tauy, <<: *atmvars}
+
+    #  -> Clouds
+    clt: {ipsl_varname: cldt, <<: *atmvars}
+
+    #  -> Radiative up at TOA
+    rlut: {ipsl_varname: topl, <<: *atmvars}
+    rsut: {ipsl_varname: SWupTOA, <<: *atmvars}
+    rlutcs: {ipsl_varname: topl0, <<: *atmvars}
+    rsutcs: {ipsl_varname: SWupTOAclr, <<: *atmvars}
```
zklaus, line 117: This seems to be a repetition of some lines above with new names. Is this section split into "IPSL style names" first and then "CMIP style names" after that? If so, I think we should consider to leave out the IPSL style names since they might lead to confusion, or otherwise add comments for the respective sub-sections.

esmvalcore/_config/extra_facets/ipslcm-mappings.yml
```diff
@@ -0,0 +1,239 @@
...
+    rsds: {ipsl_varname: SWdnSFC, <<: *atmvars}
+    rldscs: {ipsl_varname: LWdnSFcclr, <<: *atmvars}
+    rsdscs: {ipsl_varname: SWdnSFcclr, <<: *atmvars}
+
+  # ========================================================================================
+  Lmon:
+    # ========================================================================================
+    # SRF -- Land surface - ORCHIDEE
+
+    #  --> !!! This will not stay in the param file !!! --------------------
```
zklaus, line 139: What is meant here by param file? This file? Or a particular IPSL file? When will this be removed?

esmvalcore/cmor/_fixes/ipslcm/ipsl_cm6.py
```diff
@@ -0,0 +1,103 @@
...
+# The key used in extra_facets file for providing the
+# variable name (in NetCDF file) that match the CMOR variable name
+VARNAME_KEY = "ipsl_varname"
+
+
+class AllVars(Fix):
+    """Fixes for all IPSLCM variables."""
+
+    def fix_file(self, filepath, output_dir):
+        """Select IPSLCM variable in filepath, by calling CDO, if relevant.
```
zklaus, line 20:
```suggestion
        """Select IPSLCM variable in filepath.
```
The CDO implementation detail is not needed in the short summary. I would probably leave it out of the documentation completely (why does the user need to know?), but if you want to mention it a note at the end of the docstring seems better.

esmvalcore/config-developer.yml
```diff
@@ -205,7 +210,9 @@ obs4mips:
   input_dir:
     default: 'Tier{tier}/{dataset}'
     RCAST: '/'
-  input_file: '{short_name}_{dataset}_{level}_{version}_*.nc'
+    IPSL: '{realm}/{short_name}/{freq}/{grid}/{institute}/{dataset}/{latest_version}'
+  input_file:
+    default: '{short_name}_{dataset}_{level}_{version}_*.nc'
```
zklaus, line 215: This seems like a change that is rather unrelated to this PR. Was the old version broken? Or is it just for homogeneity? Either way, I would probably rather see this change in a separate PR.

esmvalcore/config-developer.yml
```diff
@@ -194,8 +195,12 @@ native6:
   cmor_strict: false
   input_dir:
     default: 'Tier{tier}/{dataset}/{latestversion}/{frequency}/{short_name}'
+    IPSL: '{root}/{account}/{model}/{status}/{exp}/{simulation}/{dir}/{freq}'
   input_file:
     default: '*.nc'
+    IPSL:
+      - '{simulation}_*_{ipsl_varname}.nc'
+      - '{simulation}_*_{group}.nc'
```
zklaus, line 203: Is this correct or a left-over from the previous attempt to treat the model data as native6?

zklaus (changes requested):
Very nice! I have reviewed the code and left a few comments. I will now turn my attention to the documentation. I am a bit concerned that there are a lot of changes to parts of the documentation that are not immediately tied to this problem and that might affect other people. For this reason I will likely suggest to separate out parts of the documentation into a separate PR to be handled after 2.3.0 with review from a wider group of people.

@senesis, please address the code related comments now, while I have a closer look at the documentation.

esmvalcore/_config/extra_facets/ipslcm-mappings.yml
```diff
@@ -0,0 +1,239 @@
+# Mapping, for ISPLCM output formats 'Analyse' and 'Output', between a
+# CMOR variable name and the labels to use by ESMValTool to find the
+# corresponding file, and the corresponding variable in the file
+#
+# For format 'Analyse', the config-development.yml file tells
+# ESMValTool to use key 'ipsl_varname' for building the filename,
+# while for format 'Output' it specifies to use key 'group'
+#
+# Specifying 'igcm_dir' here allows to avoid having to secifiy it in
```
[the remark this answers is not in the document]
senesis: Thanks

esmvalcore/_config/extra_facets/ipslcm-mappings.yml
senesis: Thanks

esmvalcore/_config/extra_facets/ipslcm-mappings.yml
senesis: The comment was fine in an earlier version, where I defined two projects.
Fixed. Thanks. The issue with providing hints at multiple place is that redudancy leads to inconsistency...

esmvalcore/_config/extra_facets/ipslcm-mappings.yml
senesis: It was a try for pre-commit issue. But it failed. Anyway, I do not use pre-commit anymore

esmvalcore/_config/extra_facets/ipslcm-mappings.yml
senesis: Thanks

esmvalcore/_config/extra_facets/ipslcm-mappings.yml
senesis: You're right. I organized a subsection commented as "ATM variables processed using their own IPSL short_name"

esmvalcore/_config/extra_facets/ipslcm-mappings.yml
senesis: Discarded. It arrived there by an automated process and I didn't re-read as carefully as you did

esmvalcore/cmor/_fixes/ipslcm/ipsl_cm6.py
senesis: The user may need to know that, in a future version, using CDO may become less useful.
I moved these few words out of the short summary

esmvalcore/config-developer.yml
senesis: You're right. It is not useful

esmvalcore/_config/extra_facets/ipslcm-mappings.yml
senesis: same issue as for rsds et al. Moved to a dedicated section

senesis:
> @senesis, please address the code related comments now, while I have a closer look at the documentation.

Done. Waiting for your advice about doc. But introducing the handling of native formats forced me to change a bit the structure of some paragraphs

zklaus:
For now, I just reduced the number of changes by reverting to the previous formatting in unrelated sections. As you can see from the diff, this is much easier to digest. Note that I don't claim that this formatting is better in any way. It just makes it easier to distinguish substantive changes, that we need to proofread, from the rest.

The main things for this were:
- Remove trailing spaces (perhaps something your editor may support you with)
- Turn tabs into spaces
- Restore the original line-breaks where possible

I will now do a proper review of the documentation changes.

doc/develop/fixing_data.rst
```diff
@@ -1,32 +1,39 @@
...
 ***********
-Dataset fix
+Fixing data
 ***********
 
-Some (model) datasets contain (known) errors that would normally prevent them
-from being processed correctly by the ESMValCore. The errors can be in
+The baseline case for ESMValCore input data is CMOR fully compliant
+data that is read using
+`Iris load_raw function <https://scitools-iris.readthedocs.io/en/latest/generated/api/iris.html#iris.load_raw>`_.
```
zklaus, line 9: Thanks to intersphinx, we can write this much more compact as
```suggestion
data that is read using Iris' :func:`iris:iris.load_raw`.
```
For more information see [intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html) and [roles](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects).

doc/develop/fixing_data.rst
```diff
@@ -1,32 +1,39 @@
...
+ESMValCore also allows for some departures with compliance (see
```
zklaus, line 10: departure *from* compliance

doc/develop/fixing_data.rst
```diff
@@ -1,32 +1,39 @@
...
 the metadata describing the dataset and/or in the actual data.
 Typical examples of such errors are missing or wrong attributes (e.g.
 attribute ''units'' says 1e-9 but data are actually in 1e-6), missing or
 mislabeled coordinates (e.g. ''lev'' instead of ''plev'' or missing
 coordinate bounds like ''lat_bnds'') or problems with the actual data
 (e.g. cloud liquid water only instead of sum of liquid + ice as specified by the CMIP data request).
 
-The ESMValCore can apply on the fly fixes to datasets that have
-known errors that can be fixed automatically.
+As an extreme case, some others data sources simply are not NetCDF
```
zklaus, line 21:
```suggestion
 As an extreme case, some data sources simply are not NetCDF
```

doc/develop/fixing_data.rst
```diff
@@ -1,32 +1,39 @@
...
+files and must go through other data load function.
```
zklaus, line 22:
```suggestion
files and must go through some other data load function.
```

doc/develop/fixing_data.rst
```diff
@@ -1,32 +1,39 @@
...
+
+The ESMValCore can apply on the fly fixes to such datasets when
+issues can be fixed automatically. This is implemented for a set
+of `Natively supported non-CMIP datasets`_. The following provide
```
zklaus, line 26:
```suggestion
of `Natively supported non-CMIP datasets`_. The following provides
```

doc/develop/fixing_data.rst
```diff
@@ -326,7 +333,11 @@ strictness to the highest:
 Natively supported non-CMIP datasets
 ====================================
 
-Fixed datasets are supported through the ``native6`` project.
+Some fixed datasets and native models formats are supported through
```
zklaus, line 336:
```suggestion
Some fixed datasets and native model formats are supported through
```

doc/develop/fixing_data.rst
```diff
@@ -354,6 +365,40 @@ For example for monthly data, place the files in the ``/Tier3/MSWEP/latestversio
...
+.. _fixing_native_models:
+
+Native models : IPSL-CM6,...
+-----------------------------
+
+The following models are natively supported through the procedure
+described above (:ref:`fix_structure`) and at
+:ref:`configure_native_models`:
+
+  - **IPSL-CM6** : both output formats (i.e. the ``Output`` and the
```
zklaus, line 377: I suggest to move to a sub-section style approach also for the models. In addition to the following suggestion, this requires a change of indentation of the whole IPSL-CM6 part.

```suggestion
IPSL-CM6
~~~~~~~~
Both output formats (i.e. the ``Output`` and the
```

doc/develop/fixing_data.rst
```diff
@@ -354,6 +365,40 @@ For example for monthly data, place the files in the ``/Tier3/MSWEP/latestversio
...
+
+      datasets:
+        - {simulation: CM61-LR-hist-03.1950, exp: piControl, freq: Analyse/TS_MO,
+           account: p86caub,  status: PROD, dataset: IPSL-CM6, project: IPSLCM,
+           root: /thredds/tgcc/store}
+        - {simulation: CM61-LR-hist-03.1950, exp: historical, freq: Output/MO,
+           account: p86caub,  status: PROD, dataset: IPSL-CM6, project: IPSLCM,
+           root: /thredds/tgcc/store}
+
+    .. _ipslcm_extra_facets_example:
```
zklaus, line 391: Is this intended as a label for the yaml block? In that case I think it has to go before the block, i.e.
```
.. _ipslcm_extra_facets_example:

.. code-block:: yaml
```

doc/develop/fixing_data.rst
```diff
@@ -354,6 +365,40 @@ For example for monthly data, place the files in the ``/Tier3/MSWEP/latestversio
...
+
+    The ``Output`` format is an example of a case where variables are
+    grouped in multi-variable files, which name cannot be computed
+    directly from datasets attributes alone but requires to use an
+    extra_facets file, which principles are explained in
+    :ref:`extra_facets`, and which content is :download:`available here
+    </../esmvalcore/_config/extra_facets/ipslcm-mappings.yml>`. These
+    multi-variable files must also undergo some data selection, which
+    may involve an external process for performance purpose.
```
zklaus, line 400: I don't think we need to mention CDO here.
```suggestion
    multi-variable files must also undergo some data selection.
```

doc/quickstart/find_data.rst
```diff
@@ -45,6 +48,40 @@ From the ESMValTool user perspective the number of data input parameters is
 optimized to allow for ease of use. We detail this procedure in the next
 section.
 
+Native model data
+---------------------------------------------------------
```
zklaus, line 52: Shorten underlining please
```suggestion
-----------------
```

doc/quickstart/find_data.rst
```diff
@@ -45,6 +48,40 @@ From the ESMValTool user perspective the number of data input parameters is
...
+Support for native model data that is not formatted according to a CMIP
+data request is quite easy using basic
+:ref:`ESMValCore fix procedure <fixing_data>` and has been implemented
+for some models :ref:`as described here <fixing_native_models>`
+
+Observational data
+---------------------------------------------------------
```
zklaus, line 59:
```suggestion
------------------
```

zklaus (changes requested):
Ok, finished the doc review. Only minor comments.

doc/develop/fixing_data.rst
senesis: No, it is a ref for the text below

doc/develop/fixing_data.rst
zklaus: Do you have reservations about this change?

zklaus:
I made a small improvement to the formatting.

I also have one more question: The atmospheric variables are now separated according to CMIP vs IPSL names. Does this not apply to the other realms? Also, there are a number of very long `#============` lines in the facets file. Do we need them? If you want to keep them, could we shorten them, maybe to 79 chars total line length or similar?

senesis (commented):
Fine

zklaus (approved):

pp-mo:
Hi @senesis @zklaus 
I've been listening in here, but it's not really my area to comment.
Except that, I just wanted to remind you that maybe the ideas at our https://github.com/SciTools/iris/pull/4176 might make the use of CDO unnecessary here ?

I was prompted to investiagte that approach by @abooton, who I think discussed it with @zklaus .
I think we will be adopting _something_ like this for our own purposes, but it's not quite clear to me which type of input (constraints) it must support, to be of use in your workflows.
So we'd really appreciate some feedback on how it could fit in with your needs here. 😄

senesis:
Hi @pp-mo, @zklaus , @abooton 

Sorry for this late reply.

I am not sure I fully understand everything in SciTools/iris#4176. 

Assuming that the question still is "would a simple (and fast) load of the single targeted variable fit the need ?" , I am afraid the answer is no, because the variables which are auxiliary coordinates for the targeted variables should be loaded too, as they might be useful at a later stage (and by the way, I agree that using CDO for selecting the variable is then not correct either). 

Further, @valeriupredoi wrote in [#ESMValTool/2141](https://github.com/ESMValGroup/ESMValTool/issues/2141#issuecomment-828562507)  that the use of iris.load_raw is a strong design principle for ESMValTool, and, if I understand well, this seems to complicate finding a working solution.

pp-mo:
> Assuming that the question still is "would a simple (and fast) load of the single targeted variable fit the need ?" , I am afraid the answer is no, because the variables which are auxiliary coordinates for the targeted variables should be loaded too, as they might be useful at a later stage (and by the way, I agree that using CDO for selecting the variable is then not correct either).

That is absolutely **_not_** what this is doing :  The approach in SciTools/iris#4176 does still analyse the whole file, but it limits which actual cube(s) are generated -- which basically is the slow bit.  
The design goal here is that **any particular constrained load will produce _exactly_ the same result as before** (only faster).

So I do think that this approach (if done right) should be **a safe bet, and a great fit for your needs as described**.

To be clear on the method, each loaded Iris cube relates to a specific CF _data_-variable, and it is those which we are selecting on -- so cubes will still have all the usual information from other (types of) file variables, such as coordinates, ancillary variables etc etc.
The equivalence also means that we don't need a control to switch this on+off

The remaining questions for me are ...
   1. whether we can interpret + speedup the _types_ of constraint expression which you need to use, and 
   2. whether the speedup works adequately for your actual usecases, and
   3. can we get this into an Iris release by the time you need

senesis:
These are good news ! From my point of view, :  
- the single constraint type would be on the NetCDF variable name
- the speed up figures you quoted [there](https://github.com/SciTools/iris/pull/4176#issue-661074769) are fine, and impressive
- the feasibility of using iris.load with a constraint instead of using iris.load_raw in ESMValTool is a question that I cannot answer, and that should be directed to @valeriupredoi  ; I understood that this should anyway take some development time, and have no idea of the relative priority of this use case in the general development plan; I ping @bouweandela and @zklaus in case they can help on this subject

bouweandela:
> the feasibility of using iris.load with a constraint instead of using iris.load_raw in ESMValTool is a question that I cannot answer, and that should be directed to @valeriupredoi ; I understood that this should anyway take some development time, and have no idea of the relative priority of this use case in the general development plan; I ping @bouweandela and @zklaus in case they can help on this subject

If this is a feature you need, could you please open an issue for it? A discussion in a merged pull request is easily lost over time. Regarding the relative priority of new features: from experience, I can tell that those features that have someone who is actively working on them, tend to get the highest priority.

pp-mo:
>  using iris.load with a constraint instead of using iris.load_raw in ESMValTool

If it helps, I can affirm that 'iris.load_raw' is almost never any different from 'iris.load', for netcdf source data.
The essential difference between the two is : 'load' attempts a CubeList.merge on the 'raw' cubes, and 'load_raw' does not.
It is actually quite unusual for there to be mergeable data in a netcdf load : it requires that there are "raw" cubes which differ **_only_** in the values of scalar coordinates (which here means coords with _no_ dimension : a coord mapped to a length-1 dimension is not a candidate).  Such cubes also can't come from the same file, because they would have to have the same var_name (each cube comes from a single data-var).

zklaus:
I have moved the discussion on Iris loading times and CDO to the new issue #1180. Let's continue there.

senesis:
By the way : thanks a lot for the support on this PR and the related issue ; nice to have that yet in V3.2

PR #1177, opened 2021-06-14
Automatically fix small deviations in vertical levels

bouweandela:
## Description

This pull request aims to make working with vertical levels that have tiny errors in the coordinate values easier. It does this by:

1) trying to automatically fix the problem
2) setting the levels equal to those requested in `extract_levels` when cube slicing is used instead of interpolation

This will make it easier to compute multi-model statististics, which now need exactly matching coordinates.

See https://github.com/ESMValGroup/ESMValCore/issues/956#issuecomment-768419704

***

***


valeriupredoi:
this is cool! maybe we should make the tolerances a user tunable parameter? I can see uses for in a configuration file

bouweandela:
I'm not sure. Do you think the values that I've chosen now make sense?

zklaus:
I think in this case it's better if we control the tolerances. Note also that it might be tempting to do a similar thing for horizontal coordinates, where e.g. in CORDEX very small deviations can occur. The difference is that there is no "right" value in CORDEX, this is decided by the model, whereas here we have an authoritative value from the CMIP tables. This is also what allows us to decide on tolerances, imho.

bouweandela:
> Note also that it might be tempting to do a similar thing for horizontal coordinates, where e.g. in CORDEX very small deviations can occur.

Indeed, this was implemented in #507 with the default tolerances. The default relative tolerance might be a bit lax (rtol 1e-5, atol 1e-8).

bouweandela:
> I think in this case it's better if we control the tolerances.

I could add that for `extract_levels`. What would you propose as the default? The current values?

zklaus:
I think the current values look fine to me, but let's have another look for the next release.

bouweandela:
The remaining Codacy issue is about too many arguments for the `extract_levels` preprocessor function, so unfortunately it cannot be solved.

bouweandela:
@zklaus I added the option to tweak the parameters as requested. Could you have another look, please?

bouweandela:
Thank you for reviewing @valeriupredoi! @zklaus Do you have any further comments on this?

zklaus (approved):

PR #1178, opened 2021-06-14
Set circular attribute in MCM-UA-1-0 fix

### LINKED ISSUE #1175: Masked values when regridding datasets with coord `circular=False`
**Describe the bug**
At the BSC, it was reported that when regridding dataset MCM-UA-1-0, a white line with masked values appeared for lon > 355. 
After some debugging it turns out that when the longitudes are not circular and do not span fully from [0,360], those masked values can appear after regridding. 

This dataset got fixed in #1098, because (among other issues) the last bound in atmospheric coordinates was not set properly. That issue makes iris set the `circular` attribute to `False`, and makes the dataset come out of the regridding with masked values. Once the bounds are fixed, a quick solution would be to set the circular attribute to `True` in the same fix, since the condition that iris imposes to set this attribute checks for the fixed bound.

But I was wondering  if it would be ok to check for this attribute, either during the cmor checks or during the regridding,  to make a more general fix.

sloosvel:
## Description
This PR checks and changes the circular attribute for dataset MCM-UA-1-0 after fixing the longitudes. This is done to avoid getting masked values after regridding.

Closes #1175 

Link to documentation:

***

~- [🧪][2] and [🛠][1] [Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available~
~-  [🛠][1] Any changed [dependencies have been added or removed](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#dependencies) correctly~

***

### DIFF
--- a/tests/integration/cmor/_fixes/cmip6/test_mcm_ua_1_0.py
+++ b/tests/integration/cmor/_fixes/cmip6/test_mcm_ua_1_0.py
@@ -177,7 +177,7 @@ def test_allvars_fix_lon_bounds(cubes_bounds):
             pass
         else:
             assert lon_coord.bounds[-1][-1] == 358.125
-            assert lon_coord.circular is True
+            assert lon_coord.circular
 
 
 def test_tas_fix_metadata(cubes):
--- a/tests/integration/cmor/_fixes/cmip6/test_mcm_ua_1_0.py
+++ b/tests/integration/cmor/_fixes/cmip6/test_mcm_ua_1_0.py
@@ -96,7 +96,8 @@ def cubes_bounds():
                                              bounds=[[-1.875, 1.875],
                                                      [354.375, 358.125]],
                                              var_name='lon',
-                                             standard_name='longitude')
+                                             standard_name='longitude',
+                                             circular=True)
     wrong_lon_coord = iris.coords.DimCoord([0, 356.25],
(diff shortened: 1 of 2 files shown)


zklaus:
@sloosvel, do you consider this already ready for review? If so, could you please go through the checklist and make sure that all items are either ticked or removed/stricken out? Also, as you probably know, `_is_circular` appears to be an internal method of iris. If we really want to use it, it might be a good idea to raise the issue with the iris guys, perhaps even asking whether they think this is the best approach.

I might in this case consider simply asserting the correct attribute without further analysis on the bounds.

sloosvel:
I think it's done now

tests/integration/cmor/_fixes/cmip6/test_mcm_ua_1_0.py
```diff
@@ -176,6 +177,7 @@ def test_allvars_fix_lon_bounds(cubes_bounds):
             pass
         else:
             assert lon_coord.bounds[-1][-1] == 358.125
+            assert lon_coord.circular is True
```
zklaus, line 180: Splitting hairs, it might be simpler to say
```suggestion
            assert lon_coord.circular
```

zklaus (approved):

PR #1207, opened 2021-07-01
Suppress third party package warnings (User and DeprecationWarning) while running pytest

### LINKED ISSUE #857: Suppress third party warnings in pytest
Currently running pytest generates a large number of  warnings from third-party modules, i.e.
```
tests/functional/test_multimodel.py: 300 warnings
  /home/stef/miniconda3/envs/esmvaltool/lib/python3.8/site-packages/iris/fileformats/cf.py:221: DeprecationWarning: tostring() is deprecated. Use tobytes() instead.
    return tuple((attr, self.getncattr(attr)) for attr in
```

These are unhelpful, and frankly not our responsibility, but they distract from testing and obscure the warnings that we can fix.
I would like to add some filters to remove those warnings as described here:
https://stackoverflow.com/questions/58399870/pytest-suppress-deprecationwarning-from-specific-3rd-party-modules 

👍👍

valeriupredoi:
## Description

Makes the test output cleaner; we can add more if we wish (but make sure you don't suppress the ESMValCore's own DeprecationWarning's). Good idea in #857 @stefsmeets - please feel free to add more in the `ignore` list :+1: 

Closes #857 

Link to documentation:

***

***

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,6 +24,10 @@ markers =
     installation: Test requires installation of dependencies
     use_sample_data: Run functional tests using real data
     sequential: Run these tests sequentially, see https://github.com/ESMValGroup/ESMValCore/issues/644
+filterwarnings =
+    ignore::UserWarning:iris.*
+    ignore::DeprecationWarning:iris.*
+    ignore::DeprecationWarning:pyke.*
 
 [coverage:run]
 parallel = true


stefsmeets (approved):
Nice to finally have this! Pytest warning spam drives me crazy (not just in esmvaltool!) 🚀

setup.cfg
```diff
@@ -24,6 +24,10 @@ markers =
     installation: Test requires installation of dependencies
     use_sample_data: Run functional tests using real data
     sequential: Run these tests sequentially, see https://github.com/ESMValGroup/ESMValCore/issues/644
+filterwarnings =
+    ignore::UserWarning:iris.*
```
bouweandela, line 28: Why are you disabling these warnings?

bouweandela (changes requested):
It looks like disabling the UserWarnings will mask problems with our own code, why do you think those should be disabled.

Another question: how does this distuingish DeprecationWarnings that are issued because we are using something old from the iris library from warnings that are issued because the iris library itself uses e.g. an old numpy function?

valeriupredoi:
I am disabling `UserWarnings` only from third party packages ie `iris` - do you think we need those? All that's coming from our package and `esmvalcore` are still being displayed. Stuff that's causing a warning from within iris but comes from eg numpy usually get shown as a numpy warning, I think. By all means, hack at will @bouweandela if you think we need some warnings that are masked in here :+1:

setup.cfg
valeriupredoi: we got nothing to do with them and they're bloating the stdout?

jvegasbsc: That's maybe true for the `UserWarning:iris.*`, but the `DeprecationWarning:iris.*` probably should be removed by fixing the code that produces them.

bouweandela: iris UserWarning are there to warn you that you're using iris in a dodgy way, e.g. collapsing a coordinate without bounds. The correct way to fix these warnings would be to
1) have a critical look at our tests and see if the test data is similar enough to real data (e.g. does it have bounds?)
2) silence the warnings if they do not make sense in our code (e.g. #1215), because just silencing them in the tests means that users of our software will still get them, but we're not aware of that because we've silenced them in the tests

bouweandela:
The problem I have with this approach is that I would like to be sure that we see:
- a deprecation warning coming from iris if we use a deprecated iris feature
- a numpy deprecation warning if we use a deprecated numpy feature in our own code, but not if iris uses a deprecated numpy feature

and I'm not convinced that is what happens if you configure it like this. Could you explain a bit better why you think this works?

setup.cfg
valeriupredoi: I really don't think stuff like:
```
tests/integration/cmor/_fixes/cmip6/test_cesm2_waccm.py::test_cl_fix_file
  /usr/share/miniconda3/envs/esmvaltool/lib/python3.9/site-packages/iris/fileformats/cf.py:584: UserWarning: Missing CF-netCDF formula term variable 'ps', referenced by netCDF variable 'lev_bnds'
    warnings.warn(
```
see original out in [GA test](https://github.com/ESMValGroup/ESMValCore/runs/3785574476?check_suite_focus=true)
are in any way interesting and/or warrant fixes from us. There's a lot of these types, OK here's one that might be useful:
```
tests/unit/preprocessor/_multimodel/test_multimodel.py::test_ignore_tas_scalar_height_coord
  /usr/share/miniconda3/envs/esmvaltool/lib/python3.9/site-packages/iris/coords.py:1982: UserWarning: Collapsing a non-contiguous coordinate. Metadata may not be fully descriptive for 'height'.
    warnings.warn(msg.format(self.name()))
```
but here too, we've been seeing this sort of warning for years and yet we've not done anything about it. Let's face it guys, I don't think we've actually fixed anything related to a warning until it became a full on raised exception :grin:

zklaus:
I agree with @bouweandela. There may be some room for an approach more targeted to specific warnings, but I'd rather see those warnings that we have now addressed by solving the underlying condition.

I am inclined to close this, but I'll leave it to you, @valeriupredoi.

valeriupredoi:
I don't have a beef with the warnings, so am ok to close this, but @stefsmeets was quite happy to see these go, so it might be that some other developers might feel the same. For me it's all the same, I actually read this guff from time to time anyway, I think we found one or two problems with iris from reading them. Up to you guys :man_shrugging:

bouweandela:
I would also like a feature this, but it should be implemented in such a way that important warnings do not get lost. I'm not convinced that this pull request in it's current state does that.

zklaus:
Since this is "only" cleaning up some output, we have made strides towards that in this release already, and we seem to have a bit more discussion necessary first, I am going to bump this to the next release.

valeriupredoi:
may be closed too, Klaus, am not in love with it and people seem to like to read warnings (myself included) :grin:

zklaus:
I think the approach needs some work, but the goal is admirable. If we can address some of the underlying causes we will improve the robustness of our code, so let's keep it around and have a closer look with a little bit more time.

valeriupredoi:
sounds good, bud :beer:

schlunma:
Can we close this? In the discussion people seem to agree that this can be closed.

valeriupredoi:
yeh, it's a flightless bird :bird:

PR #1212, opened 2021-07-05
Fix alignment of daily data with inconsistent calendars in multimodel statistics

### LINKED ISSUE #1198: Multi-model mean fails for Core V2.3.0 because of difference in aux_coords
**Describe the bug**
A multi-model mean, which was running for the last core version now fails because:
> Encountered the following exception: failed to merge into a single cube.
> Coordinates in cube.aux_coords (scalar) differ by dtype or class (i.e. DimCoord vs AuxCoord).

Different to the case reported here: https://github.com/ESMValGroup/ESMValTool/issues/2198 
the recipe does not use "extract_level" but "extract_region" and "extract_season", could this be connected?

**Please attach**
  - The recipe is attached here, it belongs to the PR https://github.com/ESMValGroup/ESMValTool/pull/2153 There the multi-model mean is currently turned off.
[recipe_pv_capacity_factor.yml_coord_issue_multi_model_mean.txt](https://github.com/ESMValGroup/ESMValCore/files/6726402/recipe_pv_capacity_factor.yml_coord_issue_multi_model_mean.txt)

  - The `main_log_debug.txt` file:
[run_main_log_debug_coord_issue_multi_model_mean.txt](https://github.com/ESMValGroup/ESMValCore/files/6726397/run_main_log_debug_coord_issue_multi_model_mean.txt)

### LINKED ISSUE #1210: How should multi-model statistics handle daily data on different calendars?
More comprehensive metadata handling in the multi-model statistics recently turned up a conceptual issue for multi-model statistics on daily data with different calendars.

The issue is that in this case there are days, for which only a subset of models may provide data. This happens usually for datasets that contain leap days (gregorian or standard calendar, all_leap) vs those that don't (noleap), or with more unusual calendars (360_day, 30 days in every month); see [CF conventions 1.7, Sect. 4.4.1](http://cfconventions.org/cf-conventions/cf-conventions.html#calendar) for the full list of supported calendars).

The old version of the multi-model statistics has two modes (selected with the parameter `span`). In `overlap` mode, it discards all days that don't appear in all datasets, producing a result that leaves out those days. In `full` mode, 
However, [the documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/stable/recipe/preprocessor.html#multi-model-statistics) states that
> As the number of days in a year may vary between calendars, (sub-)daily data with different calendars are not supported.

The new version of the multi-model statistics follows the documentation by throwing an exception, albeit a cryptic one.

Alternative strategies could be
- Follow the existing `overlap`/`full` strategies
- Refuse daily data
- Introduce a calendar aligning pre-processor. It is fairly common to deal with this problem by repeating certain days to fill up the missing information, or leaving out superfluous days, for example from the all_leap calendar.

Some aspects of this have been discussed in #1201 and #1198.
This issue has previously been mentioned in #937, #744.
It is also connected to #781.

Peter9192:
## Description

Following #1210 this PR implements the following behavior for multimodel statistics, when the input data is a daily frequency with inconsistent calendars:

- `span='overlap'`: discard those days that occur only in certain calendars but not in others
- `span='full'`: use nearest-neighbour lookup to fill missing days. Missing dates outside the original date range are masked.

This is in contrast to the previous behavior, which was:
- `span='overlap'`: Find the first and last day that is present in all datasets; remove everything before the first and after the last day.
- `span='full'`: Find the very first and the very last day across all cubes; pad every cube with masked out data before and after its actual data to cover the full time range.

This only worked for data with the same number of days per year, raising an exception otherwise.

Closes #1198

Link to documentation:

***

- [ ] ~[🧪][2] and [🛠][1] [Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available~

***

### DIFF
--- a/esmvalcore/preprocessor/_multimodel.py
+++ b/esmvalcore/preprocessor/_multimodel.py
@@ -142,63 +142,18 @@ def _time_coords_are_aligned(cubes):
     return True
 
 
-def _subset(cube, time_points):
-    """Subset cube to given time range."""
-    begin = cube.coord('time').units.num2date(time_points[0])
-    end = cube.coord('time').units.num2date(time_points[-1])
-    constraint = iris.Constraint(time=lambda cell: begin <= cell.point <= end)
-    try:
-        return cube.extract(constraint)
-    except Exception as excinfo:
-        raise ValueError(
-            "Tried to align cubes in multi-model statistics, but failed for"
-            f" cube {cube} and time points {time_points}. Encountered the "
-            f"following exception: {excinfo}")
+def _map_to_new_time(cube, time_points):
+    """Map cube onto new cube with specified time points.
 
-
-def _extend(cube, time_points):
-    """Extend cube to a specified time range.
-
-    If time points are missing before the start/after the end of the
-    time range, cubes for each missing time pointwith masked data will
-    be added to pad the time range to match `time_points`. This method
-    supports lazy operation.
+    Missing data inside original bounds is filled with nearest neighbour
+    Missing data outside original bounds is masked.
     """
-    cube.coord('time').bounds = None
-    cube_points = cube.coord('time').points
-
-    begin = cube_points[0]
-    end = cube_points[-1]
-
-    pad_begin = time_points[time_points < begin]
-    pad_end = time_points[time_points > end]
-
-    if (len(pad_begin)) == 0 and (len(pad_end) == 0):
-        return cube
-
-    template_cube = cube[:1].copy()
-    template_cube.data = np.ma.array(da.zeros_like(template_cube.data),
-                                     mask=True,
-                                     dtype=template_cube.data.dtype)
-
-    cube_list = []
-
-    for time_point in pad_begin:
-        new_slice = template_cube.copy()
-        new_slice.coord('time').points = float(time_point)
-        cube_list.append(new_slice)
-
-    cube_list.append(cube)
-
-    for time_point in pad_end:
-        new_slice = template_cube.copy()
-        new_slice.coord('time').points = float(time_point)
-        cube_list.append(new_slice)
-
-    cube_list = iris.cube.CubeList(cube_list)
+    time_points = cube.coord('time').units.num2date(time_points)
+    sample_points = [('time', time_points)]
+    scheme = iris.analysis.Nearest(extrapolation_mode='mask')
 
     try:
-        new_cube = cube_list.concatenate_cube()
+        new_cube = cube.interpolate(sample_points, scheme)
     except Exception as excinfo:
         raise ValueError(
             "Tried to align cubes in multi-model statistics, but failed for"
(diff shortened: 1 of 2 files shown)


zklaus:
Thanks @Peter9192, that's great. To aid in finishing this up, could you please "do" the checklist, i.e. tick done items and remove or strike out non-applicable items? Also, could you try to put a plainer description, i.e. distinguish between filling missing and removing superfluous days and make clear if and how this relates to the `overlap` and `full` options?

katjaweigel:
@Peter9192 thanks a lot for the fix! For the version of recipe_pv_capacity_factor.yml including the multi-model mean using different calendars runs with it. But I wonder if it should without Warning? Or would if it would be even better to only allow this if an additional preprocessor like regrid_time is used?

zklaus:
I think let's roll with it for now (i.e. the release) and revisit the issue of calendars afterward. I did a review of the code here and think it is an improvement, but it does not support 360 day calendars, for instance, just like the old multi-model code did not support it.

There are a few more quirks I'd like to talk about and probably go to handling on a day-of-year basis, but for now, let's get this in as a stable platform to base future improvements on.

Peter9192:
> Or would if it would be even better to only allow this if an additional preprocessor like regrid_time is used?

I agree that would be good, also see #744.

zklaus:
@katjaweigel, if you are nevertheless happy to move forward with this for now, could you please "Approve" the PR?

katjaweigel (changes requested):
I like the method to regrid the calendar, but I would at least keep a warning and not replace the warning without note with def _map_to_new_time.
Maybe it would be better to only allow it, if regrid_time is used as preprocessor (in agreement with different horizontal grids, here the multi-model mean can also only be applied if everything is interpolated on a common grid.)?

katjaweigel:
> @katjaweigel, if you are nevertheless happy to move forward with this for now, could you please "Approve" the PR?

@zklaus I rather agree that this one could wait, because since the multi-model mean is removed from recipe_pv_capacity_factor.yml I don't think there is any recipe affected by this issue.

zklaus:
> I like the method to regrid the calendar, but I would at least keep a warning and not replace the warning without note with def _map_to_new_time.

The warning is not removed here; it remains in [lines 116ff of `_multimodel.py`](https://github.com/ESMValGroup/ESMValCore/blob/f068c0f61ace481eba23bbb40b2a49e6a79d2ec8/esmvalcore/preprocessor/_multimodel.py#L116).

> Maybe it would be better to only allow it, if regrid_time is used as preprocessor (in agreement with different horizontal grids, here the multi-model mean can also only be applied if everything is interpolated on a common grid.)?

Yes, but let's leave that to #744.

katjaweigel:
> The warning is not removed here

@zklaus you are right, the warning is there. Sorry I checked the wrong log file!

katjaweigel (approved):

zklaus:
I edited this so that #1210 stays open for a better long-term solution.

PR #1169, opened 2021-06-10
Fix test failing due to fx files chosen differently on different OS's

valeriupredoi:
## Description

The strategy is to fix #1159 properly (one way or another), but to allow for us not having failed tests during the current release procedures, I am skipping the two tests that fail conditioning on version < 2.4.0 (if we don't fix it by then, we'll have those two tests fail again). I copied over the functionality of those two tests in a couple `_incomplete` tests that run everything else that doesn't fail and we should test for (we should remove those and re-instate the actual complete, currently skipped now).

Related to **BUT IT DON'T CLOSE IT** #1159 

Link to documentation:

***

## [Before you get started](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#getting-started)

- [ ] [☝ Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do

## [Checklist](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#checklist-for-pull-requests)

It is the responsibility of the author to make sure the pull request is ready to review. The icons indicate whether the item will be subject to the [🛠 Technical][1] or [🧪 Scientific][2] review.

[1]: https://docs.esmvaltool.org/en/latest/community/review.html#technical-review
[2]: https://docs.esmvaltool.org/en/latest/community/review.html#scientific-review

- [x] [🧪][2] The new functionality is [relevant and scientifically sound](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#scientific-relevance)
- [x] [🛠][1] This pull request has a [descriptive title and labels](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-title-and-label)
- [x] [🛠][1] Code is written according to the [code quality guidelines](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#code-quality)
- [ ] [🧪][2] and [🛠][1] [Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available
- [x] [🛠][1] [Unit tests](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#tests) have been added
- [x] [🛠][1] Changes are [backward compatible](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#backward-compatibility)
[...]

PR #1173, opened 2021-06-11
Compare datetimes instead of strings in _fixes/cmip5/test_access1_X.py

valeriupredoi:
## Description

Got rid of `num2date` calls that change the outputted string formatted date everytime there's a new release of `cf_units` and compare datime objects straight out the box.

Closes #1158 

Link to documentation:

***

## [Before you get started](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#getting-started)

- [x] [☝ Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do

## [Checklist](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#checklist-for-pull-requests)

It is the responsibility of the author to make sure the pull request is ready to review. The icons indicate whether the item will be subject to the [🛠 Technical][1] or [🧪 Scientific][2] review.

[1]: https://docs.esmvaltool.org/en/latest/community/review.html#technical-review
[2]: https://docs.esmvaltool.org/en/latest/community/review.html#scientific-review

- [x] [🧪][2] The new functionality is [relevant and scientifically sound](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#scientific-relevance)
- [x] [🛠][1] This pull request has a [descriptive title and labels](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-title-and-label)
- [x] [🛠][1] Code is written according to the [code quality guidelines](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#code-quality)
- [x] [🛠][1] Changes are [backward compatible](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#backward-compatibility)
- [x] [🛠][1] All [checks below this pull request](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-checks) were successful

***

To help with the number pull requests:

- 🙏 We kindly ask you to [review](https://docs.esmvaltool.org/en/latest/community/review.html#review-of-pull-requests) two other [open pull requests](https://github.com/ESMValGroup/ESMValCore/pulls) in this repository

PR #1216, opened 2021-07-06
Fixed search for fx files when no ``mip`` is given

### LINKED ISSUE #1159: Fx files are selected differently depending on what OS the code runs on
**Bug summary**

[this line](https://github.com/ESMValGroup/ESMValCore/blob/fc87d72c3fce12a2117b4b4e42a4857045c4d404/esmvalcore/_recipe.py#L346) loops through all possible MIPs - for the test fail the explanation is that on a Debian-based OS (Ubuntu like on my machine and Github Actions) the order in which the tables hence mip list is passes is random (but fixed each time the test runs) and it happens that `IyrAnt` is the last actual mip in the loop that is valid, whereas for CentOS (JASMIN and the CI machine), the tables hence the mips are ordered by UNIX order type: numerals, capitals then lower case last, so `fx` is the last valid mip so the test passes. This, however, uncovers a potential vulnerability - this is a rather random process of selecting the mip, and is not solid in my view - iike in the test case, for `sftgif` we have both `fx` and `IyrAnt` files available, and the file selection is performed depending on the OS where the code runs on - this is dodgy. Any ideas how to select the file that we need and is consistent across OS's? This potentially impactful since `fx` is not time-dependent whereas eg `Oyr` or `IyrAnt` are time-dependent!

**Test fail that discovered the issue**

This is an odd one: that test fails **only** on Ubuntu and OSX machines. The Ubuntu fail is reported in [this closed test PR](https://github.com/ESMValGroup/ESMValCore/pull/1149) with the [test failing on both my machine and GA both Ubuntu](https://github.com/ESMValGroup/ESMValCore/runs/2759882006?check_suite_focus=true) and the [OSX fail is here](https://github.com/ESMValGroup/ESMValCore/runs/2759882084?check_suite_focus=true) - the test is failing since the `filename` key returns a file from a list of files that does not have `fx` in it. I am not sure if this is an issue with the actual preprocessor or if it's just the test object that misbehaves on different platforms. I'd like to investigate more but I need to write some slides now instead :confounded: This is, however, a sneaky bug :beetle: @zklaus the anaconda package build may fail coz of this (depending what machine the tests are run on)

schlunma:
This PR fixes the automatic retrieval of fx files when no `mip` is specified.

I agree with @sloosvel that a predefined order for the mip tables is not useful due to the many different project tables. Therefore I just implemented an alphabetic ordering, i.e., the first `mip` that has data is automatically selected. To bypass this, the user has to define a specific `mip` in the recipe. I added a paragraph to the documentation that describes this process.

This should go into 2.3.1 since this fixes a bug that (1) lets the test fails and (2) might lead to the tool wrongly complaining about missing data. #1169 only disabled the tests and didn't actually fix the bug.

Closes #1159.

Link to documentation: https://esmvaltool--1216.org.readthedocs.build/projects/ESMValCore/en/1216/recipe/preprocessor.html#fx-variables-as-cell-measures-or-ancillary-variables

***

***

### DIFF
--- a/doc/recipe/preprocessor.rst
+++ b/doc/recipe/preprocessor.rst
@@ -180,14 +180,62 @@ To get an overview on data fixes and how to implement new ones, please go to
 
 Fx variables as cell measures or ancillary variables
 ====================================================
-The following preprocessor may require the use of ``fx_variables`` 
-to be able to perform the computations:
+The following preprocessor may require the use of ``fx_variables`` to be able
+to perform the computations:
 
-    - ``area_statistics``
-    - ``mask_landsea``
-    - ``mask_landseaice``
-    - ``volume_statistics``
-    - ``weighting_landsea_fraction``
+    - area_statistics_ (``areacella``, ``areacello``)
+    - :ref:`mask_landsea<land/sea/ice masking>` (``sftlf``, ``sftof``)
+    - :ref:`mask_landseaice<ice masking>` (``sftgif``)
+    - volume_statistics_ (``volcello``)
+    - :ref:`weighting_landsea_fraction<land/sea fraction weighting>` (``sftlf``, ``sftof``)
+
+If no ``fx_variables`` are specified for these preprocessors, the fx variables
+given in brackets are used. If given, the ``fx_variables`` argument specifies
+the fx variables that the user wishes to input to the corresponding
+preprocessor function; the user may specify it calling the variables e.g.
+
+.. code-block:: yaml
+
+    fx_variables:
+      areacello:
+      volcello:
+
+or calling the variables and adding specific variable parameters (the key-value
+pair may be as specific as a CMOR variable can permit):
+
+.. code-block:: yaml
+
+    fx_variables:
+      areacello:
+        mip: Omon
+      volcello:
+        mip: fx
+
+Alternatively, the ``fx_variables`` argument can also be specified as a list:
+
+.. code-block:: yaml
+
+    fx_variables: ['areacello', 'volcello']
+
+or as a list of dictionaries:
+
+.. code-block:: yaml
+
+    fx_variables: [{'short_name': 'areacello', 'mip': 'Omon'}, {'short_name': 'volcello', 'mip': 'fx'}]
+
+The recipe parser will automatically find the data files that are associated
+with these variables and pass them to the function for loading and processing.
+
+If ``mip`` is not given, ESMValTool will search for the fx variable in all
+available tables of the specified project (in alphabetical order) and return
+the first match.
+
+.. note::
+   Some fx variables exist in more than one table (e.g., ``volcello`` exists in
+   CMIP6's ``Ofx`` and ``Omon`` table or ``sftgif`` exists in CMIP6's ``fx``
+   and ``IyrAnt`` table). If a specific table is desired, this needs to be
+   specified by ``mip``, otherwise the table is selected based on the
+   alphabetical order and data availability.
 
 The preprocessor step ``add_fx_variables`` loads the required ``fx_variables``,
 checks them against CMOR standards and adds them either as ``cell_measure``
@@ -195,7 +243,7 @@ or ``ancillary_variable`` inside the cube data. This ensures that the
 defined preprocessor chain is applied to both ``variables`` and ``fx_variables``.
 
 Note that when calling steps that require ``fx_variables`` inside diagnostic
-scripts, the variables are expected to contain the required ``cell_measures`` or 
+scripts, the variables are expected to contain the required ``cell_measures`` or
 ``ancillary_variables``. If missing, they can be added using the following functions:
 
 .. code-block::
@@ -209,7 +257,7 @@ scripts, the variables are expected to contain the required ``cell_measures`` or
     cube_with_ancillary_sftlf = add_ancillary_variable(cube, sftlf_cube)
 
     cube_with_ancillary_sftgif = add_ancillary_variable(cube, sftgif_cube)
-  
+
   Details on the arguments needed for each step can be found in the following sections.
 
 .. _Vertical interpolation:
@@ -351,8 +399,8 @@ is for example useful for climate models which do not offer land/sea fraction
 files. This arguments also accepts the special dataset specifiers
 ``reference_dataset`` and ``alternative_dataset``.
 
-Optionally you can specify your own custom fx variable to be used in cases when e.g. a certain
-experiment is preferred for fx data retrieval:
+Optionally you can specify your own custom fx variable to be used in cases when
+e.g. a certain experiment is preferred for fx data retrieval:
 
 .. code-block:: yaml
 
@@ -361,7 +409,7 @@ experiment is preferred for fx data retrieval:
         weighting_landsea_fraction:
           area_type: land
           exclude: ['CanESM2', 'reference_dataset']
-          fx_variables: 
+          fx_variables:
             sftlf:
               exp: piControl
             sftof:
@@ -377,9 +425,12 @@ or alternatively:
           area_type: land
           exclude: ['CanESM2', 'reference_dataset']
           fx_variables: [
-            {'short_name': 'sftlf', 'exp': 'piControl'}, 
+            {'short_name': 'sftlf', 'exp': 'piControl'},
             {'short_name': 'sftof', 'exp': 'piControl'}
             ]
+
+More details on the argument ``fx_variables`` and its default values are given
+:ref:`here<Fx variables as cell measures or ancillary variables>`.
 
 See also :func:`esmvalcore.preprocessor.weighting_landsea_fraction`.
 
@@ -426,11 +477,6 @@ To mask out a certain domain (e.g., sea) in the preprocessor,
 
 and requires only one argument: ``mask_out``: either ``land`` or ``sea``.
 
-The preprocessor automatically retrieves the corresponding mask (``fx: stfof``
-in this case) and applies it so that sea-covered grid cells are set to
-missing. Conversely, it retrieves the ``fx: sftlf`` mask when land needs to be
-masked out, respectively.
-
 Optionally you can specify your own custom fx variable to be used in cases when e.g. a certain
 experiment is preferred for fx data retrieval. Note that it is possible to specify as many tags
 for the fx variable as required:
@@ -442,8 +488,8 @@ for the fx variable as required:
       landmask:
         mask_landsea:
           mask_out: sea
-          fx_variables: 
-            sftlf: 
+          fx_variables:
+            sftlf:
               exp: piControl
             sftof:
               exp: piControl
@@ -458,9 +504,12 @@ or alternatively:
         mask_landsea:
           mask_out: sea
           fx_variables: [
-            {'short_name': 'sftlf', 'exp': 'piControl'}, 
+            {'short_name': 'sftlf', 'exp': 'piControl'},
             {'short_name': 'sftof', 'exp': 'piControl', 'ensemble': 'r2i1p1f1'}
             ]
+
+More details on the argument ``fx_variables`` and its default values are given
+:ref:`here<Fx variables as cell measures or ancillary variables>`.
 
 If the corresponding fx file is not found (which is
 the case for some models and almost all observational datasets), the
@@ -470,6 +519,8 @@ Natural Earth masks are much higher than any typical global model (10m for
 land and glaciated areas and 50m for ocean masks).
 
 See also :func:`esmvalcore.preprocessor.mask_landsea`.
+
+.. _ice masking:
 
 Ice masking
 -----------
@@ -487,11 +538,8 @@ losing generality. To mask ice out, ``mask_landseaice`` can be used:
 
 and requires only one argument: ``mask_out``: either ``landsea`` or ``ice``.
 
-As in the case of ``mask_landsea``, the preprocessor automatically retrieves
-the ``fx_variables: [sftgif]`` mask.
-
-Optionally you can specify your own custom fx variable to be used in cases when e.g. a certain
-experiment is preferred for fx data retrieval:
+Optionally you can specify your own custom fx variable to be used in cases when
+e.g. a certain experiment is preferred for fx data retrieval:
 
 
 .. code-block:: yaml
@@ -500,7 +548,7 @@ experiment is preferred for fx data retrieval:
       landseaicemask:
         mask_landseaice:
           mask_out: sea
-          fx_variables: 
+          fx_variables:
             sftgif:
               exp: piControl
 
@@ -513,6 +561,9 @@ or alternatively:
         mask_landseaice:
           mask_out: sea
           fx_variables: [{'short_name': 'sftgif', 'exp': 'piControl'}]
+
+More details on the argument ``fx_variables`` and its default values are given
+:ref:`here<Fx variables as cell measures or ancillary variables>`.
 
 See also :func:`esmvalcore.preprocessor.mask_landseaice`.
 
@@ -1420,40 +1471,9 @@ Note that this function is applied over the entire dataset. If only a specific
 region, depth layer or time period is required, then those regions need to be
 removed using other preprocessor operations in advance.
 
-The ``fx_variables`` argument specifies the fx variables that the user wishes to input to the function;
-the user may specify it calling the variables e.g.
-
-.. code-block:: yaml
-
-    fx_variables: 
-      areacello:
-      volcello:
-
-or calling the variables and adding specific variable parameters (the key-value pair may be as specific
-as a CMOR variable can permit):
-
-.. code-block:: yaml
-
-    fx_variables: 
-      areacello:
-        mip: Omon
-      volcello:
-        mip: fx
-
-Alternatively, the ``fx_variables`` argument can also be specified as a list:
-
-.. code-block:: yaml
-
-    fx_variables: ['areacello', 'volcello']
-
-or as a list of dictionaries:
-
-.. code-block:: yaml
-
-    fx_variables: [{'short_name': 'areacello', 'mip': 'Omon'}, {'short_name': 'volcello', 'mip': 'fx'}]
-
-The recipe parser will automatically find the data files that are associated with these
-variables and pass them to the function for loading and processing.
+The optional ``fx_variables`` argument specifies the fx variables that the user
+wishes to input to the function. More details on this are given
+:ref:`here<Fx variables as cell measures or ancillary variables>`.
 
 See also :func:`esmvalcore.preprocessor.area_statistics`.
 
@@ -1496,42 +1516,10 @@ This function takes the argument: ``operator``, which defines the operation to
 apply over the volume.
 
 No depth coordinate is required as this is determined by Iris. This function
-works best when the ``fx_variables`` provide the cell volume.
-
-The ``fx_variables`` argument specifies the fx variables that the user wishes to input to the function;
-the user may specify it calling the variables e.g.
-
-.. code-block:: yaml
-
-    fx_variables: 
-      areacello:
-      volcello:
-
-or calling the variables and adding specific variable parameters (the key-value pair may be as specific
-as a CMOR variable can permit):
-
-.. code-block:: yaml
-
-    fx_variables: 
-      areacello:
-        mip: Omon
-      volcello:
-        mip: fx
-
-Alternatively, the ``fx_variables`` argument can also be specified as a list:
-
-.. code-block:: yaml
-
-    fx_variables: ['areacello', 'volcello']
-
-or as a list of dictionaries:
-
-.. code-block:: yaml
-
-    fx_variables: [{'short_name': 'areacello', 'mip': 'Omon'}, {'short_name': 'volcello', 'mip': 'fx'}]
-
-The recipe parser will automatically find the data files that are associated with these
-variables and pass them to the function for loading and processing.
+works best when the ``fx_variables`` provide the cell volume. The optional
+``fx_variables`` argument specifies the fx variables that the user wishes to
+input to the function. More details on this are given
+:ref:`here<Fx variables as cell measures or ancillary variables>`.
 
 See also :func:`esmvalcore.preprocessor.volume_statistics`.
 
--- a/doc/recipe/preprocessor.rst
+++ b/doc/recipe/preprocessor.rst
@@ -205,16 +205,20 @@ the variables, e.g.,
       areacello:
       volcello:
 
-or by explicitly adding variable parameters (the key-value pair may be as
-specific as a CMOR variable can permit):
+or by additionally specifying further keys that are used to define the fx
+datsets, e.g.,
 
 .. code-block:: yaml
 
     fx_variables:
       areacello:
         mip: Ofx
+        exp: piControl
       volcello:
         mip: Omon
+
+This might be useful to select fx files from a specific ``mip`` table or from a
+specific ``exp`` in case not all experiments provide the fx variable.
 
 Alternatively, the ``fx_variables`` argument can also be specified as a list:
 
@@ -226,7 +230,7 @@ or as a list of dictionaries:
 
 .. code-block:: yaml
 
-    fx_variables: [{'short_name': 'areacello', 'mip': 'Ofx'}, {'short_name': 'volcello', 'mip': 'Omon'}]
+    fx_variables: [{'short_name': 'areacello', 'mip': 'Ofx', 'exp': 'piControl'}, {'short_name': 'volcello', 'mip': 'Omon'}]
 
 The recipe parser will automatically find the data files that are associated
 with these variables and pass them to the function for loading and processing.
--- a/doc/recipe/preprocessor.rst
+++ b/doc/recipe/preprocessor.rst
@@ -180,19 +180,24 @@ To get an overview on data fixes and how to implement new ones, please go to
 
 Fx variables as cell measures or ancillary variables
 ====================================================
-The following preprocessor may require the use of ``fx_variables`` to be able
+The following preprocessors may require the use of ``fx_variables`` to be able
 to perform the computations:
 
-    - area_statistics_ (``areacella``, ``areacello``)
-    - :ref:`mask_landsea<land/sea/ice masking>` (``sftlf``, ``sftof``)
-    - :ref:`mask_landseaice<ice masking>` (``sftgif``)
-    - volume_statistics_ (``volcello``)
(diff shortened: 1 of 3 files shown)


zklaus:
The issue with this approach is that there are experiments, where there are two or more fx files for the same variable and it is impossible to determine a generic right order, e.g. [here](http://clipc-services.ceda.ac.uk/dreq/u/e9b495e2-5989-11e6-a4be-ac72891c3257.html). I think https://github.com/ESMValGroup/ESMValCore/issues/1159#issuecomment-857770098 is relevant. In other words, just specify the correct table in the recipe.

zklaus (changes requested):
I don't think any generic ordering can meaningfully resolve existing ambiguity in the table selection.

zklaus:
I do like the documentation improvements. Let's work that into 2.4.0.

schlunma:
> I don't think any generic ordering can meaningfully resolve existing ambiguity in the table selection.

I agree, that's why I added this note to the documentation:

```rst
.. note::
   Some fx variables exist in more than one table (e.g., ``volcello`` exists in
   CMIP6's ``Ofx`` and ``Omon`` table or ``sftgif`` exists in CMIP6's ``fx``
   and ``IyrAnt`` table). If a specific table is desired, this needs to be
   specified by ``mip``, otherwise the table is selected based on the
   alphabetical order and data availability.
```

Do you want the tool to fail if there's is ambiguity? Or issue a warning?

I still thinks this should go into 2.3.1 since this merely fixes a bug that was introduced in 2.3.0.

zklaus:
I think the tool should fail if there is ambiguity.

As for the bug, are you referring to #1159? If so, in my view, that doesn't reflect the introduction of a new bug, but the uncovering of an existing bug in the test, namely the ambiguity. The solution seems clear: specify the correct table in the recipe used in the test. I am not sure why @sloosvel's comment there went unanswered.

If you mean another bug, I am sorry I missed it. In that case, could you point me to the right issue, please?

schlunma:
The tests were designed for the pre-#999 situation and did not fail there. This old implementation did only check tables that included `fx` in their name, so the new version is definitely an improvement here. However, the point is that the tests didn't fail on any machine before #999 and failed after it on some machines.

#1169 "solved" this by disabling the tests. I don't think that's the way to go. This PR fixes the tests by introducing a machine-independent ordering for the tables.

I agree with you that in case of ambiguity the tool should fail, but this would be a new feature and should go into v2.4 (I can open another PR for that).

zklaus:
It seems we agree that the bug is in the test. We also agree that disabling the test is not really the right way to address this. But there seems to be a clear solution: fix the test. Why would we want to change the behavior of the tool to work around a buggy test?

schlunma:
- Added [`break`](https://github.com/ESMValGroup/ESMValCore/pull/1216/files#diff-b675431b6cf759c61f7867680472aabc0304475fd666ed514ca0365bb87c2a47R366) statement: This was missing in #999 (see https://github.com/ESMValGroup/ESMValCore/issues/1159#issuecomment-856836940 for further information.
-Added [`sorted`](https://github.com/ESMValGroup/ESMValCore/pull/1216/files#diff-b675431b6cf759c61f7867680472aabc0304475fd666ed514ca0365bb87c2a47R387-R391): This was not necessary pre-#999 but is necessary now to ensure a consistent behavior across machines. This can be removed once we make the tool fail in case of ambiguity.
- Changed message in [`RecipeError`](https://github.com/ESMValGroup/ESMValCore/pull/1216/files#diff-b675431b6cf759c61f7867680472aabc0304475fd666ed514ca0365bb87c2a47R401-R407): This only changes the message in case a `mip` is given. I can remove this if you like, but I think there is no harm in having this.

sloosvel:
I think the changes make the documentation much better. 

I agree that that `break` should not have been removed, it went over my head. And in case of ambiguity (which I have yet to see an exp with the same fx var in multiple mips, but who knows), specifying the mip solves the issue so it's much simpler than trying to come up with a search order that would fit all projects.

valeriupredoi:
@sloosvel - @zklaus provides an example of exactly that - `areacellg` in two mips, [here](http://clipc-services.ceda.ac.uk/dreq/u/e9b495e2-5989-11e6-a4be-ac72891c3257.html) and there are others (I believe even `volcello` that could be `fx` or `Ofx` if it's time variant or not, Klaus correct me if I'm wrong pls). I like this approach but I think we need the run to stop if it encounters multiple mips for the same fx variable, and prompt the user to make up their mind and add valid mip. This saves us from any future headaches due to ESGF data standards variations. Note the current lazy approach to not produce these sort of variables for all exps and mips and shortshifting by making use of just a small data production eg fx vars are produced for one ensemble only or one experiment only

zklaus:
> (I believe even `volcello` that could be `fx` or `Ofx` if it's time variant or not, Klaus correct me if I'm wrong pls).

Almost right: `volcello` can be in one or more of `Ofx`, `Omon`, `Oyr`, and `Odec`.

schlunma:
All right, I will alter the code so that it raises an error if multiple `mips` are encountered that include a given fx variable. I still believe that this should go into the release 2.3.1, since it's clearly a fix of a broken feature.

valeriupredoi:
cheers @zklaus and cheers @schlunma too, and I too believe this should be part of 2.3.1 but Klaus is the release bossman :+1:

zklaus:
Alright, I am not against putting this in 2.3.1, but it still needs a bit of work (I spotted some things in the documentation at least and I think we don't have any formal review yet), so it might come down to a question of timing.

schlunma:
Alright, now the tool raises an error if an fx variable is found in multiple tables. I also adapted the documentation accordingly.

doc/recipe/preprocessor.rst
```diff
@@ -180,22 +180,68 @@ To get an overview on data fixes and how to implement new ones, please go to
...
+
+    fx_variables: [{'short_name': 'areacello', 'mip': 'Omon'}, {'short_name': 'volcello', 'mip': 'fx'}]
+
+The recipe parser will automatically find the data files that are associated
+with these variables and pass them to the function for loading and processing.
+
+If ``mip`` is not given, ESMValTool will search for the fx variable in all
+available tables of the specified project.
+
+.. note::
```
valeriupredoi, line 232:
```suggestion
.. warning::
```

doc/recipe/preprocessor.rst
```diff
@@ -180,22 +180,68 @@ To get an overview on data fixes and how to implement new ones, please go to
...
-    - ``area_statistics``
-    - ``mask_landsea``
-    - ``mask_landseaice``
-    - ``volume_statistics``
-    - ``weighting_landsea_fraction``
+    - area_statistics_ (``areacella``, ``areacello``)
+    - :ref:`mask_landsea<land/sea/ice masking>` (``sftlf``, ``sftof``)
+    - :ref:`mask_landseaice<ice masking>` (``sftgif``)
+    - volume_statistics_ (``volcello``)
+    - :ref:`weighting_landsea_fraction<land/sea fraction weighting>` (``sftlf``, ``sftof``)
```
zklaus, line 190: Perhaps format as table with columns preprocessor, applicable fx variables?
Also, why the use of `:ref:` only for some of the preprocessor?

doc/recipe/preprocessor.rst
```diff
@@ -180,22 +180,68 @@ To get an overview on data fixes and how to implement new ones, please go to
 
 Fx variables as cell measures or ancillary variables
 ====================================================
-The following preprocessor may require the use of ``fx_variables`` 
-to be able to perform the computations:
+The following preprocessor may require the use of ``fx_variables`` to be able
```
zklaus, line 183: s/preprocessor/preprocessors/

doc/recipe/preprocessor.rst
```diff
@@ -180,22 +180,68 @@ To get an overview on data fixes and how to implement new ones, please go to
...
+preprocessor function; the user may specify it calling the variables e.g.
+
+.. code-block:: yaml
+
+    fx_variables:
+      areacello:
+      volcello:
+
+or calling the variables and adding specific variable parameters (the key-value
+pair may be as specific as a CMOR variable can permit):
```
zklaus, line 204: This line doesn't read quite right. Could you double check?

doc/recipe/preprocessor.rst
```diff
@@ -180,22 +180,68 @@ To get an overview on data fixes and how to implement new ones, please go to
...
+
+.. code-block:: yaml
+
+    fx_variables:
+      areacello:
+        mip: Omon
+      volcello:
+        mip: fx
```
zklaus, line 212:
```suggestion
      areacello:
        mip: Ofx
      volcello:
        mip: Omon
```
The area of ocean cells typically stays the same, but changes in sea surface height vary the thickness and thus the volume.

doc/recipe/preprocessor.rst
```diff
@@ -180,22 +180,68 @@ To get an overview on data fixes and how to implement new ones, please go to
...
+
+.. code-block:: yaml
+
+    fx_variables: ['areacello', 'volcello']
+
+or as a list of dictionaries:
+
+.. code-block:: yaml
+
+    fx_variables: [{'short_name': 'areacello', 'mip': 'Omon'}, {'short_name': 'volcello', 'mip': 'fx'}]
```
zklaus, line 224:
```suggestion
    fx_variables: [{'short_name': 'areacello', 'mip': 'Ofx'}, {'short_name': 'volcello', 'mip': 'Omon'}]
```

doc/recipe/preprocessor.rst
```diff
@@ -180,22 +180,68 @@ To get an overview on data fixes and how to implement new ones, please go to
...
+
+The recipe parser will automatically find the data files that are associated
+with these variables and pass them to the function for loading and processing.
+
+If ``mip`` is not given, ESMValTool will search for the fx variable in all
+available tables of the specified project.
+
+.. note::
+   Some fx variables exist in more than one table (e.g., ``volcello`` exists in
+   CMIP6's ``Ofx``, ``Oyr``, ``Odec`` and ``Omon`` tables or ``sftgif`` exists
```
zklaus, line 234:
```suggestion
   CMIP6's ``Ofx``, ``Oyr``, ``Odec``, and ``Omon`` tables or ``sftgif`` exists
```
Oxford comma

doc/recipe/preprocessor.rst
```diff
@@ -180,22 +180,68 @@ To get an overview on data fixes and how to implement new ones, please go to
...
+   in CMIP6's ``fx``, ``LImon``, ``IyrGre`` and ``IyrAnt`` tables). In these
```
zklaus, line 235:
```suggestion
   in CMIP6's ``fx``, ``LImon``, ``IyrGre``, and ``IyrAnt`` tables). In these
```

valeriupredoi (approved):
excellent work, cheers @schlunma :beer: One nag for me, I'd change that from a Note to a Warning :+1:

doc/recipe/preprocessor.rst
```diff
@@ -180,22 +180,68 @@ To get an overview on data fixes and how to implement new ones, please go to
...
+   cases, ``mip`` needs to be specified, otherwise an error is raised.
```
zklaus, line 236: Perhaps it's better to say "In case of ambiguity, ..." because sometimes the variable may be in different tables for different datasets, but every dataset individually is unambiguous.

doc/recipe/preprocessor.rst
```diff
@@ -180,22 +180,68 @@ To get an overview on data fixes and how to implement new ones, please go to
...
 
 The preprocessor step ``add_fx_variables`` loads the required ``fx_variables``,
```
zklaus, line 238: I think we should either add a `:ref:` to `add_fx_variables` or avoid explicit mentioning of it (something along the lines of "...an internal preprocessor step...") to avoid confusion.

doc/recipe/preprocessor.rst
```diff
@@ -180,22 +180,68 @@ To get an overview on data fixes and how to implement new ones, please go to
...
 checks them against CMOR standards and adds them either as ``cell_measure``
```
zklaus, line 239: Link to [CF](http://cfconventions.org/cf-conventions/cf-conventions.html#cell-measures) and or [Iris](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/coords.html#iris.coords.CellMeasure) cell measures?

doc/recipe/preprocessor.rst
```diff
@@ -180,22 +180,68 @@ To get an overview on data fixes and how to implement new ones, please go to
...
 or ``ancillary_variable`` inside the cube data. This ensures that the
```
zklaus, line 240: Link to [CF](http://cfconventions.org/cf-conventions/cf-conventions.html#ancillary-data) or [Iris](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/coords.html#iris.coords.AncillaryVariable) ancillary variables?

doc/recipe/preprocessor.rst
```diff
@@ -180,22 +180,68 @@ To get an overview on data fixes and how to implement new ones, please go to
```
zklaus, line 180: Adding an explicit anchor (something like, but better than `.. _fx ancillary cell measures:`) will make it easier in the future to keep the links intact.

doc/recipe/preprocessor.rst
```diff
@@ -377,10 +423,13 @@ or alternatively:
           area_type: land
           exclude: ['CanESM2', 'reference_dataset']
           fx_variables: [
-            {'short_name': 'sftlf', 'exp': 'piControl'}, 
+            {'short_name': 'sftlf', 'exp': 'piControl'},
             {'short_name': 'sftof', 'exp': 'piControl'}
             ]
 
+More details on the argument ``fx_variables`` and its default values are given
+:ref:`here<Fx variables as cell measures or ancillary variables>`.
```
zklaus, line 431: See comment on explicit anchor at the top.

doc/recipe/preprocessor.rst
```diff
@@ -458,10 +502,13 @@ or alternatively:
         mask_landsea:
           mask_out: sea
           fx_variables: [
-            {'short_name': 'sftlf', 'exp': 'piControl'}, 
+            {'short_name': 'sftlf', 'exp': 'piControl'},
             {'short_name': 'sftof', 'exp': 'piControl', 'ensemble': 'r2i1p1f1'}
             ]
 
+More details on the argument ``fx_variables`` and its default values are given
+:ref:`here<Fx variables as cell measures or ancillary variables>`.
```
zklaus, line 510: See comment on explicit anchor at the top.

doc/recipe/preprocessor.rst
```diff
@@ -514,6 +560,9 @@ or alternatively:
           mask_out: sea
           fx_variables: [{'short_name': 'sftgif', 'exp': 'piControl'}]
 
+More details on the argument ``fx_variables`` and its default values are given
+:ref:`here<Fx variables as cell measures or ancillary variables>`.
```
zklaus, line 564: See comment about explicit anchor on top.

doc/recipe/preprocessor.rst
```diff
@@ -1420,40 +1469,9 @@ Note that this function is applied over the entire dataset. If only a specific
...
-
-.. code-block:: yaml
-
-    fx_variables: [{'short_name': 'areacello', 'mip': 'Omon'}, {'short_name': 'volcello', 'mip': 'fx'}]
-
-The recipe parser will automatically find the data files that are associated with these
-variables and pass them to the function for loading and processing.
+The optional ``fx_variables`` argument specifies the fx variables that the user
+wishes to input to the function. More details on this are given
+:ref:`here<Fx variables as cell measures or ancillary variables>`.
```
zklaus, line 1474: See comment on explicit anchor on top.

doc/recipe/preprocessor.rst
```diff
@@ -1496,42 +1514,10 @@ This function takes the argument: ``operator``, which defines the operation to
...
-.. code-block:: yaml
-
-    fx_variables: [{'short_name': 'areacello', 'mip': 'Omon'}, {'short_name': 'volcello', 'mip': 'fx'}]
-
-The recipe parser will automatically find the data files that are associated with these
-variables and pass them to the function for loading and processing.
+works best when the ``fx_variables`` provide the cell volume. The optional
+``fx_variables`` argument specifies the fx variables that the user wishes to
+input to the function. More details on this are given
+:ref:`here<Fx variables as cell measures or ancillary variables>`.
```
zklaus, line 1520: Explicit anchor comment.

esmvalcore/_recipe.py
```diff
@@ -349,26 +349,41 @@ def _add_fxvar_keys(fx_info, variable):
...
-            fx_info['mip'] = mip
-            fx_info = _add_fxvar_keys(fx_info, variable)
-            logger.debug("For fx variable '%s', found table '%s'",
-                         fx_info['short_name'], mip)
-            fx_files = _get_input_files(fx_info, config_user)[0]
-            if fx_files:
-                logger.debug("Found fx variables '%s':\n%s",
-                             fx_info['short_name'], pformat(fx_files))
-    return found_mip, fx_info, fx_files
+def _search_fx_mip(tables, variable, fx_info, config_user):
```
zklaus, line 352: Hm. I am not sure we are on the same page here. The ambiguity does not come from the variable being defined in several tables. It comes into play when a model delivers it in several tables. For example, for CMIP6 `volcello` can be delivered in `Ofx` or (`Omon` and `Odec`). If the model delivers it only in `Ofx`, that's fine and unambiguous. If it delivers it in `Ofx` and at least one of the other two, that's an error. Delivering it for both `Omon` and `Odec` is ok; I am not sure about delivering it only for, say, `Omon`.

zklaus (changes requested):
Sorry, only made it just to the actual code; have to continue later. Anyway, I think the most important comment is the last one.

doc/recipe/preprocessor.rst
schlunma: :ref: is only necessary for those that are not section titles yet. If you have a look at the compiled doc, you don't see any difference, all preprocessors are properly linked.

doc/recipe/preprocessor.rst
schlunma: No - this is exactly what the code does, see also my comment below.

doc/recipe/preprocessor.rst
schlunma: I slightly modified this by adding "internally" and "automatically". Note that I only copy-pasted these parts of the documentation, they were already present in the old version.

doc/recipe/preprocessor.rst
schlunma: I don't get this comment.There is an anchor right there which I'm also referring to? `.. _Fx variables as cell measures or ancillary variables:`

esmvalcore/_recipe.py
schlunma: Yes, I think we are and I also understand your point.

To implement it like you describe it here a whole bunch of custom rules for every fx variable for every project has to be implemented. I thought we agreed that this is not really feasible. I think simply specifying the `mip` in these cases is by far the simplest solution that works for every project.

doc/recipe/preprocessor.rst
zklaus: You may want to read https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-arbitrary-locations. To quote:
> Using ref is advised over standard reStructuredText links to sections (like `Section title`_) because it works across files, when section headings are changed, will raise warnings if incorrect, and works for all builders that support cross-references.

doc/recipe/preprocessor.rst
zklaus: This relates to the use of `:ref:` later on. Sorry, the terminology is a bit convoluted in this space: You are linking directly to the title of the section using what naked restructured text calls [an implicit hyperlink target](https://docutils.sourceforge.io/docs/user/rst/quickref.html#hyperlink-targets) and I think I saw referred to in a sphinx context as an automatic label.

The problem with this is, when at some point in the future we change the heading, all the links break. Also, you cannot refer to this from a different file in the same document.

If instead we put an extra anchor or label here, we can refer to it from anywhere and the validity of the links is ensured by sphinx.

esmvalcore/_recipe.py
zklaus: I don't think we would need a lot of custom rules. What I propose is basically to send the data finder out and do its worst and if it finds more than one candidate, then we barf.

doc/recipe/preprocessor.rst
```diff
@@ -180,22 +180,80 @@ To get an overview on data fixes and how to implement new ones, please go to
...
-    - ``volume_statistics``
-    - ``weighting_landsea_fraction``
+.. code-block:: yaml
+
+    fx_variables:
+      areacello:
+      volcello:
+
+or by explicitly adding variable parameters (the key-value pair may be as
+specific as a CMOR variable can permit):
```
zklaus, line 209: I don't understand what this sentence means. Could you elaborate?

zklaus (changes requested):
Sorry if something is very weird; I had some trouble with the github web interface.

doc/recipe/preprocessor.rst
schlunma: But there IS an anchor:

```rst
...
To get an overview on data fixes and how to implement new ones, please go to
:ref:`fixing_data`.

.. _Fx variables as cell measures or ancillary variables:

Fx variables as cell measures or ancillary variables
====================================================
The following preprocessor may require the use of ``fx_variables`` 
to be able to perform the computations:
...
```

It may be hidden in the diff since it's already present in the current code.

doc/recipe/preprocessor.rst
schlunma: I'll try to rephrase (just copy-pasted this code from the old version).

esmvalcore/_recipe.py
schlunma: The way you described it here would mean custom rules:

> For example, for CMIP6 `volcello` can be delivered in `Ofx` or (`Omon` and `Odec`). If the model delivers it only in `Ofx`, that's fine and unambiguous. If it delivers it in `Ofx` and at least one of the other two, that's an error. Delivering it for both `Omon` and `Odec` is ok; I am not sure about delivering it only for, say, `Omon`.

How else should the tool know that `Omon` and `Odec` is ok, but `Omon` and `Ofx` is not?

Raising an error when files are found in more than one should not be an issue, I can do that. But bear in mind that this could mean that for different datasets fx variables from different tables (e.g., `IyrAnt` and `fx`) might be mixed. Is this really what we want?

doc/recipe/preprocessor.rst
zklaus: You are right, sorry! I did indeed miss the anchor because it didn't show up in the diff and also because it is exactly the same text as the title. I think it would be better if it was different, but no need to change it here.

esmvalcore/_recipe.py
zklaus: You are right, the tool would not know this. I should have phrased that better. It is true that the perfect solution, where the tool understands exactly which combinations are ok would require such rules. But my goal was to get an 80% solution by complaining only when there is actual ambiguity in the available data.

As for the second point, we absolutely want to support fx files coming from different tables for different datasets. The abovementioned `volcello` is a good example. Most models will have that in `Ofx` but [some](https://esgf-data.dkrz.de/search/cmip6-dkrz/?variable_id=volcello&table_id=Omon) have it in `Omon`.

schlunma: Alright, will implement this.

schlunma:
I don't know why the tests are failing, I though these

```
TypeError: cannot compare cftime.DatetimeGregorian(300, 1, 16, 12, 0, 0, 0) and datetime.datetime(300, 1, 16, 12, 0) (different calendars)
```

errors were fixed by #1173?

valeriupredoi:
pls merge `main` in here, if you done that since yesterday afternoon then the test container needs to rebuild with the new environment from #1197 (I belive it's done that last night) - give it a day, it'll do it, if not bugger @bouweandela see if it's stuck. Note that you can run the tests locally too and see it works

schlunma:
After the merge of `main` the error got more interesting :smile: 

```
ValueError: Tag 'andela_bouwe' does not exist in section 'authors'
```

valeriupredoi:
yes, @bouweandela is famous for causing SegFaults :rofl: It's a SegFault, Manu, am rerunning it just for you :grin:

schlunma:
The error didn't disappear after 3 restarts, so I guess there is something wrong here... Tests run fine locally.

valeriupredoi:
hang on, lemme investigate :mag

valeriupredoi:
running tests locally I am getting this:
```
E       assert "Requested fx variable 'volcello' for dataset 'CanESM5' of project 'CMIP6' is available in more than one CMOR table for 'CMIP6': ['Ofx', 'Oyr', 'Odec', 'Omon']" in "Requested fx variable 'volcello' for dataset 'CanESM5' of project 'CMIP6' is available in more than one CMOR table for 'CMIP6': ['Oyr', 'Ofx', 'Odec', 'Omon']"
E        +  where "Requested fx variable 'volcello' for dataset 'CanESM5' of project 'CMIP6' is available in more than one CMOR table for 'CMIP6': ['Oyr', 'Ofx', 'Odec', 'Omon']" = <[RecursionError('maximum recursion depth exceeded while getting the repr of an object') raised in repr()] RecipeError object at 0x7fc19b08a4c0>.message

tests/integration/test_recipe.py:2799: AssertionError
```
and no SegFault after 2 runs, after updating to Python 3.9.6 too

schlunma:
Oh, that's again related to the machine-dependent ordering. Lemme change that real quick.

valeriupredoi:
you know what? That recursion depth error might have given me a massive clue why we get all those SegFaults anyway

valeriupredoi:
> Oh, that's again related to the machine-dependent ordering. Lemme change that real quick.

you not running Ubuntu too on your machine?

schlunma:
Was doing the tests on mistral which runs on Red Hat Enterprise Linux

valeriupredoi:
weird coz I think that Circle runs on CentOS

schlunma:
Thanks so much V., that seems to have done the trick! :tada: 

@zklaus Any further comments from your side? I hope I addressed them all.

valeriupredoi:
so funny - now the tests pass fine and no more SegFault - I am going to get to the bottom of those SegFaults so help me Lord Vader! :grin:

schlunma:
Haha, I hope you will!

zklaus (approved):

zklaus:
Nice work, @schlunma ! :beers:

PR #1217, opened 2021-07-06
Add an option to download missing data from ESGF

### LINKED ISSUE #788: Download subcommand for downloading of tutorial data
It would be nice to be able to ask ESMValTool to download some "getting started" data for you for the tutorial. This could for instance be in the form of a download subcommand.

```esmvaltool download example-data```

Eventually we can also add the capability to download data for a specified recipe to this new subcommand (via Synda or otherwise).

Thoughts?

### LINKED ISSUE #1130: Replace synda download with esgf-pyclient download?
Because synda is very difficult to install and configure, while esgf-pyclient is easy to install and easier to configure than synda.

### LINKED ISSUE #1232: Running a recipe through the python API fails because of missing `extra_facets_dir`
```python
from esmvalcore.experimental import get_recipe
recipe = get_recipe('examples/recipe_python.yml')
recipe.run()
```

fails with

```
~/ESMValCore/esmvalcore/_recipe.py in _initialize_variables(self, raw_variable, raw_datasets)
   1154             required_keys.update({'start_year', 'end_year'})
   1155         for variable in variables:
-> 1156             _add_extra_facets(variable, self._cfg['extra_facets_dir'])
   1157             if 'institute' not in variable:
   1158                 institute = get_institutes(variable)

KeyError: 'extra_facets_dir'
```

Running the same recipe from the command line works fine.

bouweandela:
## Description

Replaces the ``--synda-download`` command line flag with an ``--offline`` flag. If this flag is disabled, the tool will now reliably download missing data from ESGF for CMIP3, CMIP5, CMIP6, CORDEX, and obs4MIPs. By default, the tool will still run in offline mode, but automatic downloading can be enabled by setting `offline: false` in config-user.yml or adding a `--offline=False` option on the command line. 

A new setting `download_dir` is available in config-user.yml, where data will be downloaded. New default settings for `rootpath` and `drs` that match the downloaded data are also included, so users who start out with no data get a reasonable default.

This pull request also includes a few other changes:
- renames the `obs4mips` project to `obs4MIPs`, to be consistent with the naming on ESGF.
- makes the reporting of missing data years nicer, previously this was a list of years, now it are ranges e.g. 1850-1900 instead of all numbers written out
- avoid a crash when serializing provenance if there is a space in a global attribute of the netcdf file
- Fixes a problem with the experimental API if a configuration parameter is not available in the default config-user.yml (Closes #1232)

Closes #1130 
Closes #788

Link to documentation:
Command line option `--offline`: https://esmvaltool--1217.org.readthedocs.build/projects/ESMValCore/en/1217/quickstart/run.html#running
Finding data: https://esmvaltool--1217.org.readthedocs.build/projects/ESMValCore/en/1217/quickstart/find_data.html#data-retrieval
Optional configuration file: https://esmvaltool--1217.org.readthedocs.build/projects/ESMValCore/en/1217/quickstart/configure.html#esgf-configuration
API documentation: https://esmvaltool--1217.org.readthedocs.build/projects/ESMValCore/en/1217/api/esmvalcore.esgf.html

***

***


bouweandela:
Regarding the Codacy issues:

> Library stubs not installed for "requests" (or incompatible with Python 3.8) (error) 
> Library stubs not installed for "yaml" (or incompatible with Python 3.8) (error) 

This is a problem with Codacy that I cannot solve, it does not install the required stubs libraries for type checking.

> Access to a protected member _from_results of a client class 

I would prefer to keep the `esmvalcore.esgf.ESGFFile._from_results` method private, because I think it is not useful for users of the API. `ESGFFile` instances are search results that are supposed to be created by the function `esmvalcore.esgf.find_files`, not directly by users.

doc/quickstart/configure.rst
```diff
@@ -165,6 +174,155 @@ the user.
...
+
+.. code-block:: bash
+
+    pip install keyring
+
+Next, set your username and password by running the commands:
+
+.. code-block:: bash
+
+    keyring set ESGF hostname
```
jvegasbsc, line 217: I find this section a bit confusing, mostly because ESGF has a bunch of hosts and how they manage the IDs. There are a few things that are not clear at first glance 

- Should I add only one or all I want to use?
- If only one, it can be anyone or must be the one I used to get my ESGF openID?
- The username is just the username or the full ESGF openID?

An example will be very helpful.

doc/quickstart/run.rst
```diff
@@ -48,21 +48,22 @@ To run a reduced version of the recipe, usually for testing purpose you can use
...
-to use it to automatically download the requested data from ESGF if it is not
-available locally:
+For running on systems without an internet connection, it is possible to
+disable searching for files on ESGF if they are not available locally
+by running:
 
 .. code:: bash
 
-    esmvaltool run --synda_download=True recipe_python.yml
+    esmvaltool run --offline=True recipe_python.yml
```
jvegasbsc, line 66: Should we have a toggle in the config-user for this? In our main analysis machine we do not have internet access, so this toggle should be always activated

jvegasbsc (changes requested):
There is also a more general question. The automatic download should be the default option or not?

I can see a few problems with ESMValTool downloading huge quantities of data in a user laptop after forgetting to run a recipe with the `--max-years` flag.

doc/quickstart/run.rst
bouweandela: I'll add it

bouweandela:
> There is also a more general question. The automatic download should be the default option or not?

After working with the new feature for a few weeks, I think so yes. It is very convenient to have this enabled and offers a better experience for new users.

> I can see a few problems with ESMValTool downloading huge quantities of data in a user laptop after forgetting to run a recipe with the --max-years flag.

The total amount of data that will be downloaded is listed before starting to download and if the user sees that the progress is taking forever and then looks at the log messages, they will see that they are trying to download a lot of data. I experimented with adding a download progress bar using tqdm, do you think that would help to make it more clear?

I could also add a more explicit warning to the documentation, do you think that would be better?

jvegasbsc:
It should be enough, yes. After all, if we later find that it was an error to download by default, it will be straightforward to fix.

doc/quickstart/run.rst
bouweandela: Done

doc/quickstart/configure.rst
bouweandela: I added an example and more explanation

bouweandela:
@jvegasbsc Could you have another look, please? I think I addressed all your comments.

bouweandela:
Thanks for the massive review @valeriupredoi!

>  I noticed that in the linked documentation SYNDA is still mentioned, are we not removing that?

No, we're not removing that. While we no longer use Synda for the built-in download function, Synda might still be a good choice for those who would like to maintain a large collection of ESGF data, as it has many more features than our built-in download function. This particular piece of documentation is about setting up the tool for use which such an externally managed data collection.

> what are the benefits (performance, data availability) vs Synda?

Much easier to install and configure (installed as a normal Python dependency, no configuration needed for using CMIP data). More reliable downloads (the Synda option did not work for me quite regularly).

> if the question above doesn't have a clear-cut answer, have we thought about the strategic decision of giving up using a European developed (ISENES-backed) application like Synda? Not to mention the familiarity a lot of the climate scientists have with it; on the other hand, pyesgf is actually developed by ESGF boffins so it should be much closer to the data itself than Synda. I am looking for a strategic and technical answer here (and sorry if this was discussed and given else where, and totally my bad I didn't get onto the issue in May when you opened it)

We're not giving up Synda, see also my answer to the first point you raised. It quite likely still is the best tool if you plan to maintain a large collection of ESGF data. However, it was not very suitable for our use case, see also my answer to the second point you raised.

> the default option to download data - I agree with it - but might bite back badly if, for instance, is used on a shared cluster that limits the user to a certain size of $HOME or whatever their space they use; even worse, this can clog group workspaces where users are allowed to write to (eg esmeval on JASMIN, policing constant disk clogging will get to a much higher level), so I'd use it with a pinch (heck, a cup) of salt. To be honest I'd disable it if the domain name is part of a blacklisted list of domains (including JASMIN, DKRZ etc) - what do you think?

The same could be said about writing the output data of a recipe run, so I'm not too worried about it. Filling up your home directory or group workspace with downloaded data is no different than with any other data. There might be a few mistakes while the feature is still new, but it should be fine after a while. I have added a note to the documentation that the tool may download a lot of data if you're not careful and when you run it, it will also clearly indicate how much data will need to be downloaded before doing it.

> :beer:

:cheers:

> - can you pls output filesizes with 1/10 MB precision so we don't stdout this or that file with 0 MB size?
- can you pls change the logger info when it doesn't have to download anything from "Will download 0.0 GB" to something like "All needed files found locally, will not download anything"
- remove "Successfully downloaded all requested files." when there was no download happening
- "Downloaded 0 GB in 0:00:06.131342" - maybe trim the sub-millisecond precision, this is not Formula 1 laughing

Good points, I will improve this.

> Very important (and am sure it's in the docs, but am too lazy to check): how are data versions treated? You downloading off the first vailable server or you checking cross-servers which one's got the latest data version?

It searches the entire ESGF for the latest version of the data and will then try to download from the fastest server first. If that fails, it will go to the second, third, etc, until the data is successfully downloaded or there are no servers with the latest version of the data remaining and you'll get an error.

> here's an idea - will Javi's mega-PR that does the OBS data automation allow for automatic download of OBS data off the nodes where CMORized OBS data is stored? Sorry @jvegasbsc that PR faded to black to me, being in the pipes for so long grin I think not though, since there is no actual outside world access to GWS's like esmeval on JASMIN - we should link up with the pyesgf folk and make that data downloadable in the same manner as CMIP data is - like here, in this PR.

Awesome idea. Unfortunately this would only be allowed by the license for Tier2 datasets. The ESGF holds the data for the obs4MIPs dataset, which serves exactly this purpose. I'm not sure how easy it is to add data to obs4MIPs, but if you could find out and have time to actually add the data, that would be great!

> Also, is CORDEX data accessible this way too?

Yes, this pull request makes it possible to automatically download CMIP3, CMIP5, CMIP6, CORDEX, and obs4MIPs data.

> a faulty CMIP/obs4MIPS root path in the config-user file (user mistakes, or forgets to uncomment out the correct path) automatically leads to the switch to downloads -> not optimal, we need to perform a check, or two, and use downloading only if necessary

It would be great if we could add a check if you're running on a known cluster and automatically use the correct rootpaths and DRSs for that cluster. However, I feel that is something for another pull request. I think most Jasmin or Mistral users will find out soon enough that they've made a mistake when the tool tells them it needs to download 500 GB of data and their recipe takes a day to run.

> file downloading as it is implemented now doesn't work on nopw (no parallel write) FS's, as is the case for JASMIN nopw partitions - it hangs, and after a keyboard interrupt you can see the issue with the parallel bit:
> of course, one can output the downloads to a dir on JASMIN user's $HOME which is parallel-enabled, but remember the issues with max quota of 100GB there and how easily that can be filled up by using the slightest wrong config-user file
> EDIT: I did manage to download all data on a nopw FS once but it hung again when I tried again (after removing the data dir, of course), it's not clear to me what's happening

I'm not sure why this is happening. A file is downloaded by a single thread writing to a single file, the same as in the other parts of the tool (e.g. preprocessor). [As a workaround, you could download the data to /work/scratch-pw](https://help.jasmin.ac.uk/article/176-storage) and then `rsync` it to a group workspace if you want to keep it after the run.

> Why is it hitting us now since there are a few other modules that import yaml? Is it a New issue for Codacy and it's not checking the older implements? That should be raised with Codacy, since they're using silly old Python 3.8 +1

I suspect you will see this message also if you change something about other modules that import yaml. This is a relatively [new feature in mypy](https://mypy-lang.blogspot.com/2021/05/the-upcoming-switch-to-modular-typeshed.html). Would you have time to contact Codacy about this?

zklaus:
This sounds like a very useful feature, but it can quickly be about massive amounts of data. For an installation of somebody at their own computer that is kind of their own problem, but on HPC imho we really would like to have this **not** as default. It's not just about filling up discs which, as you pointed out, is a rather normal nuisance, but about data reuse. At SMHI, we certainly want to download ESGF through our own system so that it is available to everyone then. Similarly, I expect that CEDA doesn't want a lot of shadow replica popping up on the Jasmin discs.

So I would keep the default to not downloading things, perhaps it could also be a config file option in the machine-specific section.

esmvalcore/esgf/_download.py
```diff
@@ -0,0 +1,450 @@
...
+            # cmip5.output1.ICHEC.EC-EARTH.historical
+            # .mon.atmos.Amon.r1i1p1.v20121115
+            variable = file.name.split('_')[0]
+            if 'variable' in facets and facets['variable'] == variable:
+                files.append(file)
+            else:
+                logger.warning(
+                    "Ignoring file(s) %s containing wrong variable '%s' in"
+                    " found in search for variable '%s'", file.urls, variable,
+                    facets['variable'])
```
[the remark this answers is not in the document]
bouweandela: It's not very frequent, but you're right, I'll make it a debug message instead. Somehow these things always seem more important when you're developing them than they might be for the end user.

esmvalcore/esgf/_download.py
```diff
@@ -0,0 +1,450 @@
...
+            logger.warning(
+                "No checksum available, unable to check data"
+                " integrity for %s, ", url)
+        else:
+            local_checksum = hasher.hexdigest()
+            if local_checksum != checksum:
+                raise DownloadError(
+                    f"Wrong {checksum_type} checksum for file {tmp_file},"
+                    f" downloaded from {url}: expected {checksum}, but got"
+                    f" {local_checksum}. Try downloading the file again.")
```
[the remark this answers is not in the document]
bouweandela: Maybe in a follow up pull request, for now this should be enough. If you want another try, you can just re-run the recpie.

esmvalcore/esgf/_download.py
```diff
@@ -0,0 +1,450 @@
...
+    def _download(file: ESGFFile):
+        """Download file to dest_folder."""
+        file.download(dest_folder)
+
+    total_size = 0
+    start_time = datetime.datetime.now()
+
+    errors = []
+    random.shuffle(files)
+    with concurrent.futures.ThreadPoolExecutor(max_workers=n_jobs) as executor:
```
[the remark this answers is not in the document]
bouweandela: I'm not sure what the best number is, only experience can tell. However, I think that 4 is reasonable because even if you're downloading 4 files from the same host, you're probably not asking for too much data at the same time. In the future we could improve this, for example: download 2 files at the same time per host, so use (number of hosts * 2) threads.

esmvalcore/esgf/_download.py
```diff
@@ -0,0 +1,450 @@
...
+        for future in concurrent.futures.as_completed(future_to_file):
+            file = future_to_file[future]
+            try:
+                future.result()
+            except DownloadError as error:
+                logger.error("Failed to download %s, error message %s", file,
+                             error)
+                errors.append(error)
+            else:
+                total_size += file.size
```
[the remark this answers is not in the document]
bouweandela: A recursion limit seems unlikely here, because there is no recursion taking place. There is already a timeout on the downloads, currently set to time out if the hosts that you're downloading from does not respond 5 minutes:
https://github.com/ESMValGroup/ESMValCore/blob/72eee4d1f4ade5bdbe534af04326765a26b42240/esmvalcore/esgf/_download.py#L25-L26

esmvalcore/esgf/_search.py
```diff
@@ -0,0 +1,242 @@
...
+        return (dataset, file.name)
+
+    files = sorted(files, key=same_file)
+    for _, versions in itertools.groupby(files, key=same_file):
+        versions = sorted(versions, reverse=True)
+        latest_version = versions[0]
+        result.append(latest_version)
+        if len(versions) > 1:
+            logger.debug("Only using the latest version %s, not %s",
+                         latest_version, versions[1:])
```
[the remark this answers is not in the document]
bouweandela: By default it is all ESGF nodes, but it is possible to configure the search by setting `distrib: false` in the search configuration, see https://esmvaltool--1217.org.readthedocs.build/projects/ESMValCore/en/1217/quickstart/configure.html#search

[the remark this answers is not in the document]
bouweandela: The code here avoids the problem that some files are incorrectly marked as latest on ESGF, so you'll find multiple versions of the same file/dataset even if you specify that you only want the latest version.

esmvalcore/esgf/_download.py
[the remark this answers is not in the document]
bouweandela: Changed now

esmvalcore/esgf/_download.py
```diff
@@ -0,0 +1,450 @@
...
+            local_checksum = hasher.hexdigest()
+            if local_checksum != checksum:
+                raise DownloadError(
+                    f"Wrong {checksum_type} checksum for file {tmp_file},"
+                    f" downloaded from {url}: expected {checksum}, but got"
+                    f" {local_checksum}. Try downloading the file again.")
+
+        shutil.move(tmp_file, local_file)
+        log_speed(url, self.size, duration.total_seconds())
+        logger.info("Downloaded %s (%.0f MB) in %s (%.1f MB/s) from %s",
```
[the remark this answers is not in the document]
bouweandela: Made the log message nicer using the `humanfriendly` package

esmvalcore/esgf/_download.py
```diff
@@ -0,0 +1,450 @@
...
+
+def get_download_message(files):
+    """Create a log message describing what will be downloaded."""
+    megabyte = 2**20
+    gigabyte = 2**30
+    total_size = 0
+    lines = []
+    for file in files:
+        total_size += file.size
+        lines.append(f"{file.size / megabyte:.0f} MB" "\t" f"{file}")
```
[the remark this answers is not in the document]
bouweandela: Made the log message nicer using the `humanfriendly` package

bouweandela:
@zklaus I added the offline option to the default DRS settings for Jasmin and Mistral. Unfortunately, I did not see a site specific entry in config-user for SMHI, so I was not able to add the note for SMHI. I agree that centrally managed downloads are really useful, provided that the administrators are responsive (e.g. we never got a response from the folks at DKRZ when we requested data required to run this recipe https://github.com/ESMValGroup/ESMValTool/pull/2226#issuecomment-879114437 here) and that you plan to re-use the data regularly.

I would really prefer to make downloading the missing data the default option, because I think this will be a major step forward in making the tool more accessible to new users.

If you want to enforce a certain default at specific sites, we could also look into providing a central site-specific default configuration file that administrators could put in an `/etc/esmvaltool` directory?

zklaus:
I see your point, and I think in many circumstances this is fine. I am just thinking of some of our massive recipes that would download dozens of models that could generate not only large local data, but also significant traffic at the esgf nodes. These recipes will tend to be among the interesting ones for newcomers because they give an overview of, say, CMIP6.

Perhaps we can have a threshold for the amount of data (perhaps in terms of number of global 2d fields) below which it's downloaded by default?

bouweandela:
I think that in practice the size of a user's home directory will already be this limit, e.g. the 100 GB that you're allowed to use on Jasmin or the 24 GB on Mistral. After that people will realize they need another solution and ask their system administrators to download the data to the central storage.

But if you're all really concerned that someone might accidentally download a petabyte of data and take the entire ESGF out I can add a limit.

remi-kazeroni (commented):
Great work @bouweandela, thanks! This is a very useful feature and improvement! I tested the download data option for various recipes, reran them, change the time range and everything ran smoothly :+1 
I don't have much to add to the previous reviews.

Regarding the default value of the `--offline` option, I think both possibilities are arguable. If the machine specific entries of the `config-user.yml` are used correctly, the automatic download procedure shouldn't download too much data. But that can happen as soon as there is a typo in a `drs`. A possible suggestion to warn users that they have downloaded lots of data would be to duplicate the message `INFO    [32566] Downloaded 126.61 MB in 25.69 seconds (4.9 MB/s)` at the end of the log file:
```
INFO    [32566] Ending the Earth System Model Evaluation Tool v2.3.1 at time: 2021-08-31 10:53:19 UTC
INFO    [32566] Time for running the recipe was: 0:01:20.590477
INFO    [32566] Maximum memory used (estimate): 0.8 GB
INFO    [32566] Downloaded 126.61 MB
```
but maybe it is already fine as it is now.

I noticed that the stderr contains a lengthy error message when datasets are downloaded. This is not diplayed in the `main_log*txt`. I attached the eror logfile obtained with the command: `esmvaltool run --config-file config-user.yml examples/recipe_python.yml 2> err.txt`.
[err.txt](https://github.com/ESMValGroup/ESMValCore/files/7084075/err.txt)

Finally, I haven't been able to use keyring to store my ESGF credentials. I have an OpenID and password for [esgf-data.dkrz.de](https://esgf-data.dkrz.de/projects/esgf-dkrz/) which work to login to the web service. When using `keyring set ESGF hostname` or `keyring set ESGF username`, I'm prompted for my password and always get an error after typing it. Not sure what I'm doing wrong. Note that using the `logon` section of the `~/.esmvaltool/esgf-pyclient.yml` work fine for me.

bouweandela:
Thanks for reviewing @remi-kazeroni!

> I noticed that the stderr contains a lengthy error message when datasets are downloaded. This is not diplayed in the main_log*txt. I attached the eror logfile obtained with the command: esmvaltool run --config-file config-user.yml examples/recipe_python.yml 2> err.txt.
err.txt

That a mistake in the log message that I introduced here https://github.com/ESMValGroup/ESMValCore/pull/1217/commits/91c3ed08bd505b14859acf04869c23175bbc541b and fixed here https://github.com/ESMValGroup/ESMValCore/pull/1217/commits/91c3ed08bd505b14859acf04869c23175bbc541b

> Finally, I haven't been able to use keyring to store my ESGF credentials. 

Did you try on Mistral or on your own laptop? Because on Mistral there is no suitable keyring backend installed, so maybe that could explain it?

remi-kazeroni:
> > I noticed that the stderr contains a lengthy error message when datasets are downloaded. This is not diplayed in the main_log*txt. I attached the eror logfile obtained with the command: esmvaltool run --config-file config-user.yml examples/recipe_python.yml 2> err.txt.
> > err.txt
> 
> That probably the mistake in the log message that I introduced in [91c3ed0](https://github.com/ESMValGroup/ESMValCore/commit/91c3ed08bd505b14859acf04869c23175bbc541b) and fixed in [671ccb0](https://github.com/ESMValGroup/ESMValCore/commit/671ccb0a359094b11c82ec66c35b477a14edbac5)
> 
My bad, I hadn't fetched the last commit. I have tried again and the issue is indeed solved.

> > Finally, I haven't been able to use keyring to store my ESGF credentials.
> 
> Did you try on Mistral or on your own laptop? Because on Mistral there is no suitable keyring backend installed, so maybe that could explain it?

Yes I tried on Mistral only. Let's hope such things will work better on the new machine at the end of the year.

bouweandela:
>  If the machine specific entries of the config-user.yml are used correctly, the automatic download procedure shouldn't download too much data. But that can happen as soon as there is a typo in a drs.

Typos in the `drs` might be a bit hard to catch, but I just added a warning if a rootpath is set to a directory that does not exist. That should help for cases where someone forgot or didn't know they needed to configure that.

remi-kazeroni (approved):
Thanks for all the work @bouweandela! If I understand well [this comment](https://github.com/ESMValGroup/ESMValCore/pull/1217#issuecomment-902574219), there is not much hope to pass the Codacy check for this PR. You seem to have addressed all the comments from the reviews so I can give a pre-emptive approval.

esmvalcore/esgf/_download.py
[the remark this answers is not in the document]
valeriupredoi: cool mule :+1:

esmvalcore/esgf/_download.py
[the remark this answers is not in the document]
valeriupredoi: bah humbug, after having downloaded all those files and stumbled only on an unresponsive node for exactly 5s :grin: OK fine, I guess we should think about it since 1. it'll spend a bit of time downloading, then it'll spend a bit of time finding all those files at rerun, nit picky users will be complaining

esmvalcore/esgf/_download.py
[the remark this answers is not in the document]
valeriupredoi: nice, do they have the `doggofriendly` package too? Hope that's not another dependency to be added to the deps list (yeah I sound like Scrooge McDuck :grin: )

esmvalcore/esgf/_download.py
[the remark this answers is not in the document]
valeriupredoi: OKm sounds like a reasonable number then, here's an idea for superfast downloads http://eventlet.net/ - it could be that the bottleneck is the file writing anyway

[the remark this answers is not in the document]
valeriupredoi: here's an example of a [web crawler](http://eventlet.net/doc/examples.html#producer-consumer-recursive-web-crawler)

esmvalcore/esgf/_download.py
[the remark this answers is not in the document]
valeriupredoi: thanks man, I looked at the code again, the `ESGFFile` object looks fine and looks to play will with concurrent futures - btw you have no idea how many times you may hit the max recursion depth issue even if there is no obvious recursion in sight :grin: Cool about the timer, missed that one! But 5min is prob too long don't you think?

esmvalcore/esgf/_search.py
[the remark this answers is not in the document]
valeriupredoi: yes that would be a major PITA, you just made me remember of it only now after reading your explanation - can one select a certain version? Am getting on yer nerves now, I know :grin:

tests/unit/preprocessor/_weighting/test_weighting_landsea_fraction.py
```diff
@@ -1,6 +1,7 @@
 """Unit tests for :mod:`esmvalcore.preprocessor._weighting`."""
 
 import iris
+import iris.fileformats
```
valeriupredoi, line 4: is this only for the new `iris=3.1.0`?

bouweandela: This is needed for the tests to work with iris 3.1, see [here](https://scitools-iris.readthedocs.io/en/v3.1.0/whatsnew/3.1.html#minimised-imports).

valeriupredoi: this may need to propagate to other tests and/or other modules we have? Agh, will have to give it a one off proper with iris 3.1 next week, we need to drop `pyke` too

valeriupredoi (approved):
mega PR with mega functionality :medal_military: @bouweandela

tests/unit/preprocessor/_weighting/test_weighting_landsea_fraction.py
bouweandela: All tests are passing now, so i guess it's only needed here

bouweandela:
> sys admins keep complaining they need to buy more physical disks

@valeriupredoi Do you know any of these sys admins? They might want to look into file systems that support [data dedeplication](https://en.wikipedia.org/wiki/Data_deduplication).

bouweandela:
We decided to disable the feature by default for now, to give users some time to get used to it. I changed the pull request accordingly. @jvegasbsc Could you please have another look and approve if you agree with the pull request as it is now or let me know what additional changes are required?

tests/unit/preprocessor/_weighting/test_weighting_landsea_fraction.py
zklaus: This shouldn't go into or affect this PR, but probably we should use `iris.analysis.cartography.DEFAULT_SPHERICAL_EARTH_RADIUS` instead of `iris.fileformats.pp.EARTH_RADIUS` anyway.

doc/quickstart/configure.rst
```diff
@@ -161,6 +180,199 @@ the user.
...
+    keyring set ESGF username
+    keyring set ESGF password
+
+for example
+
+.. code-block:: bash
+
+    keyring set ESGF esgf-data.dkrz.de
```
jvegasbsc, line 253: This should be like this, no?

```suggestion
    keyring set ESGF hostname esgf-data.dkrz.de
```

bouweandela: Yes, thanks, I'll change it!

zklaus: I tried it, but I got
```
$ keyring set ESGF hostname esgf-data.dkrz.de
usage: keyring [-h] [-p KEYRING_PATH] [-b KEYRING_BACKEND] [--list-backends] [--disable] [operation] [service] [username]
keyring: error: unrecognized arguments: esgf-data.dkrz.de
```

Any idea what went wrong?

remi-kazeroni: Could it be related to the current shutdown at DKRZ?

> The shutdown will start on Monday September 20th at 06:00am and we expect Mistral back on Wednesday 22nd in the afternoon. During the downtime no login is possible to9 Mistral, Eddy, ClimXtreme (miklip). Also other services like CERA and ESGF are affected by this downtime.

zklaus: No. This is a purely local change in configuration. I don't even think that keyring considers whether this is a hostname.

bouweandela: Indeed, you need to run `keyring set ESGF hostname` and then enter the hostname as the password. I'll update the documentation.

bouweandela: Sorry for the confusion, I fixed the example text. Fingers crossed that it works!

zklaus: Seems to work. Thanks!

jvegasbsc (approved):
Nice work Bouwe!

bouweandela:
Thanks for reviewing everyone!

valeriupredoi:
and big thanks to you Bouwe for implementing such a great weapon in the war against "oh I haz no data, can't run ESMValTool" :grin:

zklaus:
Very nice indeed! This may have introduced a slight regression, though. @bouweandela, could you please have a look at https://app.circleci.com/pipelines/github/ESMValGroup/ESMValCore/5681/workflows/2f674327-06e8-4cfc-b6e5-434ddade9eeb/jobs/27327?invite=true#step-103-254 ?

valeriupredoi:
good catch @zklaus - I ran that test locally to make sure the fail is not due to the SegFault crashing the CI worker, it does indeed fail with the said KeyError. On top of that 99% of my tests in that test module failed for me with `AttributeError: module 'iris' has no attribute 'cube'` so it looks we'll have to do the import dance after all

bouweandela:
That looks like the usual segfault or am I overlooking something? I'll re-run and see if there is still an error.

bouweandela:
It looks like the obs4MIPs rename is causing some collision with #1032, I'll make a pull request to complete the renaming.

valeriupredoi:
sadly no, it's a valid failed test. Also @bouweandela try running `pytest tests/unit/cmor/test_cmor_check.py` on yer machine and see all the issues arising from not explicitly importing `iris.cube` - it looks like the new iris feature has made us unable to run individual tests in these cases, without explicit imports from iris submodules :bomb:

bouweandela:
Created #1328

PR #1233, opened 2021-07-16
Add a more friendly and useful message when using default config file

valeriupredoi:
## Description

When there is no user-specified config-user file, the info stdout says Using config file None, which is both untrue and user confusing, even though the actual value is indeed `None`

Closes https://github.com/ESMValGroup/ESMValTool/issues/2235

Link to documentation:

***

***

### DIFF
--- a/esmvalcore/config-user.yml
+++ b/esmvalcore/config-user.yml
@@ -36,19 +36,21 @@ profile_diagnostic: false
 # Rootpaths to the data from different projects (lists are also possible)
 # these are generic entries to better allow you to enter your own
 # For site-specific entries, see below
-# rootpath:
-#   CMIP5: [~/cmip5_inputpath1, ~/cmip5_inputpath2]
-#   OBS: ~/obs_inputpath
-#   RAWOBS: ~/rawobs_inputpath
-#   default: ~/default_inputpath
-#   CORDEX: ~/default_inputpath
+# Comment out these when using a site-specific path
+rootpath:
+  CMIP5: [~/cmip5_inputpath1, ~/cmip5_inputpath2]
+  OBS: ~/obs_inputpath
+  RAWOBS: ~/rawobs_inputpath
+  default: ~/default_inputpath
+  CORDEX: ~/default_inputpath
 
 # Directory structure for input data: [default]/BADC/DKRZ/ETHZ/etc
 # See config-developer.yml for definitions.
-# drs:
-#   CMIP5: default
-#   CORDEX: default
-#   OBS: default
+# comment out/replace as per needed
+drs:
+  CMIP5: default
+  CORDEX: default
+  OBS: default
 
 # Site-specific entries: Jasmin
 # Uncomment the lines below to locate data on JASMIN
--- a/tests/integration/test_main.py
+++ b/tests/integration/test_main.py
@@ -13,6 +13,7 @@ from unittest.mock import patch
 import pytest
 from fire.core import FireExit
 
+import esmvalcore._recipe_checks as check
 from esmvalcore._main import Config, ESMValTool, Recipes, run
 
 
@@ -58,8 +59,8 @@ def test_run():
         run()
 
 
-def test_actual_run(tmp_path):
-    """Test run from command line in full."""
+def test_empty_run(tmp_path):
+    """Test real run with no diags."""
     recipe_file = tmp_path / "recipe.yml"
     content = dedent("""
         documentation:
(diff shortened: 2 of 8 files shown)


valeriupredoi:
OMG OMG the test coverage decreases by 0.01% :open_mouth: :rofl: I like this functionality a lot but is there any way we can set thresholds above which we accept the codecov test to pass? @bouweandela

zklaus:
Why not just improve the coverage?

jvegasbsc:
Nice touch V! I changed it a bit so we keep the actual path to the config file in the config dictionary, so we can log it directly.

I think is clearer that way and we also do not have to modify the message in the (unlikely) case that we modify the default path

valeriupredoi:
aweome, cheers Javi! Klaus, I refuse to write tests for print statements, just kidding, I should actually :grin:

zklaus:
I mean, the print statement is probably part of a function that does something. As such, it should be covered by tests because that function is covered by tests.

valeriupredoi:
@jvegasbsc I've added a (rather peasant-y) test that actually calls and runs `esmvaltool` - it occured to me that a lot of the `_main.py` functionality is not tested unless one runs the thing by calling the executable. I don't like the fact that I'm writing to cwd and then removing the output, but I can't think of any better option without passing and editing a config file. Is there any way to direct the execution straight to `tmp_dir`? BTW coverage increases by a hefty 30%

valeriupredoi:
@bouweandela we need to fix this Codecov thingie, man, the latest code changes improve test coverage for `_main.py` by 35% but the tool fails still? And caches the old result too, it's confusing

valeriupredoi:
@jvegasbsc careful that now there are two R test diags that fail in `tests/integration/test_diagnostic_run.py`

jvegasbsc:
I have changed the run test to do not overwrite the default config user and use one in the tmp path and added a new test for the lines that codecov is claiming that now are not covered.

@valeriupredoi, probably you should move the codecov config changes to its own pull request

jvegasbsc:
I messed a bit too much with this PR to be an independent reviewer. @sloosvel, can you take a look?

valeriupredoi:
cheers @jvegasbsc - I took out the codecov changes to the config and cleaned up your bad key that was making the diag test run fail :grin: Heads up that the errors reported by `esmvalcore/experimental/config/_validated_config.py` during the validation checks (at around lines 40-43) are obscure, we really need a more user-friendly error report from there :+1:

jvegasbsc:
> Why not just improve the coverage?

We finally got quite an improvement

valeriupredoi:
> > Why not just improve the coverage?
> 
> We finally got quite an improvement

yesh we did - it is biutiful :grin:

sloosvel (approved):
Seems to work fine

valeriupredoi:
cheers muchly @sloosvel - @jvegasbsc if you don't want to add more then you can just merge it, mate, I would but that'd be a bit biased :grin:

jvegasbsc:
The conda build is failing due to timeout, but that's not our fault (see #1243)

tests/integration/test_main.py
```diff
@@ -53,13 +54,41 @@ def test_version():
...
+    """)
+    recipe_file.write_text(content)
+    Config.get_config_user(path=tmp_path)
+    with pytest.raises(check.RecipeError) as exc:
+        ESMValTool.run(recipe_file, config_file=f"{tmp_path}/config-user.yml")
+    assert str(exc.value) == 'The given recipe does not have any diagnostic.'
+    log_dir = './esmvaltool_output'
+    log_file = os.path.join(log_dir,
+                            os.listdir(log_dir)[0], 'run', 'main_log.txt')
+    os.system("rm -r esmvaltool_output")
```
bouweandela, line 85: This looks very unsafe, why not use tmp_path for the output and let pytest clean up?

esmvalcore/config-user.yml
```diff
@@ -36,19 +36,21 @@ profile_diagnostic: false
...
 # these are generic entries to better allow you to enter your own
 # For site-specific entries, see below
-# rootpath:
-#   CMIP5: [~/cmip5_inputpath1, ~/cmip5_inputpath2]
-#   OBS: ~/obs_inputpath
-#   RAWOBS: ~/rawobs_inputpath
-#   default: ~/default_inputpath
-#   CORDEX: ~/default_inputpath
+# Comment out these when using a site-specific path
+rootpath:
```
bouweandela, line 40: This was commented out because we got the complaint from users that these default settings do not work ever for anyone.

PR #1240, opened 2021-07-20
Test dtype in time preproc

jvegreg:
This is a follow up from #1237.

It add checks for dtype conservation on time preprocessor tests

Closes #

## [Before you get started](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#getting-started)

- [ ] [☝ Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do

## [Checklist](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#checklist-for-pull-requests)

It is the responsibility of the author to make sure the pull request is ready to review. The icons indicate whether the item will be subject to the [🛠 Technical][1] or [🧪 Scientific][2] review.

[1]: https://docs.esmvaltool.org/en/latest/community/review.html#technical-review
[2]: https://docs.esmvaltool.org/en/latest/community/review.html#scientific-review

- [ ] [🧪][2] The new functionality is [relevant and scientifically sound](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#scientific-relevance)
- [ ] [🛠][1] This pull request has a [descriptive title and labels](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-title-and-label)
- [ ] [🛠][1] Code is written according to the [code quality guidelines](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#code-quality)
- [ ] [🧪][2] and [🛠][1] [Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available
- [ ] [🛠][1] [Unit tests](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#tests) have been added
- [ ] [🛠][1] Changes are [backward compatible](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#backward-compatibility)
- [ ] [🛠][1] Any changed [dependencies have been added or removed](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#dependencies) correctly
- [ ] [🛠][1] The [list of authors](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#list-of-authors) is up to date
- [ ] [🛠][1] All [checks below this pull request](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-checks) were successful

***

[...]

PR #1241, opened 2021-07-20
Tweak codecov configuration

### LINKED ISSUE #1234: Codecov doesn't work as expected
codecov runs once then comments (attaching the graph) for the first commit then keeps reporting the same issue (if one is found), if the issue is fixed and the coverage is actually increased via subsequent commits, it still reports the same first issue and fails to regenerate both the comment with updated graph and not to fail - see https://github.com/ESMValGroup/ESMValCore/pull/1233 - I fixed the codecov.yml file so we don't have the codecov as failed test (we don't want a failure perse but rather as information) and I asked codecov to post a comment only if the coverage changes, but this last directive doesn't really work since it still shows the result from the very first commit.

valeriupredoi:
## Description

Make codecov more robust:

- should not fail as a test, whatever the outcome of the coverage delta
- should update results everytime there is a change of coverage and not cache old results and stop displaying them
- should set a theshold for which it shouldn't even bother reporting
- should update the report only if there's a change (and delta is above the threshold)
- use `project` too?

Closes #1234 

Link to documentation:

***

***

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,4 +3,7 @@ coverage:
     project: off
     patch:
       default:
+        informational: true
         target: 100%
+comment:
+  require_changes: true  # comment only if the coverage changes


jvegasbsc:
>   should not fail as a test, whatever the outcome of the coverage delta

I like it being a failed test, as our first reaction will be to fix it. In some cases, we may decide to go on with merging while failing, but that should be discussed and approved in the PR. 

 >   should not post in-line comments in PR code, makes review very frustrating

Completely agree: its very intrusive and not very useful. Codecov page is way better to see this

 >   should update results everytime there is a change of coverage and not cache old results, and should stop displaying them, if cached (needs to refresh for every commit if that changes coverage)

Completely agree

>  should set a theshold for which it shouldn't even bother reporting

I don't think so. There are some critical features in the code that are implemented as one-liners. Setting any threshold below 100% will be unsafe. As with the failed test behaviour, it should be our call to let a PR merge with less than 100%

valeriupredoi:
good points @jvegasbsc -> totally agree with you! I am trying to see if the changes I made reflect these points

zklaus:
> > should not fail as a test, whatever the outcome of the coverage delta
> 
> I like it being a failed test, as our first reaction will be to fix it. In some cases, we may decide to go on with merging while failing, but that should be discussed and approved in the PR.

:+1: I like the failing too.


> > should not post in-line comments in PR code, makes review very frustrating
>
> Completely agree: its very intrusive and not very useful. Codecov page is way better to see this

Hmm. Not so sure. Can a PR really be considered ready for review while there are still lines that are not covered by tests? Probably not, right? I find it quite handy as a PR author to quickly see where I need to improve the tests.

> > should update results everytime there is a change of coverage and not cache old results, and should stop displaying them, if cached (needs to refresh for every commit if that changes coverage)
> 
> Completely agree

Yeah, of course. Haven't seen that behavior yet, but probably just need more experience with codecov.

> > should set a threshold for which it shouldn't even bother reporting
> 
> I don't think so. There are some critical features in the code that are implemented as one-liners. Setting any threshold below 100% will be unsafe. As with the failed test behaviour, it should be our call to let a PR merge with less than 100%

Agree. There is no bound.

valeriupredoi:
OK that option to post only if there are changes seems to work fine, same as with the inline comments being deactivated :smiley:

valeriupredoi:
> Hmm. Not so sure. Can a PR really be considered ready for review while there are still lines that are not covered by tests? Probably not, right? I find it quite handy as a PR author to quickly see where I need to improve the tests.

@zklaus no, not really, I'd argue one needs to fix the coverage, but not in ESMValTool - if we go with that there we'd get nowhere with all the diagnostics not being tested (not even the Python ones), but here, we should be strict(ish) :gun:

zklaus:
Good point about the tool. So let's keep the inline comments here in the core, but not in the tool.

jvegasbsc:
> Good point about the tool. So let's keep the inline comments here in the core, but not in the tool.

Keep them in both. Those are not comments, but annotations and can be hidden quite easily: https://twitter.com/natfriedman/status/1366417484698578953?s=20

valeriupredoi:
ok gents, I left the annotations in code in then, Javi make up your mind :laughing:

valeriupredoi:
it'd help if @bouweandela could haz a look too, after all he's the one who broght to us this marvelous tool :grin:

codecov.yml
```diff
@@ -1,6 +1,12 @@
+# github_checks:
+#     annotations: false  # remove annoying inline code comments
```
bouweandela, line 2: Did we want to keep the annotations? I have not seen them in action yet, could you point me to a pull request where they are shown?

codecov.yml
```diff
@@ -1,6 +1,12 @@
...
 coverage:
   status:
     project: off
     patch:
       default:
+        # informational: true  # turn this one on to get cov tests to always pass
```
bouweandela, line 8: I agree with @jvegasbsc and @zklaus that we should keep the failed test, it already led to a bit improvement in coverage in #1233.

codecov.yml
```diff
@@ -1,6 +1,12 @@
...
         target: 100%
+comment:
+  behavior: new  # delete old comment and post new
+  require_changes: true  # comment only if the coverage changes
```
bouweandela, line 12: I would prefer to always see the comment, so that I know that the coverage has been checked and is OK when I'm reviewing something. Then I can immediately see that I do not need to worry about that and I do not have to go through the list of changed files to check that indeed no code was changed or maybe code was changed but codecov integration is somehow broken?

codecov.yml
valeriupredoi: here's one https://github.com/ESMValGroup/ESMValCore/pull/1240/files - you can press "a" to toggle them on/off

codecov.yml
valeriupredoi: OK I agree too, after convincing myself, with help from you guys :laughing: That's whay I commented out the lines; BTW I think a few of these commented out lines and in-line comments with what they do are useful in the future if we want to re-activate/de-activate options

codecov.yml
valeriupredoi: OK, I find the coverage map and table confusing and distracting (not really well thought as a dashboard), and I'd like to see it only if the change I made introduced a change in the coverage, up to you guys

zklaus: I kind of agree with you, @valeriupredoi, in that I also have not quite wrapped my head around that map. But I would prefer to keep it for a bit just to be exposed to it to see if I learn something from it.

codecov.yml
bouweandela: Thanks! @ESMValGroup/esmvaltool-coreteam Anyone an opinion on these type of comments by codecov? Are they helpful or too much? My feeling is that they are a bit too much in your face and that it may not be obvious to everyone that it is easy to hide them when reviewing.

codecov.yml
bouweandela: Great! Personally, I'm not a big fan of leaving commented-out stuff everywhere, it looks distracting to me, like the author was working on something and then got the code merged without cleaning up afterwards.

codecov.yml
zklaus: Just in case you missed the discussion on the topic above: @jvegasbsc and I expressed support for keeping the annotations. My main argument is that they shouldn't interfere with review, because at review stage there should be no complaints about uncovered lines in the diff left. As an author, they are really helpful to easily see where the tests need to be improved before asking for review. And if they disturb you too much, just press `a`.

codecov.yml
valeriupredoi: OK sounds good, am gonna comment that line out as well, this PR is getting more extinct than Manny the mammoth :rofl:

codecov.yml
valeriupredoi: yeah pressing the key "a" does the trick, unless of course you have a French keyboard, in which case that'll mean Revolution :fr:

PR #1259, opened 2021-07-30
Move institutes from config-developer.yml to default extra facets config and add wildcard support for extra facets

bouweandela:
## Description

Simplify config-developer.yml by moving the institutes to the default extra facets configuration.

This change is not backwards compatible if you are using a custom config-developer.yml file to define additional institutes. You will have to move the institutes to a new [extra facets configuration file](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/quickstart/configure.html#extra-facets) in that case.

***

***


bouweandela:
@zklaus Do you think extra documentation is needed for this pull request?

bouweandela:
The documentation build is failing because the scipy documentation appears to be offline. All other tests are green now.

valeriupredoi:
@bouweandela why are we doing this? The config-developer is a solid place for all that jazz, and it's up there in the directory tree, rather than this info buried in three layers of dir structure. It is defo more organized this way, but names like `extra_facets` and information buried in a few subdirs is not exactly user-friendly, given that this info gets edited by others than just us

bouweandela:
@valeriupredoi The extra facets facility provides a standardized way to automatically add extra key-value pairs to datasets defined in the recipe, therefore it makes sense to use it everywhere we do this. This feature is well documented and easily accessible from config-user.yml, see [here](https://esmvaltool--1259.org.readthedocs.build/projects/ESMValCore/en/1259/quickstart/configure.html#extra-facets) for the documentation. That the defaults are buried 3 levels deep in the directory tree doesn't really matter, because we would not expect users to change them. The defaults can simply be overwritten by user-defined files if needed.

bouweandela:
@valeriupredoi Does that answer address your concerns? I really think that it would be much more clear to users if we have one way of adding extra facets instead of two or three.

@zklaus Any chance of this getting a review from you, so it can go into the upcoming release if @valeriupredoi agrees?

valeriupredoi:
cheers @bouweandela - it do make sense yes, sorry I missed yer answer which clears things for me :+1:

valeriupredoi (approved):

doc/quickstart/configure.rst
```diff
@@ -370,10 +370,9 @@ These four items here are named people, references and projects listed in the
 Extra Facets
 ============
 
-Sometimes it is useful to provide extra information for the loading of data,
-particularly in the case of native model data, or observational or other data,
-that generally follows the established standards, but is not part of the big
-supported projects like CMIP, CORDEX, obs4MIPs.
+Sometimes it is useful to automatically add extra key-value pairs to variables
+or datasets in the recipe.
+These key-value pairs are called facets.
```
zklaus, line 375: Perhaps we can find a blend of the old and the new description? I think the opening sentence here should make it immediately clear what this is and why it's useful. I think this is now missing a bit.

esmvalcore/_config/_config.py
```diff
@@ -54,7 +55,19 @@ def _load_extra_facets(project, extra_facets_dir):
...
+        return [pat for pat in patterns if fnmatch.fnmatchcase(name, pat)]
+
+    extra_facets = {}
+    for dataset_ in fnfilter(project_details, dataset):
+        for mip_ in fnfilter(project_details[dataset_], mip):
+            for var in fnfilter(project_details[dataset_], short_name):
+                facets = project_details[dataset_][mip_][var]
+                extra_facets.update(facets)
+
+    return extra_facets
```
zklaus, line 70: If I understand correctly, this adds the ability to use Unix-style filename expansion for keys in the extra facets.

First, I am a bit uneasy about applying something that is clearly meant for filenames and path to something else. I think at a minimum we should have a comment here, stating clearly that, yes this is for filenames, but we use it for something else, and that works fine and is awesome.

Second, I think this extension of the syntax should be documented.

Third, this is sufficiently different from moving the institutes that I would consider doing this in a separate PR. At a minimum, it requires a change to the PR title, IMHO.

zklaus (changes requested):
I really like this; it makes the config-developer file much clearer both in contents and conceptually. There are small open questions around documentation and versioning.

esmvalcore/_config/_config.py
bouweandela: > If I understand correctly, this adds the ability to use Unix-style filename expansion for keys in the extra facets.

Yes

> First, I am a bit uneasy about applying something that is clearly meant for filenames and path to something else. I think at a minimum we should have a comment here, stating clearly that, yes this is for filenames, but we use it for something else, and that works fine and is awesome.

I also considered using a regular expression, but I felt that the wildcards used for file selection would be more familiar to most users. For example, we also allow these for selecting ancestors in the recipe and tasks from the command line with the `--diagnostics` flag.

> Second, I think this extension of the syntax should be documented.

Good point, I'll work on improving it

> Third, this is sufficiently different from moving the institutes that I would consider doing this in a separate PR. At a minimum, it requires a change to the PR title, IMHO.

I really need a feature like this here, to be able to specify that the institute should be added to any mip and any variable. I'll update the pull request title and description.

zklaus: > > First, I am a bit uneasy about applying something that is clearly meant for filenames and path to something else. I think at a minimum we should have a comment here, stating clearly that, yes this is for filenames, but we use it for something else, and that works fine and is awesome.
> I also considered using a regular expression, but I felt that the wildcards used for file selection would be more familiar to most users. For example, we also allow these for selecting ancestors in the recipe and tasks from the command line with the `--diagnostics` flag.

I understand completely. In fact, I was rather surprised that the `fnmatch` docs are so filesystem-centric since it seems such a useful feature elsewhere. That's part of what made me weary. But I'll accept that, would just like it to be a bit more explicit for the benefit of a future maintainer/reader, i.e. rename `fnfilter` to something that doesn't imply `f`ile`n`ame and reword the one-line docstring to mention this. I, for example, needed a few minutes to convince myself that this code was not written with filenames in mind.

> > Second, I think this extension of the syntax should be documented.
> Good point, I'll work on improving it
:+1: 

> > Third, this is sufficiently different from moving the institutes that I would consider doing this in a separate PR. At a minimum, it requires a change to the PR title, IMHO.
> I really need a feature like this here, to be able to specify that the institute should be added to any mip and any variable. I'll update the pull request title and description.

If it was me, I'd probably split it up, get the extended syntax in first, and follow immediately with the institute moving. But it's ok.

bouweandela:
@zklaus Could you have another look at this please? I think I addressed all your comments.

doc/quickstart/configure.rst
```diff
@@ -585,10 +585,11 @@ These four items here are named people, references and projects listed in the
 Extra Facets
 ============
 
-Sometimes it is useful to provide extra information for the loading of data,
-particularly in the case of native model data, or observational or other data,
-that generally follows the established standards, but is not part of the big
-supported projects like CMIP, CORDEX, obs4MIPs.
+It can be useful to automatically add extra key-value pairs to variables
+or datasets in the recipe.
+These key-value pairs can for be used for :ref:`finding data <findingdata>`
```
zklaus, line 590: Duplicate `for`

zklaus (approved):
Just a mini typo. Otherwise, great, thanks!

zklaus:
One more thing: Could you please consider if you want to update the title of this PR and add one or more labels?

bouweandela:
Thanks, I updated the title but I couldn't find a relevant label other than the backwards incompatible one. Did you have a particular label in mind?

PR #1273, opened 2021-08-18
Pin NumPy to `numpy>=1.20` to avoid possible backwards incompatibilities due to change in C API

### LINKED ISSUE #1272: Pin `numpy>=1.20`: `numpy=1.20` has a change in its C API that may trigger a backwards incompatibility error
I'm back, with technical gubbins!

[Error seen and asked about here](https://stackoverflow.com/questions/66060487/valueerror-numpy-ndarray-size-changed-may-indicate-binary-incompatibility-exp) - third answer is the relevant one - this may bite us more frequently than we think since e.g Core may be installed from conda and built with a `numpy<1.20` and then when one installs a new dependency in their conda developemnt environment, conda will automatically update the resident (run) `numpy` to `numpy>=1.20`, then when the user tries running, proverbial stuff hits the fan. There are other scenarios that may lead to this issue, and I suspect they're prone to this especially if Python is old, like 3.7. I suggest we pin `numpy>=1.20` across Core and Tool - do we have any reason why not? Can't think of any myself...

valeriupredoi:
## Description

Closes #1272 

Environment builds fine for Linux and OSX, multiple Python's on [GA](https://github.com/ESMValGroup/ESMValCore/actions/runs/1143441793)

***

***

### DIFF
--- a/.github/workflows/action-install-from-source.yml
+++ b/.github/workflows/action-install-from-source.yml
@@ -22,7 +22,7 @@ on:
   push:
     branches:
     - main
-    - pin_numpy
+
   # run the test only if the PR is to main
   # turn it on if required
   #pull_request:
--- a/.github/workflows/action-install-from-source.yml
+++ b/.github/workflows/action-install-from-source.yml
@@ -22,7 +22,7 @@ on:
   push:
     branches:
     - main
-    - github-actions2
+    - pin_numpy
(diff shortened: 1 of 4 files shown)


zklaus:
As discussed in #1272, I don't think we need to do this. I'll close this PR; if you disagree let's come to an agreement in #1272 and only then return to this.

PR #1281, opened 2021-08-20
Add preprocessor `accumulate_coordinate`

jvegreg:
We are currently missing a preprocessor to change values from time fluxes to aggregated values. This PR does that.

Any CMOR variable provided in the form `X s-1` can be converted to X by using this preprocessor. It also works with `X day-1` and `X hr-1`, but those units are way less common.

***

***


valeriupredoi:
go for it if you've got time, cheers much @sloosvel :beer:

zklaus:
The names flux, total, and aggregation are a bit clunky stand-ins to say time integration. It seems better to align the terminology with existing pre-processors. Perhaps the integration (that already exists at least for depth), can simply be generalized? Basically, an `integration_1d` preprocessor would cover both and more.

sloosvel:
> Perhaps the integration (that already exists at least for depth), can simply be generalized? 

While I think it's a good idea,  that is more than what I can commit to doing right now. The `depth_integration` preproc is implemented quite differently than this one, and unifying them would probably lead to backwards incompatible changes. Would naming this `time_integration` work better for you?

zklaus:
Looking at the two implementations ([here](https://github.com/ESMValGroup/ESMValCore/blob/9d5cceb5ae4db04d09f9432a74d265983f900178/esmvalcore/preprocessor/_volume.py#L278) is the depth_integration for easy reference), I would suggest scrapping this PR, refactoring the depth_integration into integration_1d (the name can be changed of course), and implementing depth_integration as a call to integration_1d.

The reason is that this PR contains some unnecessary hard-coding (e.g. unexplained numerical constants with no reference to the units attribute of the relevant axis). The depth_integration is already almost ready for the generic 1d integration if you replace depth with a flexible coordinate reference. The only thing left todo is sort out the units, but that shouldn't be too hard and will
- fix the depth_integration preprocessor
- add this one
- give us a generic 1d integration.

zklaus:
Well, the quite part that I didn't write out loud was that then, later on, we can improve the integration by moving beyond a simple sum (effectively a coarse Riemann sum), to more sophisticated [integration rules](https://en.wikipedia.org/wiki/Numerical_integration#Methods_for_one-dimensional_integrals). That is far beyond what I'd suggest now but motivated me to keep integration and statistics separate.

jvegreg:
I'm sorry to interrupt the party, but this pull request is not a real integration, as the time coordinate nor any other are changed.

The only change in the cube metadata are the units, from X time-1 to X, but any other aspect of the cube is unchanged. This makes it quite a different beast compared to the `depth_integration` or any of the time aggregators.

sloosvel:
I think any discussion over integration or aggregation can be moved to #1589 . The thing is that this work does not collapse any coordinate, it's just multiplying the data by a factor.

bouweandela:
Thanks for explaining, I misunderstood the algorithm. In that case, I would still recommend using the features provided by iris. For example, by implementing this as follows:
```python
coord_name = 'time'
coord = cube.coord(coord_name)
factor = iris.coords.AuxCoord(
    np.diff(coord.bounds)[..., -1],
    var_name=coord.var_name,
    long_name=coord.long_name,
    units=coord.units,
)
result = cube * factor
unit = result.units.format().split(' ')[-1]
result.convert_units(unit)
result.long_name = f"{cube.long_name} * {factor.long_name}"

```

esmvalcore/preprocessor/_units.py
```diff
@@ -29,3 +30,42 @@ def convert_units(cube, units):
...
+
+    Raises
+    ------
+    ValueError
+        If the units are not supported by the operator
+    """
+    if 's-1' not in cube.units.format():
+        raise ValueError(
+            f'Units {cube.units} do not contain a supported flux definition.'
+            )
```
bouweandela, line 58: This is no longer needed

bouweandela, line 58: This is no longer needed

bouweandela (commented):
Thanks for making the requested changes @sloosvel. Do you have any ideas for a better name for this preprocessor function? Would it make sense to be able to specify the coordinate over which to accumulate?

esmvalcore/preprocessor/_units.py
sloosvel: I am not sure if it makes sense to be able to multiply data with any unit by a factor that comes from any coordinate value (`percentage * degrees_east` for instance). 

The way I see this is that this is a preprocessor that allows to go from a variable representing a rate to a "fixed" quantity by multiplying the data by the unit in the denominator. That is why it works for the time, because the units of the time coordinate can match the units of variables that represent a rate. I don't think there are any variables using `degrees_east` or `degrees_north` as units.

But of course I am not aware of all possible applications.

zklaus: What is the problem you see with `% * degrees_east`? It may be a bit unusual, but I don't see anything wrong with it from a mathematical or physical point of view, hence I see no reason to forbid it; we can leave that decision up to the user.

bouweandela: In the case of longitude, it would probably make more sense to multiply by the cell area, combining latitude and longitude. As I understand it now, this is the same preprocessor as the integration preprocessors, just without the final summation. Does that sound right?

zklaus: I wouldn't think of this as a special case. Rather, let's disallow multi-dimensional coordinates for now and possibly extend this later.

sloosvel: @bouweandela does this last suggestion sound good to you?

bouweandela: Sure, no need to implement everything at the same time

sloosvel: However I can't think of a decent name for this

bouweandela: @ESMValGroup/esmvaltool-developmentteam Anyone a suggestion for a good name for this preprocessor function? At the moment, it can multiply a variable by (1D) weights computed from the coordinate bounds. Right now only for the time coordinate, but this could trivially be extended to any coordinate.

In the future other weights might be possible (maybe cell area, volume, ..) if anyone has a use case and is interested in implementing that, though that could also go into a different preprocessor.

sloosvel: I'll just name it `accumulate_coordinate`. If anyone strongly opposes to that let us know.

sloosvel:
I think I already addressed all the comments, let me know if you have further ones

bouweandela (approved):
Looks fine to me, though I'm still not sure if this is a good name. Better than before though. I was just waiting to give the others a chance to respond to the question about the name.

sloosvel:
>Looks fine to me, though I'm still not sure if this is a good name.

I am not a fan either but I can't come up with anything better. If at any point anyone has a better idea I will update it.

PR #1284, opened 2021-08-23
Run a nightly Github Actions workflow to monitor tests memory per test (configurable for other metrics too)

### LINKED ISSUE #1283: Dodgy process makes the memory used by running tests go up the roof
`pytest -n 2 -m "not installation and not sequential"` may take up more than 4GB as we can see from *some* failed CI tests eg [this one](https://app.circleci.com/pipelines/github/ESMValGroup/ESMValCore/5481/workflows/c652d3f7-3d2a-4afb-8961-3a8fcda97444/jobs/26270) - there is something dodgy that we have to catch...

valeriupredoi:
## Description

Small nightly Github Actions workflow that installs and uses [pytest-monitor](https://github.com/CFMTech/pytest-monitor) as `pytest` plugin to output a `.pymon` sqlite database where performance metrics have been recorded for each pytest test we run (memory, CPU time, usage etc); the wflow runs a wee script I wrote that currently outputs the name of the test and amount of RES memory it used if more than 1000 MB (this can be extended to output a whole lot other metrics). Note that this works only for pytests and not for unittests; also note that I wouldn't turn it on as a mainstream pytest plugin since it slows down the tests (not sure by how much but it's quite visible).

Closes #1283 

***

***


parse_pymon.py
```diff
@@ -0,0 +1,22 @@
+"""
```
bouweandela, line 1: Maybe move this to the tests directory, so as not to clutter the main directory too much? And please put code that runs stuff in a function and the add a line
```python
if __name__ == '__main__':
    run_my_function()
```
so just importing the module does not execute it, because that will lead to problems.

bouweandela: Or put it in the github/workflows directory if it's only relevant for this workflow?

valeriupredoi: ah right, that's true, I added it to the ignore list :grin:

valeriupredoi: well, we could use it manually at any point, outside the GA dir, so prob best in `tests`

valeriupredoi:
OK good news - none of our tests exceeds 1GB resident memory for a normal run, no matter what Python or OS, yay! I don't understand why the Circle CI tests die from time to time running out of the 4GB per machine, especially if we do reinforce `-n 2` so that's mean avg mem per test of 2GB. I noticed those tests are part of the conda build workflow, so it might be that a big chunk of the memory is held up by a conda build-specific process, and tests come on top of it, but that's conda build's fault and not our tests' :bear:

parse_pymon.py
```diff
@@ -0,0 +1,31 @@
...
```
bouweandela, line 1: You forgot to remove this file, now there's two copies included

valeriupredoi: seems that the series of schoolboy mistakes continues in this PR, cheers @bouweandela  :man_facepalming:

bouweandela (approved):
Looks good now! I would also be interested in seeing a list printed of the 100 or so slowest tests, do you think it could be done with this tool?

valeriupredoi:
a very good idea @bouweandela - let me try plug that in here :+1:

valeriupredoi:
@bouweandela  dunnit, you can see them eg [here](https://github.com/ESMValGroup/ESMValCore/pull/1284/checks?check_run_id=3421645057) - test run tasks and multimodel regression are the slowest, interestingly Python 3.7 and 3.8 tests are faster than 3.9 (considering and maybe wrong tho, that the machines they run on are similar)

bouweandela:
Awesome, that `test_run_tasks` looks needlessly slow. Any chance of also listing the file that the test is in? It's a bit of a search now.

valeriupredoi:
yep. that's easily doable, lemme do it now

valeriupredoi:
K, buds - added paths too - ready to merge this? and go kick some test behinds next :grin:

tests/parse_pymon.py
```diff
@@ -0,0 +1,61 @@
...
+            row[0], row[1], row[2], row[3]
+        timed_tests.append((test_name, test_var, test_path, time_used))
+
+    timed_tests = sorted(timed_tests, reverse=True, key=itemgetter(3))
+    hundred_slowest_tests = timed_tests[0:100]
+    print("List of 100 slowest tests (name, variant: duration [s] and path)")
+    if hundred_slowest_tests:
+        for test_name, test_var, pth, test_duration in hundred_slowest_tests:
+            print(test_name + " " + test_var + ": " + "%.2f" % test_duration +
+                  "s\n" + pth)
```
bouweandela, line 46: Would it be possible to print the result as
```
34.43s    tests/integration/test_task.py::test_run_tasks[16-spawn]
```
so I can just copy paste the line and run
```
pytest tests/integration/test_task.py::test_run_tasks[16-spawn]
```
to run the slow test locally?
Now it looks like this
```
test_run_tasks test_run_tasks[16, spawn]: 34.43s
tests.integration.test_task
```
and that's still confusing, because the result is spread out over two lines and some things are repeated

valeriupredoi: man, you're harder to please than a French revolutionary in 1789 :rofl: OK, let me plug that in

bouweandela:
It looks like some of the test paths are missing, I'll try adding some `__init__.py` files and see if helps:
```
29.13s tests/integration/test_task.py::test_run_tasks[16-spawn]
27.11s test_multimodel.py::test_multimodel_regression_day[full-365_day]
```
while test_multimodel.py is probably one of these files
```
./tests/sample_data/multimodel_statistics/test_multimodel.py
./tests/unit/preprocessor/_multimodel/test_multimodel.py
```

valeriupredoi:
oh yeah, good call, I have a different timing table on my laptop (having run just a subset of tests) and din't notice this, and I skimmed only the GA output - good catch!

valeriupredoi:
those init files worked a treat, we have full test paths on the GA machine :beer:

PR #1299, opened 2021-09-01
Run the multimodel statistics preprocessor last

### LINKED ISSUE #1221: Move multimodel statistics preprocessor to after spatial/temporal statistics in default preprocessor order
At the moment, the multi_model_statistics preprocessor function is by default executed before any temporal/spatial statistics preprocessor functions. This is computationally inefficient and possibly less accurate, because it introduces the need for extra regridding/vertical interpolation before the multimodel statistics can be computed. Therefore I propose to move `multi_model_statistics` all the way to the end of the default order, after `convert_units`. Does anyone have an opinion on this @ESMValGroup/esmvaltool-developmentteam?

To explain the above a bit more clearly, here are some examples of preprocessing chains that users could build:

Current implementation:

    2D input -> regrid -> multimodel stats -> area stats
    3D input -> extract levels -> regrid -> multimodel stats -> volume stats
    daily input -> multimodel stats (fails because of 360 day + 365 day calendars) -> monthly statistics

Whereas if we would move multimodel stats to the end, the above examples would simplify to

    2D input -> area stats -> multimodel stats
    3D input -> volume stats -> multimodel stats
    daily input -> monthly statistics -> multimodel stats (works because all calendars have 12 months)

so we would lose the regrid and extract levels functions that introduce inaccuracies.

In all recipes currently in the esmvaltool, the [`custom_order` option](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/overview.html#recipe-section-preprocessors) is used to change the order to the new default order proposed above. This are recipes:
- recipe_kcs.yml
- recipe_ocean_Landschuetzer2016.yml
- recipe_ocean_example.yml
- recipe_ocean_bgc.yml

The only exception to this is recipe_carvalhais14nat.yml, with
```yaml
preproc_meanRegrid:
  custom_order: true
  regrid:
    target_grid: 0.5x0.5
    scheme: nearest
  mask_landsea:
    mask_out: sea
  multi_model_statistics:
    span: overlap
    statistics: [median]
  climate_statistics:
    operator: mean
    period: full
```
but that already uses `custom_order` to move `regrid` before `mask_landsea`, so it would not be affected by this change.

For reference: the current default preprocessor order:
https://github.com/ESMValGroup/ESMValCore/blob/695fc48caca5cbce5848769f286dea0804b178a5/esmvalcore/preprocessor/__init__.py#L99-L161

bouweandela:
Move multimodel statistics preprocessor function to the end of the default order.

Some formatting issues and the use of a deprecated Python standard library function are also fixed in this pull request.

## Description

Closes #1221

Link to documentation: https://esmvaltool--1299.org.readthedocs.build/projects/ESMValCore/en/1299/api/esmvalcore.preprocessor.html

***

***

### DIFF
--- a/esmvalcore/preprocessor/__init__.py
+++ b/esmvalcore/preprocessor/__init__.py
@@ -134,7 +134,6 @@ __all__ = [
     # 'average_zone': average_zone,
     # 'cross_section': cross_section,
     'detrend',
-    'multi_model_statistics',
     # Grid-point operations
     'extract_named_regions',
     'depth_integration',
@@ -158,7 +157,10 @@ __all__ = [
     'timeseries_filter',
     'linear_trend',
     'linear_trend_stderr',
+    # Convert units
     'convert_units',
+    # Multi model statistics
+    'multi_model_statistics',
     # Remove fx_variables from cube
     'remove_fx_variables',
     # Save to file
@@ -182,6 +184,9 @@ TIME_PREPROCESSORS = [
 ]
 
 DEFAULT_ORDER = tuple(__all__)
+"""
+By default, preprocessor functions are applied in this order.
+"""
 
 # The order of initial and final steps cannot be configured
 INITIAL_STEPS = DEFAULT_ORDER[:DEFAULT_ORDER.index('add_fx_variables') + 1]
@@ -231,7 +236,8 @@ def check_preprocessor_settings(settings):
                 "function {}".format(missing_args, step))
         # Final sanity check in case the above fails to catch a mistake
         try:
-            inspect.getcallargs(function, None, **settings[step])
+            signature = inspect.Signature.from_callable(function)
+            signature.bind_partial(**settings[step])
         except TypeError:
             logger.error(
                 "Wrong preprocessor function arguments in "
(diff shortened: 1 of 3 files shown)


esmvalcore/preprocessor/__init__.py
```diff
@@ -158,7 +157,10 @@
     'timeseries_filter',
     'linear_trend',
     'linear_trend_stderr',
+    # Convert units
     'convert_units',
```
valeriupredoi, line 161: wait - shouldn't be convert units last?

valeriupredoi (changes requested):

esmvalcore/preprocessor/__init__.py
valeriupredoi: computationally it makes sense it to be last, since you'd convert units only of a single multimodel cube no?

bouweandela: What if you want to compute both the multimodel mean and variance? They wouldn't have the same unit so you couldn't use convert_units.

bouweandela: > computationally it makes sense it to be last, since you'd convert units only of a single multimodel cube no?

No, because the multimodel statistics function also keeps all the input cubes, doing it after the multimodel statistics will mean that you have to convert the units of the number of input cubes + the number of multimodel statistics, whereas if you do it before you only need to convert the input cubes.

valeriupredoi: solid points, ignore me comment, man :+1: :beer:

valeriupredoi (approved):

Issue #1314, opened 2021-09-14
Dataset problem: models with non-trivial problems with Ofx or Omon.


ledm:
I generated a recipe that looks at every `tos` (Temperature at the ocean surface) dataset in the historical and Tier1 SSP datasets. Pretty much every model in CMIP6 has provided this dataset. 

Elsewhere in the  the recipe, I take a spatial average using `area_statistics` preprocessor, `operator: mean`, which of course, requires the `areacello` Ofx dataset. 

The preprocessor is: 
```
    area_statistics:
      fx_variables:
      - {activity: CMIP,  grid: gn, mip: Ofx, short_name: areacello, exp: '*', ensemble: '*'}
      operator: mean
```
(which requires https://github.com/ESMValGroup/ESMValCore/pull/1082).

I've been working through these and a handful of models just need some minor fixes. These models needs the `fix_ocean_depth_coord` on all Omon data.

- cnrm_cm6_1
- noresm2_lm.py
- noresm2_mm.py

However, some models simply need to be excluded at the moment. The following models can not be used for the reason supplied: 

```
skip_models  = [      # because
    'AWI-CM-1-1-MR',  # triangular grid
    'AWI-ESM-1-1-LR', # triangular grid
    'BCC-CSM2-MR',    # No fx files.
    'CAMS-CSM1-0',    # No fx files.
    'CMCC-CM2-HR4',   # No fx files.
    'GISS-E2-1-G-CC', # No fx files.
    'GISS-E2-1-H',    # No fx files.
    'GFDL-CM4',       # High resolution  (1000x1000 grid). - also problem with lat/lon in areacello.
    'GFDL-ESM4',      #  problem with lat/lon in areacello.
    'IITM-ESM',       # No longitude bounds.
    'FGOALS',         # No longitude bounds in irregular grid.
    'MIROC-ES2L',     # SSP data contains loads of other variables.
    'NorESM2-LM',     # Depth provided in weird units (density)
    'NorESM2-MM',     # Depth provided in weird units (density)
    'r13i1p1f2',      # UKESM r13 also doesn't work as it's got the wrong grid.
    ]
```
I already wrote this as a discussion here, but @valeriupredoi suggested creating an issue as well: https://github.com/ESMValGroup/ESMValTool/discussions/2308

I have a feeling that some of this could be fixed, but others are fairly intractable at the moment. Happy to be proven wrong.

zklaus:
[...]

Issue #1282, opened 2021-08-20
Optional flag that if on stops the run if any of data files not found


valeriupredoi:
Was chatting to @ledm and he raised this point, with the following used case: if a certain fx file needed for a preprocessor (area, volume etc) is not found, the user should be able to turn a button on and have the run exit without continuing (in this case, the code goes on since it may be able to use the NE masks instead of the fx data). We do have the code exit if there is Missing Data, but if one uses the `--skip-nonexistent` flag it'll go on so we'd need something like `--die-hard-fx` or something so the run exits nonetheless, if certain data is not found. From my recollection it's only the fx data that doesn't stop the run if not present, is there any other type of that that has the same behavior?

ledm:
Is there any progress on this one? It's holding up one of my analyses.

zklaus:
I don't understand the issue. It seems at most it would stop a run. If you want to stop a run, just don't run it?

ledm:
At the moment, if the FX files aren't found, the run proceeds anyway and breaks elsewhere (silently and ambiguously).

zklaus:
Sure. I am not saying that this shouldn't be improved. What I don't understand is how it holds you up, since once you have understood the problem you can just move on, no?

ledm:
The solution is to figure out exactly which models ESMValTool can't find the right fx file, then comment out all the places that those models are called. It's a manual solution to an automated problem. 

I learned very early on in my phd that anything you have to do once, you'll have to do 50 times, so automation is the only way to not waste time.

valeriupredoi:
this conversation cracks me up :laughing: We could exit on warning, but that might exit the thing earlier, or we could define a specific exception related to FX files not found and add a flag that allows the code to exit on that exception; we could also have a proper dry-run run that finds all this guff before one runs a heavy recipe with lots of preprocessors. I like the dry-run myself, something that I implemented and have a PR open many pre-pandemic moons ago, but it seems a lot of the chieftains don't like that. Up to you to say how you'd think of a working solution @ledm then I can give you a hand implementing it :+1:

valeriupredoi:
> I don't understand the issue. It seems at most it would stop a run. If you want to stop a run, just don't run it?

:rofl: This is the classical beer drinking paradigm: you want to drink but you don't want to get drunk

ledm:
The dry run seems like the best option for me. I want a list of missing files whose absence will break the run. 

[...]

Issue #1138, opened 2021-05-19
Using wildcards in recipes


bouweandela:
The philosophy behind recipes is that they are as explicit as possible, so e.g. all datasets should be fully written out so it is clear what the recipe actually computes (and with which datasets it actually works).

However, this is not very user-friendly for people writing recipes and can become difficult to read too. See e.g. the requests for the use of wildcards `'*'` for specifying datasets in https://github.com/ESMValGroup/ESMValTool/issues/671, #589, and #1082. A ['recipe filler'](https://docs.esmvaltool.org/en/latest/utils.html#recipe-filler) tool was also developed to deal with this.

Maybe we could compromise by allowing this kind of feature, but only accepting recipes that do not contain wildcards in the ESMValTool repository? For easy reproducibility, we could write out a copy of the recipe with all wildcards replaced by actual datasets in the output directory? Any opinions @ESMValGroup/esmvaltool-coreteam?

Maybe this is something for discussion at the next monthly meeting https://github.com/ESMValGroup/ESMValTool/issues/2173 or one of the upcoming @ESMValGroup/technical-lead-development-team or @ESMValGroup/scientific-lead-development-team meetings?

thomascrocker:
I can't comment on the other use cases mentioned. But in the case of finding FX variables for preprocessor activities (#1082) the use of wildcards is necessary in order for ESMValTool to find the necessary files for the preprocessor. This is because for projects outside of CMIP5 (CORDEX and CMIP6 are the examples I'm working with) the location of fx files varies by institute. Some replicate them across all ensembles and experiments, whereas others do not, and in those cases the single ensemble and experiment the files are stored in is not consistent. Wildcards in the recipe, and then choosing the first file found if there are multiple, is one way around this.

I am hoping to publish a recipe soon that replicates (and expands on) some published research looking at differences between multimodel ensembles of GCM and RCM data. I ultimately am working with 30+ model datasets from each project, but don't want to have to write separate preprocessors in my recipe for each dataset to deal with the different locations of fx files. If we're having a strict "no wildcards" policy then I suppose my recipe would fall foul of this using the fix I developed in #1082.
I would argue this case is slightly different to matching multiple datasets using wildcards, since in my case the wildcard is used to facilitate searching, but only one FX file is used.

[...]

PR #1318, opened 2021-09-15
Fx files finding: CMIP6: look in experiment=`piControl` as a last measure if fx data is not found

valeriupredoi:
## Description

This is a sort of a lifeboat approach: many CMIP6 models don't have fx data in any other experiment BUT `piControl` - fx data is produced only once and is experiment-agnostic, so they don't bother producing multiple copies; also the ESGF data storing rules are a bit lax here so they don't ask the uploaders to provide multiple copies of fx data for multiple experiments. So if one wants `exp=historical, var: tas` and then asks for some fx var to go with it, ESMValTool will not find that var, hence the lifeboat search in `piControl` right before exiting and declaring fx data not found. This **helps** with #1314 but it don't close it by no means.

Link to documentation: Q: do you think we should add documentation for this?

***

***

### DIFF
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -437,12 +437,10 @@ def _get_fx_files(variable, fx_info, config_user):
     # Before completely giving up, try looking for fx files in piControl:
     # a lot of CMIP6 models have fx data only in piControl
     if not fx_files:
-        logger.warning("Missing data for fx variable '%s'",
-                       fx_info['short_name'])
         if fx_info['project'] == 'CMIP6':
-            logger.debug(
+            logger.warning(
                 "Missing data for fx variable %s "
-                "and experiment %s; since CMIP6, "
+                "and experiment %s; since data is CMIP6, "
                 "will try experiment piControl, most fx "
                 "stuff is there.", fx_info['short_name'], fx_info['exp'])
             fx_info = dict(fx_info)
@@ -453,6 +451,9 @@ def _get_fx_files(variable, fx_info, config_user):
                     "Failed to find data for CMIP6 fx variable "
                     "%s with experiment "
                     "piControl as well.", fx_info['short_name'])
+        else:
+            logger.warning("Missing data for fx variable '%s'",
+                           fx_info['short_name'])
 
     # If frequency = fx, only allow a single file
     if fx_files:
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -441,16 +441,17 @@ def _get_fx_files(variable, fx_info, config_user):
                        fx_info['short_name'])
         if fx_info['project'] == 'CMIP6':
             logger.debug(
-                f"Missing data for fx variable {fx_info['short_name']} "
-                f"and experiment {fx_info['exp']}; since CMIP6, "
-                f"will try experiment piControl, most fx stuff is there.")
+                "Missing data for fx variable %s "
+                "and experiment %s; since CMIP6, "
+                "will try experiment piControl, most fx "
+                "stuff is there.", fx_info['short_name'], fx_info['exp'])
             fx_info['exp'] = 'piControl'
             fx_files = _get_input_files(fx_info, config_user)[0]
             if not fx_files:
                 logger.warning(
-                    f"Failed to find data for CMIP6 fx variable "
-                    f"{fx_info['short_name']} with experiment "
-                    f"piControl as well.")
+                    "Failed to find data for CMIP6 fx variable "
+                    "%s with experiment "
+                    "piControl as well.", fx_info['short_name'])
 
     # If frequency = fx, only allow a single file
     if fx_files:
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -433,9 +433,23 @@ def _get_fx_files(variable, fx_info, config_user):
         fx_files = _get_input_files(fx_info, config_user)[0]
 
     # Flag a warning if no files are found
+    # CMIP6 special case:
+    # Before completely giving up, try looking for fx files in piControl:
+    # a lot of CMIP6 models have fx data only in piControl 
     if not fx_files:
         logger.warning("Missing data for fx variable '%s'",
                        fx_info['short_name'])
+        if fx_info['project'] == 'CMIP6':
+            logger.warning(
+                f"Missing data for fx variable {fx_info['short_name']} "
+                f"and experiment {fx_info['exp']}; "
+                f"will try experiment piControl")
+            fx_info['exp'] = 'piControl'
+            fx_files = _get_input_files(fx_info, config_user)[0]
+            if not fx_files:
+                logger.warning(
+                    f"Failed to find data for fx variable "
+                    f"{fx_info['short_name']} with experiment piControl as well.")
 
     # If frequency = fx, only allow a single file
     if fx_files:
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -440,15 +440,15 @@ def _get_fx_files(variable, fx_info, config_user):
         logger.warning("Missing data for fx variable '%s'",
                        fx_info['short_name'])
         if fx_info['project'] == 'CMIP6':
-            logger.warning(
+            logger.debug(
                 f"Missing data for fx variable {fx_info['short_name']} "
-                f"and experiment {fx_info['exp']}; "
-                f"will try experiment piControl")
+                f"and experiment {fx_info['exp']}; since CMIP6, "
+                f"will try experiment piControl, most fx stuff is there.")
             fx_info['exp'] = 'piControl'
             fx_files = _get_input_files(fx_info, config_user)[0]
             if not fx_files:
                 logger.warning(
-                    f"Failed to find data for fx variable "
+                    f"Failed to find data for CMIP6 fx variable "
                     f"{fx_info['short_name']} with experiment piControl as well.")
 
     # If frequency = fx, only allow a single file
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -435,7 +435,7 @@ def _get_fx_files(variable, fx_info, config_user):
     # Flag a warning if no files are found
     # CMIP6 special case:
     # Before completely giving up, try looking for fx files in piControl:
-    # a lot of CMIP6 models have fx data only in piControl 
+    # a lot of CMIP6 models have fx data only in piControl
     if not fx_files:
         logger.warning("Missing data for fx variable '%s'",
                        fx_info['short_name'])
@@ -449,7 +449,8 @@ def _get_fx_files(variable, fx_info, config_user):
             if not fx_files:
                 logger.warning(
                     f"Failed to find data for CMIP6 fx variable "
-                    f"{fx_info['short_name']} with experiment piControl as well.")
+                    f"{fx_info['short_name']} with experiment "
+                    f"piControl as well.")
 
     # If frequency = fx, only allow a single file
     if fx_files:
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -400,21 +400,9 @@ def _search_fx_mip(tables, variable, fx_info, config_user):
     return fx_info, fx_files
 
 
-def _get_fx_files(variable, fx_info, config_user):
-    """Get fx files (searching all possible mips)."""
-    # assemble info from master variable
+def _get_correct_fx_mip(fx_info, project_tables, variable, config_user):
+    """Set the correct mips for fx variables."""
     var_project = variable['project']
-    # check if project in config-developer
-    try:
-        get_project_config(var_project)
-    except ValueError:
-        raise RecipeError(f"Requested fx variable '{fx_info['short_name']}' "
-                          f"with parent variable '{variable}' does not have "
-                          f"a '{var_project}' project in config-developer.")
-    project_tables = CMOR_TABLES[var_project].tables
-
-    # If mip is not given, search all available tables. If the variable is not
-    # found or files are available in more than one table, raise error
     if not fx_info['mip']:
         fx_info, fx_files = _search_fx_mip(project_tables, variable, fx_info,
                                            config_user)
@@ -432,6 +420,28 @@ def _get_fx_files(variable, fx_info, config_user):
         fx_info = _add_fxvar_keys(fx_info, variable)
         fx_files = _get_input_files(fx_info, config_user)[0]
 
+    return fx_info, fx_files
+
+
+def _get_fx_files(variable, fx_info, config_user):
+    """Get fx files (searching all possible mips)."""
+    orig_fx_info = fx_info
+    # assemble info from master variable
+    var_project = variable['project']
+    # check if project in config-developer
+    try:
+        get_project_config(var_project)
+    except ValueError:
+        raise RecipeError(f"Requested fx variable '{fx_info['short_name']}' "
+                          f"with parent variable '{variable}' does not have "
+                          f"a '{var_project}' project in config-developer.")
+    project_tables = CMOR_TABLES[var_project].tables
+
+    # If mip is not given, search all available tables. If the variable is not
+    # found or files are available in more than one table, raise error
+    fx_info, fx_files = _get_correct_fx_mip(fx_info, project_tables,
+                                            variable, config_user)
+
     # Flag a warning if no files are found
     # CMIP6 special case:
     # Before completely giving up, try looking for fx files in piControl:
@@ -443,9 +453,10 @@ def _get_fx_files(variable, fx_info, config_user):
                 "and experiment %s; since data is CMIP6, "
                 "will try experiment piControl, most fx "
                 "stuff is there.", fx_info['short_name'], fx_info['exp'])
-            fx_info = dict(fx_info)
+            fx_info = dict(orig_fx_info)
             fx_info['exp'] = 'piControl'
-            fx_files = _get_input_files(fx_info, config_user)[0]
+            fx_info, fx_files = _get_correct_fx_mip(fx_info, project_tables,
+                                                    variable, config_user)
             if not fx_files:
                 logger.warning(
                     "Failed to find data for CMIP6 fx variable "
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -445,6 +445,7 @@ def _get_fx_files(variable, fx_info, config_user):
                 "and experiment %s; since CMIP6, "
                 "will try experiment piControl, most fx "
                 "stuff is there.", fx_info['short_name'], fx_info['exp'])
+            fx_info = dict(fx_info)
             fx_info['exp'] = 'piControl'
             fx_files = _get_input_files(fx_info, config_user)[0]
             if not fx_files:
--- a/tests/integration/test_recipe.py
+++ b/tests/integration/test_recipe.py
@@ -2199,7 +2199,13 @@ def test_landmask_no_fx(tmp_path, patched_failing_datafinder, config_user):
         fx_variables = product.settings['add_fx_variables']['fx_variables']
         assert isinstance(fx_variables, dict)
         fx_variables = fx_variables.values()
-        assert not any(fx_variables)
+        # usable fx data is found for CMIP6 in piControl
+        # the test is to check if all those found instances are only CMIP6's
+        if fx_variables:
+            count_cmip6 = [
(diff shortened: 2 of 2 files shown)


valeriupredoi:
you guys think we should have a one-liner in the docs describing this feature? :beer:

esmvalcore/_recipe.py
```diff
@@ -433,9 +433,25 @@ def _get_fx_files(variable, fx_info, config_user):
         fx_files = _get_input_files(fx_info, config_user)[0]
 
     # Flag a warning if no files are found
+    # CMIP6 special case:
+    # Before completely giving up, try looking for fx files in piControl:
+    # a lot of CMIP6 models have fx data only in piControl
     if not fx_files:
         logger.warning("Missing data for fx variable '%s'",
                        fx_info['short_name'])
+        if fx_info['project'] == 'CMIP6':
```
schlunma, line 442: I think you need to change the logic here:

```py
if not fx_files and fx_info['project'] == 'CMIP6':
    # do your thing
    # do NOT log warning if not successful
if not fx_files:
    # log the existing warning
```

With the current version, the warning about missing data is logged, even if the in the following iteration files are found for `piControl`, which is very confusing!

esmvalcore/_recipe.py
```diff
@@ -433,9 +433,25 @@ def _get_fx_files(variable, fx_info, config_user):
...
+            logger.debug(
+                "Missing data for fx variable %s "
+                "and experiment %s; since CMIP6, "
+                "will try experiment piControl, most fx "
+                "stuff is there.", fx_info['short_name'], fx_info['exp'])
+            fx_info['exp'] = 'piControl'
```
```suggestion
            fx_info = dict(fx_info)
            fx_info['exp'] = 'piControl'
```

To avoid unwanted side effects.

esmvalcore/_recipe.py
```diff
@@ -433,9 +433,25 @@ def _get_fx_files(variable, fx_info, config_user):
...
+            fx_files = _get_input_files(fx_info, config_user)[0]
+            if not fx_files:
+                logger.warning(
+                    "Failed to find data for CMIP6 fx variable "
+                    "%s with experiment "
+                    "piControl as well.", fx_info['short_name'])
```
schlunma, line 454:
```suggestion
```

I think this is not necessary, see comment above.

tests/integration/test_recipe.py
```diff
@@ -2199,7 +2199,13 @@ def test_landmask_no_fx(tmp_path, patched_failing_datafinder, config_user):
...
         assert isinstance(fx_variables, dict)
         fx_variables = fx_variables.values()
-        assert not any(fx_variables)
+        # usable fx data is found for CMIP6 in piControl
+        # the test is to check if all those found instances are only CMIP6's
+        if fx_variables:
+            count_cmip6 = [
+                p for p in fx_variables if p.get('project', None) == 'CMIP6'
+            ]
+            assert len(fx_variables) == len(count_cmip6)
```
schlunma, line 2208: I'm not sure why this change is necessary. The `patched_failing_datafinder` that is used your is not dependent on the `exp` right now, so your changes here shouldn't affect the outcome of this change. Can you double-check?

tests/integration/test_recipe.py
```diff
@@ -2397,6 +2403,73 @@ def test_fx_vars_mip_search_cmip6(tmp_path, patched_datafinder, config_user):
...
+
+    # Check add_fx_variables
+    fx_variables = product.settings['add_fx_variables']['fx_variables']
+    assert isinstance(fx_variables, dict)
+    assert len(fx_variables) == 5
+    assert '_fx_' in fx_variables['areacella']['filename']
+    assert '_Ofx_' in fx_variables['areacello']['filename']
+    assert '_Efx_' in fx_variables['clayfrac']['filename']
+    assert '_fx_' in fx_variables['sftlf']['filename']
+    assert '_Ofx_' in fx_variables['sftof']['filename']
```
schlunma, line 2470: I'm pretty sure this test is not working as it should. The `patched_datafinder` that is used here does not care about `exp`, so it also happily returns date for your `exp: historical-222`.

Please check if the files are truly from your exp by something like

```py
assert '_historical-222_' in fx_variables['areacella']['filename']
```

I'm pretty sure you need to extend these lines and use the `patched_failing_datafinder` for this to work:

https://github.com/ESMValGroup/ESMValCore/blob/671dd3799e84a0ed033e1ae66a5021eb25d60c09/tests/integration/test_recipe.py#L296-L303

schlunma (changes requested):

schlunma:
> you guys think we should have a one-liner in the docs describing this feature? beer

Yes! I think [this section](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/recipe/preprocessor.html?highlight=fx#fx-variables-as-cell-measures-or-ancillary-variables) would be a good place.

zklaus:
> you guys think we should have a one-liner in the docs describing this feature?

I think we probably should not have this feature.

I see that you removed the "Before you get started" section of the PR template, but perhaps that does serve a purpose?
We already have means to address this problem, e.g. the wildcard specifier "*" for the experiment of an fx file. This seems to try to scratch the same itch.

I think the next step is for @ledm to clarify the problems in ESMValGroup/ESMValTool#2308.

In my view, we don't have the collective time to cook up half-solutions to ill-described problems. If you have time for ESMValTool, please work on PRs in the [2.4.0 milestone](https://github.com/ESMValGroup/ESMValCore/milestone/7) instead.

valeriupredoi:
@zklaus 

> We already have means to address this problem, e.g. the wildcard specifier "*" for the experiment of an fx file. This seems to try to scratch the same itch.

That's very good point! But this solves two things, and it actually complements the wildcard implementation:

- this solves the problem when an exp is actually specified but fx files are not found there
- and the fact that the code should first look in the exp defined by user (if any), then piControl, and only then in `*` exps

valeriupredoi:
> In my view, we don't have the collective time to cook up half-solutions to ill-described problems. If you have time for ESMValTool, please work on PRs in the 2.4.0 milestone instead.

The problem is very well defined in this PR's description - and is independent of us - fx data is sparse and most of the times only in piControl

esmvalcore/_recipe.py
valeriupredoi: yes exactly, imo the warning needs to be there since we switch experiments, if the user defines a preferred experiment they need to be aware that fx files were not found for that one, and instead the code tried to look for piControl ones

esmvalcore/_recipe.py
valeriupredoi: why do you want a copy of that, it'll get destroyed at return point anyway?

tests/integration/test_recipe.py
valeriupredoi: yes I double checked - the assertion fails for CMIP6 since it finds fx files in piControl, with it passing for CMIP5 since there were no fx files found there

tests/integration/test_recipe.py
valeriupredoi: no it does not, it returns no files for that exp for CMIP5

esmvalcore/_recipe.py
schlunma: Ok, how about:


```py
if not fx_files and fx_info['project'] == 'CMIP6':
    # do your thing
    # if files found, output warning that exp has changed
if not fx_files:
    # log the existing warning
```

As I said, right now it gives a warning that no files are found, then only says in debug that `piControl` is searched now, and if actually files are found, nothing else is returned. So if your log level is not `debug`, you just get the message about missing data, nothing else (even though the `piControl` data has been found)

esmvalcore/_recipe.py
valeriupredoi: ok am silly, that actually gets returned, good call!

schlunma: `dict` are mutable, so of you change that in the function it's also changed in the code the calls the function,

esmvalcore/_recipe.py
valeriupredoi: true, forgot those are debug! yeah, lemme change it :+1:

tests/integration/test_recipe.py
schlunma: If this is the case then I think something weird happens in your new code or the test - the failing data finder is designed to fail in this situation, regardless of `exp`...

valeriupredoi: the original test was `assert not any(fx_variables)` which was passing since there were no fx files for either CMIP5 or CMIP6; now, with the lookup in piControl, it finds the piControl fx file so the original test was failing for CMIP6, but still passing for CMIP5, nonetheless it was failing at exit

schlunma: But due the use of the [`patched_failing_datafinder`](https://github.com/ESMValGroup/ESMValCore/blob/d75eb9102bf1938bb33da2034c8a25973651532d/tests/integration/test_recipe.py#L279-L306), this test shouldn't find any data, regardless of CMIP5 or CMIP6 or the `exp` of them! I'm really confused why it suddenly finds data for `piControl` now...

valeriupredoi: gah! blast, you're right, it finds whatever you want it to find, it did find `historical-222` instead of `piControl`  :man_facepalming:

tests/integration/test_recipe.py
valeriupredoi: gah! blast, you're right, it finds whatever you want it to find, it did find historical-222 instead of piControl man_facepalming

tests/integration/test_recipe.py
valeriupredoi: it finds the piControl data all too well, man! Here's the output if I force fail the test:
```
dict_values([])
dict_values([{'preprocessor': 'landmask', 'project': 'CMIP6', 'mip': 'Lmon', 'exp': 'piControl', 'start_year': 2000, 'end_year': 2005, 'ensemble': 'r1i1p1f1', 'variable_group': 'sftof', 'short_name': 'sftof', 'diagnostic': 'diagnostic_name', 'dataset': 'CanESM5', 'grid': 'gn', 'recipe_dataset_index': 1, 'institute': ['CCCma'], 'activity': 'CMIP', 'alias': 'CMIP6', 'original_short_name': 'sftof', 'standard_name': 'sea_area_fraction', 'long_name': 'Sea Area Percentage', 'units': '%', 'modeling_realm': ['ocean'], 'frequency': 'fx', 'filename': '/tmp/pytest-of-valeriu/pytest-96/popen-gw0/test_landmask_no_fx0/input/sftof_Lmon_CanESM5_piControl_r1i1p1f1_gn_1990_1999.nc'}])
--------------------------------------------------------------------- Captured log call ----------------------------------------------------------------------
WARNING  esmvalcore._recipe:_recipe.py:455 Missing data for fx variable 'sftlf'
WARNING  esmvalcore._recipe:_recipe.py:455 Missing data for fx variable 'sftof'
WARNING  esmvalcore._recipe:_recipe.py:441 Missing data for fx variable sftlf and experiment historical; since data is CMIP6, will try experiment piControl, most fx stuff is there.
WARNING  esmvalcore._recipe:_recipe.py:450 Failed to find data for CMIP6 fx variable sftlf with experiment piControl as well.
WARNING  esmvalcore._recipe:_recipe.py:441 Missing data for fx variable sftof and experiment historical; since data is CMIP6, will try experiment piControl, most fx stuff is there.
WARNING  esmvalcore._recipe:_recipe.py:455 Missing data for fx variable 'sftlf'
```
exactly what I want :grin:

schlunma: Not exactly what I want :smile: : `sftof_Lmon_CanESM5_piControl_r1i1p1f1_gn_1990_1999.nc`. There is no `sftof` in `Lmon`.

The main problem here is that `fx_info['mip']` is set to `Lmon` after the automatic mip search, which causes the data finder to not fail (since the `mip` is not `fx` anymore). In this case you need to repeat the following logic (i.e. the automatic mip search) also for the new `exp` (after removing `mip` again from `fx_info` if necessary):

https://github.com/ESMValGroup/ESMValCore/blob/d75eb9102bf1938bb33da2034c8a25973651532d/esmvalcore/_recipe.py#L418-L433

tests/integration/test_recipe.py
valeriupredoi: OK I fixed this now :beer:

zklaus:
Let's put some numbers to the problem. Looking only at `(Ofx, areacello)` in `CMIP` experiments using the data pool at Jasmin, I find the following statistics for the published fx files. Here, number gives the number of unique (source_id, ensemble) combinations that publish these fx files under the given experiments.

|exp|number|
|-|-|
|historical|23|
|piControl,1pctCO2,historical|14|
|abrupt-4xCO2,historical|13|
|piControl,historical|12|
|piControl|10|
|amip,historical|9|
|piControl,1pctCO2,esm-piControl,historical|3|
|abrupt-4xCO2|3|
|1pctCO2,abrupt-4xCO2,historical|2|
|1pctCO2,historical|2|
|1pctCO2|2|
|amip,piControl,1pctCO2,abrupt-4xCO2,historical|1|
|esm-piControl-spinup,esm-hist,piControl,piControl-spinup,esm-piControl,1pctCO2,abrupt-4xCO2,historical|1|
|piControl,1pctCO2|1|
|piControl,1pctCO2,abrupt-4xCO2,historical|1|
|piControl,amip,1pctCO2,historical|1|
|piControl,piControl-spinup,1pctCO2,abrupt-4xCO2,historical|1|

In other words, `historical` is more prevalent than `piControl`, being absent from only 16 entries, but really there is no clear system. In light of this, I again strongly suggest that we continue the already ongoing discussions about how to approach this instead of moving forward with this PR.

zklaus (changes requested):
The assumption that FX data is generally available in the piControl experiment is not borne out by the data, hence further discussion is required.

valeriupredoi:
wait, I don;t understand the numbers:


historical | 23
-- | --
piControl,1pctCO2,historical | 14

Does this mean that 23 models have (Ofx, areacello) in historical and 14 models have it in piControl,1pctCO2, and historical?

zklaus:
Yes, to be precise 23 (model, ensemble) combinations have the data only in `historical`, 14 (model, ensemble) combinations have it in all three of piControl, 1pctCO2, and historical.

tests/integration/test_recipe.py
valeriupredoi: aye! but why is Lmon added as a valid mip for sftof in the first place? All my new bit of code is doing is jiggling the exp for a set of otherwise fixed and hopefully valid data parameters?

valeriupredoi:
OK but that's not a relevant count since some models have a large number of ensembles, those 23 there might be from a single model, which is probably the IPSL one :grinning: The fact that you see piControl in every row of your table is telling you why this PR is in :beer:

valeriupredoi:
what's interesting from your table is that historical is prob as well populated with fx data as piControl, so we can do a search there too

tests/integration/test_recipe.py
schlunma: That's why:

https://github.com/ESMValGroup/ESMValCore/blob/d75eb9102bf1938bb33da2034c8a25973651532d/esmvalcore/_recipe.py#L389-L391

I think this was necessary since code at a later stage needs the `mip` attribute.

valeriupredoi: OK but this is not good since we need to set the correct data/file parameters right before we call  `_get_input_files(fx_info, config_user)[0]` - why are we running that function with invalid data descriptors, like sftof-Lmon?

schlunma: We didn't do that before the changes here :wink: As I said, you need to make sure to use same logic as for the original `exp`: If `mip` exists -> use this; if not -> call `_search_fx_mip`

valeriupredoi: as in - the moment we call the data finder function we need to make sure that `fx_info` dict is correct and not massage it afterwards if we don't find files, unless one changes the exp to piControl which is a CMOR table-independent parameter (like I'm doing here) - what am I missing here?

valeriupredoi: arite, that whole block needs to be a function on its own, and an annoying one too, since that needs to be called once (ideally) and the `fx_info` needs to be set correctly before calling the data finder file fetter bleh :grin:

zklaus:
> OK but that's not a relevant count since some models have a large number of ensembles, those 23 there might be from a single model, which is probably the IPSL one.

There might be some of that going on, but manual inspection suggests that is not the main reason. But of course, feel free to suggest a better metric, or better yet, let us know which numbers you used to justify this approach.

> The fact that you see piControl in every row of your table is telling you why this PR is in.

You mean the fact that we **don't** see it in every row and indeed missing in many and some of the most important rows let's us take a step back and re-evaluate the approach, right?

tests/integration/test_recipe.py
valeriupredoi: OK Manu, I have made that mip jiggling a function now, that gets applied before we look for piControl, rehabilitated the old test that needed to stay the same and added two failsafe tests that should fail without using the `patched_datafinder` - these cover the new functionality with piControl, but they cover it synthetically, I don't want to start using yaml files containing actual paths to filesso I can write a test that doesn't find eg historical but finds piControl, that's overkill for this PR, what do you think?

valeriupredoi:
OK @zklaus so what do you suggest right now? @ledm (and possibly other, less vocal, users) are faced with a long-term issue that fx data is in all sorts of dirs where esmvaltool is not looking for, and the wildcard PR is stagnant still - and to be honest with you I'd rather search in a subset of exps than in **all** exps - one, less time consuming and two - God only knows what crazy fx files are in an `abrupt-4xCO2-monkeysOnAPlane` experiment, at least we know piControl is solid data, OK maybe historical too. Anyway, am open to suggestions, but am also keen on a rather sooner than later solution, even if it's a bit kluge-y :beer:

schlunma:
Cheers @valeriupredoi for all the changes :beer: I have some additional comments, but let's continue the review once we've sorted out how to proceed with this!

ledm:
From my side, now that there's the chance to use wildcards (#1082) to find the FX files, I'm not sure that we need to have a default location. (Sorry @valeriupredoi).

zklaus:
@valeriupredoi, I mainly suggest two things:

First, following the standard procedure, that is
1. Start from an actual problem. If you don't immediately have an error to go off, but still suspect a problem, collect the information that shows the problem.
2. Open an issue to discuss the problem and how to address it. Rarely is one person in a position of complete awareness of implications and boundary conditions.
3. If and when reasonable consensus is reached, start working on a draft PR. Pay attention to the checklist; it's there for a reason and we should not think that as core or lead or whatever developers we don't have to follow it, but rather exemplify how and why following it improves the quality.

In this part, the main problem with this PR is that it is lacking in step 1 because it starts from an unsubstantiated claim. Whether that claim turns out to be true or not is almost immaterial, but given the data, I wouldn't call it completely true.

Step 2 was just left out completely.
This is particularly troubling because there is actually a lot of discussion on this precise topic ongoing, see for example https://github.com/ESMValGroup/ESMValCore/issues/1282#issuecomment-919030036, https://github.com/ESMValGroup/ESMValCore/issues/1138#issuecomment-845202897, and ESMValGroup/ESMValTool#2308, with a lot of good insights and ideas.
This PR just sidesteps all of that information and all the aspects covered in the discussion.
Obscuring this sidestepping of the community, the checklist for step 3 was modified to remove the issue requirement.


Ok, so that was the first thing. The second thing is that we need to prioritize. Why do you reckon the wildcard PR (ESMValGroup/ESMValCore#1082) is stagnant? I think it has something to do with us being more eager to start new things than to finish old ones. Don't get me wrong. I think this is very human and I certainly have the same urge. I firmly believe that we need to get better here and using milestones to highlight imminent work is one tool for doing this. Of course, there are other tools, and we can always discuss how to do this, but any way you slice it, we need to close old issues and PRs, preferably by solving and finishing them.

valeriupredoi:
@schlunma cheers! @ledm all good by me, as long as you think you can proceed with work @zklaus nice pep talk and analysis, I am aware of all the procedures, but sometimes if you stick too much to procedures you end up with a procedurally sound, half-working machinery - I am more of a dirty fix before we fix it proper kind of guy :grin: So consensus is to close this and go on with #1082 ? Are there any known risks of searching all the experiments - my only concern is that some crazy experiment might have some crazy fx data that's non-standard - if this is not the case, then we can defo forget about this PR :+1:

zklaus:
Well, in this case, it helped us avoid unnecessary changes. Unfortunately only after wasting a lot of resources.

PR #1323, opened 2021-09-21
Fix tests to import explicitly `iris.fileformats` and remove `esmpy` from `setup.py` since not on PyPi

valeriupredoi:
## Description

Iris 3.1.0 needs you to explicitly import any of its submodules and don't use them as iris attributes in the code; also, I had to remove `esmpy` from `setup.py` since pip can't find it anymore - that's a bit of an odd one - has it gone off PyPi?

Closes #1322 

***

## [Before you get started](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#getting-started)

- [x] [☝ Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do

## [Checklist](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#checklist-for-pull-requests)

It is the responsibility of the author to make sure the pull request is ready to review. The icons indicate whether the item will be subject to the [🛠 Technical][1] or [🧪 Scientific][2] review.

[1]: https://docs.esmvaltool.org/en/latest/community/review.html#technical-review
[2]: https://docs.esmvaltool.org/en/latest/community/review.html#scientific-review

- [x] [🛠][1] This pull request has a [descriptive title and labels](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-title-and-label)
- [x] [🛠][1] Changes are [backward compatible](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#backward-compatibility)
- [x] [🛠][1] Any changed [dependencies have been added or removed](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#dependencies) correctly
- [ ] [🛠][1] All [checks below this pull request](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-checks) were successful

***

To help with the number pull requests:

- 🙏 We kindly ask you to [review](https://docs.esmvaltool.org/en/latest/community/review.html#review-of-pull-requests) two other [open pull requests](https://github.com/ESMValGroup/ESMValCore/pulls) in this repository

PR #1324, opened 2021-09-21
explicitly imports of iris submodules instead of using iris attributes (`iris=3.1.0` incompatibility change)

valeriupredoi:
## Description

Due to [this note](https://scitools-iris.readthedocs.io/en/v3.1.0/whatsnew/3.1.html#minimised-imports) in the iris=3.1.0 release notes, it looks like it's not safe to use `iris.cube.Cube` or `iris.cube.CubeList` any more without the explicit import of `iris.cube` - ie use it as imported submodule and not as attribute. There may be more instances that I missed, also @bouweandela @bjlittle do we need to change all calls to `iris` attributes to explicit imports (eg `iris.Constraint` etc)?

Closes #issue_number

Link to documentation:

***

***

### DIFF
--- a/esmvalcore/cmor/_fixes/cmip5/ec_earth.py
+++ b/esmvalcore/cmor/_fixes/cmip5/ec_earth.py
@@ -1,5 +1,6 @@
 """Fixes for EC-Earth model."""
 import iris
+import iris.cube
 import numpy as np
 from dask import array as da
 
--- a/esmvalcore/cmor/_fixes/cmip6/cnrm_cm6_1.py
+++ b/esmvalcore/cmor/_fixes/cmip6/cnrm_cm6_1.py
@@ -1,5 +1,6 @@
 """Fixes for CNRM-CM6-1 model."""
 import iris
+import iris.cube
 
 from ..common import ClFixHybridPressureCoord
 from ..fix import Fix
--- a/esmvalcore/cmor/_fixes/native6/era5.py
+++ b/esmvalcore/cmor/_fixes/native6/era5.py
@@ -3,6 +3,8 @@ import datetime
 import logging
 
 import iris
+import iris.cube
+
 import numpy as np
 
 from ..fix import Fix
--- a/esmvalcore/preprocessor/_ancillary_vars.py
+++ b/esmvalcore/preprocessor/_ancillary_vars.py
@@ -4,6 +4,7 @@ import logging
 
 import dask.array as da
 import iris
+import iris.cube
 
 from esmvalcore.cmor.check import cmor_check_data, cmor_check_metadata
 from esmvalcore.cmor.fix import fix_data, fix_metadata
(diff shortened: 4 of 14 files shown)


bouweandela:
I think there is no need to worry about this, because as soon as you call a function that uses `iris.cube`, e.g. `iris.load` (the very first thing we do when starting the preprocessor), it will import `iris.cube` so we don't need to do it anymore. Did you actually encounter problems related to this? Similar story for #1323.

valeriupredoi:
no, not at test run stage, but I've not run anything else though, like an actual recipe. That's good if we don't need it - less guff, but I'd still like to have a closer look and maybe see if there are other stuff that we should import - how about `iris.Constraint`? I guess we can do it on a "let's fix it if it's broken" basis anyway, and close this - for the other PR, I've seen only issues with the weighted landsea mask test which is fixed in #1217 but I assumed other tests could be at risk too. We can close that too now that @zklaus has fixed the issue with esmpy (by fixing it proper upstream) and we can open an issue to fix the import of the Earth radius (not from that pp antiquated way). What you say @bouweandela and @zklaus - seems like this week am closing three of my own PRs :sob:

zklaus:
Hm. I have to say I am not sure whether I agree with @bouweandela here.

If we use a library in our code we should declare it as a dependency, even if it is also the dependency of one of our other dependencies. For example, we should declare dask, even though it is already being pulled in by iris.

Similarly, one should perhaps do the relevant imports, even if they already happen by a side-effect of another line of code.

I need to think about this a bit.

valeriupredoi:
another thing we don't import but use as attr is `iris.analysis` - I reckon @zklaus is right, and we should get all these imported explicitly, but alas, am not that familiar with why the actual need to import them in the new iris, I just took that comment in the iris 3.1 release notes as a must-do :+1:

bouweandela:
@zklaus I think this is a very hairy problem in general, because it is never clear in Python what submodules are loaded as a side effect and which ones need to be loaded separately. It would be a huge task to check this for all imports. It is also very difficult to unit test this, because you might be loading libraries in your test code that are required to run your production code and you'll never notice that your production code only works because of a side effect of your test. That's why I proposed to not worry about it until it actually becomes a problem. I'm fine with merging this, but I wouldn't spend too much time on it until it becomes a real problem that our users run into.

zklaus:
Agree that we should spend time only on actual problems, so if @valeriupredoi agrees I suggest/agree to drop this.

As a matter of principle, I'm still on the fence. I don't think you need to check if a module you need (accidentally) gets imported by side-effect. If you write `package.module.something` in your code, you write `import package.module` at the top, no matter what; no checking involved.

But yeah, let's be economical and spend time where we have problems.

@valeriupredoi, please reopen if you think that's better.

valeriupredoi:
all good, let's fix it if it's broken, it ain't broken now :grin:

PR #1345, opened 2021-10-06
cleaned up a bit Github Actions workflows

valeriupredoi:
## Description

- Changed some GA test titles so they are a bit more readable and some other minor guff
- Changed the actual names of workflows (removed `action-` prefix and gave them more friendly names)
***

***

### DIFF
--- a/.github/workflows/run-tests-monitor.yml
+++ b/.github/workflows/run-tests-monitor.yml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
     - main
-    - cleanup_github_actions
   # run the test only if the PR is to main
   # turn it on if required
   #pull_request:


.github/workflows/action-conda-publish.yml
```diff
@@ -51,6 +52,7 @@ jobs:
         run: |
           export BUILD_FOLDER=/tmp/esmvalcore/_build
           # use --label test before actual release and check all works well
+          # esmvalgroup is not the main channel anymore, conda-forge superseded it
```
zklaus, line 55: This comment confuses me. Shall we just scrap the whole action?

zklaus (changes requested):
I think we also said to remove the `action-` prefix from the files. Might as well do that at the same time?

.github/workflows/action-conda-publish.yml
valeriupredoi: I was thinking about that too - but what if we decide we really need to test something related to the package build (or even more, we want to test the package build) - we already have a working GA workflow rather than fishing around the interwebs for one :grin:

zklaus: We are using this thing called git that should allow us to get it back even in that case :wink:

valeriupredoi: sifting through git history is more painful than finding pictures of your ex-gf's in your photo album, man :rofl: But OK, I can remove it for good if yout hink it's a good idea

zklaus: That depends on how you maintain your history. If you can't use it to find your own work, you are probably doing it wrong. But that is a topic for a different time.

I think we can remove this.

zklaus (approved):
Looking good! Could you please make sure the checklist is tended to?

PR #1367, opened 2021-10-19
Fix individual pytest runs broken due to missing explicit imports from `iris`

### LINKED ISSUE #1329: Running individual tests with `pytest` fails when not all `iris` submodules are imported inside test module
Related to #1322 (but that's closed since it served its purpose). Running tests collectively with `pytest` or `pytest -m MSSG` works fine since, as @bouweandela mentioned, imports of iris submodules are done at top level, even if not explicitly in the test module itself. But running individual test modules for debugging and development purposes, fails when `iris` submodules are not imported explictly inbthe test module. Example:

```
pytest tests/unit/cmor/test_cmor_check.py
```

dies spectacularly with multiple:
```
FAILED tests/unit/cmor/test_cmor_check.py::TestCMORCheck::test_warning_fail_on_error - AttributeError: module 'iris' has no attribute 'cube'
```

valeriupredoi:
## Description

Some of the tests we have can not be run individually (for dev/debugging cases) since some of `iris`'s submodules need explicit importing during the test session.

All tests that fail at individual run point are fixes-related, here is the rerun script that I used to check I fixed them all OK 
[failsrerun.sh.txt](https://github.com/ESMValGroup/ESMValCore/files/7373364/failsrerun.sh.txt)

Closes #1329 

***

***


valeriupredoi:
this is a bit tumbleweed-y, you guys reckon I can get a review here :beer: 
tumbleweed

tests/integration/cmor/_fixes/__init__.py
```diff
@@ -0,0 +1,3 @@
+import iris
+import iris.aux_factory
+import iris.cube  # noqa
```
bouweandela, line 3: Why do you need the #noqa?

valeriupredoi: flake8 complains the import is not used, and throws an error

bouweandela: Why do you need the import if it's not used?

valeriupredoi: because it's needed in a bunch of tests, and rather than importing them in each of the test modules it's better to put it in the init - flake is silly and doesn't think outside the box

valeriupredoi:
@schlunma you been around recently, you mind having a look at this one pls, mate - a short PR that's been sidelined by most everybody :cry:

schlunma:
Sure thing! But I actually share Klaus' opinion from [this](https://github.com/ESMValGroup/ESMValTool/pull/2455#pullrequestreview-850908888) comment here that imports should be put into the files where they are needed.

For example, this only works when you are running all tests, but fails when you are running individual tests that do not import the  parent modules.

valeriupredoi:
> For example, this only works when you are running all tests, but fails when you are running individual tests that do not import the parent modules.

Wait what? Can you give me a concrete example pls? This is the whole point of the PR - to fix individual runs of tests, say test `cow.py` works when run in the whole suite, but `pytest cow.py` fails due to missing explicit import of iris submodule `bull` in `cow.py`

schlunma:
I tried to find a test in the `tests/integration/cmor/_fixes/` subdirectory, but actually didn't find any (even in the `main` branch!). Can you tell me which test failed for you before this PR?

I guess the reason these tests work without the explicit import of something like `import iris.cube` is that in some module that is imported by the tests, e.g., `from esmvalcore.cmor.fix import Fix` we have the necessary imports of the submodules of iris (e.g., `import iris.cube`), and after that a simple `import iris` in the corresponding test enables us to use `iris.cube.xxx`.

To be honest I'm not quite sure how python handles these kind of nested imports, but from what I can tell these changes here do not really do anything (since the module `tests.integration.cmor._fixes` is never imported as far as I can tell).

schlunma:
By the way, I got a `mypy` error in basically every test that looks like this:

```
4: error: Skipping analyzing "cf_units": module is installed, but missing library stubs or py.typed marker
```

(Using a fresh environment that was created yesterday)

valeriupredoi:
do a `pip install -e .[develop]` to install locally, I had a mypy error before I did that too :+1:

schlunma:
It already is installed locally, but I could fix it by running `pytest` in the `ESMValCore` root directory instead of `ESMValCore/tests/integration/cmor/_fixes`...weird!

valeriupredoi:
> It already is installed locally, but I could fix it by running `pytest` in the `ESMValCore` root directory instead of `ESMValCore/tests/integration/cmor/_fixes`...weird!

ah yeah `pytest` needs to be run in the root dir

schlunma:
I ran all of them and also didn't find an error; I think we can safely close this :+1:

valeriupredoi:
thanks a lot dude! Yes, I'll close now, the question is though, what has changed to have those tests pass now? @bouweandela or @zklaus any ideas?

PR #1384, opened 2021-11-04
Update version number for third release candidate 2.4.0rc3

zklaus:
Update the version number for the upcoming release candidate.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -169,11 +169,11 @@ authors:
 
 
 cff-version: "1.0.3"
-date-released: 2021-10-27
+date-released: 2021-11-04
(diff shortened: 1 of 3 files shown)


valeriupredoi (approved):
fire up, Klaus! But don't you reckon it's a good time to get the stable release out rather than yet another rc?

zklaus:
That would be great. But with the removal of the AtmosphereSigmaFactory, I feel we need to do another round of testing first.

valeriupredoi:
fair do's, but let's not get [RC135](https://en.wikipedia.org/wiki/Boeing_RC-135) :laughing:

PR #1400, opened 2021-11-26
Update documentation on running the tool

bouweandela:
## Description

Update the documentation on running a recipe. The examples given were confusing, because they suggested that it was possible to use the command `esmvaltool run recipe_python.yml`.

Link to documentation:

***

***

### DIFF
--- a/doc/quickstart/run.rst
+++ b/doc/quickstart/run.rst
@@ -6,24 +6,47 @@ Running
 The ESMValCore package provides the ``esmvaltool`` command line tool, which can
 be used to run a :doc:`recipe <../recipe/index>`.
 
-To run a recipe, call ``esmvaltool run`` with the desired recipe:
+To list the available commands, run
 
 .. code:: bash
 
-	esmvaltool run recipe_python.yml
+    esmvaltool --help
+
+It is also possible to get help on specific commands, e.g.
+
+.. code:: bash
+
+    esmvaltool run --help
+
+will display the help message with all options for the ``run`` command.
+
+To run a recipe, call ``esmvaltool run`` with the path to the desired recipe:
+
+.. code:: bash
+
+	esmvaltool run recipe_example.yml
+
+The ``esmvaltool run recipe_example.yml`` command will first look if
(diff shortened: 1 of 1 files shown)


bouweandela:
Requires pull request ESMValGroup/ESMValTool#2432 for the links to work, so after that is merged, the build on CicleCI should be successful.

doc/quickstart/run.rst
```diff
@@ -6,32 +6,57 @@ Running
...
+
+will display the help message with all options for the ``run`` command.
+
+To run a recipe, call ``esmvaltool run`` with the path to the desired recipe:
+
+.. code:: bash
+
+	esmvaltool run recipe_example.yml
+
+The ``esmvaltool run recipe_example.yml`` command will first look if
```
valeriupredoi, line 29: here too, we should probably use the name of an existing recipe or the vaguely geeky `$RECIPE_NAME` - some users will actually try running with `recipe_example.yml` :grin:

valeriupredoi (changes requested):
was about to approve but then I saw you are calling that `recipe_example.yml` everywhere - did I miss a trick and you added that? Otherwise we shouldn't post run command examples with non-existing recipes, in a month we'll have ten Issues titled "Recipe recipe_example.yml can't be found" :rofl:

doc/quickstart/run.rst
bouweandela: I'll add a note that this is just an example and the recipe does not actually exist. I would prefer not to use a recipe from ESMValTool because that may not even be installed.

valeriupredoi: good call, how about we copy `recipe_python.yml` in Core - that'll run without actually installing Tool

bouweandela: It wouldn't work, because you'd also need the diagnostic script and all the shared code that the diagnostic script uses

bouweandela:
Hi @valeriupredoi, is there anything left to do on this? I'm not sure I can make it any better with the current suggestions.

PR #1404, opened 2021-12-06
remove `profile_diagnostic` from diagnostic settings

### LINKED ISSUE #1392: profile_diagnostic option is passed to the diagnostic script
The `profile_diagnostic` option from config-user.yml is passed on to the diagnostic script as a parameter in settings.yml. This is not necessary, because this option only affects how the diagnostic script is run by the (Python) interpreter. I don't see how this option would need to be used by the diagnostic script itself, so I propose we stop passing it on.

valeriupredoi:
## Description

Closes #1392 

Link to documentation:

***

***

### DIFF
--- a/esmvalcore/_recipe.py
+++ b/esmvalcore/_recipe.py
@@ -1392,7 +1392,6 @@ class Recipe:
             for key in (
                     'output_file_type',
                     'log_level',
-                    'profile_diagnostic',
                     'auxiliary_data_dir',
             ):
                 settings[key] = self._cfg[key]


bouweandela:
It looks like `profile_diagnostic` will still be written to the NCL interface file. Could you have a look?

If you increase the test coverage I'll buy you a :beer: next time we meet :wink:

valeriupredoi:
yesh man, on it today! I'll have a look at the NCL hiccup too :+1:

valeriupredoi:
OK @bouweandela I fixed the NCL bit, and now we are testing for `profile_diagnostic` not being a dig setting, note however that I also turned back on the diagnostic test on Python - why did we comment that out?

valeriupredoi:
@bouweandela I don't understand why codecov is complaining - those lines are nice and green (checked for) in my copy of the tests coverage (local run) - apart from that, this is RFR :beer:

valeriupredoi:
agh nevermind! I figured it out - we're running a very low number of `task` tests in CircleCI - that's not good - coverage for `_task.py` goes down from 84% to 67% the way we run the tests with "no installation" flag - I removed it, it should be fine, it's not slowing the tests down visibly

valeriupredoi:
OK am converting this to a Draft until I get to the bottom of the issue of running a Python diagnostic in a test env on Circle - that's a bit disconcerting at the mo, we should be able to test the full top to bottom task infrastructure via a Py diag

valeriupredoi:
OK finally happy with this, and we get a nice boost in coverage for `_task.py` and overall decent coverage boost. I've added some test output to CirceCI Artifacts so we can see what's happening when things go south, the Artifacts bit really needs a cleanup and us storing different things in different directories, but that's for another PR, @bouweandela she's all yours, mate :beer:

schlunma:
@valeriupredoi I can have a look at this today, can you merge the current `main` into it, please? Cheers!

valeriupredoi:
many thanks Manu! I have just merged the `main`, knock yerself out :grin:

.circleci/config.yml
```diff
@@ -34,7 +34,7 @@ jobs:
             . /opt/conda/etc/profile.d/conda.sh
             conda activate esmvaltool
             pip install .[test]
-            pytest -n 2 -m "not installation and not sequential"
+            pytest -n 2 -m "not sequential"
```
schlunma, line 37: Can you briefly explain why these changes in this file are relevant for this PR?

tests/integration/test_diagnostic_run.py
```diff
@@ -52,23 +51,22 @@ def check(result_file):
...
+    'diagnostic.py':
+    dedent("""
+        import os
+
+        def main():
+            os.system('cp settings.yml ../../../../../result.yml')
+
+        if __name__ == '__main__':
+            main()
+        """),
```
schlunma, line 69: That looks a bit odd. Why is this change needed in this PR and do you check this script somewhere? I would expect something like `assert file_exists('result.yml')` somewhere in the code.

schlunma (changes requested):
I just tested this for python and NCL. For python it worked fine, but for NCL the option `profile_diagnostic` is still in the settings file:

```ncl
diag_script_info@profile_diagnostic = False
```

Apart from that I have some comments regarding the automatic tests.

.circleci/config.yml
valeriupredoi: thanks for looking into this, man! Yes, good point, I should add this to the PR description: the `not installation` message tells pytest not to run the bits that need NCL, R, Julia etc if the executables are not on the system, so that those tests don't fail; that reduces task testing by quite a bit, and `_task.py` test coverage is pretty bad that way, so I went about and made it do those checks anyway, without the tests failing if the executables are not found

tests/integration/test_diagnostic_run.py
valeriupredoi: it's a hack - it needs that `result.yml` to be checked, without running the actual diagnostic, so am just copying it from where the monkeypatched test ran to where the the file should be in the case of an actual run - maybe @bouweandela can think of a more elegant solution, I didn't want to change too much of his original test, hence the hacky bit here

.circleci/config.yml
valeriupredoi: also see my comment here (I actually did bother commenting on that, but I didn't include it in the PR description) https://github.com/ESMValGroup/ESMValCore/pull/1404#issuecomment-988168030

valeriupredoi:
@schlunma cheers dude! Good spot about the NCL settings still having that, it was a nested dict, fixed it now :laughing:

tests/integration/test_diagnostic_run.py
valeriupredoi: well aren't a big ray of sunshine today :rofl: OK seriously now, I'll look into getting rid of this hack :+1:

valeriupredoi: OK am a total loss here - @bouweandela how do you get a Python diagnostic to write a `results.yml` file with diag env variables like `run_dir` etc without accessing neither the environment (I don't want to add `import os` stuff in the diag) nor by importing `run_diagnostic` from esmvaltool since that package is not supposed to be installed?

valeriupredoi: or @schlunma - don't matter who answers me, as long as we have an answer to fix that ugly hack :grin:

schlunma: Looks like `settings.yml` is located in the cwd, so just load this with `yaml.safe_load()` and then simply write a file to `run_dir` or copy `settings.yml` using `shutil.copy()`.

bouweandela: The point I was trying to make in my previous comment is that you do not need to simulate a diagnostic run to test the feature you're implementing in this pull request. Why are you editing this test at all? The only thing you need to test is that the interface file (settings.yml) gets written correctly.

schlunma:
This fails with a

```
KeyError: 'diag_script_info'
```
now.

[main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/8017422/main_log_debug.txt)

@valeriupredoi I agree with @bouweandela here that it might make sense remove your changes here to files that the tests that are not relevant for this PR. You think we can include this in the first release candidate (tomorrow morning)?

tests/integration/test_diagnostic_run.py
valeriupredoi: no absolutely not, you are right there Bouwe, but we need to run a sim diag to test the tasking mechanism, that mecahnism is not tested otherwise, where do you think that big bump in coverage percentage comes from?

valeriupredoi:
> This fails with a
> 
> ```
> KeyError: 'diag_script_info'
> ```
> 
> now.
> 
> [main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/8017422/main_log_debug.txt)
> 
> @valeriupredoi I agree with @bouweandela here that it might make sense remove your changes here to files that the tests that are not relevant for this PR. You think we can include this in the first release candidate (tomorrow morning)?

no, the tests are very much relevant to the second main scope of this PR - first is to get rid of that profile settings arg, and second is to increase the coverage for tasking which was poor previously. After a relatively major kerfuffle with git (I was sound asleep when I merged only God knows what I merged in here, took me an hour to fix it) I finally fixed the test to be more secure than previously (and fixed the issue with the NCL setting key too, cheers Manu!). If you don't run the Python mock test you don't test the tasking proper, and Bouwe made sure that he wrote the test in a rather "scratch your left ear with your right hand" way, so my hand were tied and couldn't go around it too much - why do you need a `result.yml` file in the main output dir is beyond me :grin: Anyway, am done now, take it or leave it :laughing:

valeriupredoi:
thanky much, Manu! :beer:

valeriupredoi:
yes, I knew it'll complain - the bugger is now complaining that the increase in coverage dropped from 19% to 18% :rofl: But that's an extra 18%, ya codecov dummer! The problem is that I would have to write another test case for when the `profile_diagnostic` is not inside the `diag_settings_info` dict, but I don't think running yet another mock diagnostic would justify testing only for those two lines only - what you think?

valeriupredoi:
cheers, Manu! That's only for the NCL case and it's not immediately straightforward to me how to do that without running another mock diagnostic (which I don't see the need for, it just slows down the test) - any ideas? You speak NCL much better than me :grin:

schlunma:
Haha, please don't quit just yet, I need your review for a fix for #1480 in a couple of minutes :smile:

valeriupredoi:
am still in, dude :vulcan_salute:

schlunma (approved):
Just tested everything again, works fine now! Thanks V!!

valeriupredoi:
thoughts and prayers for my hero Manu :us: :eagle: Cheers much dude!!

PR #1434, opened 2022-01-18
Pin `flake8<4` to keep getting relevant error traces when tests fail with FLAKE8 issues

### LINKED ISSUE #1405: Pin `flake8<4` - running `pytest` tests with `--flake8` turned on may mask print statements that are necessary if test fails
A bunch of native Python errors - like undefined name/variable - are now picked up by Flake8 when running a pytest test - if those errors can be debugged via print statements that the test developer has purposefully put in the test for debugging, those print statements don't get flushed to stdout even if you change the `--capture` value when running pytest, because flake8 runs right at the very beginning and once it fails it discards every other call that follows in the test. I'd like to change this somehow, without having to turn off the `--flake8` flag, but I don't know how - any ideas?

valeriupredoi:
## Description

@bouweandela and I have decided to pin flake8, so we stop getting garbage instead of actual FLAKE8 issues when tests fail with FLAKE8 issues

Closes #1405 

Link to documentation:

***

***

### DIFF
--- a/package/meta.yaml
+++ b/package/meta.yaml
@@ -72,7 +72,7 @@ test:
     - tests
     - setup.cfg
   requires:
-    - flake8<4
+    - flake8<4  # https://github.com/ESMValGroup/ESMValCore/issues/1405
     - pytest>=3.9,!=6.0.0rc1,!=6.0.0
     - pytest-cov
     - pytest-env
--- a/package/meta.yaml
+++ b/package/meta.yaml
@@ -48,6 +48,7 @@ requirements:
     - esmpy!=8.1.0  # see github.com/ESMValGroup/ESMValCore/issues/1208
     - fiona
     - fire
+    - flake8<4  # see https://github.com/ESMValGroup/ESMValCore/issues/1405
     - humanfriendly
     - isodate
     - jinja2
--- a/package/meta.yaml
+++ b/package/meta.yaml
@@ -48,7 +48,6 @@ requirements:
     - esmpy!=8.1.0  # see github.com/ESMValGroup/ESMValCore/issues/1208
     - fiona
     - fire
-    - flake8<4  # see https://github.com/ESMValGroup/ESMValCore/issues/1405
     - humanfriendly
     - isodate
     - jinja2
@@ -73,6 +72,7 @@ test:
     - tests
     - setup.cfg
   requires:
+    - flake8<4
     - pytest>=3.9,!=6.0.0rc1,!=6.0.0
     - pytest-cov
     - pytest-env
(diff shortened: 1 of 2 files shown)


zklaus:
This PR adds flake8 in places where it was not present before. That is surprising since we have been using it for a long time. Could you please double-check that there are no other places to add it and explain how it is being pulled in at the moment?

valeriupredoi:
> This PR adds flake8 in places where it was not present before. That is surprising since we have been using it for a long time. Could you please double-check that there are no other places to add it and explain how it is being pulled in at the moment?

looking at this right now, man, am a bit confused but I'll figure my way round in a jiffy :+1:

valeriupredoi:
OK Ive tested the syntax 'pytest-flake8>=1.0.6;flake8<4' and that don't work, now it's pinned in the test section, I think it's all fine now :+1:

schlunma:
We already have a 2.5 milestone, right? I don't think we need a dedicated project for that.

valeriupredoi:
ah gut then, no not a project, used the wrong term :grin:

valeriupredoi:
we should get this in the sooner the better, that `_io.String` error instead of the actual flake8 problem is starting to get me :rage:

zklaus (approved):
Thanks, looking good. Could you please also:
- Check ESMValTool for the same requirement
- Read https://conda-forge.org/docs/maintainer/adding_pkgs.html#constraining-packages-at-runtime and let me know what you think how we should proceed on the feedstock
- Possibly open an issue and/or PR at the feedstock?

Thanks! :beers:

valeriupredoi:
ace @zklaus :beer: The Tool PR for this is here https://github.com/ESMValGroup/ESMValTool/pull/2492 - I'll have a look at the feedstock situation now :+1:

valeriupredoi:
this https://conda-forge.org/docs/maintainer/adding_pkgs.html#constraining-packages-at-runtime returns a Unable to connect in FF, in fact https://conda-forge.org/ seems to be down - even their Status is not loading, hope it be back :grin:

valeriupredoi:
OK so the question is: do we want to restrict `flake8` in the conda-forge package - do we distribute the tests with the package and if so, do they actually get run? I'd say we should, and for that I think we just have to add the pin in the `test: requires:` section or maybe just add the recommended `run_constraint:` section in `test:` - I am not entirely sure what the difference would be - do you know, Klaus?

valeriupredoi:
anyway I took the discussion here https://github.com/conda-forge/esmvalcore-feedstock/issues/15

PR #1444, opened 2022-01-28
Add new `JULIA_LOAD_PATH` to diagnostic task test

### LINKED ISSUE #1443: [tests failing] `JULIA_LOAD_PATH` has changed and is causing test failures
`JULIA_LOAD_PATH` has changed its formatting and it now gets a bunch of garble at the end eg `'JULIA_LOAD_PATH': '/tmp/pytest-of-valeriu/pytest-137/popen-gw1/test_initialize_env__jl_0/esmvaltool/install/Julia:@:@whynumba:@stdlib'` where the first bit `whynumba` is my conda environment name and `stdlib` is guff; it causes the task tests to fail eg [here](https://github.com/ESMValGroup/ESMValCore/runs/4969579136?check_suite_focus=true) - the env variable is set in `_task.py` at l.396:
```python
            # Set the julia virtual environment
            env['JULIA_LOAD_PATH'] = "{}:{}".format(
                DIAGNOSTICS.path / 'install' / 'Julia',
                os.environ.get('JULIA_LOAD_PATH', ''),
            )
```
and the garble is not localized to my machine or other specific machines, indeed if one runs this small script:
```python
import os

vg = os.environ.get('JULIA_LOAD_PATH')
print(vg)
```
gets that garble: `@:@whynumba:@stdlib` in my case. Note that the `@whynumba` is an env variable on its own:

`'JULIA_PROJECT': '@whynumba'`

no idea what `@stdlib` is in this case.

Julia installed:
```
(whynumba) valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ conda list julia
# packages in environment at /home/valeriu/miniconda3/envs/whynumba:
#
# Name                    Version                   Build  Channel
julia                     1.7.1                h989b2f6_3    conda-forge
```
reverting to a previous `_2` build is not fixing this. Currently I have no idea what is causing this, I'll investigate more tomorrow.

valeriupredoi:
## Description

See https://github.com/ESMValGroup/ESMValCore/issues/1443 for reference and specifically our Julia friend's @ngam comments https://github.com/ESMValGroup/ESMValCore/issues/1443#issuecomment-1023503304 and https://github.com/ESMValGroup/ESMValCore/issues/1443#issuecomment-1023506785 - cheers for that pointer! Basically, when Julia is fully installed from conda-forge the path encapsulated in `JULIA_LOAD_PATH` is no longer just `Julia` but also `@:@$CONDA_ENV:@stdlib`; the test I fix here fails only when there is a Julia installation, otherwise it goes through.

Closes #1443 

Link to documentation:

***

***


zklaus:
I'm not sure it's wise to rely on the exact form of the additional conda-forge stuff in the Julia path. As you have heard from the other guys, this is still prone to change, plus, if the user uses a different way of installing Julia, this might be different anyhow.

Why don't we just check that the first component (`"...".split(":")[0]`) is the right thing?

valeriupredoi:
> I'm not sure it's wise to rely on the exact form of the additional conda-forge stuff in the Julia path. As you have heard from the other guys, this is still prone to change, plus, if the user uses a different way of installing Julia, this might be different anyhow.
> 
> Why don't we just check that the first component (`"...".split(":")[0]`) is the right thing?

good call actually! Prob more agnostic, as you say :+1:

tests/unit/task/test_diagnostic_task.py
```diff
@@ -39,20 +38,17 @@ def test_initialize_env(ext, tmp_path, monkeypatch):
     env = {}
     if ext in ('.jl', '.py'):
         env['MPLBACKEND'] = 'Agg'
+    if ext in ('.ncl', '.R'):
+        env['diag_scripts'] = str(diagnostics_path)
```
zklaus, line 42: Let's not move this around. Unless there is a reason for that that I am missing?

tests/unit/task/test_diagnostic_task.py
```diff
@@ -39,20 +38,17 @@ def test_initialize_env(ext, tmp_path, monkeypatch):
...
-        julia_project = os.environ.get('JULIA_PROJECT')
-        if julia_project is not None:
-            extras = ':@:' + julia_project + ':@stdlib'
-            julia_path = julia_path + extras
-
-        env['JULIA_LOAD_PATH'] = julia_path
-    if ext in ('.ncl', '.R'):
-        env['diag_scripts'] = str(diagnostics_path)
+        task.env['JULIA_LOAD_PATH'] = \
+            task.env['JULIA_LOAD_PATH'].split(":")[0]
```
zklaus, line 51: I would try to avoid modifying task.env in-place.

zklaus (changes requested):

tests/unit/task/test_diagnostic_task.py
valeriupredoi: well you can't check for the truncated `JULIA_LOAD_PATH` otherwise, since the `task.env` dict contains the one with the crazy `@:@$CONDA_ENV:@stdlib`

tests/unit/task/test_diagnostic_task.py
valeriupredoi: ok let me move it back, shouldn't matter too much

tests/unit/task/test_diagnostic_task.py
zklaus: Sure you can. Here are three ways:
```
test_env = copy.deep_copy(task.env)
...
```
```
env = ":".join(["..."] + task.env[...].split(":")[1:]
```
```
assert task.env[...].split(":")[0] == "..."
assert task.env["MPLBACKEN"] == "Agg"
...
```
My personal favorite is number three.

tests/unit/task/test_diagnostic_task.py
valeriupredoi: moved it back :+1:

tests/unit/task/test_diagnostic_task.py
valeriupredoi: ah with copying, yeh I don't like copying stuff, but I guess it saves us from changing it in place...OK then, I'll do it this way :+1:

valeriupredoi: what you reckon, Klaus? The PITA Monitor test is dying coz of the pymon crappe, I still need to go back to that PR and figure out why those ancestors failures when using the `--db` flag, all else looks fine tho

valeriupredoi:
what you reckon, Monsieur Klaus @zklaus ? :beer:

zklaus (approved):

valeriupredoi:
thanky, sir!

PR #1448, opened 2022-01-31
Generic regridding preprocessor

zklaus:
## Description
This adds the long-promised iris-esmf-regrid based regridding to the regrid preprocessor.
It does so in a generic way, meaning the preprocessor is not actually adapted to the iris-esmf-regrid package, but rather gets the ability to load any kind of Iris regridding scheme by specifying the corresponding factory, also allowing for the passing of arbitrary arguments.

I'd like to solicit some early feedback!

Do you think this is feasible, i.e. using a generic loading mechanism?

If you want to try it out, don't forget to create a new environment that contains the iris-esmf-regrid conda package as mentioned in the environment.yml in this PR (of course you can also just install it in your environment).
You can use the corresponding branch iris-esmf-regrid and the `recipe_ocean_multimap.yml` in it to try things out.

Documentation:
- Preprocessor docs: https://esmvaltool--1448.org.readthedocs.build/projects/ESMValCore/en/1448/recipe/preprocessor.html#regridding-interpolation-extrapolation-schemes
- API docs: https://esmvaltool--1448.org.readthedocs.build/projects/ESMValCore/en/1448/api/esmvalcore.preprocessor.html#esmvalcore.preprocessor.regrid

***

### DIFF
--- a/doc/recipe/preprocessor.rst
+++ b/doc/recipe/preprocessor.rst
@@ -931,7 +931,7 @@ example of its usage in an ESMValTool preprocessor is:
         regrid:
           target_grid: 2.5x2.5
           scheme:
-            reference: esmf_regrid:ESMFAreaWeighted
+            reference: esmf_regrid.schemes:ESMFAreaWeighted
             mdtol: 0.7
 
 .. TODO: Remove the following warning once things have settled a bit.
--- a/doc/recipe/preprocessor.rst
+++ b/doc/recipe/preprocessor.rst
@@ -829,6 +829,16 @@ The arguments are defined below:
 Regridding (interpolation, extrapolation) schemes
 -------------------------------------------------
 
+ESMValTool has a number of built-in schemes, which are presented in
+:ref:`built-in regridding schemes`. Additionally, it is also possible to use
+third party regridding schemes designed for use with :doc:`Iris
+<iris:index>`. This is explained in :ref:`generic regridding schemes`.
+
+.. _built-in regridding schemes:
+
+Built-in regridding schemes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
 The schemes used for the interpolation and extrapolation operations needed by
 the horizontal regridding functionality directly map to their corresponding
 implementations in :mod:`iris`:
@@ -866,6 +876,67 @@ See also :func:`esmvalcore.preprocessor.regrid`
    degrees should not produce any memory-related issues, but be advised that
    for resolutions of ``< 0.5`` degrees the regridding becomes very slow and
    will use a lot of memory.
+
+.. _generic regridding schemes:
+
+Generic regridding schemes
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:ref:`Iris' regridding <iris:interpolation_and_regridding>` is based around the
+flexible use of so-called regridding schemes. These are classes that know how
+to transform a source cube with a given grid into the grid defined by a given
+target cube. Iris itself provides a number of useful schemes, but they are
+largely limited to work with simple, regular grids. Other schemes can be
+provided independently. This is interesting when special regridding-needs arise
+or when more involved grids and meshes need to be considered. Furthermore, it
+may be desirable to have finer control over the parameters of the scheme than
+is afforded by the built-in schemes described above.
+
+To facilitate this, the :func:`~esmvalcore.preprocessor.regrid` preprocessor
+allows the use of any scheme designed for Iris. The scheme must be installed
+and importable. To use this feature, the ``scheme`` key passed to the
+preprocessor must be a dictionary instead of a simple string that contains all
+necessary information. That includes a ``reference`` to the desired scheme
+itself, as well as any arguments that should be passed through to the
+scheme. For example, the following shows the use of the built-in scheme
+:class:`iris.analysis.AreaWeighted` with a custom threshold for missing data
+tolerance.
+
+.. code-block:: yaml
+
+    preprocessors:
+      regrid_preprocessor:
+        regrid:
+          target_grid: 2.5x2.5
+          scheme:
+            reference: iris.analysis:AreaWeighted
+            mdtol: 0.7
+
+The value of the ``reference`` key has two parts that are separated by a
+``:`` with no surrounding spaces. The first part is an importable Python
+module, the second refers to the scheme, i.e. some callable that will be called
+with the remaining entries of the ``scheme`` dictionary passed as keyword
+arguments.
+
+One package that aims to capitalize on the :ref:`support for unstructured
+meshes introduced in Iris 3.2 <iris:ugrid>` is
+:doc:`iris-esmf-regrid:index`. It aims to provide lazy regridding for
+structured regular and irregular grids, as well as unstructured meshes. An
+example of its usage in an ESMValTool preprocessor is:
+
+.. code-block:: yaml
+
+    preprocessors:
+      regrid_preprocessor:
+        regrid:
+          target_grid: 2.5x2.5
+          scheme:
+            reference: esmf_regrid:ESMFAreaWeighted
+            mdtol: 0.7
+
+.. warning::
+   Just as the mesh support in Iris itself, this new regridding package is
+   still considered experimental.
 
 .. _ensemble statistics:
 
--- a/doc/recipe/preprocessor.rst
+++ b/doc/recipe/preprocessor.rst
@@ -829,7 +829,7 @@ The arguments are defined below:
 Regridding (interpolation, extrapolation) schemes
 -------------------------------------------------
 
-ESMValTool has a number of built-in schemes, which are presented in
+ESMValTool has a number of built-in regridding schemes, which are presented in
 :ref:`built-in regridding schemes`. Additionally, it is also possible to use
 third party regridding schemes designed for use with :doc:`Iris
 <iris:index>`. This is explained in :ref:`generic regridding schemes`.
--- a/doc/recipe/preprocessor.rst
+++ b/doc/recipe/preprocessor.rst
@@ -934,6 +934,7 @@ example of its usage in an ESMValTool preprocessor is:
             reference: esmf_regrid:ESMFAreaWeighted
             mdtol: 0.7
 
+.. TODO: Remove the following warning once things have settled a bit.
 .. warning::
    Just as the mesh support in Iris itself, this new regridding package is
    still considered experimental.
--- a/esmvalcore/preprocessor/_regrid.py
+++ b/esmvalcore/preprocessor/_regrid.py
@@ -1,5 +1,6 @@
 """Horizontal and vertical regridding module."""
 
+import importlib
 import logging
 import os
 import re
@@ -302,7 +303,7 @@ def _regional_stock_cube(spec: dict):
 
 def _attempt_irregular_regridding(cube, scheme):
     """Check if irregular regridding with ESMF should be used."""
-    if scheme in ESMF_REGRID_METHODS:
+    if isinstance(scheme, str) and scheme in ESMF_REGRID_METHODS:
         try:
             lat_dim = cube.coord('latitude').ndim
             lon_dim = cube.coord('longitude').ndim
@@ -503,7 +504,14 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
     --------
     extract_levels : Perform vertical regridding.
     """
-    if HORIZONTAL_SCHEMES.get(scheme.lower()) is None:
+    if isinstance(scheme, dict):
+        module_name, scheme_name = scheme.pop("scheme").rsplit(".", 1)
+        scheme_module = importlib.import_module(module_name)
+        raw_scheme = getattr(scheme_module, scheme_name)
+        loaded_scheme = raw_scheme(**scheme)
+    else:
+        loaded_scheme = HORIZONTAL_SCHEMES.get(scheme.lower())
+    if loaded_scheme is None:
         emsg = 'Unknown regridding scheme, got {!r}.'
         raise ValueError(emsg.format(scheme))
 
@@ -562,7 +570,7 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
         if _attempt_irregular_regridding(cube, scheme):
             cube = esmpy_regrid(cube, target_grid, scheme)
         else:
-            cube = cube.regrid(target_grid, HORIZONTAL_SCHEMES[scheme])
+            cube = cube.regrid(target_grid, loaded_scheme)
 
         # Preserve dtype and use masked arrays for 'unstructured_nearest'
         # scheme (see https://github.com/SciTools/iris/issues/4463)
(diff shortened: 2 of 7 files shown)


schlunma:
I just successfully tested this PR with a couple if `iris` schemes (e.g. `iris.analysis.Linear`) :tada: It works really well and just like expected.

Three small comments:
- If an invalid scheme is provided, the error message can be really cryptic. Might be useful to catch the original exception and use a `raise from` with a more user-friendly message.
- Similarly, if `scheme` is not given in the `dict`, this leads to a `KeyError: 'scheme'`. This message should be improved.
- I think we should rename the `scheme` key in the `dict` to something else to avoid duplicating the word "scheme". Maybe something like `name`, `module`, or `method`?

```yml
  regrid:
    regrid:
      target_grid: 0.5x0.5
      scheme:
        name/module/method: iris.analysis.Linear
        extrapolation_mode: mask
```

esmvalcore/preprocessor/_regrid.py
```diff
@@ -503,7 +504,14 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
     --------
     extract_levels : Perform vertical regridding.
     """
-    if HORIZONTAL_SCHEMES.get(scheme.lower()) is None:
+    if isinstance(scheme, dict):
+        module_name, scheme_name = scheme.pop("scheme").rsplit(".", 1)
+        scheme_module = importlib.import_module(module_name)
```
valeriupredoi, line 509: I would sneak in an `ImportError` exception catcher here

valeriupredoi (changes requested):
nice idea, I likes! One quick prelim suggestion from me, and of course, as Lenin used to say - tests, tests and tests again :grin: (lemme know I can help with the tests)

zklaus:
Alright, thanks @schlunma and @valeriupredoi. I'll take that as an initial go-ahead and will come back with a more fleshed out version.

schlunma:
@zklaus You think this will be ready in time for v2.5.0? I think it would be nice to include this.

zklaus:
Now on to writing some tests.

zklaus:
The main purpose of this PR is to allow the use of iris-esmf-regrid to enable faster, lazy regridding all around and to facilitate regridding of unstructured grids with the new iris mesh support. However, the mechanism introduced here is more general, and hence I introduce neither iris-esmf-regrid as a dependency, nor extensive specific documentation. For documentation we rely on the documentation of iris-esmf-regrid itself; as a dependency this should be introduced alongside the recipe that makes use of it.

Of course, this can be debated. If you prefer to integrate things more tightly here, we can take this now, or in a later PR for the next version. That would still allow us to do more experimentation with the released core in the next development cycle of the tool.

schlunma:
Great, thanks Klaus!

I think it would have been nice to already include the `iris-esmf-regrid` dependency with this PR, but I'm also fine with just adding this framework. Please make sure to open an issue about this (with the label v2.6) so we don't forget about this.

Will have a look at this PR now!

schlunma:
One further thought about this after reading the doc: I noticed that you use `iris-esmf-regrid` quite often in the doc. If `iris-esmf-regrid` is mentioned there, I think we need to include this.

Would it make sense to include it as a dependency, add a very small test, e.g.,

```py
def test_regrid_generic_regridding(self):
    regrid(self.src_cube, '1x1', {"reference": "esmf_regrid.schemes:ESMFAreaWeighted"})
```

and add a note to the doc that this feature is still "experimental"?

esmvalcore/preprocessor/_regrid.py
```diff
@@ -478,13 +479,12 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
...
-        ``linear``,
-        ``linear_extrapolate``,
-        ``nearest``,
-        ``area_weighted``,
+    scheme : str or dict
+        The regridding scheme to perform. If both source and target grid are
+        structured (regular or irregular), can be one of the built-in schemes
+        ``linear``, ``linear_extrapolate``, ``nearest``, ``area_weighted``,
         ``unstructured_nearest``.
+        Alternatively, a `dict` that specifies generic regridding (see below).
```
schlunma, line 487:
```suggestion
        Alternatively, a ``dict`` that specifies generic regridding (see below).
```

esmvalcore/preprocessor/_regrid.py
```diff
@@ -502,8 +502,71 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
     See Also
     --------
     extract_levels : Perform vertical regridding.
+
+    Notes
+    -----
+    This preprocessor allows for the use of arbitrary Iris regridding schemes,
```
schlunma, line 508:
```suggestion
    This preprocessor allows for the use of arbitrary :mod:`iris` regridding schemes,
```

esmvalcore/preprocessor/_regrid.py
```diff
@@ -502,8 +502,71 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
...
+    that is anything that can be passed as a scheme to
+    :meth:`iris.cube.Cube.regrid` is possible. This enables the use of further
+    parameters for existing schemes, as well as the use of more advanced
+    schemes for example for unstructured meshes.
+    To use this functionality, a dictionary must be passed for the scheme with
+    a mandatory entry of `reference` in the form specified for the object
```
schlunma, line 514:
```suggestion
    a mandatory entry of ``reference`` in the form specified for the object
```

esmvalcore/preprocessor/_regrid.py
```diff
@@ -502,8 +502,71 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
...
+    reference of the `entry point data model <https://packaging.python.org/en/
+    latest/specifications/entry-points/#data-model>`_,
+    i.e. `importable.module:object.attr`. This is used as a factory for the
```
schlunma, line 517:
```suggestion
    i.e. ``importable.module:object.attr``. This is used as a factory for the
```

esmvalcore/preprocessor/_regrid.py
```diff
@@ -502,8 +502,71 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
...
+
+        my_preprocessor:
+          regrid:
+            target: 1x1
+            scheme:
+              reference: iris.analysis:Linear
+              extrapolation_mode: nanmask
+
+    To use the area weighted regridder available in
+    :class:`esmf_regrid.ESMFAreaWeighted`, make sure that `iris-esmf-regrid` is
```
schlunma, line 534:
```suggestion
    :class:`esmf_regrid.ESMFAreaWeighted`, make sure that ``iris-esmf-regrid`` is
```

schlunma (commented):
A couple of minor comments on the doc. It would be great if you can also expand the doc [here](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/recipe/preprocessor.html#regridding-interpolation-extrapolation-schemes).

About the failing test: I just installed ESMValTool and ESMValCore in development mode on my machine without any problems. I also successfully ran a recipe and have a working version of `iris-esmf-regrid` (ran your test successfully). No idea what's going wrong in the CI machine...

esmvalcore/preprocessor/_regrid.py
zklaus: Numpy style docstrings like we use should use single backticks here, I believe (see https://numpydoc.readthedocs.io/en/latest/format.html#common-rest-concepts).

esmvalcore/preprocessor/_regrid.py
zklaus: Hm. It's not realy about the `iris` module, is it? How about
```suggestion
    This preprocessor allows for the use of arbitrary :doc:`iris:index` regridding schemes,
```

zklaus:
I know what's wrong. There is no PyPI package for iris-esmf-regrid yet. Consequently, any attempt to install it directly via pip fails. I don't think that is much of a limitation in reality, but I will remove the package from `setup.py` for now and start the process of getting a PyPI package for iris-esmf-regrid up.

zklaus:
Actually forget my last comment (that I have now hidden). The package is there. I'll investigate. I will say that I got a response of "currently the PyPI servers are overloaded" when trying `pip search esmf_regrid` locally before, but I am not completely sure that was the problem for the CI as well.

valeriupredoi:
can't find it, man - use the web based search https://pypi.org/search/?q=iris-esmf-regrid

valeriupredoi:
this one maybe? Says it's the bridge between Iris and ESMF (A Bridge Too Far, if you asked me :laughing: ) https://pypi.org/project/esmf-regrid/

zklaus:
Yeah, of course that's the one. As named in setup.py. But what are you trying to say?

valeriupredoi:
nothing, I basically just landed in this discussion and still need to figure my way around - maybe I should not comment until then :laughing:

esmvalcore/preprocessor/_regrid.py
schlunma: I'm not sure how this is rendered, so please just make the change. You can also ignore this comment if you like :smile:

zklaus: I made a change in the spirit and [here](https://esmvaltool--1448.org.readthedocs.build/projects/ESMValCore/en/1448/api/esmvalcore.preprocessor.html#esmvalcore.preprocessor.regrid) is how it renders. Let me know if you would like a further change.

schlunma:
Ahh...well that's bad. I don't think there is an easy way to solve this apart from dropping support for python 3.7, right?

In that case it's probably best to remove it from the environment again (and also remove the test) and explicitly mention in the doc that it needs to be installed manually and requires python 3.8.

valeriupredoi:
Klaus opened an issue here https://github.com/SciTools-incubator/iris-esmf-regrid/issues/161 - I've been snooping around (mostly looking at their [conda lock file](https://github.com/SciTools-incubator/iris-esmf-regrid/blob/main/requirements/nox.lock/py38-linux-64.lock)) and it's not obvious to me at all why Python<3.9 - all deps seem to be at their latest versions but built specifically against Python=3.8

schlunma:
Ahh, it's python=3.8, not python>=3.8, misunderstood that :grimacing:

valeriupredoi:
:snake: < :three: :nine: Manu

zklaus:
I think in any case it is reasonable to remove it from the environment. After all, it really is only an example (from the point of view of the generic regridder).

schlunma:
Agreed, but then we need to make it crystal clear in the documentation that `iris-esmf-regrid` needs to be installed manually and that only python<3.8 is supported.

zklaus:
I have removed the dependency and improved the tests.

I am not so sure we need to be very forceful about how to install iris-esmf-regrid.

First, it really is only an example. If you want to use a different regridder or advanced features of the Iris regridders (perhaps some that will be added in the future), this works just as well, so I don't want to focus too much on this particular package.
Second, practically, it will not be much of an issue. Most people will not insist on a Python version and just ask for things to be installed, which will work out fine. We run into the problem in the CI because we test all versions.
Third, it is likely to change and since we also state that it is experimental, people should not be surprised with hiccups like this.

So imho it's better to have simple, accessible documentation and to refer to the iris-esmf-regrid documentation for further details.


Now on to writing the extended documentation requested by @schlunma.

zklaus:
Done. One last round of reviews, @schlunma? @valeriupredoi do you want to have a look too and see if your blocking change request is sufficiently addressed?

valeriupredoi:
cheers, Klaus! I'll uncork meself in a minute :+1:

doc/recipe/preprocessor.rst
```diff
@@ -867,6 +877,67 @@ See also :func:`esmvalcore.preprocessor.regrid`
...
+          target_grid: 2.5x2.5
+          scheme:
+            reference: iris.analysis:AreaWeighted
+            mdtol: 0.7
+
+The value of the ``reference`` key has two parts that are separated by a
+``:`` with no surrounding spaces. The first part is an importable Python
+module, the second refers to the scheme, i.e. some callable that will be called
+with the remaining entries of the ``scheme`` dictionary passed as keyword
+arguments.
```
valeriupredoi, line 919:
```suggestion
argument.
```

esmvalcore/preprocessor/_regrid.py
```diff
@@ -502,8 +502,71 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
...
+
+        my_preprocessor:
+          regrid:
+            target: 1x1
+            scheme:
+              reference: esmf_regrid:ESMFAreaWeighted
+
+    .. note::
+
+        Note that :doc:`iris-esmf-regrid:index` is still experimental.
```
valeriupredoi, line 547: maybe add a TODO so we remove this when it's not experimental anymore?

valeriupredoi (approved):
very nice, great work, Klaus! Juat a couple typos from me, all good :beers:

esmvalcore/preprocessor/_regrid.py
zklaus: Yeah, I was thinking the same thing. In fact, I had even typed it out and then removed it again. Do we have a standard way to note this kind of TODO? I think up to now we only have used issues for that purpose, no?

valeriupredoi: don't think we have any (only for deprecated stuff), maybe just plug one simple one and in 4 years we'll remove it if we don't get nuked in the meantime :grin:

doc/recipe/preprocessor.rst
valeriupredoi: well, it's one dictionary = one keyword argument, but am not fussed, keep it as it is, me and grammatical correctness are not friends anyway

valeriupredoi: ah gotcha, cheers, man! All clear for take off :airplane:

schlunma (commented):
Cheers Klaus, very nice work!

I tested this with actual data and it seems to work fine.

One very tiny comment on the doc.

esmvalcore/preprocessor/_regrid.py
```diff
@@ -502,8 +502,71 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
...
+
+        my_preprocessor:
+          regrid:
+            target: 1x1
+            scheme:
+              reference: iris.analysis:Linear
+              extrapolation_mode: nanmask
+
+    To use the area weighted regridder available in
+    :class:`esmf_regrid.ESMFAreaWeighted`, make sure that ``iris-esmf-regrid``
```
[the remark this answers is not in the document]
zklaus: Good catch. The reason is that the scheme really is in `esmf_regrid.schemes.ESMFAreaWeighted` but made available via some import into `esmf_regrid`. For now, I have made the module explicit; maybe we can simplify it again later when things have settled.

schlunma (approved):

PR #1453, opened 2022-02-01
Synchronized config-user.yml with version from ESMValTool

schlunma:
## Description

Apply changes from ESMValGroup/ESMValTool#2507

***

***

### DIFF
--- a/esmvalcore/config-user.yml
+++ b/esmvalcore/config-user.yml
@@ -1,14 +1,32 @@
 ###############################################################################
-# User's configuration file for the ESMValTool
+# Example user configuration file for ESMValTool
+###############################################################################
+#
+# Note for users:
+# --------------
+# Site-specific entries for different HPC centers are given at the bottom of
+# this file. Comment out/replace as needed.
+#
+###############################################################################
+#
+# Note for developers:
+# -------------------
+# Two identical copies of this file (``ESMValTool/config-user-example.yml`` and
+# ``ESMValCore/esmvalcore/config-user.yml``) exist. If you change one of it,
+# make sure to apply the changes to the other.
+#
 ###############################################################################
 ---
 
 # Directory where results are stored
-output_dir: ./esmvaltool_output
+output_dir: ~/esmvaltool_output
+
 # Directory for storing downloaded climate data
 download_dir: ~/climate_data
+
 # Auxiliary data directory (used for some additional datasets)
 auxiliary_data_dir: ~/auxiliary_data
+
 # Run at most this many tasks in parallel [null]/1/2/3/4/..
 # Set to null to use the number of available CPUs.
 # If you run out of memory, try setting max_parallel_tasks to 1 and check the
@@ -17,25 +35,34 @@ auxiliary_data_dir: ~/auxiliary_data
 # can increase the number of parallel tasks again to a reasonable number for
 # the amount of memory available in your system.
 max_parallel_tasks: null
+
 # Disable the automatic download of missing CMIP3, CMIP5, CMIP6, CORDEX,
 # and obs4MIPs data from ESGF by default. This is useful if you are working
 # on a computer without an internet connection.
 offline: true
+
 # Log level of the console
 log_level: info
+
 # Exit on warning (only for NCL diagnostic scripts)? true/[false]
 exit_on_warning: false
+
 # Plot file format? [png]/pdf/ps/eps/epsi
 output_file_type: png
+
 # Remove the preproc dir if all fine
-remove_preproc_dir: false
+remove_preproc_dir: true
+
 # Use netCDF compression true/[false]
 compress_netcdf: false
+
 # Save intermediary cubes in the preprocessor true/[false]
 save_intermediary_cubes: false
+
 # Path to custom config-developer file, to customise project configurations.
 # See config-developer.yml for an example. Set to None to use the default
 config_developer_file: null
+
 # Get profiling information for diagnostics
 # Only available for Python diagnostics
 profile_diagnostic: false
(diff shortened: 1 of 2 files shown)


valeriupredoi (approved):
cheers, Manu!

esmvalcore/config-user.yml
```diff
@@ -45,19 +72,30 @@ profile_diagnostic: false
...
+  CMIP3: [~/cmip3_inputpath1, ~/cmip3_inputpath2]
+  CMIP5: [~/cmip5_inputpath1, ~/cmip5_inputpath2]
+  CMIP6: [~/cmip6_inputpath1, ~/cmip6_inputpath2]
+  OBS: ~/obs_inputpath
+  OBS6: ~/obs6_inputpath
+  obs4MIPs: ~/obs4mips_inputpath
+  ana4mips: ~/ana4mips_inputpath
+  native6: ~/native6_inputpath
+  RAWOBS: ~/rawobs_inputpath
+  default: ~/default_inputpath
```
bouweandela, line 84: What was the reason for changing this back to the old default from before #1217? I liked the new default value here because it would work for the data that gets automatically downloaded with the ``--offline=False`` method.

I think the old example was confusing because it contains a mix of lists of paths and paths.

esmvalcore/config-user.yml
```diff
@@ -45,19 +72,30 @@ profile_diagnostic: false
...
-  CMIP3: ESGF
-  CMIP5: ESGF
-  CMIP6: ESGF
-  CORDEX: ESGF
-  obs4MIPs: ESGF
+  CMIP3: default
+  CMIP5: default
+  CMIP6: default
+  CORDEX: default
+  obs4MIPs: default
```
bouweandela, line 94: What was the reason for changing this back to the old default from before #1217? I liked the new default value, because it encourages storing the data in such a structured way, that enables you to find out what version of the files you have. If you download CMIP data and store it with the `default` (i.e. no subdirectories) format, it becomes very hard to tell what version of the files you have.

esmvalcore/config-user.yml
```diff
@@ -1,14 +1,32 @@
...
+# Two identical copies of this file (``ESMValTool/config-user-example.yml`` and
+# ``ESMValCore/esmvalcore/config-user.yml``) exist. If you change one of it,
+# make sure to apply the changes to the other.
+#
 ###############################################################################
 ---
 
 # Directory where results are stored
-output_dir: ./esmvaltool_output
+output_dir: ~/esmvaltool_output
```
bouweandela, line 22: While I agree that this is a sensible default (it's what I always use in my own configuration file), it is not necessarily the most expected. Most tools write to the current working directory, if you do not tell them where to write.

bouweandela:
Note that the defaults are also set here: https://github.com/ESMValGroup/ESMValCore/blob/7d430722af8a3aee206287edb210eedc87ca4fcb/esmvalcore/_config/_config.py#L105-L123

This pull request changes the defaults only in config-user.yml, leading to the situation where not specifying a value will give you a different default than using the default setting from the example, except if you're using the experimental API which uses the default values from the example.

esmvalcore/config-user.yml
valeriupredoi: well, we don't, or, we write wherever the config tells it to, if that default may lead to spawning a large number of output dirs, the user will probably blame us first upon finding out they have 300 `esmvaltool_output` dirs because they don't know how to do `pwd` and figure out where they are. Also, they may get lost and say they never ran that particular run :grin:

valeriupredoi:
I think @schlunma just wanted to make the example config file more user-friendly, and not operate on any actual defaults, Bouwe

esmvalcore/config-user.yml
valeriupredoi: yes but that assumes an ESGF-like structure (ie you have downloaded data) - we should NOT encourage people to download data

esmvalcore/config-user.yml
schlunma: To be honest I didn't really think about this and used the options that were use in ESMValTool. Showcasing the possibility to use multiple paths and also list all available projects seemed like a good idea. If a user wants to use this file with automatic downloads they need to change the default value of `offline` anyway, and in the file there is a comment:

```
Uncomment the lines below to locate data that has been automatically
# downloaded from ESGF (using "offline: false")
```

I'm also fine with the old one, but don't really see the use. For me this is an example file which needs to be altered anyway.

esmvalcore/config-user.yml
schlunma: See comment above

esmvalcore/config-user.yml
schlunma: I agree with V here, writing to a path that doesn't depend on the `pwd` seems like the better option here.

schlunma:
> Note that the defaults are also set here:
> 
> https://github.com/ESMValGroup/ESMValCore/blob/7d430722af8a3aee206287edb210eedc87ca4fcb/esmvalcore/_config/_config.py#L105-L123
> 
> This pull request changes the defaults only in config-user.yml, leading to the situation where not specifying a value will give you a different default than using the default setting from the example, except if you're using the experimental API which uses the default values from the example.

Actually that is not 100% true. The only options that I changed (apart from the `rootpath` which does not have a default in `ESMValCore/esmvalcore/_config/_config.py`) are `drs`, `remove_preproc_dir` and `output_dir`. `drs` already differed in the old version, `remove_preproc_dir` now actually matches the entry (it didn't before) and `output_dir` doesn't match its entry now. However, there are other entries (e.g. `auxiliary_data`) which do not match their entries in the new and old version.

I agree with you that we should be consistent here, but this PR didn't worsen this. I will open address this in another PR.

esmvalcore/config-user.yml
bouweandela: The idea behind the changes I made in #1217 was to finally make it possible to run the tool for the first time without editing the example config-user.yml, as documented [here](https://docs.esmvaltool.org/en/latest/quickstart/running.html#running-a-recipe). That failed slightly because people objected to automatically downloading the required data, but if you specify `--offline=False` on the command line it still works.

Showcasing per project paths and lists of paths is also good, but may we could add a few more commented-out examples instead to showcase that?

schlunma: Sure, will add that to #1457

PR #1455, opened 2022-02-02
Moved extraction of reference datasets for (horizontal and vertical) regridding into preprocessor functions

### LINKED ISSUE #56: Regridding to reference dataset fails sometimes
Not an urgent issue, but something I noticed:

Regridding to a reference dataset, e.g.

```yaml
regrid_to_ref:
  regrid:
    target_grid: GFDL-CM3
    scheme: linear
```

invokes a simple `iris.load_cube`

https://github.com/ESMValGroup/ESMValTool/blob/43598ee48a327fc3ef2a6e53768324d02ad61863/esmvaltool/preprocessor/_regrid.py#L169

on the **raw** nc file, which may fail in rare cases if the file contains multiple variables (e.g `rsdt` of GFDL-CM3), which leads to errors like this:

```
  File "<stdin>", line 1, in <module>
  File "~/anaconda3/envs/esmvaltool/lib/python3.6/site-packages/iris/__init__.py", line 376, in load_cube
    raise iris.exceptions.ConstraintMismatchError(str(e))
iris.exceptions.ConstraintMismatchError: failed to merge into a single cube.
  cube.long_name differs: 'Length of average period' != 'End time for average period'
  cube.var_name differs: 'average_DT' != 'average_T2'
  cube.units differs: Unit('days') != Unit('days since 1860-01-01 00:00:00', calendar='gregorian')
```

I think we need an advanced loading function here.

### LINKED ISSUE #1454: Vertical regridding to reference levels fails when reference dataset needs to be downloaded
**Describe the bug**

A recipe with the following preprocessor

```yml
  preprocessor_1:
    extract_levels:
      levels: bcc-csm1-1
      scheme: linear
```

fails when the reference dataset (here `bcc-csm1-1`) needs to be downloaded (with `offline=False`) since the loading of this dataset already happens in `_recipe.py` prior to downloading:

https://github.com/ESMValGroup/ESMValCore/blob/7d430722af8a3aee206287edb210eedc87ca4fcb/esmvalcore/_recipe.py#L164-L172

**Please attach**
[recipe_test_reference_levels.yml.txt](https://github.com/ESMValGroup/ESMValCore/files/7985530/recipe_test_reference_levels.yml.txt)
[main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/7985531/main_log_debug.txt)

schlunma:
## Description

This PR moves the extraction of reference datasets used for horizontal and vertical regridding to the dedicated preprocessor functions. This allows us to specify reference datasets that need to be (automatically) downloaded first.

In addition, it adds fixes prior to the loading step of the reference dataset for horizontal regridding, similar to what is already done for the target levels for vertical regridding.

Closes #1454
Closes #56

Link to documentation:

***

***

### DIFF
--- a/esmvalcore/preprocessor/_regrid.py
+++ b/esmvalcore/preprocessor/_regrid.py
@@ -450,10 +450,10 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
 
     Note that the target grid can be a cube (:py:class:`~iris.cube.Cube`),
     path to a cube (``str``), a grid spec (``str``) in the form
-    of `MxN`, or a ``dict`` specifying the target grid.
+    of `MxN`, a ``dict`` specifying the target grid, or a ``dict`` specifying a
+    reference dataset.
 
-    For the latter, the ``target_grid`` should be a ``dict`` with the
-    following keys:
+    To specify a target grid with a ``dict``, the following keys are necessary:
 
     - ``start_longitude``: longitude at the center of the first grid cell.
     - ``end_longitude``: longitude at the center of the last grid cell.
@@ -462,6 +462,26 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
     - ``start_latitude``: latitude at the center of the first grid cell.
     - ``end_latitude``: longitude at the center of the last grid cell.
     - ``step_latitude``: constant latitude distance between grid cell centers.
+
+    To specify a reference dataset with a ``dict``, the following keys are
+    necessary:
+
+    - ``filename``: path to the file that contains the reference grid.
+    - ``project``: project of the reference dataset.
+    - ``dataset``: name of the reference dataset.
+    - ``short_name``: name of the variable in the reference dataset.
+    - ``mip``: name of the MIP table used.
+    - ``frequency``: temporal frequency of the reference dataset.
+    - ``fix_dir``: path to the directory where the new file is ssaved in case
+      fixes are necessary.
+
+    Note
+    ----
+    If ``levels`` is a ``dict`` and it does not contain the key ``filename``,
+    it is automatically assumed that you specified the target grid with
+    ``start_longitude``, ``end_longitude``, etc. If a valid reference dataset
+    is specified, fixes are applied to the reference dataset prior to loading
+    the data.
 
     Parameters
     ----------
@@ -475,9 +495,9 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
         of the form ``MxN``, which specifies the extent of the cell, longitude
         by latitude (degrees) for a global, regular target grid.
 
-        Alternatively, a dictionary with a regional target grid may
-        be specified (see above).
-
+        Alternatively, a dictionary with a regional target grid may or a
+        dictionary describing the reference dataset may be provided (see
+        above).
     scheme : str
         The regridding scheme to perform, choose from
         ``linear``,
@@ -498,6 +518,13 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
     -------
     :py:class:`~iris.cube.Cube`
         Regridded cube.
+
+    Raises
+    ------
+    KeyError
+        If ``target_grid`` is a :obj:`dict` describing a reference dataset
+        (i.e., it contains a ``filename``) and at least one of the other
+        necessary keys is missing.
 
     See Also
     --------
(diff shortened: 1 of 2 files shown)


schlunma:
@valeriupredoi @zklaus Before I adapt the tests here, can I briefly get your feedback on this implementation? Does it make sense in your opinion?

valeriupredoi:
thanks a bunch @schlunma - looking at it now, sorry it escaped my wits yesterday!

esmvalcore/preprocessor/_regrid.py
```diff
@@ -463,6 +463,26 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
...
+    To specify a reference dataset with a ``dict``, the following keys are
+    necessary:
+
+    - ``filename``: path to the file that contains the reference grid.
+    - ``project``: project of the reference dataset.
+    - ``dataset``: name of the reference dataset.
+    - ``short_name``: name of the variable in the reference dataset.
+    - ``mip``: name of the MIP table used.
+    - ``frequency``: temporal frequency of the reference dataset.
+    - ``fix_dir``: path to the directory where the new file is ssaved in case
```
valeriupredoi, line 475:
```suggestion
    - ``fix_dir``: path to the directory where the new file is saved in case
```

esmvalcore/preprocessor/_regrid.py
```diff
@@ -463,6 +463,26 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
...
+      fixes are necessary.
+
+    Note
+    ----
+    If ``levels`` is a ``dict`` and it does not contain the key ``filename``,
+    it is automatically assumed that you specified the target grid with
+    ``start_longitude``, ``end_longitude``, etc. If a valid reference dataset
```
valeriupredoi, line 482: wait what? A grid is specified by an MxN specification, not by defining a box, am confusado here

valeriupredoi: ah nevermind, it's the `_spec_to_latlonvals()` stuff - can you add a pointer to that func here maybe, that confused me, prob gonna confuse somebody else too

valeriupredoi (changes requested):
draft request changes to a draft PR, not gonna be anal before it's actually RfR :grin:

valeriupredoi:
so here's the thing - as I understand it, your are now passing all the basic ingredients to the regrid preprocessor, rather than have some bits done first in recipe and then others (like the reference level extraction) be done via the regridder preprocessor - I like it! I like that stuff gets done by the module that needs to do it, and not outside it. What happens if there's an issue with the reference dataset, other than a CMOR issue, that might not even be picked up if the user chooses to run with all CMOR shields down? Will that surface only when the regridder is called? That might be after a lot of stuff has already been computed

schlunma:
> What happens if there's an issue with the reference dataset, other than a CMOR issue, that might not even be picked up if the user chooses to run with all CMOR shields down? Will that surface only when the regridder is called? That might be after a lot of stuff has already been computed

It will surface once the reference dataset has entered the preprocessor chain, so probably a little bit earlier than the actual regridding. I agree that this is not ideal because it also increases the computation times by duplicating the loading for every dataset, but I don't see another way right now.

Something that might be possible is to add an additional step between the check for data availability (and the downloading) and the actual start of the preprocessing, but this is something I can't do for v2.5.

I'm not sure how to proceed. Should I try to get this in or leave it to v2.6 and come up with a cleaner solution?

valeriupredoi:
thanks for the note, Manu! The question is what's the immediate benefit of this vs the cost it introduces and vs a longer mulling over and possible implementation of a download me/check me/use me function for datasets before the preprocessor starts? I'd also want to hear what @zklaus and @bouweandela think about this :beer:

schlunma:
The short-term benefit is that the bug mentioned in #1454 is fixed.

But after giving this some more thoughts I think this should be implemented properly, e.g., by adding a function `_load_reference_datasets()` between the `download` and `run` here:

https://github.com/ESMValGroup/ESMValCore/blob/3a3e5f537c03a96027f5f91650d9e0c1d405fe42/esmvalcore/_recipe.py#L1752-L1763

Maybe I can address this for v2.5, but if not we can do a bugfix or even postpone this to v2.6, I don't think it's that urgent.

valeriupredoi:
thanks, Manu! I am of the personal opinion that this should not be hurried up, but rather done in a take it easy, test proper manner - up to you and the others, am not gonna block progress :cn: :grin:

bouweandela:
This would also be addressed (differently) in #1609.

PR #1507, opened 2022-02-21
Update README.md

thomascrocker:
Readme updated with details of why this fork exists and the release tag that was used to run the recipes.


thomascrocker:
Hi @Peter9192 I made changes to the readme file in main, and have now switched the default branch to my one (I think it is probably clearer overall that way) so want to merge those readme changes in... but the repo won't let me merge without an approved review!

thomascrocker:
Ah, I have just noticed this has created the PR in the ESMVal group and not my fork... Will delete.

PR #1508, opened 2022-02-23
Added tests for duplicates in changelog and removed duplicates

schlunma:
## Description

This PR copies the test for duplicate entries from ESMValTool and removes duplicates that have been found by it.

***

### DIFF
--- a/tests/unit/test_docs.py
+++ b/tests/unit/test_docs.py
@@ -1,0 +1,34 @@
+"""Tests for the documentation."""
+import collections
+import os
+import re
+
+# For entries in the following list, two appearances (but not more) are
+# allowed. This can be useful if a pull request needs to appear in multiple
+# sections.
+ALLOWED_EXCEPTIONS = []
+
+
+def test_duplications_in_changelog():
+    changelog_path = os.path.join(os.path.dirname(__file__), '../..',
+                                  'doc/changelog.rst')
+    with open(changelog_path) as changelog:
+        changelog = changelog.read()
+
+    # Find all pull requests
+    pr_links = re.compile(
+        "<https://github.com/ESMValGroup/ESMValCore/pull/[0-9]+>")
+    links = pr_links.findall(changelog)
+
+    # Remove one entry for the allowed exceptions
+    for pr_link in ALLOWED_EXCEPTIONS:
+        if pr_link in links:
+            links.remove(pr_link)
+
+    # Check for duplicates
+    if len(links) != len(set(links)):
+        print('The following PR are duplicated in the changelog:')
+        print('\n'.join((link
+                         for link, count in collections.Counter(links).items()
+                         if count > 1)))
+        assert False
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -191,7 +191,6 @@ Preprocessor
 Automatic testing
 ~~~~~~~~~~~~~~~~~
 
--  Pin Python to 3.9 in environment.yml on CircleCI and skip mypy tests in conda build (`#1176 <https://github.com/ESMValGroup/ESMValCore/pull/1176>`__) `Bouwe Andela <https://github.com/bouweandela>`__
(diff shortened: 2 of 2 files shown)


schlunma:
Why is this failing now?

```
Need type annotation for "ALLOWED_EXCEPTIONS" (hint: "ALLOWED_EXCEPTIONS: List[<type>] = ...")
```

There are 100s of other tests withouth type annotations...

tests/unit/test_docs.py
```diff
@@ -0,0 +1,35 @@
+"""Tests for the documentation."""
```
valeriupredoi, line 1: cheers Manu! Here's a general point about this test module: I am happy to see a start of actually testing the documentation outside the CI test that just builds it, but I'd be happier to see a dedicated test dir created where we put different test modules for the documentation so e.g. for this case `tests/unit/documentation/test_changelog.py` - what do you think?

tests/unit/test_docs.py
```diff
@@ -0,0 +1,35 @@
...
+import collections
+import os
+import re
+from typing import List
+
+# For entries in the following list, two appearances (but not more) are
+# allowed. This can be useful if a pull request needs to appear in multiple
+# sections.
+ALLOWED_EXCEPTIONS: List[str] = []
```
valeriupredoi, line 10: quite baffled by the need for this variable :confused:

valeriupredoi (changes requested):
requesting structural changes, nothing code-related from me. Also pinging @bouweandela on this

tests/unit/test_docs.py
schlunma: Sure, can do that, but note that's the exact location the [corresponding ESMValTool test](https://github.com/ESMValGroup/ESMValTool/blob/main/tests/unit/test_docs.py) lives.

tests/unit/test_docs.py
schlunma: See discussion here: ESMValGroup/ESMValTool#2554. I thought it might be good to add this feature to this test as well if we need to allow duplicates at some point.

zklaus:
I have to say, I am not so convinced about this approach. Partly because I think it should be enough to have a PR show up in one section. This also aligns with our idea of having exactly one section-related label per PR. Partly because the writing is on the wall that we want to move to a commonly available changelog generator rather than make our little script more elaborate.

schlunma:
My sole intention with this PR was to add a similar test to ESMValCore which was already available in ESMValTool. I was quite surprised that is test isn't available in ESMValCore yet. And it actually found duplicate elements!

I don't want to start a discussion about potential alternatives to our changelog generator here. I just copy-pasted the test from ESMValTool to ESMValCore.

I agree that in a perfect world we don't need to account for duplicates, but I think the argumentation by @axel-lauer in https://github.com/ESMValGroup/ESMValTool/pull/2554 to allow duplicates in certain cases is more than reasonable. This test also doesn't allow duplicates per se, they need to be added manually. Not sure why this is a big deal to be honest.

valeriupredoi:
@zklaus I think this will help with duplicate detection, for now, until we have that modern changelog generator :beer:

zklaus:
Sorry, yes, the test in and of itself is useful and should be added :+1:.
My scepticism is only about the `ALLOWED_DUPLICATES`, which indeed is missing from the corresponding ESMValTool test.

schlunma:
> My scepticism is only about the `ALLOWED_DUPLICATES`, which indeed is missing from the corresponding ESMValTool test.

This is added here: https://github.com/ESMValGroup/ESMValTool/pull/2554. Maybe we should move the discussion to this PR. If we add it there, I suggest we also add it here (and vice versa).

valeriupredoi:
more :baby_chick: :egg: instances :laughing:

schlunma:
I removed the list to allow duplicate entries.

valeriupredoi (approved):
thought I had approved this, sorry, Manu! OK - we should start thinking how to use Github API for this stuff, in the future, just so we become total nerds :grin:

PR #1542, opened 2022-03-16
Install all `install` packages from `setup.py` from conda-forge instead of pip

### LINKED ISSUE #1541: Install all dependecies from conda-forge
Copy all dependencies from `setup.py` to `environment.yml` so that all packages are installed from conda-forge.

All dependencies: https://github.com/conda-forge/esmvalcore-feedstock/blob/main/recipe/meta.yaml

@ESMValGroup/esmvaltool-coreteam I'm pretty sure we discussed that somewhere, but I didn't find an issue to this. Please correct me if I'm wrong, then we can close this.

valeriupredoi:
## Description

Closes #1541 

This the move of all packages that were previously in `setup.py: install` and were installed from PyPi to `environment.yml` and hence installed from and only from conda-forge. @zklaus suggested we use the `- nodefaults` channel option so that we get all those deps only from the specified channel (conda-forge in this case) and it's a good suggestion since I had already encountered an issue with a secondary dep being poorly installed from `main`. Ready for your consideration, folks :+1: 

***

***


valeriupredoi:
OK the OSX import error is actually not related to MyProxyClient import per se, but rather an issue with SSL:
```
from myproxy.client import MyProxyClient

  File "/home/valeriu/ESMValCore/testimp.py", line 1, in <module>
    from myproxy.client import MyProxyClient
  File "/home/valeriu/miniconda3/envs/experimental-all-conda/lib/python3.10/site-packages/myproxy/client/__init__.py", line 42, in <module>
    from OpenSSL import crypto, SSL
  File "/home/valeriu/miniconda3/envs/experimental-all-conda/lib/python3.10/site-packages/OpenSSL/__init__.py", line 8, in <module>
    from OpenSSL import crypto, SSL
  File "/home/valeriu/miniconda3/envs/experimental-all-conda/lib/python3.10/site-packages/OpenSSL/crypto.py", line 11, in <module>
    from OpenSSL._util import (
  File "/home/valeriu/miniconda3/envs/experimental-all-conda/lib/python3.10/site-packages/OpenSSL/_util.py", line 5, in <module>
    from cryptography.hazmat.bindings.openssl.binding import Binding
  File "/home/valeriu/miniconda3/envs/experimental-all-conda/lib/python3.10/site-packages/cryptography/hazmat/bindings/openssl/binding.py", line 14, in <module>
    from cryptography.hazmat.bindings._openssl import ffi, lib
ImportError: libssl.so.1.1: cannot open shared object file: No such file or directory
```

zklaus:
@valeriupredoi, if pollution from the main channel is the problem, this indicates a problem in your local installation because the main channel shouldn't even be known to your local conda/mamba.

But the best mitigation might be the addition of the `- nodefaults` channel to the channel list of the `environment.yml` to prevent any packages from there.

zklaus:
It's documented [here](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#creating-an-environment-file-manually) for conda and definitely works for mamba.

valeriupredoi:
sweet! it's so goin in here then :beer:

valeriupredoi:
trying to push some refs (incl changes to the doc and GA tests turned off) but GH is hosed :face_with_thermometer:

setup.py
```diff
@@ -29,36 +29,8 @@
...
-        'psutil',
-        'pybtex',
-        'pyyaml',
-        'requests',
-        'scipy>=1.6',
-        'scitools-iris>=3.1.0,<3.2.0',
-        'shapely[vectorized]',
-        'stratify',
-        'yamale',
-    ],
```
zklaus, line 61: I think removing this stuff here is a mistake. The `setup.py` should absolutely have all the dependency information. This will be crucial to guarantee a correct installation when someone chooses an alternate method of installation. But it should be duplicated in the `environment.yml` file. That way, using conda/mamba is one way to install all the dependencies, using apt/yum/whatever is another, and on `pip install .` pip can verify that everything is in place as it should be.

valeriupredoi: @zklaus this is very good point and I also thought about it but I think we need to remove them because:
- we don't want to run into those odd cases both you and I have seen when perfectly conda-forgeable dependencies were installed from PyPi instead (at the `pip install` stage)
- we don't want to maintain two separate lists that may diverge - note that quite a few of the dependencies are not on PyPi but are only on conda-forge; there are a few that have slightly different names and versions PyPi vs conda-forge too
- apt/yum works installations works only for the packages that are on Synaptic repo/RPM's, there are again those that are only on conda-forge (is there a way to install a purely conda-forge package with eg apt?)

BUT you do make a very good point about `pip install -e .` which is a good safeguard and test that all packages have been installed correctly (bar the wiggle from my first bullet point up). What if we read `environment.yml` in `setup.py` and checked the packages are indeed installed correctly and if not, prompt the installer to try again **with** mamba? Or, can you think of a better way to check at `pip install` stage?

zklaus: Here is how I think about this; perhaps we need to have a conversation in the technical lead team.

The source distribution on PyPI is *not* just another way to install things. It is *the release*. This is our product, it must contain correct information about dependencies, but it makes no claim about how to install the dependencies.

On conda-forge, there is a package of the PyPI release. Coincidentally, it happens to be produced by some of the same people that are working on the project as well. Similar thinking applies to modules. If we manage to straighten out our dependency handling, it is imaginable that further install methods be added in the future. Why not an ESMValTool package in debian+derivatives like ubuntu? Why not in [Spack](https://spack.readthedocs.io/en/latest/)? All of these projects, including conda-forge, will take the information about dependencies from the PyPI source distribution.


Now, can we do something to reduce duplication? Perhaps. One thing we should probably consider is moving to an essentially empty setup.py with all of the information migrated to setup.cfg, as Iris did, following the official encouragement of the Python packaging crowd. Also, perhaps additionally, we could look at how other projects handle this, see, e.g., [Sparse](https://github.com/pydata/sparse/blob/master/setup.py).

One tricky bit is that project names on PyPI and conda-forge sometimes differ. Perhaps the best way is to be clear that both lists must be maintained, to come up with a good way of handling the extras (test, doc, ...), and to simplify things as much as possible by avoiding dependencies of dependencies and having clear rules on what to include.

valeriupredoi: all very good points, Klaus! I have re-converted the PR to a Draft, I will reinstate our `setup.py` as it was before I made the changes, and I support the idea we talk this at the next @ESMValGroup/technical-lead-development-team technical meeting which I will be circulating a Doodle poll for next week :+1:

znicholls: Two cents from an outsider (although I've done packaging a bit), but maybe this conversation is already finished (in which case I'd be interested in the solution).

I think you basically just have to maintain multiple lists. As far as I can tell, it's impossible to only have a single source of truth (e.g. I have no idea how you could auto-compile your conda dependencies for the conda forge feedstock repo from the dependencies in your source repository). That's annoying, but I think just means you end up relying on really good testing (test it can be installed from conda, test it can be installed from pypi, test it can be installed from some other source etc.).

zklaus: I think @znicholls is mostly right, though the conda-forge bot helps us part of the way.

valeriupredoi:
bumping this to Milestone 2.8 since I am more than booked up with other stuff this week, including Milestone 2.7/freeze

valeriupredoi:
superseded by https://github.com/ESMValGroup/ESMValCore/pull/1786

PR #1545, opened 2022-03-17
Improve performance of `volume_statistics`

### LINKED ISSUE #1498: Performance of volume_statistics
**Describe the bug**
@ESMValGroup/esmvaltool-coreteam do I have time to quickly fix volume statistics for 2.5? The current implementation performs quite poorly and I have been running the function in one year of data with 75 levels for half an hour and it's not done yet. I think that any approach such as da.average(axis=[all axis except time], weights) would be better than the double loop that is implemented right now.

**Please attach**

sloosvel:
## Description

This PR tries to improve the performance of `volume_statistics` using iris and dask functions.
Closes #1498 

Link to documentation:

***

~- [ ] [🛠][1] Any changed [dependencies have been added or removed](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#dependencies) correctly~

***


zklaus:
The CI seems confused. I'll close and reopen to trigger a restart.

sloosvel:
Just merged with main to see if it helps. In any case I wanted to add an extra test to see if the weighting works but I am not sure how to do so.

sloosvel:
Is there anything pending to be adressed otherwise? Maybe it would be good if someone could double check that the changes do not affect the diagnostics.

sloosvel:
@ESMValGroup/esmvaltool-coreteam Anyone with time to review this?
@ledm do you have any example for a small test as you mentioned in https://github.com/ESMValGroup/ESMValCore/issues/1498#issuecomment-1072369792? 

I think if the changes are reasonable it would be an nice improvement for 2.6, and if this needs further work at least there is still time to correct whatever is needed.

Many thanks in advance

sloosvel:
I tried implementing the following test to take into account the weights:

```python
    def test_volume_statistics_weights(self):
        data = np.ma.arange(1, 25).reshape(2, 3, 2, 2)
        self.grid_4d.data = data
        measure = iris.coords.CellMeasure(
            data,
            standard_name='ocean_volume',
            units='m3',
            measure='volume'
        )
        self.grid_4d.add_cell_measure(measure, range(0, measure.ndim))
        result = volume_statistics(self.grid_4d, 'mean')
        expected = np.ma.array(
            [8.333333333333334, 19.144144144144143],
            mask=[False, False])
        self.assert_array_equal(result.data, expected)
```
This passes for both implementations. 

However there is a difference when the input data is defined as regular array instead of a masked array, even if the mask of the previous example is set to `False` :

```python
    def test_volume_statistics_weights(self):
        data = np.arange(1, 25).reshape(2, 3, 2, 2)
        self.grid_4d.data = data
        measure = iris.coords.CellMeasure(
            data,
            standard_name='ocean_volume',
            units='m3',
            measure='volume'
        )
        self.grid_4d.add_cell_measure(measure, range(0, measure.ndim))
        result = volume_statistics(self.grid_4d, 'mean')
        expected = np.ma.array(
            [8.333333333333334, 19.144144144144143],
            mask=[False, False])
        self.assert_array_equal(result.data, expected)

```

This passes the test for the new implementation, but not for the current one. Even though the content of the arrays is the same, the only difference is that in the first example the array is defined as a masked array object with the mask set to False at every point. This difference is coming from this part of the code in `main`:

```python
            try:
                layer_vol = np.ma.masked_where(cube[time_itr, z_itr].data.mask,
                                               grid_volume[time_itr,
                                                           z_itr]).sum()

            except AttributeError:
                # ####
                # No mask in the cube data.
                layer_vol = grid_volume.sum()

```
When the data in the cube is defined as a masked array object, `layer_vol` is computed by using a slice of grid_volume and summing it. Whereas when the data is defined as an array and has no mask in it, layer_volume sums over the whole `grid_volume` array, instead of a slice. I guess this is a bug? 

I think that since this preprocessor is mostly used in ocean variables that tend to have a mask, the change in implementations should not affect the results though.

zklaus:
Good catch, @sloosvel. Indeed, I think at least with recent Iris versions all cubes loaded from netcdf files have a mask. In any case, I would suggest to replace the code in main that you posted, i.e.
```
            try:
                layer_vol = np.ma.masked_where(cube[time_itr, z_itr].data.mask,
                                               grid_volume[time_itr,
                                                           z_itr]).sum()

            except AttributeError:
                # ####
                # No mask in the cube data.
                layer_vol = grid_volume.sum()
```
with 
```
            layer_vol = np.ma.masked_where(np.ma.getmask(cube[time_itr, z_itr].core_data()),
                                           grid_volume[time_itr, z_itr]).sum()
```

To reiterate what we all know:
- Avoid `cube.data` in favor of `cube.core_data()` to keep things lazy where possible
- Avoid `masked_array.data` and `masked_array.mask` in favor of `np.ma.getdata` and (`np.ma.getmask` or `np.ma.getmaskarray`).

zklaus:
Gotcha. Yes, I think that was a bug.

sloosvel:
Is there anything missing in this PR for it to get approved and merged?

zklaus:
It was not clear to me that this is ready for review now, since new commits were added. Will have a look later.

sloosvel:
> Will have a look later.

Unfortunately, I can't do later because things are starting to pile up for me, and with the holidays and everything I would rather not die trying to get this release out.

Please @ESMValGroup/esmvaltool-coreteam reviews are needed here!

sloosvel:
Ran out of ideas as for how can I get this reviewed. As you can see @kserradell, @pabretonniere, I tried. Guess we will have to tell users running in HR that ESMValTool is not ready to handle their data, as the current implementation dies before finishing the computations.

schlunma (commented):
This is probably a very stupid question, but why does a simple


```py
result = cube.collapsed(
        [cube.coord(axis='Z'), cube.coord(axis='Y'), cube.coord(axis='X')],
        iris.analysis.MEAN,
        weights=grid_volume)
```

instead of first collapsing X,Y and then Z not work?

sloosvel:
I think I got confused with the double loop. You can collapse all coordinates at once, but you need to mask the volume first. If you don't mask the volume and collapse all coordinates at once, you don't get the same results. 

Maybe doing it in two steps helps with the memory use though? After all this implementation is much faster, but the previous one uses less memory.

schlunma:
I think if collapsing all coordinates at once works (with masking the volume first), you should go for it. It (1) makes the code much easier and (2) there is one step less for the dask scheduler.

schlunma:
> However, the tool we have at the BSC collapses all coordinates at once so I don't think it's a wrong approach, these must be precision differences.

Yes, I think so, too.

PR #1549, opened 2022-03-21
Extended ICON CMORizer

schlunma:
## Description

I could get my hands on further ICON data which allowed me to extend the ICON CMORizer (the current version wasn't quite able to read this new data). In particular, the following features have been implemented:

- Added variables `areacello` and `siconc`.
- Added support for true height coordinates. If a true height coordinate (i.e., one with units of m) is already present in the file, the CMORizer does not try to convert that to a general height coordinate anymore. In addition, `air_pressure` coordinates are only added now if possible (i.e., if the corresponding variable is present in the input files).
- Added support for CMOR variables that only use `latitude` and not `longitude` by adding checks a la `if 'longitude' in self.vardef.dimensions:` before fixing coordinates (and vice versa). In these cases no spatial index coordinate is added.
- Added automatic addition of time dimension if required by the CMOR standard but not provided by the model output. An example is the variable `zg` which is time-invariant for ICON but a time dimension is required by CMOR.
- Added automatic addition of grid information (in particular latitude and longitude coordinates) to files which do not include this. This is done by downloading the corresponding ICON grid file (using the attribute `grid_file_uri` given by each ICON file). If present and not older than 7 days, this downloaded grid file is reused.
- Added automatic addition of `positive` attribute to variables (based on `self.vardef.positive`).
- Fixed bounds of `air_pressure` coordinate (they were reversed along the trailing dimension).

Doc: https://esmvaltool--1549.org.readthedocs.build/projects/ESMValCore/en/1549/quickstart/find_data.html#icon

I tested the CMORizer on the old and new ICON data I have. It works well in both cases. I also added lots of new tests, but the old ones weren't changed.

***


schlunma:
Is there anything I can do about that Codacy issue (`Library stubs not installed for "requests" (or incompatible with Python 3.9) (error)`)?

valeriupredoi:
OK the Codacy issue has been reported before, most recently by @bouweandela and we don't really have a solid fix for it in mind, I reckon we can brush it under the proverbial carpet :+1:

esmvalcore/cmor/_fixes/icon/_base_fixes.py
```diff
@@ -0,0 +1,146 @@
...
+        if grid_path.exists():
+            mtime = grid_path.stat().st_mtime
+            now = datetime.now().timestamp()
+            age = now - mtime
+            if age < self.CACHE_VALIDITY:
+                logger.debug("Using cached ICON grid file '%s'", grid_path)
+                self._horizontal_grids[grid_name] = self._load_cubes(grid_path)
+                return self._horizontal_grids[grid_name]
+            logger.debug("Existing cached ICON grid file '%s' is outdated",
+                         grid_path)
```
valeriupredoi, line 119: there are a couple loose ends here - what if the first `if` doesn't get through? The second `if` throws only a debug message - anything you'd want to return instead maybe?

esmvalcore/iris_helpers.py
```diff
@@ -1,8 +1,73 @@
...
+import iris.util
 import numpy as np
 
 
+def add_leading_dim_to_cube(cube, dim_coord):
+    """Add new leading dimension to cube.
+
+    An input cube with shape ``(x, ..., z)`` will be transformed to a cube with
+    shape ``(w, x, ..., z)`` where ``w`` is the length of ``dim_coord``. Note
+    that the data is broadcasted to the new shape.
```
valeriupredoi, line 14: maybe add `dim_coord` needs to be scalar 1-d rather than 2-d etc

valeriupredoi (approved):
looking good and a lot of work too, just a couple minor comments from me, of course I couldn't test anything, this was just a walk through the code :beer:

esmvalcore/iris_helpers.py
schlunma: Great idea! I just added this to the doc and also added a check in the code for this.

esmvalcore/cmor/_fixes/icon/_base_fixes.py
schlunma: I think that's fine to not use `else` statements here:

```python
l.94         if self.GRID_FILE_ATTR not in cube.attributes:
```
An error is raised immediately after the `if`, so no `else` necessary.

```python
l.104         if grid_name in self._horizontal_grids:
```
The function `return`s immediately after the `if`, so no `else` necessary.

```python
l.110         if grid_path.exists():
```
If the grid file does not exist, the function starts immediately downloading it (l.122).

```python
l.114             if age < self.CACHE_VALIDITY:
```
If the file age is younger, it is returned immediately after the `if`, so no `else` necessary. If it is older, the debug message is printed and the function continues to download the file again (l.122), so no return is necessary.

esmvalcore/iris_helpers.py
valeriupredoi: cool, cheers, bud!

esmvalcore/cmor/_fixes/icon/_base_fixes.py
valeriupredoi: aha, sounds good, wasn't too aware of what was happening there, just thought I'd flag it just in case :+1:

Issue #312, opened 2019-10-14
Move EMAC cmorizers to fixes


bjoernbroetz:
The code implemented as EMAC cmorizers need to go to be made `fixes`.

mattiarighi:
Why not under `cmorizers` (but this will not be ESMValCore any more)?

bouweandela:
Note that there is also code in the derive section that can be moved to a fix, if the input is only a single variable. If there is more than one input variable to the derive process, it should be implemented as a normal derive function.

Issue #309, opened 2019-10-14
Move EMAC mapping to standard structure  


bjoernbroetz:
When parsing the recipe, augment the variable entries with project/model specific keys. The mapping files should be specified in config-developer.yml

Mapping files will contain CMIP to source project mapping information, e.g.

```
tas:
  source_name: temp2_ave
  channel: Amon
```

Issue #139, opened 2019-07-03
[EMAC] CMORize-light data


axel-lauer:
Read and convert ("CMORize-light") EMAC data

**Input**
* filename(s) of file containing required EMAC variable(s)
* mapping table (EMAC variable(s) --> CMOR variable)

**Output**
* Iris cube containing CMORized-light EMAC data

**Tasks**
- [ ] add meta data (e.g. longitudes, latitudes, time, units, etc.)
- [] scripts for deriving CMOR variable from EMAC variable(s)
- [ ] custom CMOR tables for non-existing CMOR variables

schlunma:
Run first successful tests with dummy data including:

- CMIP5 data for comparison (to check if something broke)
- Simple CMOR variable with only one entry in mapping table and no EMAC derive script
- Complex CMOR variable with two entries in mapping table (from different channels) and an EMAC derive script
- Derived CMOR variable consisting of the two CMOR variable mentioned above

Note: **All** CMOR checks were successful :tada:

schlunma:
Tested the following variables:

- [ ] clivi: {xivi_ave: Amon}
- [ ] clt: {aclcov_ave: Amon} #recipe
- [ ] clwvi: {xlvi_ave: Amon, xivi_ave: Amon} #recipe
- [ ] co2mass: {MP_CO2_ave: tracer_pdef_gp}
- [ ] evspsbl: {evap_ave: Amon} #recipe
- [ ] hfss: {ahfs_ave: Amon}
- [ ] hfls: {ahfl_ave: Amon}
- [ ] lnox: {NOxcg_ave: lnox_PaR_T_gp, NOxic_ave: lnox_PaR_T_gp} #recipe
- [ ] pr: {aprl_ave: Amon, aprc_ave: Amon, aprs_ave: Amon} #recipe
- [ ] prc: {aprc_ave: Amon}
- [ ] prl: {aprl_ave: Amon} #not cmor 
- [ ] prsn: {aprs_ave: Amon}
- [ ] prw: {qvi_ave: Amon}
- [ ] siconca: {seaice_ave: Amon} (**CMOR table missing**)
- [ ] sithick: {siced_ave: Amon}
- [ ] rlds: {flxtbot_ave: Amon, tradsu_ave: Amon} #recipe
- [ ] rlus: {tradsu_ave: Amon}
- [ ] rlut: {flxttop_ave: Amon}
- [ ] rlutcs: {flxtftop_ave: Amon}
[...]

Issue #66, opened 2019-06-11
Add model-specific reformatters to _reformat.py


mattiarighi:
The tool has the capability to convert raw model output of specific models to the CF/CMOR standard.

This is currently available in NCL for a few models:

- [ ] EMAC
- [ ] GFDL
- [ ] EC-EARTH
- [ ] GO

This has to be rewritten using IRIS and the functionalities already available in `_reformat.py`.

A supporting partner for each model is needed (I can help for EMAC).

mattiarighi:
Assigned to the guys who discussed this at the workshop.

zklaus:
As for EC-Earth, I think this is not so high on the priority list anymore: For CMIP6 we developed a pretty flexible cmorization tool that can provide us with the necessary output even online. Plus there is a good indication that the output format will change substantially in the not-so-far future.

mattiarighi:
I think this is a high priority for EMAC only, since it's a deliverable in one of our projects.
I changed the title and edited the text.

valeriupredoi:
@mattiarighi @axel-lauer you guys want me to write the script?

mattiarighi:
@bjoernbroetz at the workshop suggested to use existing tools for this purpose, I leave it to him for further details.

bjoernbroetz:
At the workshop I brought up the cdo CMOR operator:

- [slides](https://www.dkrz.de/c6de/dicad/tp-1/data-workflows-1/data-workflows-during-production-phase/cmip-compliant-processing/functionality-of-cdo-cmor)
- [docu](https://code.mpimet.mpg.de/projects/cdo/wiki/CDO_CMOR_Operator) 
- [pdf](https://code.mpimet.mpg.de/attachments/download/19411/cdo_cmor.pdf)

The mapping from model specific vocabulary to CMIP6 is made available at https://c6dreq.dkrz.de (Variable Mapping). There you can get json-files with the information. The info was placed there manually by the scientist related to the models (EMAC included).

At the workshop if I remember right we were skeptical to use `cdo` for this purpose. However, the mapping table from the website above might be of value to us, even if we choose to go for our own implementation.

zklaus:
Just to clarify since we talked about it at the workshop in the slightly wider context, I think for models like EC-Earth or IPSL where separate means of cmorization have been developed it doesn't make sense to replicate that effort with cdo.

[...]

PR #1554, opened 2022-03-28
Added EMAC CMORizer

schlunma:
## Description

This PR adds a CMORizer for native EMAC model output.

Closes https://github.com/ESMValGroup/ESMValCore/issues/312
Closes https://github.com/ESMValGroup/ESMValCore/issues/309
Closes https://github.com/ESMValGroup/ESMValCore/issues/139
Closes https://github.com/ESMValGroup/ESMValCore/issues/66

Link to documentation: 

***

***


schlunma:
I encountered a rather awkward problem while implementing this:

EMAC data has lots of variables with `units = "0 - 1"`. While these files can be read by iris (which simply assigns an `invalid_units` attribute), each time a file with this is loaded the message

```bash
ut_scale(): NULL factor argument
ut_are_convertible(): NULL unit argument
```

is printed to the screen. These messages make the log [basically unreadable](https://github.com/ESMValGroup/ESMValCore/files/8652579/out.txt)  if multiple years and/or multiple variables are read. Unfortunately, these messages are not issued via pythons `warnings` module, so they cannot be simply ignored with this.

I really want to get rid of these messages, but I'm not quite sure how. I thought of 3 possible options:

1. Wait until `cf_units` fixes this (see https://github.com/SciTools/cf-units/issues/240).
2. Fix this by redirecting `stderr` in our `load` preprocessor. Unfortunately, [`contextlib.redirect_stderr`](https://docs.python.org/3/library/contextlib.html#contextlib.redirect_stderr) does not work here. I don't fully understand why, but [this answer](https://stackoverflow.com/a/22434262/6864675) here suggests that this function does not work "at file descriptors level". This post gives a rather lengthy alternative that indeed successfully suppresses the message.
3. Fix this by using `fix_file` to remove the faulty units with `netCDF4`.

Option 1  would obviously be the optimal choice, but I'm not sure how long this will take. Option 2 is fast, but I'm not sure if adding this workaround to the heart of our loading mechanism is a good idea. Option 3 is definitely cleaner than option 2, but super slow since it involves copying many files.

Any ideas/recommendations/comments on this? @bouweandela @valeriupredoi @zklaus

schlunma:
Thanks @bouweandela, that's awesome!! I just tested this and works perfectly fine. I will put this into `load()` as a temporary workaround!

schlunma:
> don't think so - the root is `Path(__file__).resolve().parent / 'test_data'` and this is used for at least 4 more tests - cesm2, gfdl etc - why are those not hitting the fan?

`__file__` should be `ESMValCore/tests/integration/cmor/_fixes/conftest.py`, so `Path(__file__).parent` should be `ESMValCore/tests/integration/cmor/_fixes`, right? I think that's correct. And the test is also running fie locally with the correct path...

valeriupredoi:
let's figure it out :grin: I am failing one cesm2 test and forcing to print out where and what's in the test data path

bouweandela:
This message from the checkout step warns about the problem:

> Either git or ssh (required by git to clone through SSH) is not installed in the image. Falling back to CircleCI's native git client but the behavior may be different from official git. If this is an issue, please use an image that has official git and ssh installed.

somehow the emac.nc data file does not get checked out. I will make a pull request to install git and ssh before checkout.

valeriupredoi:
scratch that! am a dummer - it's sequential

valeriupredoi:
aye, @bouweandela is right - no sign of the emac.nc file in the (correct) test data root:

- TEST DATA PATH CESM2 /root/project/tests/integration/cmor/_fixes/test_data
- dir /root/project/tests/integration/cmor/_fixes/test_data contains 'cesm2_cl.nc', 'cesm2_waccm_cl.nc', 'cnrm_cm6_1_cl.nc', 'common_cl_a.nc', 'common_cl_ap.nc', 'common_cl_hybrid_height.nc', 'create_test_data.py', 'gfdl_cm4_cl.nc', 'icon_2d.nc', 'icon_3d.nc', 'icon_grid.nc'
_____________________________ test_sample_data_tas __________

bouweandela:
The problem can indeed be fixed by installing git and ssh before checkout on CircleCI. See #1601 for more info.

axel-lauer (changes requested):
I looked through the plots produced by the EMAC example recipe `recipe_emac_new.yml` and most things look good - as far as I can tell. There are, however, a few items that might need to be double-checked:

* map of precip (given in kg/m2/s): the color scale ranges from 0-10, i.e. the plot is empty and evaluation is not possible.
* map of zg at 500 hPa: the values over Antarctica and over Greenland look very strange, I would expect the values to range somewhere between 5000 and 6000 m but not between 0 and 3000 m as in the plot.
* time series of surface pressure: assuming that the values shown are global mean values, the absolute values and the variability look wrong

axel-lauer (approved):
I successfully ran `recipe_emac_new.yml` and the output looks fine to me. Three quick answers to my previous comments:

* pr: OK, the preprocessor output looks fine, but I am not sure why the diagnostic expects mm/day instead of simply using the units provided in the metadata. But that's not a problem of the EMAC related additions.
* ps: also OK, the preprocessor output looks fine but I do not fully understand why the diagnostic creates such a strange looking time series. But again, this is not related to the EMAC extensions and thus not relevant here.
* zg @ 500 hPa: OK, the original EMAC output already looks weird with the 500 hPa geopotential heitght in Antarctica and Greenland being lower than the surface height! We should ask our EMAC modelers about this, but again, this is not related to the EMAC additions and thus fine.

So in summary, all fine and green light from the science review.

remi-kazeroni (approved):
This PR looks good to me: code, documentation, tests, recipe run with `recipe_emac_new.yml`. Thanks for your work @schlunma! It's nice to have support for EMAC data as part of the on-the-fly cmorizers.

schlunma:
Awesome, great!! Many thanks for the quick review @axel-lauer @remi-kazeroni :tada:

PR #1561, opened 2022-04-15
Try searching multiple ESGF index nodes

### LINKED ISSUE #1333: `test_real_search` fails on circleci with read timeout
**Describe the bug**
The `test_real_search`, introduced recently in #1217, fails on circleci with read timeout. This just may be a fluke, but even so, two thoughts:
- We might want to check network accessibility from circleci and possibly skip the test there or explore circleci configuration
- We might want to catch the exception and give better feedback to the user.

The test result can be inspected here:
https://app.circleci.com/pipelines/github/ESMValGroup/ESMValCore/5713/workflows/c1f2d784-4702-4ec4-ba56-862f387e2636/jobs/27482?invite=true#step-103-159
the test is located here:
https://github.com/ESMValGroup/ESMValCore/blob/43e6b16ed7f84adbf611f07d3c535e3a3b9a2791/tests/integration/esgf/test_search_download.py#L131

bouweandela:
## Description

Closes #1333

Link to documentation: https://esmvaltool--1561.org.readthedocs.build/projects/ESMValCore/en/1561/quickstart/configure.html#search

***

***

### DIFF
--- a/tests/unit/config/test_esgf_pyclient.py
+++ b/tests/unit/config/test_esgf_pyclient.py
@@ -99,6 +99,35 @@ def test_read_config_file(monkeypatch, tmp_path):
     assert cfg == reference
 
 
+def test_read_v25_config_file(monkeypatch, tmp_path):
+    """Test function read_config_file for v2.5 and earlier.
+
+    For v2.5 and earlier, the config-file contained a single `url`
+    instead of a list of `urls` to specify the ESGF index node.
+    """
+    cfg_file = tmp_path / 'esgf-pyclient.yml'
+    monkeypatch.setattr(_esgf_pyclient, 'CONFIG_FILE', cfg_file)
+
+    cfg_file_content = {
+        'search_connection': {
+            'url': 'https://some.host/path'
+        },
+    }
+    with cfg_file.open('w') as file:
+        yaml.safe_dump(cfg_file_content, file)
+
+    reference = {
+        'search_connection': {
+            'urls': [
+                'https://some.host/path',
+            ]
+        }
+    }
+
+    cfg = _esgf_pyclient.read_config_file()
+    assert cfg == reference
+
+
 @pytest.mark.parametrize('with_keyring_creds', [True, False])
 def test_default_config(monkeypatch, mocker, tmp_path, with_keyring_creds):
     """Test that load_esgf_pyclient_config returns the default config."""
--- a/tests/unit/config/test_esgf_pyclient.py
+++ b/tests/unit/config/test_esgf_pyclient.py
@@ -13,8 +13,16 @@ DEFAULT_CONFIG: dict = {
         'bootstrap': True,
     },
     'search_connection': {
-        'url':
-        'http://esgf-node.llnl.gov/esg-search',
+        'urls': [
+            'https://esgf-index1.ceda.ac.uk/esg-search',
+            'https://esgf-node.llnl.gov/esg-search',
+            'https://esgf-data.dkrz.de/esg-search',
+            'https://esgf-node.ipsl.upmc.fr/esg-search',
+            'https://esg-dn1.nsc.liu.se/esg-search',
+            'https://esgf.nci.org.au/esg-search',
+            'https://esgf.nccs.nasa.gov/esg-search',
+            'https://esgdata.gfdl.noaa.gov/esg-search',
+        ],
         'distrib':
         True,
         'timeout':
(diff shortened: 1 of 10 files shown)


bouweandela:
The remaining Codacy issues are not easily solved:
1) Use of `global`: we actually want the search function to remember which index node is online, so it doesn't need to wait for a timeout with every new search. Using a global variable to store this information in seems a good way to do it.
2) Missing mypy stubs: this is something we can only solve by running the Codacy analysis on CircleCI or with a GitHub action that first installs the required dependencies. Something to be considered, but not in this pull request.

tests/unit/config/test_esgf_pyclient.py
```diff
@@ -13,16 +13,24 @@
...
+        'urls': [
+            'https://esgf-index1.ceda.ac.uk/esg-search',
+            'https://esgf-node.llnl.gov/esg-search',
+            'https://esgf-data.dkrz.de/esg-search',
+            'https://esgf-node.ipsl.upmc.fr/esg-search',
+            'https://esg-dn1.nsc.liu.se/esg-search',
+            'https://esgf.nci.org.au/esg-search',
+            'https://esgf.nccs.nasa.gov/esg-search',
+            'https://esgdata.gfdl.noaa.gov/esg-search',
+        ],
```
valeriupredoi, line 25: cheers @bouweandela - I think it looks good, but I've not tested it, I trust you did, and probably best if you turn on the GA tests and rerun the tests a few times (if you expect a much less frequent failure due to node not being reachable). I do think we should put this list of nodes someplace centrally and just load it from there both for the actual code and for the testing environment, I have seen this list repeated at least three times in different places, and that's prone to introducing issues if we change one node somewhere but not everywhere. What you reckon?

valeriupredoi (changes requested):

valeriupredoi:
> The remaining Codacy issues are not easily solved:
> 
>     1. Use of `global`: we actually want the search function to remember which index node is online, so it doesn't need to wait for a timeout with every new search. Using a global variable to store this information in seems a good way to do it.
> 
>     2. Missing mypy stubs: this is something we can maybe solve by running the Codacy analysis on CircleCI or with a GitHub action that first installs the required dependencies. Something to be considered, but not in this pull request.

The missing stubs for `mypy` I've seen in other places too - it's becoming a bit annoying - can we not introduce a `#noqa` and specify why the that in an inline comment?

tests/unit/config/test_esgf_pyclient.py
bouweandela: Unfortunately, I have not been able to test this in real life either, because you need an index node that is timing out for that and I don't think it's that common.

The risk of introducing an error because of the duplication is not that big, the worst thing that can happen is that the documentation goes out of date. The unit test ensures that the default configuration and test code are identical. I agree that this may not be the most elegant unit test, but reading the list from the same source in production and test code is asking for trouble.

bouweandela:
> The missing stubs for mypy I've seen in other places too - it's becoming a bit annoying - can we not introduce a #noqa and specify why the that in an inline comment?

I don't think it's that easy. When you run prospector locally, things work fine because the stubs are installed as development dependencies. However, this doesn't work on Codacy because Codacy does not install our dependencies. A potential solution could be to run the Codacy command line client in a GitHub action and upload the results, but I've looked at that option for about half an hour and did not immediately see how to get that to work in a maintainable way.

tests/unit/config/test_esgf_pyclient.py
valeriupredoi: > Unfortunately, I have not been able to test this in real life either, because you need an index node that is timing out for that and I don't think it's that common.

Yeah that's why I was thinking we might pick that up from running a bunch of GA tests - but it's prob too much hassle - we'll have to just wait and see on nightly/PR merge tests I reckon

valeriupredoi:
> > The missing stubs for mypy I've seen in other places too - it's becoming a bit annoying - can we not introduce a #noqa and specify why the that in an inline comment?
> 
> I don't think it's that easy. When you run prospector locally, things work fine because the stubs are installed as development dependencies. However, this doesn't work on Codacy because Codacy does not install our dependencies. A potential solution could be to run the Codacy command line client in a GitHub action and upload the results, but I've looked at that option for about half an hour and did not immediately see how to get that to work in a maintainable way.

hahah, that's asking for trouble IMHO. Is Codacy installing the env deps from a container? Coz then that should be configurable. Again, not a biggie for me - am gonna approve, cheers, man!

valeriupredoi (approved):

bouweandela:
> Is Codacy installing the env deps from a container? 

I think they're using this to run prospector: https://github.com/codacy/codacy-prospector (i.e. they do not install any dependencies, just `prospector[with_everything]`) and they have this GitHub action available: https://github.com/codacy/codacy-analysis-cli-action#integration-with-codacy-for-client-side-tools

zklaus (approved):

PR #1584, opened 2022-05-16
Tweak `extract_point` preprocessor: explain what it returns if one point coord outside cube and add explicit test 

valeriupredoi:
## Description

I panicked a little bit while reviewing one of @sloosvel 's PRs dealing with yet another extraction scenario since I thought the extraction process can do extrapolation on top of the known interpolation and we'd end up extracting a point in Italy when the cube only contains Nigeria, in which case data would be absolutely bonkers. It proves out that in that case the code don't fail, but returns fully masked data of the extracted point cube - good - but this is mentioned nowhere! So here it is, mentioned here, and added a test for good measure - even though there was a test for outside points extraction, this one checks specifically for the mask to be full.

***

***

### DIFF
--- a/doc/recipe/preprocessor.rst
+++ b/doc/recipe/preprocessor.rst
@@ -1654,6 +1654,11 @@ match the respective latitude and longitude coordinate to those of the
 input coordinates. If the input coordinate is a scalar, the dimension
 will be missing in the output cube (that is, it will be a scalar).
 
+If the point to be extracted has at least one of the coordinate point
+values outside the interval of the cube's same coordinate values, then
+no extrapolation will be performed, and the resulting extracted cube
+will have fully masked data.
+
 Parameters:
   * ``cube``: the input dataset cube.
   * ``latitude``, ``longitude``: coordinates (as floating point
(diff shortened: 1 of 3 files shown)


schlunma:
Yeah, the extrapolation options are hidden here:

https://github.com/ESMValGroup/ESMValCore/blob/b087bf37e727fbcb5a5d71bb93f014fa066c22f6/esmvalcore/preprocessor/_regrid.py#L56-L59

It might make sense at some point to allow custom kwargs for these, just like we did for `regrid` in https://github.com/ESMValGroup/ESMValCore/pull/1448. Sometimes it makes indeed sense to allow extrapolation. But this can be done in another PR, maybe open an issue?

valeriupredoi:
I'd steer clear from that, Manu, or add a big Hazard sign - extrapolation can easily be misused by the user and their results could be absolutely mental, but valid numbers, and even not 1e20's - actual regular Joe data. At any rate, we can talk about that more in an issue, what you think of this PR for now?

schlunma:
> I'd steer clear from that, Manu, or add a big Hazard sign - extrapolation can easily be misused by the user and their results could be absolutely mental, but valid numbers, and even not 1e20's - actual regular Joe data. At any rate, we can talk about that more in an issue, what you think of this PR for now?

This definitely shouldn't be the default option, but might actually be useful for experienced users. But yes, let's discuss this in another issue :+1:

schlunma (approved):

valeriupredoi:
cheers, Manu! :beer:

PR #1589, opened 2022-05-18
Add `axis_statistics` and improve `depth_integration`

### LINKED ISSUE #1583: Adding more operators to `depth_integration`
A while back, Javi worked on generalising the `depth_integration` preprocessor to accept other operators. It would be useful to have this changes officially in the core. 

Additionally, it could be an opportunity to improve some lines to make the preprocessor lazy, namely:

```python
            if depth.ndim == 1:
                slices = [None for i in cube.shape]
                coord_dim = cube.coord_dims(cube.coord(axis='z'))[0]
                slices[coord_dim] = slice(None)
                thickness = np.abs(thickness[tuple(slices)])
            ones = np.ones_like(cube.data)


```

yes

sloosvel:
## Description

This PR adds a new preprocessor called `axis_statistics` that collapses a cube's axis using iris operators. The axis bounds are used as weights in case the operator accepts weights. This preprocessor also gets called inside `depth_integration`.
Closes #1583 

Link to documentation:

***

***


doc/recipe/preprocessor.rst
```diff
@@ -1819,6 +1820,19 @@ cell measures or ancillary variables`.
...
 
+``axis_statistics``
+---------------------
+
+This function operates over a given axis, and removes it of the
+output cube. 
+
+Takes arguments:
+  * axis: direction over which the statistics will be performed.
+  * operator: defines the operation to apply over the axis
```
zklaus, line 1831: I think we need a list of supported operators here.

esmvalcore/preprocessor/_volume.py
```diff
@@ -275,6 +277,51 @@ def volume_statistics(cube, operator):
...
+    they are computed using the coordinate bounds.
+
+    Arguments
+    ---------
+    cube: iris.cube.Cube
+        input cube.
+    axis: str
+        direction over where to apply the operator.
+    operator: str
+        statistics to perform.
```
zklaus, line 293: Here, too, we need a list of supported operators.

esmvalcore/preprocessor/_volume.py
```diff
@@ -275,6 +277,51 @@ def volume_statistics(cube, operator):
...
+        coord = cube.coord(axis=axis)
+    except iris.exceptions.CoordinateNotFoundError as err:
+        raise ValueError(
+            'Axis {} not found in cube {}'.format(
+                axis,
+                cube.summary(shorten=True))) from err
+    operation = get_iris_analysis_operation(operator)
+    weights = None
+    if operator_accept_weights(operator):
+        coord_dim = cube.coord_dims(cube.coord(axis=axis))[0]
```
zklaus, line 310:
```suggestion
        coord_dims = cube.coord_dims(coord)
        assert len(coord_dims) == 1
        coord_dim = coord_dims[0]
```
Note that also the weights calculation below is not meaningful if the coordinate is multi-dimensional.

zklaus (changes requested):

esmvalcore/preprocessor/_volume.py
sloosvel: > if the coordinate is multi-dimensional

What should the behaviour be if the coordinate is multidimensional? Because for instance in the case of wanting to collapse the 'X' axis for a multidimensional coordinate, this implementation would collapse the 'Y' axis instead because the value of `coord_dim` would always be 1

zklaus: Actually, in that case `len(coord_dims)` would be 2 and thus this would raise an `AssertionError`. This is the crudest way to say we only support 1d coordinates right now. Since we are shooting for a quick solution, I think supporting only 1d coordinates is sensible, but it would be better to be clearer about it, for example by raising a `NotImplemented` exception.

I would probably go so far as to do that even in the case of weight free operators, at least to start with.

sloosvel: Ok, added the raise of the NotImplementedError

esmvalcore/preprocessor/_volume.py
sloosvel: added as well, please let me know if the changes work for you

doc/recipe/preprocessor.rst
```diff
@@ -1819,6 +1820,21 @@ cell measures or ancillary variables`.
...
 
 
+``axis_statistics``
+---------------------
+
+This function operates over a given axis, and removes it of the
+output cube. 
+
+Takes arguments:
+  * axis: direction over which the statistics will be performed.
```
bouweandela, line 1830: Could you add a list of valid choices for axis? Or link to the relevant iris documentation?

sloosvel: Sure!

doc/recipe/preprocessor.rst
```diff
@@ -1819,6 +1820,27 @@ cell measures or ancillary variables`.
 See also :func:`esmvalcore.preprocessor.volume_statistics`.
 
 
+``axis_statistics``
+---------------------
+
+This function operates over a given axis, and removes it of the
```
zklaus, line 1826:
```suggestion
This function operates over a given axis, and removes it from the
```

esmvalcore/preprocessor/_volume.py
```diff
@@ -275,6 +277,59 @@ def volume_statistics(cube, operator):
...
+        raise NotImplementedError(
+            'axis_statistics not implemented for '
+            'multidimensional coordinates.')
+    operation = get_iris_analysis_operation(operator)
+    weights = None
+    if operator_accept_weights(operator):
+        coord_dim = coord_dims[0]
+        expand = list(range(cube.ndim))
+        expand.remove(coord_dim)
+        weights = np.abs(coord.bounds[..., 1] - coord.bounds[..., 0])
```
zklaus, line 321: Let's keep it lazy:
```suggestion
        bounds = coord.core_bounds()
        weights = np.abs(bounds[..., 1] - bounds[..., 0])
```

esmvalcore/preprocessor/_volume.py
```diff
@@ -293,23 +348,7 @@ def depth_integration(cube):
         collapsed cube.
     """
     # ####
```
zklaus, line 350: What's the purpose of this line? (I know that you didn't introduce it here.) If nobody knows, let's remove it?

tests/unit/preprocessor/_volume/test_volume.py
```diff
@@ -7,7 +7,8 @@
 from cf_units import Unit
 
 import tests
-from esmvalcore.preprocessor._volume import (volume_statistics,
+from esmvalcore.preprocessor._volume import (axis_statistics,
+                                             volume_statistics,
                                              depth_integration,
                                              extract_trajectory,
                                              extract_transect,
```
zklaus, line 14: Should we sort this list alphabetically? (All of it, of course, but github doesn't allow me to extend the comment context so far.)

tests/unit/preprocessor/_volume/test_volume.py
```diff
@@ -78,6 +79,70 @@ def setUp(self):
...
+            self.assert_array_equal(result.data, expected)
+
+        for operator in ['std_dev', 'variance']:
+            result = axis_statistics(self.grid_4d, 'z', operator)
+            expected = np.ma.zeros((2, 2, 2))
+            self.assert_array_equal(result.data, expected)
+
+        result = axis_statistics(self.grid_4d, 'z', 'sum')
+        expected = np.ma.ones((2, 2, 2)) * 250
+        self.assert_array_equal(result.data, expected)
```
zklaus, line 96: This test is carefully set up such that the result is the same for many different operations. This is a bit dangerous, imho, since it will miss many types of errors, such as wrong selection of operator, mixing of wrong axis, etc.
It would be better to have slightly different input (e.g. from linrange) and unique outputs. The programming effort can be controlled using paramtrized tests, see https://docs.pytest.org/en/6.2.x/parametrize.html. This way, even the two loops can be removed in favor of a single one of their bodies and a list of `(operator, expected) pairs for the parametrization.

zklaus (changes requested):

tests/unit/preprocessor/_volume/test_volume.py
sloosvel: I had some issues when trying to parametrize a test inside a class, so I wrote it in a loop

zklaus: Could you try it one more time? If that doesn't work, we'll call the resulting method `_inner_test_something` and implement the loop explicitly in a proper test method of the class.

Regardless of the approach, the point about the dangerous tests stands.

esmvalcore/preprocessor/_volume.py
sloosvel: I cannot call `expand_dims` in dask, so the wouldn't the bounds get realised later on anyway? They are always going to be a 1D array though, so I thought it wouldn't be a big deal if they are not lazy.

zklaus: The `expand_dims` function has been added in [recent dask versions](https://docs.dask.org/en/stable/changelog.html#id45), but don't worry about it. You should be able to use `np.expand_dims` and have one of NumPy's dispatch mechanisms (e.g. [NEP-18](https://numpy.org/neps/nep-0018-array-function-protocol.html)) figure out the right implementation.

sloosvel:
Anything else missing here? I updated the tests.

tests/unit/preprocessor/_volume/test_volume.py
sloosvel: Yes, at the end I went for writing each test case separately. Hope that works for the time being.

znicholls: Looks nice and matches what is done in https://github.com/ESMValGroup/ESMValCore/blob/4a840fb76d89453643758210f3aab9ad5caa4632/tests/unit/preprocessor/_time/test_time.py#L526 (can be changed another time)

schlunma:
I will have a look at this on Monday or early next week.

esmvalcore/preprocessor/_volume.py
```diff
@@ -275,6 +277,60 @@ def volume_statistics(cube, operator):
...
+    operation = get_iris_analysis_operation(operator)
+    weights = None
+    if operator_accept_weights(operator):
+        coord_dim = coord_dims[0]
+        expand = list(range(cube.ndim))
+        expand.remove(coord_dim)
+        bounds = coord.core_bounds()
+        weights = np.abs(bounds[..., 1] - bounds[..., 0])
+        weights = np.expand_dims(weights, expand)
+        weights = da.broadcast_to(weights, cube.shape)
```
schlunma, line 324:
```suggestion
        weights = np.abs(
            coord.core_bounds()[..., 1] - coord.core_bounds()[..., 0]
        )
        weights = iris.util.broadcast_to_shape(weights, cube.shape, coord_dims)
```

I think this can be simplified with[ `iris.util.broadcast_to_shape`](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris/util.html#iris.util.broadcast_to_shape).

schlunma (commented):
One small suggestion to simplify the code, will test afterwards

esmvalcore/preprocessor/_volume.py
schlunma: I never had performance problems with this, as it as doesn't copy the data but rather returns repeated views. I think it will work fine.

schlunma (approved):
You were right @sloosvel, the old version is about ~10% faster than `iris.util.broadcast_to_shape`, which really matters for large 4D datasets. I restored it, sorry for the mess!

Tested `axis_statistics` with all kinds of options and `depth_integration` , everything worked fine :tada:

Cheers :rocket:

sloosvel:
@zklaus Have the changes you requested been adressed?

PR #1596, opened 2022-05-23
DO NOT MERGE toss in some really silly test fails see if output still ok with `flake8>4`

valeriupredoi:
DO NOT MERGE for Pete's sake!


valeriupredoi:
OK @bouweandela all good with the new flake8 - nice and explicit fail stdout :+1:

PR #1607, opened 2022-05-30
Fix ocean region coordinate in msft datasets (CMIP6)

dhohn:
## Description

Fixes coordinate issues for some mass stream function (msft) variables in several CMIP6 datasets. Described in this comment: https://github.com/ESMValGroup/ESMValTool/issues/2645#issuecomment-1121161408

Related to #1587

Link to documentation:

***

## [Before you get started](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#getting-started)

- [ ] [☝ Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do

## [Checklist](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#checklist-for-pull-requests)

It is the responsibility of the author to make sure the pull request is ready to review. The icons indicate whether the item will be subject to the [🛠 Technical][1] or [🧪 Scientific][2] review.

[1]: https://docs.esmvaltool.org/en/latest/community/review.html#technical-review
[2]: https://docs.esmvaltool.org/en/latest/community/review.html#scientific-review

- [ ] [🧪][2] The new functionality is [relevant and scientifically sound](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#scientific-relevance)
- [ ] [🛠][1] This pull request has a [descriptive title and labels](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-title-and-label)
- [ ] [🛠][1] Code is written according to the [code quality guidelines](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#code-quality)
- [ ] [🧪][2] and [🛠][1] [Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available
- [ ] [🛠][1] [Unit tests](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#tests) have been added
- [ ] [🛠][1] Changes are [backward compatible](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#backward-compatibility)
- [ ] [🛠][1] Any changed [dependencies have been added or removed](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#dependencies) correctly
- [ ] [🛠][1] The [list of authors](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#list-of-authors) is up to date
[...]

PR #1612, opened 2022-06-01
add CMOR check exception for a basin coord named sector

dhohn:
## Description

Add a CMOR check exception for an auxcoord named basin that has an out_name == sector. As discussed in PR #1607.

Closes Issue #1587

Link to documentation:

***

***

### DIFF
--- a/esmvalcore/cmor/check.py
+++ b/esmvalcore/cmor/check.py
@@ -437,7 +437,6 @@ class CMORCheck():
                                 coord.var_name,
                                 coordinate.out_name,
                             )
-                            # coord.var_name = coordinate.out_name
                         else:
                             self.report_error(
                                 'Coordinate {0} has var name {1} '
--- a/esmvalcore/cmor/check.py
+++ b/esmvalcore/cmor/check.py
@@ -426,6 +426,18 @@ class CMORCheck():
                                 coordinate.out_name,
                             )
                             coord.var_name = coordinate.out_name
+                        elif (self._cmor_var.table_type in 'CMIP6' and
+                              coordinate.name == "basin" and
+                              coord.var_name == "sector"):
+                            self.report_debug_message(
+                                'Coordinate {0} has var name {1} '
+                                'instead of {2}. '
+                                "But that's considered OK and ignored.",
+                                coordinate.name,
+                                coord.var_name,
+                                coordinate.out_name,
+                            )
+                            # coord.var_name = coordinate.out_name
                         else:
                             self.report_error(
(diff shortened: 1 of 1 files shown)


zklaus:
Thanks, @dhohn. I think maybe it would be better to tie the exception to the standard name being `region` or `area_type`. Those are the two standard names that require strings as values. As a consequence, the coordinate variable needs to add another dimension for the string length (a technical netcdf quirk) and thus cannot be a traditional coordinate variable with the same name as its own, single dimension. That's really why the `var_name` can in principle be arbitrary and we don't even need to bring CMIP into this. In principle, it would be nice to check the dimension name, but I don't think that information is available to us in Iris. The good news is that standard names `region` and `area_type` really is the complete list for this case, since those are all additional [controlled vocabularies in CF](http://cfconventions.org/standard-names.html).

Does that sound ok?

dhohn:
I've added a test. Took me a while to figure out that mock interface ;)

zklaus:
Yeah, that can be confusing. Thanks, LGTM! @ESMValGroup/esmvaltool-coreteam, could somebody else have a look too?

bouweandela (changes requested):
Code looks fine, but some explanation would be nice

esmvalcore/cmor/check.py
```diff
@@ -426,6 +426,16 @@ def _check_dim_names(self):
                                 coordinate.out_name,
                             )
                             coord.var_name = coordinate.out_name
+                        elif coord.standard_name in ['region', 'area_type']:
+                            self.report_debug_message(
+                                'Coordinate {0} has var name {1} '
+                                'instead of {2}. '
+                                "But that's considered OK and ignored. "
+                                'See also https://github.com/ESMValGroup/ESMValCore/discussions/1587',  # noqa: line-too-long
```
bouweandela, line 434:
```suggestion
                                'But that's considered OK and ignored. See also '
                                'https://github.com/ESMValGroup/ESMValCore/discussions/1587',
```
maybe you could try if this works? Usually if you put the URL on it's own line the linter does not complain

bouweandela (approved):

esmvalcore/cmor/check.py
dhohn: I tried. Still too long and flake doesn't like it either.

sloosvel: I added the link as a separate variable at the beginning of the function. Maybe it's not the cleanest thing to do but the code in the checker is quite idented. I will merge this if all tests pass.

PR #1656, opened 2022-07-06
Adapt tests to new version of cf-units

### LINKED ISSUE #1655: Test failure due to change in cf-units
For a couple of days now, we have seen our tests fail. The problem stems from a change in the latest cf-units release where the handling of the `gregorian` and `standard` calendars was changed, namely SciTools/cf-units#205. This creates the failure in our tests where they find now the calendar to be `standard` instead of `gregorian`. The quick fix is to just change the tests to expect `standard`, but I have a lingering concern that other places like diag scripts may be surprised to find `standard` now.

This is similar to ESMValGroup/ESMValTool#2705.

zklaus:
## Description

This is a bugfix to address recent test failures due to a change in cf-units.

Closes #1655 

Link to documentation:

***

***

### DIFF
--- a/tests/integration/cmor/_fixes/cmip5/test_access1_0.py
+++ b/tests/integration/cmor/_fixes/cmip5/test_access1_0.py
@@ -13,7 +13,7 @@ from esmvalcore.cmor.table import get_var_info
 from esmvalcore.iris_helpers import date2num
 
 
-class TestAllVars(unittest.TestCase):
+class TestAllVars:
     """Test all vars fixes."""
 
     def setUp(self):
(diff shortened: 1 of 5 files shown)


zklaus:
This is a start to address recent test failures for the core similar to those in the tool that arise from the changes in cf-units.

I need help with two things: How to convert the `unittest.TestCase` derived classes in the cmip5 fix tests, and how to deal with the inevitable regression. @bouweandela or @schlunma, any ideas?

schlunma:
Why do you want to get rid of `unittest.Testcase`? As an alternative for setup and teardown I would use [`fixtures`](https://docs.pytest.org/en/latest/explanation/fixtures.html#about-fixtures).

schlunma:
Ah I see.

I think you could just define the fixture outside of the class and then use it inside the class like you would normally do. In addition you could add a [`scope='class'`](https://docs.pytest.org/en/latest/how-to/fixtures.html#scope-sharing-fixtures-across-classes-modules-packages-or-session) parameter to the fixture, but I don't think that's useful here. To mimic the behavior of `setUp` the default scope (fixture is destroyed after every test) should be the one to go.

sloosvel:
Is this worth another release candidate?

zklaus:
I would say so, yes. But it's your call. I am also not fully aware of other changes that may have happened since the last release candidate, but generally speaking, the release is expected to be identical to the last release candidate.

valeriupredoi:
I think this needs to be included in the RC rather than straight into the release since all this calendars business is a source of major pain in the rear and can lead to some unwanted changes in the actual diagnostics' results. @zklaus would you have some time to polish this a bit more? If not I can take over from here, no problemo :beer:

sloosvel:
No worries, I got a new branch to work on this. I will close this one.

bouweandela:
For the upcoming relaese and to avoid a lot of last minute work, would it be an option to just pin cf-units to the version that we know works and that @sloosvel already ran all the tests with?

valeriupredoi:
@bouweandela I think that's a very good idea indeed!

zklaus:
> > Alright, I'm stepping off here. Release managers, feel free to continue with this branch or to start your own fixes.
> 
> And what is is that is left to be finished in here? It's quite surprising to see these kind of answers considering the kind of comments that I usually have to hear about my own "unfinished work".

It's the responsibility of the release managers to take care of the nightly tests. They have been broken for a week now, so I did the digging for the cause and how to fix it. I'm sorry if this came across as anything but helpful.

sloosvel:
>It's the responsibility of the release managers to take care of the nightly tests. They have been broken for a week now, so I did the digging for the cause and how to fix it. I'm sorry if this came across as anything but helpful.

I know, but I had to finish running recipes for a release candidate that could have been avoided. In any case, everyone in favour of pinning then?

valeriupredoi:
@sloosvel am pinning it as we speak :beer:

valeriupredoi:
about the tests - we should think of a stricter policy while we're working on a RC or stable release, something like "thou shall not pass" if the tests fail

zklaus:
@valeriupredoi, I am not sure what you mean, but perhaps that's a new issue/discussion?

valeriupredoi:
> @valeriupredoi, I am not sure what you mean, but perhaps that's a new issue/discussion?

what I mean is to have a clause in the push to pypi that all GA tests have passed - ok maybe too strict - have the tests open an issue automatically if one test fails, and maybe turn that feature on only during release periods so we don't overkill?

bouweandela:
> have the tests open an issue automatically if one test fails

I like that, there is probably a github action for that

valeriupredoi:
@bouweandela there's a github action even for when you forget to turn on that github action :rofl:

PR #1661, opened 2022-07-11
update conda lock file manually - test failed due to cf-units issue which is now fixed

valeriupredoi:


valeriupredoi:
quick one for the reviewers - I wonder what happened to my admin rights - this was an easy merge w/o needing review :confused:

sloosvel (approved):

zklaus:
I see nothing wrong with the lock file, but it doesn't actually touch cf-units. Is this still needed or how does it relate to the cf-units issue?

sloosvel:
I would like to get the new release candidate out today, so if this could get sorted out it would be great!

valeriupredoi:
it is now, thanks, Saskia!

valeriupredoi:
oh shoot, completely missed that one, Klaus (too hot to scroll) - it's not changed coz we pinned it, the lock file uses the current `environment` file so it had initially failed its scheduled 10-daily re-creation run coz of cf-units, so I had to rebuild it manually just to stick to the schedule - no actual fix needed, just sticking to the scheduled rebuild

PR #1722, opened 2022-09-15
Remove `pytest_flake8` plugin and use `flake8` instead

### LINKED ISSUE #1699: Consider dropping `pytest-flake8` plugin
It seems there's not much love between `flake8` and `pytest-flake8` (see eg https://github.com/tholo/pytest-flake8/issues/87#issuecomment-1216704377 ) and if you look up that issue there are people saying `pytest-flake8` is going its own way proper, and it's not condoned by the Pytest community - we've already had issues before and even now, and asking the `flake8` devs to fix stuff only worsens the situation, since they are not communicating with the plugin devs. Can we suggest alternatives for the plugin plss @ESMValGroup/technical-lead-development-team

valeriupredoi:
## Description

Discussed this and @bouweandela provided a working solution to the problem of the `pytest_flake8` plugin being more and more at odds with the mother package `flake8`. I think we should change the documentation too, tell users to run `flake8` before a push, but maybe not in this PR, what do you think @bouweandela ? We should also visit the containers :package: 

Closes #1699 

Link to documentation:

***

***

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ commands:
           command: |
             . /opt/conda/etc/profile.d/conda.sh
             conda activate esmvaltool
-            flake8 -j 4
+            flake8
             pytest -n 4 -m "not sequential" --junitxml=report-1.xml
             pytest -n 0 -m "sequential" --cov-append --junitxml=report-2.xml
             esmvaltool version
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ commands:
           command: |
             . /opt/conda/etc/profile.d/conda.sh
             conda activate esmvaltool
-            flake8
+            flake8 -j 4
             pytest -n 4 -m "not sequential" --junitxml=report-1.xml
             pytest -n 0 -m "sequential" --cov-append --junitxml=report-2.xml
             esmvaltool version
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,6 +27,7 @@ commands:
           command: |
             . /opt/conda/etc/profile.d/conda.sh
             conda activate esmvaltool
+            flake8 -j 4
             pytest -n 4 -m "not sequential" --junitxml=report-1.xml
             pytest -n 0 -m "sequential" --cov-append --junitxml=report-2.xml
             esmvaltool version
--- a/.github/workflows/create-condalock-file.yml
+++ b/.github/workflows/create-condalock-file.yml
@@ -64,7 +64,7 @@ jobs:
       - shell: bash -l {0}
         run: esmvaltool version
       - shell: bash -l {0}
-        run: flake8 -j 4
+        run: flake8
       - shell: bash -l {0}
         run: pytest -n 2 -m "not installation and not sequential"
       - shell: bash -l {0}
--- a/.github/workflows/create-condalock-file.yml
+++ b/.github/workflows/create-condalock-file.yml
@@ -64,6 +64,8 @@ jobs:
       - shell: bash -l {0}
         run: esmvaltool version
       - shell: bash -l {0}
+        run: flake8 -j 4
+      - shell: bash -l {0}
         run: pytest -n 2 -m "not installation and not sequential"
       - shell: bash -l {0}
         run: pytest -n 0 -m "sequential"
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ REQUIREMENTS = {
         'jinja2',
         'nc-time-axis',  # needed by iris.plot
         'nested-lookup',
-        'netCDF4',
+        'netCDF4<1.6.1',  # segfaults galore
         'numpy',
         'pandas',
         'pillow',
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,6 @@ REQUIREMENTS = {
         'pytest>=3.9,!=6.0.0rc1,!=6.0.0',
         'pytest-cov>=2.10.1',
         'pytest-env',
-        'pytest-flake8>=1.0.6',
         'pytest-html!=2.1.0',
         'pytest-metadata>=1.5.1',
         'pytest-mypy',
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ REQUIREMENTS = {
         'jinja2',
         'nc-time-axis',  # needed by iris.plot
         'nested-lookup',
-        'netCDF4<1.6.1',  # segfaults galore
+        'netCDF4!=1.6.1',  # github.com/ESMValGroup/ESMValCore/issues/1723
         'numpy',
         'pandas',
         'pillow',
--- a/.github/workflows/install-from-condalock-file.yml
+++ b/.github/workflows/install-from-condalock-file.yml
@@ -59,7 +59,7 @@ jobs:
       - shell: bash -l {0}
         run: esmvaltool version 2>&1 | tee source_install_linux_artifacts_python_${{ matrix.python-version }}/version.txt
       - shell: bash -l {0}
-        run: flake8 -j 4
+        run: flake8
       - shell: bash -l {0}
         run: pytest -n 2 -m "not installation and not sequential"
       - shell: bash -l {0}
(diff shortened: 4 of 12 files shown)


valeriupredoi:
so the good news is that `flake8` runs fine on me local machine (as well as the CI machines) but produces no progress bar/dots/flying cats etc - when it finishes it just exists the prompt - the user won't know what the heck's going on, @bouweandela know a way to show some progress on screen? Like mamba - shooting lasers, now that's cool :star2:

valeriupredoi:
Ha! Now that's an odd one - get rid of a pytest plugin, immediately run into a HDF5 segfault, well that escalated quickly :rofl:

schlunma:
> Ha! Now that's an odd one - get rid of a pytest plugin, immediately run into a HDF5 segfault, well that escalated quickly rofl

:smiley: 

@taly-pereira ran into the same problem today with a recipe that ran fine for me. We think that a new `hdf5` version might be the problem, as I am using

```
hdf5                     1.12.1         mpi_mpich_h08b82f9_4   conda-forge
```

whereas she uses (freshly installed environment from today)

```
hdf5                      1.12.2          mpi_mpich_h08b82f9_0    conda-forge
```

We are currently downgrading her version to check if that fixes the problem...

schlunma:
Weird. I haven't found anything on their GitHub...it also looks like 1.12.2 is not the latest version

valeriupredoi:
it is https://anaconda.org/conda-forge/hdf5/files - but it's fairly old and we've not seen anything weird until today?

valeriupredoi:
it's netCDF4 that's the problem - updated 5h ago https://anaconda.org/conda-forge/netcdf4

schlunma:
Ahh, yes, that makes sense!

schlunma:
Maybe pin it to `!=1.6.1` so that fixed version gets pulled in automatically in the future?

valeriupredoi:
yis, doing that now :beer:

.github/workflows/create-condalock-file.yml
```diff
@@ -63,6 +63,8 @@ jobs:
         run: esmvaltool --help
       - shell: bash -l {0}
         run: esmvaltool version
+      - shell: bash -l {0}
+        run: flake8 -j 4
```
bouweandela, line 67: You'll probably want to set the number of jobs on GitHub actions to 2 for Linux and 3 for osx: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

bouweandela:
> I think we should change the documentation too, tell users to run flake8 before a push, but maybe not in this PR, what do you think @bouweandela ?

We have pretty extensive advice here aleady: https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#code-quality, including the advice to use pre-commit, which runs flake8 as well as a bunch of other useful checks.

.github/workflows/create-condalock-file.yml
valeriupredoi: does it matter? I got two cores too on me trusted Toshiba, but I can thread as many pytest processes as I want - well, as many threads as I have

valeriupredoi:
> > I think we should change the documentation too, tell users to run flake8 before a push, but maybe not in this PR, what do you think @bouweandela ?
> 
> We have pretty extensive advice here aleady: https://docs.esmvaltool.org/projects/esmvalcore/en/latest/contributing.html#code-quality, including the advice to use pre-commit, which runs flake8 as well as a bunch of other useful checks.

oh yeah I knew about that - very nice indeed - but was wondering if it'd be worth it to tell an (advanced) developer how to check the code standards quality across the package with `flake8 -j X`?

bouweandela:
No: we're already recommending `prospector`, which will probably tell you about all flake8 issues and more, and then we're recommending pre-commit, so I think that's enough. People already complain that our docs are too lengthy.

.github/workflows/create-condalock-file.yml
bouweandela: Yes, if you have more jobs running than CPU cores, the computer will start [context switching](https://en.wikipedia.org/wiki/Context_switch), which will slow things down. You should either not specify the number of jobs at all (flake8 tries to automatically detect the correct number) or set it to the number of CPU cores.

valeriupredoi:
OK fair enough. That who complains the docs are too long has not read the fsspec documentation :grin:

.github/workflows/create-condalock-file.yml
valeriupredoi: OK I'll let flake8 decide but am not changing the pytest -n numbers, coz that's too much work :grin: Also is that thread context switching or process context switching coz if it's thread then we should be fine - 2 cores with at least two threads each should be enough for -n 4

valeriupredoi: nevermind my blabber - we have `pytest -n 2` set everywhere and the OSX tests are slow just because of OSX not because they don't use all the 3 allocated cores, also 2 or 3 cores? What's this, 1998? :laughing:

bouweandela: > OK I'll let flake8 decide but am not changing the pytest -n numbers, coz that's too much work

:+1: 

> Also is that thread context switching or process context switching coz if it's thread then we should be fine - 2 cores with at least two threads each should be enough for -n 4

If you think this is something worth spending time on, then I would recommend running some experiments with different settings

valeriupredoi: nah man, see my latest comment - I thinks we're all good, Max Verstappen Red Bull performance :cow2: :netherlands:

.circleci/config.yml
```diff
@@ -27,6 +27,7 @@ commands:
           command: |
             . /opt/conda/etc/profile.d/conda.sh
             conda activate esmvaltool
+            flake8
```
bouweandela, line 30: Could you please use
```suggestion
            flake8 -j 4
```
because [automatically detecting the number of CPUs does not work well on CircleCI](https://support.circleci.com/hc/en-us/articles/360005442714-Your-test-tools-are-smart-and-that-s-a-problem-Learn-about-when-optimization-goes-wrong-) machines.

setup.py
```diff
@@ -42,7 +42,7 @@
         'jinja2',
         'nc-time-axis',  # needed by iris.plot
         'nested-lookup',
-        'netCDF4',
+        'netCDF4!=1.6.1',  # github.com/ESMValGroup/ESMValCore/issues/1723
```
bouweandela, line 45: These changes look unrelated to this pull request?

valeriupredoi: eagle-eyed Bouwe :eagle: but also "Man, you weren't there!" (Vietnam reference) - check the whole panic from yesterday when @schlunma and myself were trying to figure out why the tests are dying spectacularly with a flurry of SegFaults - also this https://github.com/Unidata/netcdf4-python/issues/1192 and this https://github.com/ESMValGroup/ESMValCore/pull/1724

valeriupredoi: BTW remember our old SegFaults that eventually almost totally died out like Covid? Back with the vengeance :laughing:

.circleci/config.yml
valeriupredoi: OK good find! I think we're both spending a bit too much time on this very technical, minute matter already, must be Friday :grin:

setup.py
bouweandela: Awesome detective work you both. Since #1724 has now been merged, could you merge `main` into this branch, so the diff shows only the changes you actually want to make here?

valeriupredoi: done, bud :beer:

bouweandela:
Looks good to me now! Only when I run `flake8` on my computer I get a lot of errors coming from files in the `.eggs` directory. Do you think it would make sense to add that to the list of directories to ignore?

valeriupredoi:
cheers Bouwe! I don't have any eggs, nor does the CI - where do you get your eggs from, bro? :egg:

bouweandela:
It is created by setuptools, e.g. if you ever ran `python setup.py develop` or similar, you'll have it. Let's leave if for now.

bouweandela (approved):

PR #1730, opened 2022-09-22
[TESTING ONLY] Unpin `netCDF4` and run a bunch of hourly tests

valeriupredoi:
This is for **TESTING ONLY** - do not mind it! I have unpinned netCDF4 to pick up the latest (problematic) 1.6.1 and run hourly tests to be sure that the only one that craps the bed is the sample data cache one. FYI @bouweandela


valeriupredoi:
pretty clear by now all the failing tests originate in the test sample data module, gonna close

PR #1765, opened 2022-10-20
Fix rotated coordinate grids and `tas` and `pr` for CORDEX datasets

### LINKED ISSUE #1764: Dataset problem: CORDEX metadata issues
**Describe the dataset issue**
While @zklaus works on fixing grid issues with CORDEX datasets, there is also the need to add fixes for metadata issues or time coordinate issues in CORDEX datasets.

I will open a PR with the fixes that we noticed are needed so far.

FYI @pepcos @nhsavage 

**Assign the right people**
If you are already a member of the ESMValTool GitHub project, please assign Valeriu Predoi (valeriupredoi) and
Klaus Zimmermann (zklaus) to the issue. They will then check the issue raised and propagate it further to the
data model developers.

**Please attach**
  - Attach a text file with the output of `ncdump -h /path/to/file.nc`

Cheers :beer:

sloosvel:
## Description

This PR adds metadata fixes for CORDEX data, that may break the checker or the concatenation of historical+scenario data, as well as fixing a bug that prevented the loading of the X-hourly tables in the CORDEX project. It also solves issues with rotated coordinate grids using the [`py-cordex`](https://github.com/euro-cordex/py-cordex) package to access the correct grid information per each domain.

Closes #1764 

Link to documentation:

***

~- [ ] [🧪][2] and [🛠][1] [Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available~
~- [ ] [🛠][1] Any changed [dependencies have been added or removed](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#dependencies) correctly~

***


sloosvel:
Any news here? Or on the grid fixes for that matter? This seems to be getting stalled and we have work to do that requires this. Thanks!

esmvalcore/cmor/_fixes/cordex/mohc_hadgem2_es/smhi_rca4.py
```diff
@@ -0,0 +1,7 @@
+"""Fixes for rcm SMHI-RCA4 driven by MOHC-HadGEM2."""
```
pepcos, line 1:
```suggestion
"""Fixes for rcm SMHI-RCA4 driven by MOHC-HadGEM2-ES."""
```

esmvalcore/cmor/_fixes/cordex/mohc_hadgem2_es/mohc_hadrem3_ga7_05.py
```diff
@@ -0,0 +1,7 @@
+"""Fixes for rcm MOHC-HadREM3-GA7-05 driven by MOHC-HadGEM2."""
```
pepcos, line 1:
```suggestion
"""Fixes for rcm MOHC-HadREM3-GA7-05 driven by MOHC-HadGEM2-ES."""
```

sloosvel:
@ESMValGroup/esmvaltool-coreteam would anybody have time to review this? It's been one month and a half of no news :shrug: 

It looks like a lot but it's just a set of very basic data fixes. There is also a small fix that prevented the loading of the X-hourly tables. The most relevant part is that the coordinates for datasets in CORDEX that are in a rotated coord system are set according to the [grid specifications](https://cordex.org/wp-content/uploads/2012/11/CORDEX-domain-description_231015.pdf) that are stored in the external package `py-cordex`. This does not fix datasets in a Lambert Conformal systems.

For example there are several datasets defined in Europe that have wrong coordinates that set them in Africa. With this fix problems like that can be avoided because the grid information is substituted and the resulting coordinates after the fix are properly set in Europe.

image (1)


@remi-kazeroni this is relevant for IS-ENES as it covers the work that should have been delivered a year ago but no one cared to do.

I can provide a recipe to test once Levante is back from maintenance.

remi-kazeroni:
> @remi-kazeroni this is relevant for IS-ENES as it covers the work that should have been delivered a year ago but no one cared to do.
> 
> I can provide a recipe to test once Levante is back from maintenance.

@sloosvel, thanks a lot for picking that up and working on that! 👍  As I've no experience working with CORDEX data, I don't feel qualified to review this PR myself. @zklaus, since you were also involved in the IS-ENES3 task related to CORDEX, would you maybe have time to review this PR?

sloosvel:
>As I've no experience working with CORDEX data, I don't feel qualified to review this PR mysel

No worries! that should not be a problem! Most of these fixes don't require any knowledge in CORDEX data, they are just regular fixes that change metadata.

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
```diff
@@ -0,0 +1,168 @@
...
+        domain_info = cx.domain_info(data_domain)
+        for dim_coord in ['rlat', 'rlon']:
+            old_coord = cube.coord(domain[dim_coord].standard_name)
+            old_coord_dims = old_coord.cube_dims(cube)
+            points = domain[dim_coord].data
+            coord_system = iris.coord_systems.RotatedGeogCS(
+                grid_north_pole_latitude=domain_info['pollat'],
+                grid_north_pole_longitude=domain_info['pollon']
+            )
+            new_coord = iris.coords.DimCoord(
```
pepcos, line 112: It would be good to log the differences between the original grid and the pycordex one. Maybe we could raise an error for the time being when the differences are larger than a threshold of 10e-4 deg (~11km). The same for the geographical coords. We could later discuss how to tackle the grids that do not conform to the standards in an other pull request

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
```diff
@@ -0,0 +1,168 @@
...
+            bounds = domain[f'{aux_coord}_vertices'].data
+            new_coord = iris.coords.AuxCoord(
+                points,
+                var_name=aux_coord,
+                standard_name=domain[aux_coord].standard_name,
+                long_name=domain[aux_coord].long_name,
+                units=Unit(domain[aux_coord].units),
+                bounds=bounds
+            )
+            cube.add_aux_coord(new_coord, (1, 2))
```
pepcos, line 141: This will fail for fx variables such as areacella, since the axis would be 0 and 1.

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
```diff
@@ -0,0 +1,202 @@
...
+            new_coord = iris.coords.AuxCoord(
+                points,
+                var_name=aux_coord,
+                standard_name=domain[aux_coord].standard_name,
+                long_name=domain[aux_coord].long_name,
+                units=Unit(domain[aux_coord].units),
+                bounds=bounds
+            )
+            logger.info(old_coord.units)
+            self._check_grid_differences(old_coord, new_coord)
```
sloosvel, line 168: A value conversion for longitudes should be added before this check, because if the longitudes are given in different ranges, the check will report an error when it may not be the case

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
sloosvel: Done as well, I generalised the dimension values

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
pepcos: looks great to me

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
sloosvel: Actually this case only happens for one dataset and driver and I think the check is right to be raised, because the longitude values are quite wrong.

sloosvel:
Ok, tests are failing due to #1870, but I addressed all your comments @jcos. Here is a recipe to test the datasets that are being fixed in this PR, which are available in levante: 

[recipe_cordex.yml.txt](https://github.com/ESMValGroup/ESMValCore/files/10229321/recipe_cordex.yml.txt)

The datasets that are commented are the ones that have grid differences larger than the threshold that we added, which will make stop the run. Those datasets should be fixed in upcoming work.

pepcos (approved):
I've ran the recipe with the CORDEX data in levante and everything expected to work worked. As the lambert comformal grids are not yet supported I'd add a Warning saying so if such a grid is found in case another error arises after.

sloosvel:
Other than the tests failing due to #1765 , I would say this is ready. Anyone wants to double check @ESMValGroup/technical-lead-development-team ?

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
```diff
@@ -0,0 +1,205 @@
...
+            "points and standard %s domain points "
+            "for dataset %s and driver %s is: %s.",
+            new_coord.var_name,
+            self.extra_facets['domain'],
+            self.extra_facets['dataset'],
+            self.extra_facets['driver'],
+            str(diff)
+        )
+
+        if diff > 10e-4:
```
zklaus, line 116: I think we might want to have this threshold as a parameter. For different projections, the difference in real meters corresponding to the difference in projection units can be vastly different. Even when those projection units are "meters", which often then means "meters in the projection plane".

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
```diff
@@ -0,0 +1,205 @@
...
+            cube.remove_coord(old_coord)
+            points = domain[aux_coord].data
+            bounds = domain[f'{aux_coord}_vertices'].data
+            new_coord = iris.coords.AuxCoord(
+                points,
+                var_name=aux_coord,
+                standard_name=domain[aux_coord].standard_name,
+                long_name=domain[aux_coord].long_name,
+                units=Unit(domain[aux_coord].units),
+                bounds=bounds
```
zklaus, line 163:
```suggestion
                bounds=bounds,
```
Adding a comma here at the end will ensure that our auto formatter yapf doesn't mess it up in the future.

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
```diff
@@ -0,0 +1,205 @@
...
+        Returns
+        -------
+        iris.cube.CubeList
+
+        """
+        for cube in cubes:
+            coord_system = cube.coord_system()
+            if isinstance(coord_system, RotatedGeogCS):
+                self._fix_rotated_coords(cube)
+                self._fix_geographical_coords(cube)
```
zklaus, line 197: Since both use the same `domain` dataset, consider creating it here and passing it to the functions. Timing the creation of that dataset shows that it takes half a second, so not deadly, but longer then expected, particularly if used with a large number of datasets. Perhaps even consider adding a function for its creation with a cache, something like
```python
@lru_cache
def _get_domain(domain_name):
    return cx.cordex_domain(domain_name, add_vertices=True)
```
This you could use in both functions unmodified thanks to the cache. If you opt for this, don't forget to add the
```python
from functools import lru_cache
```
at the top.

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
```diff
@@ -0,0 +1,205 @@
...
+            if isinstance(coord_system, LambertConformal):
```
zklaus, line 198: Perhaps it would be better as an `elif`? In any case, I would suggest also adding an `else` branch to make sure other grids don't just fall through the cracks. This will allows us to search through the logs when someone complains.

esmvalcore/cmor/check.py
```diff
@@ -1081,6 +1081,8 @@ def _get_cmor_checker(table,
                 table, ', '.join(CMOR_TABLES)))
 
     cmor_table = CMOR_TABLES[table]
+    if table == 'CORDEX' and mip.endswith('hr'):
+        mip = mip.replace('hr', 'h')
```
zklaus, line 1085: Add short explanatory comment?

zklaus (commented):
I haven't looked at the fixes yet, but this looks great! Thanks for the work. I left a few minor comments and I have one more that I couldn't attach to any code:

Most `__init__.py` files have a short oneliner at the top, but `tests/integration/cmor/_fixes/cordex/__init__.py` is empty. Should we add the usual one-liner also there?

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
```diff
@@ -0,0 +1,205 @@
...
+        -------
+        iris.cube.CubeList
+
+        """
+        for cube in cubes:
+            if cube.coord('time').units.calendar == 'standard':
+                cube.coord('time').units = Unit(
+                    'days since 1850-1-1 00:00:00',
+                    calendar='proleptic_gregorian'
+                )
```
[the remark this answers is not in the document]
sloosvel: Because some files had standard calendars and some other proleptic gregorian and that broke the concatenation. The reference date was the same in that case, yes.

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
sloosvel: I am not sure what you fully mean here. With @pepcos we tested this value with -11 and -44 domains and the datasets that passed the checks were quite decent. In any case I would leave it like that for now and if we see that with that threshold we are allowing datasets that are not good enough maybe we can find a way to tune it.

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
sloosvel: Ok, I will move the domain retrieval out of the loop

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
sloosvel: Ok, will do

esmvalcore/cmor/_fixes/cordex/mohc_hadgem2_es/dmi_hirham5.py
```diff
@@ -0,0 +1,25 @@
...
+            Input cubes.
+
+        Returns
+        -------
+        iris.cube.CubeList
+
+        """
+        for cube in cubes:
+            cube.coord('latitude').attributes = {}
+            cube.coord('longitude').attributes = {}
```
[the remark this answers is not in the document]
sloosvel: The historical exp had random attributes in the coords, whereas the scenario ones didn't (or the other way around) and that broke the concatenation

esmvalcore/cmor/check.py
sloosvel: Ok, will do

tests/integration/cmor/_fixes/cordex/test_cordex_fixes.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+def test_mohchadrem3ga705_fix_metadata(cubes):
+    fix = MOHCHadREM3GA705(None)
+    out_cubes = fix.fix_metadata(cubes)
+    assert cubes is out_cubes
+    for cube in out_cubes:
+        for coord in cube.coords():
+            right_coord = cubes[0].coord(coord)
+            assert coord == right_coord
```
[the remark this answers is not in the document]
sloosvel: Yes but cubes[0] is always the correct one, so I am comparing the output of the check to a cube that has not been changed because it's correct from the beginning.

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
sloosvel: Done

esmvalcore/cmor/check.py
sloosvel: Done

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
sloosvel: Done

sloosvel:
Ok, I think I addressed everything @bouweandela

esmvalcore/cmor/_fixes/cordex/mohc_hadgem2_es/dmi_hirham5.py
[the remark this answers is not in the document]
bouweandela: It would have been nice to add that as a comment, but let's leave it for now since you don't remember exactly.

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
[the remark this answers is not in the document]
bouweandela: How about using the [`change_calendar`](https://cf-units.readthedocs.io/en/latest/unit.html#cf_units.Unit.change_calendar) method, would that be safer?

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
bouweandela: Tuning it later sounds fine to me. Out of curiosity: If I understand it correctly, `-11` and `-44` refers to specific resolutions, but wouldn't you expect the projection and region to have a larger effect on this? Did you test with all CORDEX regions and all supported projections?

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
bouweandela: Did you add caching too? I couldn't find it in the code. Half a second per file can become time-consuming when loading many files.

tests/integration/cmor/_fixes/cordex/test_cordex_fixes.py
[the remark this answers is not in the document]
bouweandela: It would be good to add a comment to the code to explain that. Like that, the information is available for future developers.

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
sloosvel: I am not sure about that because I see there are a bunch of issues with caching inside classes

tests/integration/cmor/_fixes/cordex/test_cordex_fixes.py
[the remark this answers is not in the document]
sloosvel: I see, I have updated the test then. It's parametrising now

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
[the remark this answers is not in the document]
sloosvel: Done, I did not know about change_calendar, that will be useful!

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
sloosvel: Nevermind, this can be sorted if defining the functions outside the class

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
```diff
@@ -0,0 +1,206 @@
...
+            "Maximum difference between original %s"
+            "points and standard %s domain points  "
+            "for dataset %s and driver %s is: %s.", new_coord.var_name,
+            self.extra_facets['domain'], self.extra_facets['dataset'],
+            self.extra_facets['driver'], str(diff))
+
+        if diff > 10e-4:
+            raise RecipeError(
+                "Differences between the original grid and the "
+                f"standardised grid are above 10e-4 {new_coord.units}.", )
```
bouweandela, line 125:
```suggestion
                f"standardised grid are above 10e-4 {new_coord.units}.")
```

esmvalcore/cmor/_fixes/cordex/cordex_fixes.py
```diff
@@ -0,0 +1,206 @@
...
+            elif isinstance(coord_system, LambertConformal):
+                logger.warning(
+                    "Support for CORDEX datasets in a Lambert Conformal "
+                    "coordinate system is ongoing. Certain preprocessor "
+                    "functions may fail.")
+            else:
+                raise RecipeError(
+                    f"Coordinate system {coord_system.grid_mapping_name} "
+                    "not supported in CORDEX datasets. Must be "
+                    "rotated_latitude_longitude or lambert_conformal_conic.", )
```
bouweandela, line 204:
```suggestion
                    "rotated_latitude_longitude or lambert_conformal_conic.")
```

bouweandela:
Thanks for the great work @sloosvel @pepcos and @zklaus :partying_face:  Really awesome that we're finally getting CORDEX to work!

PR #1770, opened 2022-10-25
Unpin cf-units

### LINKED ISSUE #1655: Test failure due to change in cf-units

bouweandela:
## Description

Closes #1655. Continuing from #1656.

It would be nice to unpin cf-units, to make working with recent iris versions easier.

***

***


bouweandela:
There is one remaining test failure, but it looks like it is caused by caching intermediate results on CircleCI. The cache key contains the date and branch, so this this should solve itself by tomorrow. Running an extra check here: https://app.circleci.com/pipelines/github/ESMValGroup/ESMValCore?branch=unpin-cf-units-fix-cache

bouweandela:
Still failing, will have another look on Thursday..

zklaus:
I think the issue is that https://github.com/ESMValGroup/ESMValTool_sample_data doesn't contain any data with calendar `standard`, but only with the equivalent calendar `gregorian`, that has been deprecated in CF. Consequently, the change in the first commit in this PR means that the test is asking for a non-existent category.

Solutions might be to add `standard` data to the sample data or to add an alias for the existing `gregorian` data via `cube_dict["standard"] += cube_dict["gregorian"]` or similar [here](https://github.com/ESMValGroup/ESMValCore/blob/3149d07ebdb314c5bd2c17323f1d8c5b208dea4d/tests/sample_data/multimodel_statistics/test_multimodel.py#L169).

valeriupredoi:
whoa that's a hefty one! Cheers, gents! @bouweandela could you pls fire up the GA tests too before merging (and unfire them up once things pass) :beer:

bouweandela:
> I think the issue is that https://github.com/ESMValGroup/ESMValTool_sample_data doesn't contain any data with calendar standard, but only with the equivalent calendar gregorian, that has been deprecated in CF. Consequently, the change in the first commit in this PR means that the test is asking for a non-existent category.

@zklaus Thanks for the pointer! Any idea why tests that are using a fresh installation are passing just fine with the same data?

bouweandela:
> why tests that are using a fresh installation are passing just fine with the same data?

To answer my own question: the tests work fine with newer versions of cf-units, but fail with 3.0.1 which is still used in the test that start from the docker container. I'll try to fix things so the tests pass with both versions.

valeriupredoi:
> whoa that's a hefty one! Cheers, gents! @bouweandela could you pls fire up the GA tests too before merging (and unfire them up once things pass) beer

@bouweandela could you pls merge main here, then fire up the GA tests, then fire them off if all goes to plan? Let me know, I can do that too, am not just passing orders :grin:

bouweandela:
@valeriupredoi All tests passed

tests/sample_data/multimodel_statistics/test_multimodel.py
```diff
@@ -24,7 +25,7 @@
         marks=pytest.mark.skip(
             reason='Cannot calculate statistics with single cube in list')),
     '365_day',
-    'gregorian',
+    'standard' if cf_units.__version__ >= '3.1' else 'gregorian',
```
valeriupredoi, line 28: this is a spanner in the works - we shouldn't test like this, we should mark them as xfailed if the version is lower than 3.1 since ideally that'll not happen

valeriupredoi (approved):
thanks a lot @bouweandela :beer: You sure you don't want to pin the lower end to >3.1 - that'd be harmless and we'd be on the safe way too?

bouweandela:
> You sure you don't want to pin the lower end to >3.1 - that'd be harmless and we'd be on the safe way too?

@zklaus carefully wrote all the tests so they work with both versions, so I just continued with that idea. The nice thing about that is that if someone does `git pull main`, things will just keep working without them having to install new software. The downside is of course that it makes the code slightly more complicated.

tests/sample_data/multimodel_statistics/test_multimodel.py
bouweandela: Note that the other tests are also written to support both the new and older versions of cf-units

valeriupredoi: the tests are, but passing is not the point of a test, is the functionality in our code dual as well? If yes, I is happy and no more nagging :grin:

bouweandela: It looks like `'gregorian'` as a calendar was removed in v1.9 of the CF-conventions. I'll update the code too so it does something sensible..

bouweandela: I think the code works as is with both versions, it's just that versions of cf-units >= 3.1 will automatically convert the `'gregorian'` calendar to `'standard'`.

valeriupredoi: OK sounds good, thanks, bud :beer:

bouweandela:
@ESMValGroup/technical-lead-development-team Could someone please have a final look at this and merge?

sloosvel:
Just to be annoying and nitpicky, I did a search on `gregorian` and the docstrings in _unify_time_coordinate (in the multimodel module) still mention it. Maybe it could be updated.

bouweandela:
Thanks for checking @sloosvel, I just changed it.

bouweandela:
OK, now the tests are failing due to https://github.com/ESMValGroup/ESMValTool/issues/2924

bouweandela:
It would be best to merge #1805 before this pull request.

sloosvel:
Sine #1805 is merged, I will merge this

PR #1772, opened 2022-10-26
Fixed time points of native CESM2 output

schlunma:
## Description

I noticed that the time points of native CESM2 output are located at the end of the time interval. This PR introduces a fix that moves these to the center of the interval.

***


schlunma:
In my specific use case, I had problems with this data when calculating monthly climatologies from monthly data with the preprocessor. Since the points are located at the end of the periods (= first day of the following month), this effectively shifts all data by 1 month and leads to shifted climatology.

E.g., the time points and bounds look like this

```
points = [1-2-2000, 1-3-2000, 1-4-2000, ...]
bounds = [[1-1-2000, 1-2-2000], [1-2-2000, 1-3-2000], [1-3-2000, 1-4-2000], ...]
```

Here, the first points is identified as February, but it's actually January, and so on.

zklaus:
I see. I guess for monthly and lower-frequency data it is not likely that the data is pointwise (in time), though checking the cell method might be good. I would be more careful with daily and sub-daily data.

schlunma:
So what would you recommend? Something like this?


```py
if cell_methods == "time: point":
    # do nothing
else:
    # fix point = mean(bounds)
```

or additionally check for the frequency of the data? Or maybe even check for `cell_methods == "time: mean/sum"` and only fix if that's true?

zklaus:
I am a bit unsure. Since in principle there is nothing wrong with the time coordinate, might it be better to just use an explicit `regrid_time` preprocessor where needed?

schlunma:
`regrid_time` does not work since it sets the time point to the center of the given month:

https://github.com/ESMValGroup/ESMValCore/blob/3149d07ebdb314c5bd2c17323f1d8c5b208dea4d/esmvalcore/preprocessor/_time.py#L915-L918

In this case here, it sets `1-2-2022` (which should be January) to `15-2-2022`, which is even more problematic.

I understand that technically it might not be wrong to set the time point to the end of the interval, but given that literally every other monthly dataset I've seen uses a point roughly in the center of the interval, I think it would be a good idea if this CMORizer did the same for CESM2, too (at least for monthly data when `cell_method` is not `time: point`).

zklaus:
Alright, I think that is a bug in `regrid_time`, but in order to move forward here, I agree to do this as you suggest for monthly data when `cell_method` is not `point`.

I do belief it's reasonably common to have the point at the end of the interval when the quantity in question is an accumulation over the interval, particularly where it is the only variable from a data provider such as in observations from stations.

bouweandela:
Indeed we have a similar fix for ERA5: https://github.com/ESMValGroup/ESMValCore/blob/3666bfa2380c53467cbb0474fba0a49426ff5e7f/esmvalcore/cmor/_fixes/native6/era5.py#L38. There we compute means from aggregated quantities and then give the means the mid time point instead of the end.

schlunma:
@zklaus would be great if you could have a look at that, thanks! :+1:

esmvalcore/cmor/_fixes/cesm/cesm2.py
```diff
@@ -49,3 +49,27 @@ def fix_metadata(self, cubes):
         self.fix_var_metadata(cube)
 
         return CubeList([cube])
+
+    def _fix_time(self, cube):
+        """Fix time coordinate of cube.
+
+        For monthly data that does not correspond to point measurements, move
+        time points to center of time period given by time bounds (currently
+        the points are located at the end of the interval).
```
valeriupredoi, line 58: please provide this in the docstring: an example of old data, and example of new data as per desired behaviour

tests/integration/cmor/_fixes/cesm/test_cesm2.py
```diff
@@ -255,6 +275,38 @@ def test_only_longitude(monkeypatch):
...
+
+
+def test_fix_time_day(monkeypatch, cube_1d_time):
+    """Test `_fix_time``."""
+    fix = get_allvars_fix('Amon', 'mon', 'tas')
+    monkeypatch.setitem(fix.extra_facets, 'frequency', 'day')
+    fix._fix_time(cube_1d_time)
+    time_coord = cube_1d_time.coord('time')
+    np.testing.assert_array_equal(time_coord.points, [2, 4, 6])
+    np.testing.assert_array_equal(time_coord.bounds, [[0, 2], [2, 4], [4, 6]])
```
valeriupredoi, line 307: but where was the fix applied? These are identical to the input in `cube_1d_time`? Or am I blind :eyeglasses:

valeriupredoi (changes requested):
I am not sure I can be the right person to approve this PR simply since I don't have experience with CESM2 data - but from a technical pov I have two little wiggles. Maybe it's best you and @zklaus settle the argument first, then I can greenlight this one. Cheers, Manu! :beer:

tests/integration/cmor/_fixes/cesm/test_cesm2.py
schlunma: Yes, the fix does indeed nothing here, since it looks like daily data. As agreed with @zklaus, the fix is only supposed to change monthly data, all other data is unchanged.

esmvalcore/cmor/_fixes/cesm/cesm2.py
schlunma: Done :+1:

tests/integration/cmor/_fixes/cesm/test_cesm2.py
valeriupredoi: great, cheers, Manu!

valeriupredoi (approved):

valeriupredoi:
@ESMValGroup/technical-lead-development-team this is ready to be merged, can one of you peps pls have a look and do so - unless @zklaus would like to add more comments/suggestions - Klaus, please chime in if so :beer:

zklaus (approved):

PR #1830, opened 2022-11-28
Xfail esgf search tests if they fail with HTTP error from requests

valeriupredoi:
## Description

The ESGF node at CEDA has been down for a few good hours now, and our tests are [failing](https://github.com/ESMValGroup/ESMValCore/actions/runs/3560815574/jobs/5988323814) and rightly so! But this is the last thing we'd want to see when, say, we make a release and the package fails to build due to an annoying, and independent of us issue. I marked those tests as xfailed only when the exception occurs; most of the time it doesn't so we  are still testing correctly - this id done like this:

- if test fails with requests HTTP error then it's marked as xfailed
- if test fails with any other exception then it's marked as red failed
- if test passes (normal behaviour) then it's marked as xpassed

***

***

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -21,7 +21,6 @@ on:
   push:
     branches:
       - main
-      - xfail_esgf_searh_test
   # run the test only if the PR is to main
   # turn it on if required
   #pull_request:


bouweandela:
Good idea to fix this, but shall we fix the code instead of the tests? I was doing a tutorial on the tool today with some people and we had to modify everyone's [~/.esmvaltool/esgf-pyclient.yml](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/quickstart/configure.html#search) file so it uses a different order of the index nodes.

bouweandela:
Created an issue to report the problem: #1833

valeriupredoi:
> Good idea to fix this, but shall we fix the code instead of the tests? I was doing a tutorial on the tool today with some people and we had to modify everyone's [~/.esmvaltool/esgf-pyclient.yml](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/quickstart/configure.html#search) file so it uses a different order of the index nodes.

Indeed, a code fix is better, but I think we should have a failsafe in place in testing too. Xpassed is better than failed 😁

valeriupredoi:
> Created an issue to report the problem: #1833

Cool! I also reported the index node issue to CEDA 👍

bouweandela:
Just created a pull request to fix the issue: 1834

bouweandela:
> we should have a failsafe in place in testing too. Xpassed is better than failed grin

Every user of ESMValCore with the default configuration is affected by this problem, so to me it seems really good that our unit test picked this up.

Issue #651, opened 2020-05-26
Anomalies preprocessor with period 'full' removes important metadata from cube


Peter9192:
**Describe the bug**
When the `anomalies` preprocessor function is applied with `period=full`, it removes important metadata such as the variable name, all attributes, etc. This doesn't happen for other periods like `month` or `day`. Consequently, subsequent preprocessors or diagnostics fail.

I think it's happening [here](https://github.com/ESMValGroup/ESMValCore/blob/8306a6561a011f61dec949dc56889568b1af815e/esmvalcore/preprocessor/_time.py#L463):
```
cube = cube - reference
```
 and this seems to be the intended behaviour in Iris, as it is documented [here](https://scitools.org.uk/iris/docs/latest/userguide/cube_maths.html#calculating-a-cube-anomaly). 

Perhaps EMSValTool should only subtract the data instead of the full cube?

**Example**
`print(cube)`:

```
air_temperature / (K)               (time: 1668)
     Dimension coordinates:
          time                           x
     Auxiliary coordinates:
          day_of_year                    x
          month_number                   x
     Scalar coordinates:
          height: 1.5 m
          latitude: 0.0 degrees, bound=(-90.0, 90.0) degrees
          longitude: 179.0625 degrees, bound=(-0.9375, 359.0625) degrees
     Attributes:
          Conventions: CF-1.7
          associated_files: baseURL: http://cmip-pcmdi.llnl.gov/CMIP5/dataLocation gridspecFile: gridspec_atmos_fx_ACCESS1-0_rcp45_r0i0p0.nc...
          branch_time: 732311.0;109207.0;732311.0
          cmor_version: 2.8.0
          contact: The ACCESS wiki: http://wiki.csiro.au/confluence/display/ACCESS/Home. Contact...
          experiment: RCP4.5;historical;RCP4.5
          experiment_id: rcp45;historical;rcp45
          forcing: GHG, Oz, SA, Sl, Vl, BC, OC, (GHG = CO2, N2O, CH4, CFC11, CFC12, CFC113,...
          frequency: mon
          initialization_method: 1
          institute_id: CSIRO-BOM
          institution: CSIRO (Commonwealth Scientific and Industrial Research Organisation, Australia),...
          model_id: ACCESS1-0
          modeling_realm: atmos
          parent_experiment: historical;pre-industrial control;historical
          parent_experiment_id: historical;piControl;historical
          parent_experiment_rip: r1i1p1
          physics_version: 1
[...]

PR #1837, opened 2022-11-29
Update product attributes and `metadata.yml` with cube metadata before saving files

### LINKED ISSUE #505: Units reported in metadata.yml may differ from actual units

schlunma:
## Description

This PR adds a function that updates the product attributes (and thus the `metadata.yml` files) with cube metadata before saving files. This ensures that the entries in `metadata.yml` match the cube metadata.

Currently modified metadata:

- `standard_name` from `cube.standard_name`
- `long_name` from `cube.long_name`
- `var_name` from `cube.short_name`
- `units` from `cube.units`
- `frequency` from `cube.attributes['frequency']`

Closes #505

Documentation: https://esmvaltool--1837.org.readthedocs.build/projects/ESMValCore/en/1837/develop/preprocessor_function.html#metadata

***

***


schlunma:
@bouweandela this is possible solution to #505.

In your [comment](https://github.com/ESMValGroup/ESMValCore/issues/505#issuecomment-1326187115) you mentioned also adapting frequencies accordingly. I thought about this some time now and found several possible solutions for this:

1. Use a similar function than the [one](https://github.com/ESMValGroup/ESMValCore/blob/320a275d1730c229dce1e9921d03cb3ec1b18d97/esmvalcore/cmor/check.py#L884-L938) used in the CMOR checks. This one checks **all** time points and might be very expensive, especially if we do not make any assumptions about the cube's frequency and let the code check everything.
2. Only check the first two points. This is much cheaper, but might lead to wrong results.
3. Somehow couple this to the preprocessors used. E.g., if someone uses `annual_statistics`, we know that the data is yearly afterwards. This might not be straightforward to implement and might get confusing if multiple of the units-changing preprocessors are used.

In addition, I have some questions:

- How do we deal with data that does not have a time dimensions anymore (e.g., after running `climate_statistics` with `period=full`). Use `fx`? Use `None`?
- How do we deal with climatologies (e.g., after running `climate_statistics` with `period=monthly`? For monthly climatologies, do we simply use `mon`? I am not entirly sure as this is different to a standard monthly time series.

All these problems lead me to the conclusion that it might be best to tackle the frequency update at a later stage. This PR solves an actual existing problem, whereas no one complained about wrong frequencies before. So maybe just open an issue about this and merge this without adapting the frequencies?

@ESMValGroup/esmvaltool-coreteam any opinions?

bouweandela:
My idea wasn't nearly as advanced as that: i thought we could just update the frequency in metadata.yml using the value of the `frequency` or `time_frequency` attribute on the cube.

Prepeocessors who change the frequency or any of the other attributes should then take care of updating the appropriate cube attributes.

bouweandela:
> Should I update all time preprocessors with the frequency attribute in this PR? If yes, I would need an answer to my questions above.

I would leave that for a new pull request.

> Are you proposing to update the product attributes from *all* cube attributes? I guess this can be a little bit dangerous, for example, if a dataset uses a key that we also use in the raw data (e.g., project) then we might accidentally incorrectly overwrite the product attributes.

No, just the ones that are wrong after preprocessing.

schlunma:
All right, implemented.

One remaining question: Currently, if a cube property (e.g., `cube.standard_name`) is `None`, the metadata is update with an empty string. This is in line with our current procedure in the CMOR checks.

However, in the light of preprocessors that completely remove cube metadata (https://github.com/ESMValGroup/ESMValCore/issues/651), this might lead to unexpected results. Would it be safer to just not update the corresponding values if it's `None`?

schlunma:
Corresponding frequency issue: https://github.com/ESMValGroup/ESMValCore/issues/1840

tests/unit/preprocessor/test_preprocessor_file.py
```diff
@@ -0,0 +1,163 @@
...
+
+    assert product.attributes == {
+        'filename': 'file.nc',
+        'standard_name': 'air_temperature',
+        'long_name': 'Near-Surface Air Temperature',
+        'short_name': '',
+        'units': 'K',
+        'frequency': 'day',
+    }
+    assert isinstance(product.attributes['units'], str)
```
bouweandela, line 105: This would be easier to maintain if it was parametrized using pytest

bouweandela (changes requested):
Looks good @schlunma! I tested this with the `convert_units` preprocessor function and that works fine now.

esmvalcore/preprocessor/__init__.py
```diff
@@ -463,10 +465,34 @@ def save(self):
...
         if self._cubes is not None:
+            self.update_attributes()
             self.save()
             self._cubes = None
             self.save_provenance()
 
+    def update_attributes(self):
+        """Update product attributes from cube metadata."""
+        if not self._cubes:
+            return
```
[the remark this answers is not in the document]
schlunma: I would rather leave that here. It's a public function, so in theory it could also be called at other places. Also, the check ensures that the CubeList is not empty.

tests/unit/preprocessor/test_preprocessor_file.py
schlunma: So do we want to keep the current behavior regarding `None` names? Set them to `''`?

esmvalcore/preprocessor/__init__.py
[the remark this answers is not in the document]
bouweandela: I'm not too happy about PreprocessorFile as part of the public API because I think it isn't well enough designed for that. I never intended it to be public, this is an accident, that's why it's not in `__all__` and not documented. Could you just add the new method as a private method, so at least we're not extending this unwanted public API?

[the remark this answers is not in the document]
bouweandela: I created an issue to discuss this: #1847

tests/unit/preprocessor/test_preprocessor_file.py
bouweandela: There's probably some NCL diagnostics depending on that.. so I think yes?

schlunma: My question is rather: do we update the names at all if they are `None` or do we keep the original value in this case?

bouweandela: I imagine this could be useful if a preprocessor modifies a variable in such a way that the `standard_name` is no longer valid? For the other attributes, I'm less sure if it is useful to have a mechanism to unset them.

schlunma: I also think having this in there is good, but we have to keep in mind that this might lead to unexpected changes (https://github.com/ESMValGroup/ESMValCore/issues/651).

bouweandela: That looks like a bug in a preprocessor function, right? This change will mean that bug is going to propagate into the metadata.yml file, but if the bug was fixed things should work fine again.

schlunma: Yes, that's definitely a bug, just wanted to mention that this might cause problems when testing actual recipes.

esmvalcore/preprocessor/__init__.py
```diff
@@ -463,10 +465,34 @@ def save(self):
...
+
+        # Names
+        names = {
+            'standard_name': 'standard_name',
+            'long_name': 'long_name',
+            'var_name': 'short_name',
+        }
+        for (name_in, name_out) in names.items():
+            cube_val = getattr(ref_cube, name_in)
+            self.attributes[name_out] = '' if cube_val is None else cube_val
```
valeriupredoi, line 487: are we 100% sure this will never fail with a `KeyError`?

esmvalcore/preprocessor/__init__.py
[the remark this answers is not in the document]
valeriupredoi: Manu marked this as resolved, which one is it - all good by you @bouweandela and things will be treated in a separate PR to close that issue?

tests/unit/preprocessor/test_preprocessor_file.py
```diff
@@ -0,0 +1,142 @@
...
+    product._update_attributes()
+
+    assert product.attributes == {
+        'filename': 'file.nc',
+        'standard_name': 'air_temperature',
+        'long_name': 'Near-Surface Air Temperature',
+        'short_name': 'tas',
+        'units': 'K',
+        'frequency': 'day',
+    }
```
valeriupredoi, line 56: we should start thinking using dedicated tools that do comparisons like [deepdiff](https://pypi.org/project/deepdiff/)

valeriupredoi (commented):
cheers guys - am the merger, so I am not requesting any changes, but I have a couple questions before I merge this piece of art :grin:

tests/unit/preprocessor/test_preprocessor_file.py
schlunma: Looks definitely interesting, but maybe a bit of an overkill here (since it's only a flat dict)?

valeriupredoi: yeah not here, more like a me being trying to be a showoff remark :laughing: Just a pointer for future, on a serious note

esmvalcore/preprocessor/__init__.py
valeriupredoi: OK cool :beer:

valeriupredoi:
thanks a lot @schlunma and @bouweandela :beer:

PR #1838, opened 2022-11-29
Update CEDA ESGF index node hostname

valeriupredoi:
## Description

Kind info provided by @alaniwi - many thanks, Alan!

***

***


tests/unit/esgf/test_search.py
```diff
@@ -119,7 +119,7 @@ def get_mock_connection(mocker, search_results):
     cfg = {
         'search_connection': {
             'urls': [
-                'https://esgf-index1.example.com/esg-search',
+                'https://esgf.ceda.ac.uk/esg-search',
```
bouweandela, line 122: This looks like a mistake?

tests/unit/esgf/test_search.py
```diff
@@ -208,7 +208,7 @@ def test_esgf_search_files(mocker):
     files = _search.esgf_search_files(facets)
 
     SearchConnection.assert_called_once_with(
-        url='https://esgf-index1.example.com/esg-search')
+        url='https://esgf.ceda.ac.uk/esg-search')
```
bouweandela, line 211: This looks like a mistake?

tests/integration/esgf/search_results/Amon_r1i1p1_historical,rcp85_INM-CM4_CMIP5_tas.json
```diff
@@ -133,7 +133,7 @@
         "format": [
             "netCDF, CF-1.4"
         ],
-        "index_node": "esgf-index1.ceda.ac.uk",
+        "index_node": "esgf.ceda.ac.uk",
```
bouweandela, line 136: Is it necessary to update the index node in these old search results used for testing?

bouweandela (changes requested):
Thanks for fixing this @valeriupredoi!

tests/unit/esgf/test_search.py
valeriupredoi, line 211:
```suggestion
        url='https://esgf.example.com/esg-search')
```

valeriupredoi: :man_facepalming:

bouweandela: It's still different from what it was, can we just keep it the same?

tests/integration/esgf/search_results/Amon_r1i1p1_historical,rcp85_INM-CM4_CMIP5_tas.json
bouweandela: If you think that's best, it's fine with me.

tests/unit/esgf/test_search.py
valeriupredoi: why keep the `index1` if that has completely vanished from the internets?

bouweandela: There are two index nodes in the mock search connection configuration used for testing in this file. Since these index nodes aren't real anyway (they use the example.com domain), I don't understand why this needs to be changed. But if you're really keen on changing them anyway, then please change both (now you're only changing the first one) and keep the numbering in the names because it makes the test code easier to understand if there is a '1' in the name of the first one and a '2' in the name of the second one.

tests/unit/esgf/test_search.py
```diff
@@ -119,7 +119,7 @@ def get_mock_connection(mocker, search_results):
     cfg = {
         'search_connection': {
             'urls': [
-                'https://esgf-index1.example.com/esg-search',
+                'https://esgf.example.com/esg-search',
                 'https://esgf-index2.example.com/esg-search',
```
bouweandela, line 123: What I meant is this: why rename `index1` and leave `index2`? At least rename both if you want to get rid of the word `index`, e.g.
```suggestion
                'https://esgf-1.example.com/esg-search',
                'https://esgf-2.example.com/esg-search',
```
the numbering makes sense if you look at the tests in the file, so please keep it.

tests/unit/esgf/test_search.py
bouweandela: It looks like we're misunderstanding each other. I wrote a new comment below, is that any more clear?

valeriupredoi:
hi @bouweandela sorry for the back and forthing, I have now reverted those two changes in test search, bud :beer:

bouweandela (approved):
All good now, thanks!

PR #1846, opened 2022-12-02
Removed unnecessary test that fails in iris 3.4.0

schlunma:
## Description

This PR removes a test that [fails ](https://github.com/ESMValGroup/ESMValCore/actions/runs/3593600178/jobs/6050760969)after a bugfix that has been introduced with iris 3.4.0 (https://github.com/SciTools/iris/pull/5062). The tests that check that `dtype` is preserved are now located in iris and can be removed from our test suite.

## [Checklist](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#checklist-for-pull-requests)

It is the responsibility of the author to make sure the pull request is ready to review. The icons indicate whether the item will be subject to the [🛠 Technical][1] or [🧪 Scientific][2] review.

[1]: https://docs.esmvaltool.org/en/latest/community/review.html#technical-review
[2]: https://docs.esmvaltool.org/en/latest/community/review.html#scientific-review

- [x] [🧪][2] The new functionality is [relevant and scientifically sound](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#scientific-relevance)
- [x] [🛠][1] This pull request has a [descriptive title and labels](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-title-and-label)
- [x] [🛠][1] Code is written according to the [code quality guidelines](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#code-quality)
- [x] [🧪][2] and [🛠][1] [Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available
- [x] [🛠][1] [Unit tests](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#tests) have been added
- [x] [🛠][1] Changes are [backward compatible](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#backward-compatibility)
- [x] [🛠][1] Any changed [dependencies have been added or removed](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#dependencies) correctly
- [x] [🛠][1] The [list of authors](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#list-of-authors) is up to date
- [ ] [🛠][1] All [checks below this pull request](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-checks) were successful

***

[...]

PR #1848, opened 2022-12-02
Dry run release v2.7.1 bugfix

valeriupredoi:
@ESMValGroup/technical-lead-development-team these are the planned changes for the bugfix release that we talked about earlier today: they include:

- unpinning `cf-units` and `netCDF4` in environment
- unpinning those two in setup.py and pinning mypy (since otherwise hell will break loose with the new standards)
- cherrypicking the test that @schlunma deleted in #1846 

Please have a look - this PR will be merged (note the target for merge: release v2.7.0 branch) only when we agree all is fine, and then after I change the version number.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -2,20 +2,19 @@
 name: esmvaltool
 channels:
   - conda-forge
+  - nodefaults
 
 dependencies:
   - cartopy
-  - cf-units>=3.0.0,<3.1.0  # github.com/ESMValGroup/ESMValCore/issues/1655
+  - cf-units
   - cftime
   - compilers
-  # 1.8.18/py39, they seem weary to build manylinux wheels
-  # and pypi ver built with older gdal
   - fiona
   - esmpy!=8.1.0  # see github.com/ESMValGroup/ESMValCore/issues/1208
   - geopy
   - iris>=3.2.1
   - nested-lookup
-  - netcdf4!=1.6.1  # github.com/ESMValGroup/ESMValCore/issues/1723
+  - netcdf4
   - pandas
   - pillow
   - pip!=21.3
--- a/setup.py
+++ b/setup.py
@@ -28,8 +28,7 @@ REQUIREMENTS = {
     # Use with pip install . to install from source
     'install': [
         'cartopy',
-        # see https://github.com/SciTools/cf-units/issues/218
-        'cf-units>=3.0.0,<3.1.0,!=3.0.1.post0',  # ESMValCore/issues/1655
+        'cf-units',
         'dask[array]',
         'esgf-pyclient>=0.3.1',
         'esmpy!=8.1.0',  # see github.com/ESMValGroup/ESMValCore/issues/1208
@@ -42,7 +41,7 @@ REQUIREMENTS = {
         'jinja2',
         'nc-time-axis',  # needed by iris.plot
         'nested-lookup',
-        'netCDF4!=1.6.1',  # github.com/ESMValGroup/ESMValCore/issues/1723
+        'netCDF4',
         'numpy',
         'pandas',
         'pillow',
@@ -59,7 +58,8 @@ REQUIREMENTS = {
     ],
     # Test dependencies
     'test': [
-        'flake8<5.0',  # github.com/ESMValGroup/ESMValCore/issues/1696
+        'flake8',
+        'mypy<0.990',
         'pytest>=3.9,!=6.0.0rc1,!=6.0.0',
         'pytest-cov>=2.10.1',
         'pytest-env',
--- a/tests/integration/preprocessor/_regrid/test_regrid.py
+++ b/tests/integration/preprocessor/_regrid/test_regrid.py
@@ -241,11 +241,6 @@ class Test(tests.Test):
         expected = np.array([[[1]], [[1]], [[1]]])
         np.testing.assert_array_equal(result.data, expected)
 
-        # Make sure that dtype is not preserved (since conversion from float to
-        # int would be necessary)
-        assert np.issubdtype(self.unstructured_grid_cube.dtype, np.integer)
-        assert result.dtype == np.float64
-
         # Make sure that output is a masked array with correct fill value
         # (= maximum int)
         np.testing.assert_allclose(result.data.fill_value,


schlunma:
Thanks for tackling this!! I think you'll need to cherry-pick https://github.com/ESMValGroup/ESMValCore/pull/1770 as well.

valeriupredoi:
> Thanks for tackling this!! I think you'll need to cherry-pick #1770 as well.

indeed! That's a bummer - not an easy cherrypick that one, let me get on with it see where I get :grimacing:

valeriupredoi:
first stop is see if the envs (all of them) pick up iris 3.4 - will be back when the GA's finish so I can inspect the envs

bouweandela:
It would be a lot cleaner if you could create the v2.7.1 release by only cherry-picking stuff that is already in `main` instead of adding your own commits. I suspect this is going to cause trouble again with the automatic versioning if you do it like this. Do you think that would be possible?

valeriupredoi:
@bouweandela - `main` is way ahead of the release branch, cherrypicking is a disaster that way - plus, there is stuff that there is in main that we really must not use (like some other pins, the env file containing all packages etc); what I am getting in here has already been merged in main, and yes, we may get some weirdness at commit table, but that's the least of our problems IMHO

valeriupredoi:
OK guys I reckon we good to go, can you have a look please and do me a review? Also, what do people think it's the best way to release this - merge it into the 2.7.0 branch and up the version number there to trigger a release or up the version number here, in this branch, with out a merge? :beer:

valeriupredoi:
@bouweandela we can not merge this safely into `main` due to the issue with mypy and all the baloney about standards; Core is already at a 2.7.1 development version `ESMValCore: 2.7.1.dev38+g5180b4f` so all we need to do is to bring the PyPI and conda-forge packages to a 2.7.1 so that @schlunma and others can bring in iris=3.4. For that we can just let loose a bugfix 2.7.1 off this branch, build the packages, and then have a floating bugfix release until we release stable 2.8. If you can provide me with a better solution, am gonne buy you a pint (of radler), and implement it :grin:

valeriupredoi:
@bouweandela so this is how it looks if we really want to merge this into `main` - it's a bomb :bomb: I honestly don't know how we'd get about this, as I see it we have two alternatives:

- we either create a 2.7.1 off the latest `main` but that means we have to test run all the recipes
- or, we keep the local changes before [3b0f1f6](https://github.com/ESMValGroup/ESMValCore/pull/1848/commits/3b0f1f67a5092f32ba80a30bf6d6608405a20632) and we create a bugfix release candidate off this branch, then we either don't update the version in `main` since that's already at 2.7.1dev by default, or we update the version manually but don't release anything

Opinions? :beer: 

Oh and BTW the test_multimodel in sample data is an absolute nightmare to prune, since it contains a lot of stuff that deals with updates by @schlunma in main - those are beautiful, but are causing the poopstorm wrt conflicts

schlunma:
With all the changes since 2.7.0 (especially in the multimodel module and its tests) I guess merging this branch back into main without changes is not possible (some tests from 2.7.0 fail in the current main, and some tests from the current main fail in the 2.7.0 branch).

Is there a way to hardcode the version number somehow? If we don't merge this back into main, will this also mess up the version numbers of future releases?

valeriupredoi:
> With all the changes since 2.7.0 (especially in the multimodel module and its tests) I guess merging this branch back into main without changes is not possible (some tests from 2.7.0 fail in the current main, and some tests from the current main fail in the 2.7.0 branch).


exactly - it's a bit of a no-go either way; we fought ourselves into a corner, and we won't get away without serious recipes testing if we want to keep the things off `main` 
 
> Is there a way to hardcode the version number somehow? If we don't merge this back into main, will this also mess up the version numbers of future releases?

No, it shouldn't mess it up; plus the current version of `main` is already at 2.7.1devXXX

I'd very much appreciate a bit of feedback from @bouweandela before I revert commits and prepare a release with the changes before [3b0f1f6](https://github.com/ESMValGroup/ESMValCore/pull/1848/commits/3b0f1f67a5092f32ba80a30bf6d6608405a20632) - also bit of feedback from you Manu after I revert the commits would be greatly appreciated :beer:

bouweandela:
We need something that can be merged back into `main` to get the versioning working properly. However, we can of course add commits to that to undo any unwanted changes after making the release but before merging it back into `main`. This will make a bit of a mess of our version history, but we if make an effort to keep the `git history` somewhat clean it should hopefully not be too bad.

The (least bad) way forward I see is the following:
- start by branching off the `v2.7.0` tag
- add 1 commit with a clear description per feature that will be added to `v2.7.1`, keeping in mind that we want minimal differences with `main`
- create the release
- add 1 commit to undo any changes we do not want in `main` (such as pinning the mypy version)
- merge the release branch back into `main`

valeriupredoi:
good call @bouweandela - I did that before [3b0f1f6](https://github.com/ESMValGroup/ESMValCore/pull/1848/commits/3b0f1f67a5092f32ba80a30bf6d6608405a20632) (even if my commits are a bit too many) - I will revert the commits later than that commit, keep this branch for reference and will work on a new branch (off the 27 release) and include the changes from here on that branch with more condensed and descriptive commits

bouweandela:
To make the release, we'll need to create a branch called v2.7.x and add the commits that we want in the bugfix release. Do you want to manually create meaningful commits? Or would you prefer to make pull requests so we can squash-merge them and get a clean git history like that?

valeriupredoi:
how about we rename `release_270stable` to `2.7.x` then squash merge this in it, then we up the version there to 2.7.1 and create a release off it?

bouweandela:
After some discussion offline, @valeriupredoi and I decided to make one pull request per added feature for v2.7.1 against the v2.7.x branch.

valeriupredoi:
indeed! and I will close this once *all* the PRs to 2.7.x have been initiated. :bear: with me now...

valeriupredoi:
OK piecemeal PR's (this broken down) have been created:

- #1855 
- #1856 
- #1857 
- #1858 

Please close this (or I close it) once all those PR's have been merged in v2.7.x :beer:

bouweandela:
Would it be possible to include #1838 as well? That will make sure everyone has an up-to-date default ESGF configuration again when they install the tool.

valeriupredoi:
> Would it be possible to include #1838 as well? That will make sure everyone has an up-to-date default ESGF configuration again when they install the tool.

I think we're good to include that too, but unfortunately am very hungry now, so I'll have dinner, will do that tomorrow, after you've plugged in all the other component PRs, man :beer:

valeriupredoi:
changed this to Draft so there is really no accidental push to v2.7.x (which I am also gonna protect via commit restrictions only for myself and @bouweandela ). v2.7.x looks to be just about ready for the bugfix release anyway, so we'll close this soon. Very many thanks for your review work @bouweandela :beer:

valeriupredoi:
we now have a bugfix release https://pypi.org/project/ESMValCore/2.7.1/ so I'm closing this :+1:

PR #1609, opened 2022-05-31
Support wildcards in the recipe and improve support for ancillary variables and dataset versioning

bouweandela:
## Description

Add a dataset class that simplifies dealing with datasets as defined in the recipe.

Goals of this pull request are to:

- Support wildcards in the dataset definitions in the recipe
- Add a new syntax to the recipe for specifying supplementary datasets for attaching ancillary variables and cell measures to the main variable
- Write out the versions of datasets to a 'filled' copy of the recipe, this will make it much easier to reproduce results from previous runs because this stops the data from changing under our feet, taking at least one changing factor out that makes testing from one version to the next difficult. The filled recipe is saved in the `run` directory, e.g. `run/recipe_example_filled.yml` when running a recipe called `recipe_example.yml`.
- Refactor esmvalcore/recipe/_recipe.py so it uses the `esmvalcore.dataset.Dataset` class added in #1877 

### Example recipe demonstrating the new wildcard features

```yaml
preprocessors:
  global_mean:
    area_statistics:
      operator: mean

datasets:
  - dataset: '*'
    institute: '*'
    project: CMIP6
    exp: historical
    ensemble: '*'
    grid: '*'

diagnostics:
  example_diagnostic:
    description: Global mean temperature.
    variables:
      tas:
        mip: Amon
        preprocessor: global_mean
        timerange: '1990/2000'
    scripts: null
```
this will expand to all CMIP6 datasets and attach the `areacella` variable that is most similar to the main variable.

### Documentation

- [Automatically populating a recipe with all available datasets](https://esmvaltool--1609.org.readthedocs.build/projects/ESMValCore/en/1609/recipe/overview.html#automatically-populating-a-recipe-with-all-available-datasets)
- [Specifying supplementary variables in the recipe](https://esmvaltool--1609.org.readthedocs.build/projects/ESMValCore/en/1609/recipe/overview.html#adding-ancillary-variables-and-cell-measures)
- [Preprocessor functions that use ancillary variables and cell measures](https://esmvaltool--1609.org.readthedocs.build/projects/ESMValCore/en/1609/recipe/preprocessor.html#ancillary-variables-and-cell-measures)

### Temporary `use_legacy_supplementaries` command-line/config-user option

[...]

PR #1877, opened 2022-12-19
Add `esmvalcore.dataset` module

bouweandela:
## New features

- Add the [`esmvalcore.dataset.Dataset`](https://esmvaltool--1877.org.readthedocs.build/projects/ESMValCore/en/1877/api/esmvalcore.dataset.html#esmvalcore.dataset.Dataset) class, which can be used to define datasets from Python. It can be used to find and load the data needed for using the [`esmvalcore.preprocessor`](https://esmvaltool--1877.org.readthedocs.build/projects/ESMValCore/en/1877/api/esmvalcore.preprocessor.html) functions.

- Add the function [`esmvalcore.dataset.datasets_to_recipe`](https://esmvaltool--1877.org.readthedocs.build/projects/ESMValCore/en/1877/api/esmvalcore.dataset.html#esmvalcore.dataset.datasets_to_recipe), which can be used to create or update recipes with `Dataset`s. This can be a convenient alternative to the [recipe filler utility](https://docs.esmvaltool.org/en/latest/utils.html#recipe-filler).

- Adds a new option to config-user.yml called `always_search_esgf`. This will cause the tool to always search ESGF for files, even if some files are available locally. This feature can be used when you want to 1) create a recipe with all data available on ESGF or 2) download the latest versions of files from ESGF when there is already an older version available.

- Move the files that do things related to loading a recipe (`esmvalcore/_recipe.py`, `esmvalcore/_recipe_checks.py`, and `esmvalcore/recipe_schema.yml`) into a new `esmvalcore/_recipe` directory. This makes it possible to add functionality without growing the file `esmvalcore/_recipe.py` even further. -> Should this be done in a separate pull request?

Example Jupyter notebooks that demonstrate how to use the new features are available in the [`notebooks`](https://github.com/ESMValGroup/ESMValCore/tree/add-dataset/notebooks) directory and also shown on readthedocs in a new section [Example notebooks](https://esmvaltool--1877.org.readthedocs.build/projects/ESMValCore/en/1877/example-notebooks.html).

This is part of #1609. The next (and hopefully final) step towards getting #1609 merged will be to use the new `Dataset` in `esmvalcore/_recipe.py` to read the datasets in the recipe. This pull request introduces some duplicate code because it already transfers some of the functionality from `esmvalcore/_recipe.py` to the `esmvalcore.dataset.Dataset`, but it cannot be removed yet because it is still used by the old dataset reading code.

Link to documentation: https://esmvaltool--1877.org.readthedocs.build/projects/ESMValCore/en/1877/api/esmvalcore.dataset.html

***

***


bouweandela:
The remaining Codacy issues fall into three categories:
- Prospector does not like the use of private methods because this could cause trouble if someone subclasses the `Dataset` class and implements a different version of these methods. I'm not sure how to fix this without making too much of the internals of the `Dataset` class public, so I would suggest we ignore this for now unless anyone has a better suggestion.
- Codacy doesn't seem to respect `from __future__ import annotations`, so there are some complaints about wrong types. Running `prospector` locally these problems are not shown.
- There was a bug in pydocstyle (https://github.com/PyCQA/pydocstyle/issues/531) so it complained about the docstrings of properties. It is fixed in the latest release, but Codacy is not using that yet.

doc/quickstart/configure.rst
```diff
@@ -62,6 +62,10 @@ omitted in the file.
   # on a computer without an internet connection.
   offline: true
 
+  # Search ESGF for files even when files are available locally. This option is
+  # useful to make sure you have the latest version of all files.
```
schlunma, line 66: It would be great if you could use the same syntax here as for other options with a fixed set of possible values, e.g., for `offline`:

```yaml
# Disable automatic downloads --- [true]/false
# Disable the automatic download of missing CMIP3, CMIP5, CMIP6, CORDEX,
# and obs4MIPs data from ESGF by default. This is useful if you are working
# on a computer without an internet connection.
```

Also, if I understand the code correctly this only works if `offline=False`. This may be obvious, but would be nice to add here:

So maybe something like

```suggestion
  # Search ESGF for files even when files are available locally --- true/[false]
  # This option is useful to make sure you have the latest version of all files.
  # Remember to set ``offline: false`` if this option is set to ``true``.
```

esmvalcore/config-user.yml
```diff
@@ -48,6 +48,10 @@ max_parallel_tasks: null
 # on a computer without an internet connection.
 offline: true
 
+# Search ESGF for files even when files are available locally. This option is
+# useful to make sure you have the latest version of all files.
```
schlunma, line 52: It would be great if you could use the same syntax here as for other options with a fixed set of possible values, e.g., for `offline`:

```yaml
# Disable automatic downloads --- [true]/false
# Disable the automatic download of missing CMIP3, CMIP5, CMIP6, CORDEX,
# and obs4MIPs data from ESGF by default. This is useful if you are working
# on a computer without an internet connection.
```

Also, if I understand the code correctly this only works if `offline=False`. This may be obvious, but would be nice to add here:

So maybe something like

```suggestion
# Search ESGF for files even when files are available locally --- true/[false]
# This option is useful to make sure you have the latest version of all files.
# Remember to set ``offline: false`` if this option is set to ``true``.
```

setup.cfg
```diff
@@ -16,13 +16,16 @@ markers =
...
 [flake8]
 exclude =
+    .eggs/
     doc/conf.py
 
 [coverage:run]
 parallel = true
 [coverage:report]
 exclude_lines =
+    pragma: no cover
```
schlunma, line 26: Could you briefly explain why we need this?

schlunma (commented):
Thanks for this addition Bouwe!! :rocket:  I had a look at the code that changes the current behavior of the tool. Looks really good, I only have very small comments.

I will now play around with the notebooks.

Note: I did NOT look at the code of the new features.

environment.yml
```diff
@@ -41,6 +41,8 @@ dependencies:
   - yamale
   # Python packages needed for building docs
   - autodocsumm>=0.2.2
+  - ipython
```
schlunma, line 44: This should probably also be added to `setup.py`?

schlunma: BTW, why do we need `ipython`? Wouldn't `ipykernel` be the correct package to run notebooks?

setup.cfg
bouweandela: This excludes this code
https://github.com/ESMValGroup/ESMValCore/blob/d91371ca97e786f088a978b55e907415079ad1cb/esmvalcore/_recipe/from_datasets.py#L15-L16
from coverage reporting. The import needs to be done like this to be able to use type hints while avoiding circular imports. You can read more about it [here](https://coverage.readthedocs.io/en/7.0.0/excluding.html#excluding-code-from-coverage-py).

environment.yml
bouweandela: It's for building the documentation. Without ipython the jupyter notebooks cannot be converted to .rst files.

bouweandela: For running the notebooks I would install jupyterlab, but that is not needed to just build the documentation.

schlunma:
I just tested the notebooks. This is really amazing stuff, Bouwe! Especially the usage of wildcards in combination with the `Dataset` -> recipe function is a game changer!! :rocket:

A couple of comments:

1. I think it would be nice to add a comment to each of the notebooks how to load a user's custom configuration file, e.g., `CFG.load_from_file('path/to/config.yml')`.
2. The link in the Discovering data notebook to `configurable duration` does not work.
3. Cell `[7]` in the Loading, processing, and visualizing data notebook does not download anything if you actually have ESGF data, it needs to be

```py
files = list(tas.files)
for anc in tas.ancillaries:
    files.extend(anc.files)
files = [file for file in files if isinstance(file, ESGFFile)]
download(files, CFG['download_dir'])
tas.find_files()
print(tas.files)
for anc in tas.ancillaries:
    print(anc.files)
```

4. Adding ancillaries after calling `augment_facets()` needs to wrong facets for the ancillaries, e.g.,

```py
tas = Dataset(
    short_name='tas',
    mip='Amon',
    project='CMIP5',
    dataset='GFDL-ESM2G',
    ensemble='r1i1p1',
    exp='historical',
    timerange='1990/2100',
)
tas.augment_facets()
tas.add_ancillary(short_name='areacella', mip='fx', ensemble='r0i0p0')
print(tas.ancillaries)
```

gives

```bash
[Dataset:
 {'dataset': 'GFDL-ESM2G',
  'project': 'CMIP5',
  'mip': 'fx',
  'short_name': 'areacella',
  'ensemble': 'r0i0p0',
  'exp': 'historical',
  'frequency': 'mon',
  'institute': ['NOAA-GFDL'],
  'long_name': 'Near-Surface Air Temperature',
  'modeling_realm': ['atmos'],
  'original_short_name': 'tas',
  'product': ['output1', 'output2'],
  'standard_name': 'air_temperature',
  'timerange': '1990/2100',
  'units': 'K'}
 session: 'session-d9f21999-2ea5-43b7-932d-29a9045fde5a_20230110_130835']
```

Notice the wrong `long_name` and `units`. This leads to problems when trying to download the data (i.e., it does not work).

5. I could not run all cells of the Loading, processing, and visualizing data notebook due to a concatenation error. For

```py
tas = Dataset(
    short_name='tas',
    mip='Amon',
    project='CMIP5',
    dataset='CanESM2',
    ensemble='r1i1p1',
    exp=['historical', 'rcp85'],
    timerange='1850/2100',
)
```

the tool found fx files for the historical and rcp85 simulation. This leads to a concatenation error when calling `cube = tas.load()` since the tool tries to concatenate two fx files in time (which it shouldn't do):

```bash
ERROR:esmvalcore.preprocessor:Failed to run preprocessor function 'concatenate' on the data
```

6. I noticed that `find_files` ignores time ranges. I.e., if a time range with no data is selected, a `load()` gives an `IndexError: list index out of range`; if some years of the time range are not available the missing years are simply ignored (this should at least raise a warning in my opinion).

bouweandela:
> 2. The link in the Discovering data notebook to configurable duration does not work.

It should be fixed now

bouweandela:
> Cell [7] in the Loading, processing, and visualizing data notebook does not download anything if you actually have ESGF data, ..

Thanks, I changed it

> I could not run all cells of the Loading, processing, and visualizing data notebook due to a concatenation error.

Good find! I changed the experiment to just `historical` to keep the notebook simple.

bouweandela:
> 4. Adding ancillaries after calling augment_facets() needs to wrong facets for the ancillaries

Indeed, it does not work if you do it in that order. I don't think it can be made to work reliably, so I added some documentation to the docstring and notebook to explain more clearly how the ancillaries are added. Do you think it is better now?

bouweandela:
> 6. I noticed that find_files ignores time ranges. I.e., if a time range with no data is selected, a load() gives an IndexError: list index out of range; if some years of the time range are not available the missing years are simply ignored (this should at least raise a warning in my opinion).

I added a check if there are any input files so at least there is no weird `IndexError`. I tried to use `esmvalcore._recipe.check.data_availability`, but the error message came out not looking very applicable because it refers to a log file that isn't so easily accessible from the API. I think that gaps in the resulting time coordinate should be found by the CMOR check anyway.

schlunma:
Thanks Bouwe for the changes, all the notebooks run fine now!

5. Thanks for fixing the notebook, but shouldn't this also fixed in the code? I think this is clearly a bug. In the tool this is solved by [using just the first fx file](https://github.com/ESMValGroup/ESMValCore/blob/155335a4b16ea911f546f9d6b4909bd427ae0af2/esmvalcore/_recipe.py#L447-L450) if the variable has `frequency=fx`. This might not be the perfect solution though..
6. Perfect, for missing files this raises a nice error now. You are probably right about data with gaps, this will probably raise an error in the frequency check. However, I was thinking about missing data at the start/end of the interval, e.g., using `timerange: 2000/3000` for the historical experiment. In this case the tool finds data for the years 2000-2005 and happily returns just theses five years without even issuing a warning. I find this highly problematic, e.g., if only a very small of years are missing or a multi-model mean is calculated and not all input datasets span the same time period.

I am running some recipes, all looks fine so far :+1:

bouweandela:
> 5. Thanks for fixing the notebook, but shouldn't this also be fixed in the code? I think this is clearly a bug. In the tool this is solved by [using just the first fx file](https://github.com/ESMValGroup/ESMValCore/blob/155335a4b16ea911f546f9d6b4909bd427ae0af2/esmvalcore/_recipe.py#L447-L450) if the variable has frequency=fx. This might not be the perfect solution though..

I think this should be left to the user of the API. For recipes, there isn't much flexibility for programming things, but for the API there is. You could manually choose which dataset you want to use as an ancillary dataset or if you have many you could compare the facets and choose the best match automatically. E.g. for use from the recipe I implemented https://github.com/ESMValGroup/ESMValCore/blob/8cdd9cc82ecc9f91fa8845706d8cfcd26560530c/esmvalcore/_recipe/datasets.py#L255-L289.

bouweandela:
> 6. Perfect, for missing files this raises a nice error now. You are probably right about data with gaps, this will raise an error in the frequency check. However, I was thinking about missing data at the start/end of the interval, e.g., using timerange: 2000/3000 for the historical experiment. In this case the tool finds data for the years 2000-2005 and happily returns just these five years without even issuing a warning. I find this highly problematic, e.g., if only a very small number of years is missing or a multi-model mean is calculated and not all input datasets span the same time period.

Would it be OK with you to postpone implementing such a feature to another pull request? I can see that it would be useful, but it's not trivial to implement. I could compare the bounds of the time coordinate (if there are any and fall back to points otherwise and do something clever to compensate for the fact that time points can be anywhere in the interval) with the requested period in the `timerange` facet, but I'm not sure how well this will work in practice.

schlunma:
> I think this should be left to the user of the API. For recipes, there isn't much flexibility for programming things, but for the API there is. You could manually choose which dataset you want to use as an ancillary dataset or if you have many you could compare the facets and choose the best match automatically. E.g. for use from the recipe in #1609 I implemented

Not too happy about this to be honest, but since this is such an edge case and it will be handled when used within ESMValTool I guess that's fine.

> Would it be OK with you to postpone implementing such a feature to another pull request? I can see that it would be useful, but it's not trivial to implement. I could compare the bounds of the time coordinate (if there are any and fall back to points otherwise and do something clever to compensate for the fact that time points can be anywhere in the interval) with the requested period in the timerange facet, but I'm not sure how well this will work in practice.

Couldn't we use the same code we use in the tool for this? If the time range of the files doesn't match the requested time range an error like the following is raised:

```bash
2023-01-13 08:59:17,614 UTC [3178263] ERROR   Missing data for preprocessor test/tas:
- No input data available for years 2101-2150 in files:
/work/bd0854/DATA/ESMValTool2/CMIP5_DKRZ/NOAA-GFDL/GFDL-ESM2G/rcp85/mon/atmos/Amon/r1i1p1/v20120412/tas/tas_Amon_GFDL-ESM2G_rcp85_r1i1p1_208601-209012.nc                                                                                                                    
/work/bd0854/DATA/ESMValTool2/CMIP5_DKRZ/NOAA-GFDL/GFDL-ESM2G/rcp85/mon/atmos/Amon/r1i1p1/v20120412/tas/tas_Amon_GFDL-ESM2G_rcp85_r1i1p1_209101-209512.nc                                                                                                                    
/work/bd0854/DATA/ESMValTool2/CMIP5_DKRZ/NOAA-GFDL/GFDL-ESM2G/rcp85/mon/atmos/Amon/r1i1p1/v20120412/tas/tas_Amon_GFDL-ESM2G_rcp85_r1i1p1_209601-210012.nc
```

Again, not too happy about this, but if it's too much work and you make sure that this is handled when this is used within the tool I am also fine with it.

All the recipes I tested ran fine with this change, so I think this is good to go :+1: Thanks again for this great feature, @bouweandela! :tada:

bouweandela:
Awesome, thanks a lot for reviewing @schlunma!

> Couldn't we use the same code we use in the tool for this?

There are two main issues with the current code used for checking when running the recipe:
1) It only works if there is a recognizable time range in the filename
2) It only checks for entire years (not months, days, or hours)
so I feel that adding this will give users a false sense of security because it only works sometimes.

> make sure that this is handled when this is used within the tool I am also fine with it.

Thanks, I will make sure the behavior of the tool when running a recipe does not change in #1609 and I created an issue to describe the feature here #1892.

schlunma:
> There are two main issues with the current code used for checking when running the recipe:
> 
>     1. It only works if there is a recognizable time range in the filename
> 
>     2. It only checks for entire years (not months, days, or hours)
> 
> 
> so I feel that adding this will give users a false sense of security because it only works sometimes.

But since we use the [`clip_timerange` preprocessor ](https://github.com/ESMValGroup/ESMValCore/blob/02c4e8e9966c8d8d35315643328b8c144ebbb8c6/esmvalcore/dataset.py#L523-L526) while loading the files I guess we should be fine? This selection would only be a course first step, the actual selection is done in `clip_timerange`.

esmvalcore/_recipe/from_datasets.py
```diff
@@ -0,0 +1,327 @@
...
+def datasets_to_recipe(
+    datasets: Iterable[Dataset],
+    recipe: dict | None = None,
+) -> dict:
+    """Create or update a recipe from datasets.
+
+    Parameters
+    ----------
+    datasets
+        Datasets to use in the recipe.
```
valeriupredoi, line 278: give one example pls

esmvalcore/_recipe/from_datasets.py
```diff
@@ -0,0 +1,327 @@
...
+    recipe
```
valeriupredoi, line 279: mention data type (`dict`)

esmvalcore/_recipe/recipe.py
```diff
@@ -593,23 +567,28 @@ def _get_input_files(variable, config_user):
...
             # Only look on ESGF if files are not available locally.
+            try:
+                check.data_availability(
+                    input_files,
+                    variable,
+                    globs,
+                    log=False,
+                )
+            except RecipeError:
+                search_esgf = True
```
valeriupredoi, line 581: this is too much under the hood, I think a debug message would be appropriate here, what say you?

esmvalcore/cmor/table.py
```diff
@@ -29,6 +29,41 @@
...
+    project = facets['project']
+    mip = facets['mip']
+    short_name = facets['short_name']
+    derive = facets.get('derive', False)
+    table = CMOR_TABLES.get(project)
+    if table:
+        table_entry = table.get_variable(mip, short_name, derive)
+    else:
+        table_entry = None
+    if table_entry is None:
```
valeriupredoi, line 52:
```suggestion
```

setup.cfg
valeriupredoi: thanks, was about to ask the same :laughing:

tests/integration/recipe/test_check.py
```diff
@@ -1,13 +1,14 @@
-"""Integration tests for :mod:`esmvalcore._recipe_checks`."""
```
valeriupredoi, line 1: I'd keep the original test module name `test_recipe_checks`

valeriupredoi (approved):
I went through the code too, and I think this is super solid! But I have a few minor suggestions; am approving this since it's up to @bouweandela to implement my suggestions, nothing make or break. Note that I've not looked at the iPython notebooks :beer:

esmvalcore/_recipe/from_datasets.py
```diff
@@ -0,0 +1,327 @@
...
+
+    recipe = {'diagnostics': diagnostics}
+    return recipe
+
+
+def _datasets_to_recipe(datasets: Iterable[Dataset]) -> Recipe:
+    """Convert datasets to a condensed recipe dict."""
+    for dataset in datasets:
+        if 'diagnostic' not in dataset.facets:
+            raise RecipeError(f"'diagnostic' facet missing from {dataset},"
```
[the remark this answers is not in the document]
bouweandela: The dataset string description starts with the word `Dataset`, that's why I did not add it here because then the text will read `'diagnostic' facet missing from dataset Dataset: ..,`

esmvalcore/_recipe/from_datasets.py
```diff
@@ -0,0 +1,327 @@
...
+    # Remove common datasets from groups
+    for name, datasets in dataset_mapping.items():
+        group = groups[name]
+        var_datasets = set(datasets) - common_datasets
+        if var_datasets:
+            group['additional_datasets'] = [
+                v for k, v in datasets.items() if k in var_datasets
+            ]
+        else:
+            group.pop('additional_datasets')
```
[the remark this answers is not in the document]
bouweandela: That wouldn't work: the point here is to remove an empty `additional_datasets` list from the recipe.

esmvalcore/_recipe/from_datasets.py
```diff
@@ -0,0 +1,327 @@
...
+        else:
+            for ensemble in _group_ensemble_names(ensembles):
+                facets = dict(group_facets)
+                facets['ensemble'] = ensemble
+                result.append(facets)
+    return result
+
+
+def _group_ensemble_names(ensemble_names: Iterable[str]) -> list[str]:
+    """Group ensemble names.
```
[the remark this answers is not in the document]
bouweandela: It could be useful standalone, but you can already use it as part of the `datasets_to_recipe` function so maybe that is enough. I am not sure where it would fit in the API at the moment. Once we have a public `esmvalcore.recipe` module (possibly starting out from `esmvalcore.expirimental.recipe` it might fit in there..

esmvalcore/_recipe/from_datasets.py
bouweandela: The data types are shown in the function signature based on the type hints: https://esmvaltool--1877.org.readthedocs.build/projects/ESMValCore/en/1877/api/esmvalcore.dataset.html#esmvalcore.dataset.datasets_to_recipe

It seems duplicate to then repeat them in the docstrings.

esmvalcore/_recipe/from_datasets.py
bouweandela: Examples are available [here](https://esmvaltool--1877.org.readthedocs.build/projects/ESMValCore/en/1877/notebooks/composing-recipes.html), I'll add a link

esmvalcore/cmor/table.py
bouweandela: I would be hesitant to remove that because it is probably in the current code for a reason:
https://github.com/ESMValGroup/ESMValCore/blob/155335a4b16ea911f546f9d6b4909bd427ae0af2/esmvalcore/_recipe.py#L94-L97

tests/integration/recipe/test_check.py
bouweandela: That would be confusing because module is no longer called that, it has been moved to [esmvalcore/_recipe/check.py](https://github.com/ESMValGroup/ESMValCore/blob/add-dataset/esmvalcore/_recipe/check.py)

esmvalcore/_recipe/from_datasets.py
bouweandela: The updated documentation, with the link to the examples, is now available [here](https://esmvaltool--1877.org.readthedocs.build/projects/ESMValCore/en/1877/api/esmvalcore.dataset.html#esmvalcore.dataset.datasets_to_recipe).

esmvalcore/_recipe/from_datasets.py
bouweandela: The updated documentation, with an improved description of what the `recipe` argument and the return value are, is now available [here](https://esmvaltool--1877.org.readthedocs.build/projects/ESMValCore/en/1877/api/esmvalcore.dataset.html#esmvalcore.dataset.datasets_to_recipe).

esmvalcore/_recipe/from_datasets.py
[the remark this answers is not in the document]
valeriupredoi: ah gotcha!

esmvalcore/_recipe/from_datasets.py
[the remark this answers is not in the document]
valeriupredoi: mm OK that's fine, but you can redefine `recipe` - nay bother, not gonna shed tears over this, man

esmvalcore/_recipe/from_datasets.py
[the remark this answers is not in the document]
valeriupredoi: OK fine by me :+1:

esmvalcore/_recipe/from_datasets.py
valeriupredoi: sweet, cheers, bud

esmvalcore/_recipe/from_datasets.py
valeriupredoi: perfect

esmvalcore/cmor/table.py
valeriupredoi: ha OK I was looking about too, couldn't find any breakage point but heyho, not gonna improve complexity if we remove one single line, but that might break stuff downstream :laughing:

tests/integration/recipe/test_check.py
valeriupredoi: OK up to you, bud

esmvalcore/_recipe/recipe.py
valeriupredoi: aha! Happy to see one, then we skip this

PR #1218, opened 2021-07-06
Draft intake use

bouweandela:
## Description

Closes #31 

Link to documentation:

***

## [Before you get started](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#getting-started)

- [ ] [☝ Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do

## [Checklist](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#checklist-for-pull-requests)

It is the responsibility of the author to make sure the pull request is ready to review. The icons indicate whether the item will be subject to the [🛠 Technical][1] or [🧪 Scientific][2] review.

[1]: https://docs.esmvaltool.org/en/latest/community/review.html#technical-review
[2]: https://docs.esmvaltool.org/en/latest/community/review.html#scientific-review

- [ ] [🧪][2] The new functionality is [relevant and scientifically sound](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#scientific-relevance)
- [ ] [🛠][1] This pull request has a [descriptive title and labels](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-title-and-label)
- [ ] [🛠][1] Code is written according to the [code quality guidelines](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#code-quality)
- [ ] [🧪][2] and [🛠][1] [Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available
- [ ] [🛠][1] [Unit tests](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#tests) have been added
- [ ] [🛠][1] Changes are [backward compatible](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#backward-compatibility)
- [ ] [🛠][1] Any changed [dependencies have been added or removed](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#dependencies) correctly
- [ ] [🛠][1] The [list of authors](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#list-of-authors) is up to date
- [ ] [🛠][1] All [checks below this pull request](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-checks) were successful

***

To help with the number pull requests:

[...]

Issue #1165, opened 2021-06-09
Simplify DRS and rootpath in user configuration by using defaults or lists


bsolino:
**Is your feature request related to a problem? Please describe.**
This is not a problem yet, but it's something we are anticipating it may become. Currently we are working on reading datasets natively, and the original idea was to group them on one project, `native6`. However, in practice many of these datasets have their own folder structure or name convention, and it seems it makes more sense to implement as separate projects (see #494 for discussion on this topic). We anticipate that this policiy could increase the number of datasets enough that it clutters the configuration files. Furthermore, it requires users to keep track of the datasets and update their configuration file accordingly.

However, in practice many users share a configuration, as they work in a few HPC's. We are already anticipating this and handling it by having different DRS's for these machines in the `config-developer.yml` file, and by providing by default configurations that users can uncomment in `config-user.yml`.

This process could be simplified if users had only to define once in which machine they are working, so that DRS would be used by default. Any exception that a user would need could be specified separatedly.

E.g, from:

```
# Site-specific entries: Jasmin
# Uncomment the lines below to locate data on JASMIN
drs:
  CMIP6: BADC
  CMIP5: BADC
  CMIP3: BADC
  CORDEX: BADC
  OBS: BADC
  OBS6: BADC
  obs4mips: BADC
  ana4mips: BADC
```

to

```
drs:
# Site-specific entries:
# Uncomment the lines below to locate data on JASMIN
  default: BADC
```

Exceptions could be handled by using the current format, which also would make it backwards compatible

```
drs:
  default: DKRZ
  CORDEX: BADC
```

A similar approach could be used with rootpaths, although it would require defining the routes somewhere else.

An alternative to the "default" field would be to use lists. For example, DKRZ currently has one main roothpath, two main DRS and a few exceptions:

```
rootpath:
  CMIP6: /mnt/lustre02/work/ik1017/CMIP6/data/CMIP6
  CMIP5: /mnt/lustre02/work/bd0854/DATA/ESMValTool2/CMIP5_DKRZ
  CMIP3: /mnt/lustre02/work/bd0854/DATA/ESMValTool2/CMIP3
  CORDEX: /mnt/lustre02/work/ik1017/C3SCORDEX/data/c3s-cordex/output
  OBS: /mnt/lustre02/work/bd0854/DATA/ESMValTool2/OBS
  OBS6: /mnt/lustre02/work/bd0854/DATA/ESMValTool2/OBS
[...]

Issue #795, opened 2020-09-24
Rethinking the configuration file


stefsmeets:
Hi everyone,

Over the last week I have been working on implementing an importable config item (#785). As part of this PR, I have been discussing with @bouweandela and @peter9192 if this would be a good time to rethink how the config file is structured. I would like to start a discussion on how to change the configuration.

There are several issues on the topc. I think with #785 we have made some strides towards tackling nearly all of them. In the next post I will detail we propose to set it up.

#794: Example config file could use some improvements
#793: "Default" is not a great name for a DRS in the configuration
#494: Support per dataset DRS
#129: Specify directory stucture per path instead of per project in config-user.yml
#100: Config-user.yml rootpath.
#93: Too many unnessary settings in config-user.yml
#64: Make user configuration file more user friendly

In particular, I also want to draw your attention #93, to remove some settings from the default user config file. Which settings are really necessary, and which should be deprecated? Are there new keywords that should be added?

Suggestions from @bouweandela to be removed (#93):
`write_plots: true`
`write_netcdf: true`
`output_file_type: png` # [ps]/pdf/png/eps/epsi
`profile_diagnostic: false` -> only interesting for Python diagnostic developers
`save_intermediary_cubes`: false -> for debugging purposes, probably only interesting for diagnostic developers
`exit_on_warning: false` -> specific to NCL diagnostics, probably only interesting for diagnostic developers
`max_data_filesize: 100` -> specific to NCL diagnostics, probably only interesting for diagnostic developers

Also:
`config_developer_file: null` -> No longer needed with the PR, could be tweaked to specify the default data reference syntax

stefsmeets:
We have been working with the following scenarios:

1. As a user, the default settings should work out-of-the-box
2. As a user, I want to specify a directory with some data
3. As an advanced user, I want to tweak the directory structure to my liking
4. As a cluster administrator, I want to define custom data reference syntax
5. As a cluster user, I want to easily quickly initialize configs for different machines
6. As a cluster user, I want to add an additional data path with some extra data I downloaded
7. As a developer, the rootpath/drs/cfg-developer structure is a bit of a mess and confusing to work with

So in the PR I have done away with the `cfg-developer.yml`, and split it up into different DRS specifications per machine (DRKZ, Jasmin, etc.)

In the console you can list them:

```console
~$ esmvaltool config list
# Available site-specific configs
# Use `esmvaltool config get [config_name]` to copy them
- badc
- bsc
- cp4cds
[...]

PR #1894, opened 2023-01-13
Make DRS path templates configurable per rootpath

### LINKED ISSUE #129: Specify directory stucture per path instead of per project in config-user.yml
At the moment all drs need to be the same per project, while it would make much more sense to be able to specify the drs per rootpath.

To make this more clear, now we have this:
```yaml
rootpath:
  # Rootpath to CMIP5 data
  CMIP5:
    - /badc/cmip5/data/cmip5/output1
    - ~/some_other_path

# Directory structure for input data: [default]/BADC/DKRZ/ETHZ/etc
# See config-developer.yml for definitions.
drs:
  CMIP5: BADC
```
while I think this would be much more useful:
```yaml
# Directory structure for input data in the format
# input_data:
#   project_a:
#     /path/to/dir_1: drs_1
#     /path/to/dir_2: drs_2
#   project_b:
#     /path/to/dir_3: drs_2
# See config-developer.yml for valid drs definitions per project.
input_data:
  CMIP5:
    /badc/cmip5/data/cmip5/output1: BADC
    ~/some_other_path: default
```

bouweandela:
## Description

Add the possibility to specify the names of the directory and filename templates in config-user.yml per path instead of per project.

Example

```yaml
download_dir: /work/bd0854/DATA/ESMValTool2/download
rootpath:
  CMIP6:
    /work/bd0854/DATA/ESMValTool2/CMIP6_DKRZ: DKRZ
    /work/bd0854/DATA/ESMValTool2/download: ESGF
```

Closes #129

Link to documentation: https://esmvaltool--1894.org.readthedocs.build/projects/ESMValCore/en/1894/quickstart/find_data.html#explaining-config-user-rootpath

***

***


doc/quickstart/find_data.rst
```diff
@@ -552,27 +546,37 @@ Explaining ``config-user/rootpath:``
...
   .. code-block:: yaml
 
-    CMIP6: [/badc/cmip6/data/CMIP6/CMIP, /home/users/johndoe/cmip_data]
+    rootpath:
+      CMIP6:
+        /badc/cmip6/data/CMIP6: BADC
+        ~/climate_data: ESGF
+
+  Note that this notation combines the ``rootpath`` and ``drs`` settings, so it
+  is not necessary to specify the directory structure in under ``drs`` in this
```
valeriupredoi, line 578:
```suggestion
  is not necessary to specify the directory structure under ``drs`` in this
```

valeriupredoi: what happens if they do, and either that (`drs: ...`) or this one is wrong, or, if they differ?

valeriupredoi (changes requested):
I'd like to say awesome, this is very cool! and it is, but I am starting to quiver at every addition to the configuration. This one in particular is cumbersome, and as awesome its functionality, as bad as the way it's done in the configuration; here "bad" in the sense of user friendliness, nothing to do with how it's implemented. A few points:

- I would add a check if the user actually inputs `drs: ...` and that differs from the one in ` /path/: drs`
- I know you are mapping path to drs but the name ESGF to which the path is mapped to is confusing especially since the examples on DKRZ are not related to ESGF but, rather, to some bundled data in one rando directory
- overall, this is testing the limit of how more complicated can we make things for users, I am not at all at ease with this TBF, and I know I'm not helping by moaning but - there's gotta be an easier route. Just imagine you are a first time user of ESMValTool - how would you feel seeing such a configuration file? And not you, as a talented software engineer, but you as a PhD student just starting studying climate modelling

bouweandela:
The alternative would be to completely redesign how we configure finding data. There are some good ideas out there already, see e.g. https://github.com/ESMValGroup/ESMValCore/issues/795#issuecomment-698219457, the issues linked therein, and https://github.com/ESMValGroup/ESMValCore/issues/1165. However, none of these options would be backward compatible.

doc/quickstart/find_data.rst
bouweandela: It is not used.

valeriupredoi: well OK, but one needs to toss a warning at least, say eg "You specified drs: DRS but the other one is used... so that one gets ignored etc"

valeriupredoi:
> The alternative would be to completely redesign how we configure finding data. There are some good ideas out there already, see e.g. [#795 (comment)](https://github.com/ESMValGroup/ESMValCore/issues/795#issuecomment-698219457), the issues linked therein, and #1165. However, none of these options would be backward compatible. That's why I opted for keeping the existing functionality and extending it with yet another option.

yes! Redesign is the key here - at the end of the day, the user will still get their data, but going through much less a pain how to configure their environment. If we go on like this, even us, the devs, will start getting lost in configuration - I, for one, am starting to already, and only have about 6 years experience with the Tool :grin:

esmvalcore/config-user.yml
```diff
@@ -173,22 +183,27 @@ drs:
 # Uncomment the lines below to locate data on Levante at DKRZ.
 #auxiliary_data_dir: /work/bd0854/DATA/ESMValTool2/AUX
 #rootpath:
-#  CMIP6: /work/bd0854/DATA/ESMValTool2/CMIP6_DKRZ
-#  CMIP5: /work/bd0854/DATA/ESMValTool2/CMIP5_DKRZ
-#  CMIP3: /work/bd0854/DATA/ESMValTool2/CMIP3
-#  CORDEX: /work/ik1017/C3SCORDEX/data/c3s-cordex/output
+#  CMIP6:
+#    /work/bd0854/DATA/ESMValTool2/CMIP6_DKRZ: DKRZ
+#    /work/bd0854/DATA/ESMValTool2/download: ESGF
```
valeriupredoi, line 188: OK let's try work with this - leaving aside my rather negative comments above for now, can we at least make this a reverse key-val instance, so maybe it's a bit more user-firndly? eg:
```yaml
  CMIP6:
    - DKRZ: /work/bd0854/DATA/ESMValTool2/CMIP6_DKRZ
    - ESGF: /work/bd0854/DATA/ESMValTool2/download
```

valeriupredoi (changes requested):

remi-kazeroni:
Hi @bouweandela, sorry for being late in reviewing this. I will take a look once the comment's from @valeriupredoi are addressed, would that be ok? Please let us know what is needed to bring this PR to the finish line 🍺

esmvalcore/config-user.yml
bouweandela: Why is it more user-friendly to reverse the key and values?

bouweandela:
> Redesign is the key here

How about adding a new item to config-user.yml that can be used to directly construct [`esmvalcore.local.DataSource`](https://esmvaltool--1894.org.readthedocs.build/projects/ESMValCore/en/1894/api/esmvalcore.local.html#esmvalcore.local.DataSource) objects from? We could then easily make that extensible by allowing the user to provide custom DataSource classes, so they could instead use e.g. intake-esm #1218 or whatever they want.
 
```yml
data_sources:
  - type: esmvalcore.local.DataSource
    project: CMIP6
    rootpath: /work/bd0854/DATA/ESMValTool2/CMIP6_DKRZ
    dirname_template: '{activity}/{institute}/{dataset}/{exp}/{ensemble}/{mip}/{short_name}/{grid}/{version}'
    filename_template: '{short_name}_{mip}_{dataset}_{exp}_{ensemble}_{grid}*.nc'
```
the `type` key would only be needed if it's something else than `esmvalcore.local.DataSource`. 

This has some advantages:
- It does not hide where the path templates are defined, making it easier to understand
- No nested dicts with many possible entries, just a single list of DataSource objects
- Extensible if we provide an abstract base class for DataSource. It should have a `find_files` method and the kind of things that it can return, current examples: `esmvalcore.local.LocalFile` and `esmvalcore.esgf.ESGFFile`

There's also a few disadvantages:
- It may look a bit more intimidating at first glance
- For user-friendliness, it we would still to finally do something with the proposal of defining sites (e.g. Jasmin, Levante, etc) and having some simple way of getting a default configuration for those
- No solution for where the CMOR tables or output filenames are defined, those would still be in config-developer.yml
- I'm not sure if I have time to implement this, having this in v2.8 seems unlikely

schlunma:
I personally like this new feature very much (the way it is implemented right now)! Specifying DRS per rootpath will make my daily work much easier (I don't know how much time I wasted in the past with creating nested directories when I wanted to use data from two different sources). So I am definitely in favor of having this in v2.8.

About the key-value order: since rootpaths are unique and DRS are not (each rootpath has a single DRS, but each DRS can be used for multiple rootpaths) the order that @bouweandela proposed (e.g., `/path/to/cmip6: DKRZ`) makes much more sense to me.

About the customized data sources: I also like this idea very much! Being able to define custom file and dir templates quickly in the config-user file without messing with a custom config-developer file is great! However, I agree that this might be more intimidating  for new users. Maybe we could add some defaults in the config-developer file and the user has the possibility to add custom entries in the config-user file? This could mean that the config-developer and config-user files contain identical options with the config-developer as machine/default configuration, and the config-user as the user-defined configuration. The current distinction of these two files feels rather arbitrary to me. I think this has been proposed in the past, but I cannot find the corresponding issue. Maybe something for a v3.0?

TLDR; I am in favor of merging this as is, and think about allowing custom data sources in a future version.

valeriupredoi:
So here's the thing - just like Eisenhower once said, I like all this, as long as it's not in the config-user file 😁 Are you guys not realizing that we're slowly beginning to bena tool with such a complicated configuration that one needs less time to learn to fly a Boeing 737 than to learn to correctly configure ESMValTool? I'll bark more on the call today :beer:

schlunma:
To be honest, I think the current version of specifying DRS+rootpath in two different places (`drs` and `rootpath`) is more complicated than what Bouwe proposes here, which is "for rootpath XYZ, use DRS XZY".

valeriupredoi:
> To be honest, I think the current version of specifying DRS+rootpath in two different places (`drs` and `rootpath`) is more complicated than what Bouwe proposes here, which is "for rootpath XYZ, use DRS XZY".

this is not decommissioning the other config option, it adds to it it as an extra option. At any rate, let's talk on the call, we can add here and in #1609 an executive summary of our discussion on the call, later on

bouweandela:
The @ESMValGroup/technical-lead-development-team decided to postpone this feature to the next release.

remi-kazeroni:
Thanks @bouweandela for all your work on this already and agreeing to shift this to the next release. Let's make sure that we have a release planning meeting soon after v2.8 is released to discuss this PR. Or even a specific meeting to discuss how we could revamp the configuration of ESMValTool in order to increase the user-friendliness of the Tool

schlunma:
I'd really like to revive this discussion here. The changes proposed in this PR, i.e., being able to do something like this

```yaml
download_dir: /work/bd0854/DATA/ESMValTool2/download
rootpath:
  CMIP6:
    /work/bd0854/DATA/ESMValTool2/CMIP6_DKRZ: DKRZ
    /work/bd0854/DATA/ESMValTool2/download: ESGF
```

would be really helpful, especially when dealing with native model output. I really hope we can include this in v2.9.

bouweandela:
At the workshop it was agreed that a new configuration file format is acceptable and I'll write a proposal. Dropping this from the v2.10 milestone because it seems unlikely that we can agree on the new format before that time.

schlunma:
> At the workshop it was agreed that a new configuration file format is acceptable and I'll write a proposal. Dropping this from the v2.10 milestone because it seems unlikely that we can agree on the new format before that time.

It's been 8 months since the workshop, and there was no progress regarding the new configuration. Is there any way we can merge this feature as it is proposed in this PR? I know that it would be much cleaner to refactor our configuration, but this just doesn't seem realistic at this point.

On the other hand, this issue has been open since 2018 (!), and many people would benefit a lot from this.

bouweandela:
Sorry for being slow with the writeup for the new configuration, I'll try to get it done next week. About merging this as-is, maybe that would be something to discuss at the next @ESMValGroup/technical-lead-development-team meeting?

schlunma:
Good idea, I just added it to the agenda.

valeriupredoi (approved):
Judas has re-converted himself :grin:

valeriupredoi:
don't blame me for holding this up for too long and now we have a bazillion conflicts :grin:

schlunma:
@bouweandela I would really like to have this in v2.12. Would it be fine for you if I solve the merge conflicts and finalize this PR?

bouweandela:
Yes, that would be great!

axel-lauer (approved):
I tried this for a number of different cases and it works fine but only as long as you do not remove the DRS entry completely. If you do so while not explicitly specifying a DRS format (e.g. when reusing lines from an old config file) the tool shows some unexpected behavior. For example obs4mips is then expected to follow the ESGF DRS instead of 'default' as it was the case for old configuration files.

Maybe this behavior could be changed so the DRS section can be removed completely?

schlunma:
Thanks for testing @axel-lauer! 🚀 

The reason for this behavior is the [default entry for `drs`](https://github.com/ESMValGroup/ESMValCore/blob/405300886cfb2314eadc6ff2f9b137371ada0dfe/esmvalcore/config-user.yml#L105-L110), which says `ESGF` for the obs4mips data. If you don't specify `drs` in your own configuration file, this default is used. If you specify `drs` in your configuration file but omit obs4mips, then `default` instead of `ESGF` is used.

I agree that this behavior is confusing, but this was not introduced with this PR. With the current main branch you get the same result.

schlunma (approved):
Thanks everyone, looks great! 🚀

PR #1897, opened 2023-01-18
Fix type hints for Python versions < 3.10

bouweandela:
## Description

Python 3.8 and 3.9 do not support
```python
a = str | int
```
Python 3.8 does not support
```python
a = list[str]
```

even with `from __future__ import annotations`.

In function signatures, the syntax is supported though.

Credits to @schlunma for figuring this out.

This pull request updates the code so the older Python versions are type hinted correctly.

***

***

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -21,7 +21,6 @@ on:
   push:
     branches:
       - main
-      - mypy-python-version
   # run the test only if the PR is to main
   # turn it on if required
   #pull_request:
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -21,6 +21,7 @@ on:
   push:
     branches:
       - main
+      - mypy-python-version
   # run the test only if the PR is to main
   # turn it on if required
   #pull_request:
--- a/environment.yml
+++ b/environment.yml
@@ -47,7 +47,7 @@ dependencies:
   - sphinx_rtd_theme
   # Python packages needed for testing
   - flake8
-  - mypy  # >=0.990
+  - mypy<0.990  # >=0.990
   - pytest>=3.9,!=6.0.0rc1,!=6.0.0
   - pytest-cov>=2.10.1
   - pytest-env
--- a/environment.yml
+++ b/environment.yml
@@ -47,7 +47,7 @@ dependencies:
   - sphinx_rtd_theme
   # Python packages needed for testing
   - flake8
-  - mypy>=0.990
+  - mypy  # >=0.990
   - pytest>=3.9,!=6.0.0rc1,!=6.0.0
   - pytest-cov>=2.10.1
   - pytest-env
--- a/environment.yml
+++ b/environment.yml
@@ -47,7 +47,7 @@ dependencies:
   - sphinx_rtd_theme
   # Python packages needed for testing
   - flake8
-  - mypy<0.990  # >=0.990
+  - mypy>=0.990
   - pytest>=3.9,!=6.0.0rc1,!=6.0.0
   - pytest-cov>=2.10.1
   - pytest-env
--- a/esmvalcore/_recipe/from_datasets.py
+++ b/esmvalcore/_recipe/from_datasets.py
@@ -6,7 +6,7 @@ import logging
 import re
 from copy import deepcopy
 from functools import partial
-from typing import TYPE_CHECKING, Any, Iterable, Mapping, Sequence
+from typing import TYPE_CHECKING, Any, Dict, Iterable, Mapping, Sequence
 
 from nested_lookup import nested_delete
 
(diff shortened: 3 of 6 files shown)


valeriupredoi:
not sure this can be solved easily @bouweandela - have a look at this https://github.com/tiangolo/typer/issues/348 - am trying to pin typer see what's out of that, but until they fix it upstream I reckon we can't do much

bouweandela:
I'm not sure if typer is related to this issue. What puzzles me is that [dask](https://github.com/dask/dask/) uses a very similar configuration, so why is it not working for us?

valeriupredoi:
> I'm not sure if typer is related to this issue. What puzzles me is that [dask](https://github.com/dask/dask/) uses a very similar configuration, so why is it not working for us?

I thought typer was needed if we wanted to mess about with these type hints, I was wrong and I closed that PR - I am wondering though, maybe dask is using a different configuration of standards, or no mypy? Lemme have a look too

schlunma:
I don't understand how pinning the Python version of mypy is supposed to solve this. Code like

```
MyType = list[int] | dict
```

is simply not valid code in Python<3.10 and will always raise these errors, no mypy involved whatsoever.

schlunma:
But there is no `__future__` stuff in this PR, is there? On the other hand, dask uses that:

https://github.com/dask/dask/blob/81f771e05f57cab2838534c319a9b81f6e5a00cd/dask/highlevelgraph.py#L1

valeriupredoi:
> But there is no `__future__` stuff in this PR, is there? On the other hand, dask uses that:
> 
> https://github.com/dask/dask/blob/81f771e05f57cab2838534c319a9b81f6e5a00cd/dask/highlevelgraph.py#L1

not in this PR, Manu, since this a fixer PR - Bouwe put those _futures shtuff already, look at `esmvalcore/dataset.py` or `typing.py`

schlunma:
Ah, yes, that makes sense. Sorry, I didn't check that.

I think I know what the problem is: `from __future__ import annotations` [allows the usage of standard types in **variable and function definitions** ](https://peps.python.org/pep-0585/#implementation), i.e., something like:

```py
def f(x: list[int]) -> dict[str, float]:
    pass
```

is allowed and works with Python >= 3.7 when using `from __future__ import annotations`.

Example in our code: https://github.com/ESMValGroup/ESMValCore/blob/25e508adc14e9390fab81e412f035fe3ce607e35/esmvalcore/_recipe/from_datasets.py#L26

However, as far as I can tell, it is **not** possible to use this feature in code that is always executed (e.g., to define custom types) in Python < 3.10 like we do here:

https://github.com/ESMValGroup/ESMValCore/blob/25e508adc14e9390fab81e412f035fe3ce607e35/esmvalcore/_recipe/from_datasets.py#L20-L21

If you look at the error message we get, it's always those two lines. Other code that uses syntax like `dict[...]` in function and variable definitions does not raise errors.

valeriupredoi:
> Ah, yes, that makes sense. Sorry, I didn't check that.
> 
> I think I know what the problem is: `from __future__ import annotations` [allows the usage of standard types in **variable and function definitions** ](https://peps.python.org/pep-0585/#implementation), i.e., something like:
> 
> ```python
> def f(x: list[int]) -> dict[str, float]:
>     pass
> ```
> 
> is allowed and works with Python >= 3.7 when using `from __future__ import annotations`.
> 
> Example in our code:
> 
> https://github.com/ESMValGroup/ESMValCore/blob/25e508adc14e9390fab81e412f035fe3ce607e35/esmvalcore/_recipe/from_datasets.py#L26
> 
> However, as far as I can tell, it is **not** possible to use this feature in code that is always executed (e.g., to define custom types) in Python < 3.10 like we do here:
> 
> https://github.com/ESMValGroup/ESMValCore/blob/25e508adc14e9390fab81e412f035fe3ce607e35/esmvalcore/_recipe/from_datasets.py#L20-L21
> 
> If you look at the error message we get, it's always those two lines. Other code that uses syntax like `dict[...]` in function and variable definitions does not raise errors.

@schlunma I think that's a bingo! Since it doesn't seem to be mypy's fault here - let me see if we get the same with a lower mypy pin, if that's the case then we got us a runner

valeriupredoi:
yeah I reckon that's it - just figured it out it's not mypy that's at fault here, unless it's a dormant bug that's not been checked since a few versions ago - @schlunma you wanna make the changes you suggest directly in here and let's have us a test? I don't think @bouweandela would mind, since I've already hacked his PR quite a bit, before getting back to its original state :grin:

valeriupredoi (approved):
boom! Looking great, cheers, Manu! @bouweandela I took the liberty make this RfR and approve it while I'm at it too

valeriupredoi:
@bouweandela got a bit of time to look at this, buds?

PR #1922, opened 2023-02-09
Fix automatic download of ICON grid file and make ICON UGRIDization optional (default: true)

schlunma:
## Description

After extensive testing of the ICON on-the-fly CMORizer with the new UGRIDization (thanks @Paulinebonnet111!!) we discovered two issues with this:

1. The automatic download of the grid file can lead to corrupted nc files when initiated from multiple processes (this happens with `max_parellel_tasks` > 1) at the same time.
2. Some diagnostics cannot handle UGRID data (yet). An example is the [Psyplot diagnostic](https://docs.esmvaltool.org/en/latest/api/esmvaltool.diag_scripts.psyplot_diag.html?highlight=psyplot#module-esmvaltool.diag_scripts.psyplot_diag).

This PR fixes these issues by

1. introducing a [`filelock`](https://py-filelock.readthedocs.io/en/latest/index.html) before attempting to download the grid file. I also added a smarter caching of the mesh to the CMORizer, which reduces the number of times the lock is actually blocking processes to a minimum (I couldn't see any performance drops with ~100 ICON datasets).
2. making the UGRIDization optional with a new facet `ugrid` (can be used in the extra facets or in the recipe). The default option for this is `True`.

Sorry, the diff of this PR is really large because I copied code from `esmvalcore/cmor/_fixes/icon/icon.py` to `esmvalcore/cmor/_fixes/icon/_base_fixes.py` so that the cached mesh is also available in the base class. The three main changes are:

- Optional UGRIDization: https://github.com/ESMValGroup/ESMValCore/blob/a40c3a3ddeaed101bbf134bdec7ad220d71d7d1c/esmvalcore/cmor/_fixes/icon/icon.py#L299-L306
- Caching of meshes: https://github.com/ESMValGroup/ESMValCore/blob/a40c3a3ddeaed101bbf134bdec7ad220d71d7d1c/esmvalcore/cmor/_fixes/icon/_base_fixes.py#L287-L293
- Lock for downloading: https://github.com/ESMValGroup/ESMValCore/blob/a40c3a3ddeaed101bbf134bdec7ad220d71d7d1c/esmvalcore/cmor/_fixes/icon/_base_fixes.py#L211-L225

`filelock` was already in our environment (through other packages), but I added it anyway to make sure we get it.

Doc: https://esmvaltool--1922.org.readthedocs.build/projects/ESMValCore/en/1922/quickstart/find_data.html#icon

***

### DIFF
--- a/doc/quickstart/find_data.rst
+++ b/doc/quickstart/find_data.rst
@@ -363,7 +363,7 @@ algorithm <https://earthsystemmodeling.org/regrid/#regridding-methods>`__:
            reference: esmf_regrid.experimental.unstructured_scheme:regrid_unstructured_to_rectilinear
            method: conservative
 
-This automatic UGRIDization is enabled by default, but can be switched of with
+This automatic UGRIDization is enabled by default, but can be switched off with
 the facet ``ugrid: false`` in the recipe or the extra facets (see below).
 This is useful for diagnostics that do not support input data in UGRID format
 (yet) like the :ref:`Psyplot diagnostic <esmvaltool:recipes_psyplot_diag>` or
--- a/doc/quickstart/find_data.rst
+++ b/doc/quickstart/find_data.rst
@@ -337,33 +337,7 @@ A variable-specific default for the facet ``var_type`` is given in the extra
 facets (see next paragraph) for many variables, but this can be overwritten in
 the recipe.
 
-Similar to any other fix, the ICON fix allows the use of :ref:`extra
-facets<extra_facets>`.
-By default, the file :download:`icon-mappings.yml
-</../esmvalcore/config/extra_facets/icon-mappings.yml>` is used for that
-purpose.
-For some variables, extra facets are necessary; otherwise ESMValTool cannot
-read them properly.
-Supported keys for extra facets are:
-
-============= ============================= =================================
-Key           Description                   Default value if not specified
-============= ============================= =================================
-``latitude``  Standard name of the latitude ``latitude``
-              coordinate in the raw input
-              file
-``longitude`` Standard name of the          ``longitude``
-              longitude coordinate in the
-              raw input file
-``raw_name``  Variable name of the          CMOR variable name of the
-              variable in the raw input     corresponding variable
-              file
-``var_type``  Variable type of the          No default (needs to be specified
-              variable in the raw input     in extra facets or recipe if
-              file                          default DRS is used)
-============= ============================= =================================
-
-ESMValCore automatically makes native ICON data `UGRID
+ESMValCore can automatically make native ICON data `UGRID
 <https://ugrid-conventions.github.io/ugrid-conventions/>`__-compliant when
 loading the data.
 The UGRID conventions provide a standardized format to store data on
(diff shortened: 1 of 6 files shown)


Paulinebonnet111:
I tested it and all works fine ! Thanks :)

Paulinebonnet111:
I tested it and all works fine ! Thanks :)

Paulinebonnet111 (approved):
I tested it and all works fine ! Thanks :)

doc/quickstart/find_data.rst
```diff
@@ -337,6 +337,39 @@ A variable-specific default for the facet ``var_type`` is given in the extra
...
+
+   preprocessors:
+     regrid_icon:
+       regrid:
+         target_grid: 1x1
+         scheme:
+           reference: esmf_regrid.experimental.unstructured_scheme:regrid_unstructured_to_rectilinear
+           method: conservative
+
+This automatic UGRIDization is enabled by default, but can be switched of with
```
valeriupredoi, line 366:
```suggestion
This automatic UGRIDization is enabled by default, but can be switched off with
```

remi-kazeroni:
Hi @valeriupredoi, the start of the release process is approaching (next week). Would it be possible for you to take a look at this PR this week so that it could go into the release? Thanks a lot 🍺

esmvalcore/cmor/_fixes/icon/_base_fixes.py
```diff
@@ -27,9 +31,155 @@ def __init__(self, *args, **kwargs):
...
+            node_coords_and_axes=[(node_lat, 'y'), (node_lon, 'x')],
+            connectivities=[connectivity],
+            face_coords_and_axes=[(face_lat, 'y'), (face_lon, 'x')],
+        )
+
+        return mesh
+
+    def _get_grid_url(self, cube):
+        """Get ICON grid URL from cube."""
+        if self.GRID_FILE_ATTR not in cube.attributes:
```
valeriupredoi, line 140: hope this is defined somewhere in the unchanged code in this PR, right?

esmvalcore/cmor/_fixes/icon/_base_fixes.py
```diff
@@ -58,47 +208,105 @@ def get_horizontal_grid(self, cube):
...
+                now = datetime.now().timestamp()
+                age = now - mtime
+                if age < self.CACHE_VALIDITY:
+                    logger.debug("Using cached ICON grid file '%s'", grid_path)
+                    self._horizontal_grids[grid_name] = self._load_cubes(
+                        grid_path
+                    )
+                    return self._horizontal_grids[grid_name].copy()
+                logger.debug("Existing cached ICON grid file '%s' is outdated",
+                             grid_path)
```
valeriupredoi, line 236: this whole block should probably sit outside the lock context manager - you are just testing for the age of the file, and only initiate the locked download if the file is young enough

valeriupredoi (changes requested):
looking good @schlunma - just one typo and a question from me :beer:

esmvalcore/cmor/_fixes/icon/_base_fixes.py
schlunma: Yes, it is! https://github.com/ESMValGroup/ESMValCore/blob/a40c3a3ddeaed101bbf134bdec7ad220d71d7d1c/esmvalcore/cmor/_fixes/icon/_base_fixes.py#L28

esmvalcore/cmor/_fixes/icon/_base_fixes.py
schlunma: I thought so too at the beginning, but consider the following scenario (which I encountered in practice).

Process 1 starts the download and creates the grid file. Process 2 sees that the file exists and tries to use it, but since the download has not finished the file cannot be loaded --> Error.

Thus, the `if grid_path.exists():` needs to be inside the context manager. I would need a check for the existence of the file within the lock anyway, otherwise locked processes would still initiate the download after the lock has been released.

valeriupredoi: but are you checking anywhere for the existence of the `lock` file? Coz that's the one you don't want to have lingering about. At any rate, gonna approve now since it looks to be running fine and it actually dies from you guys' tests (not tested myself) :+1:

valeriupredoi (approved):
:beer:

esmvalcore/cmor/_fixes/icon/_base_fixes.py
schlunma: No, I am not, but that would be `filelock`'s job in the context manager :smile: However, the lock does not rely on the existence of the file, but rather uses a OS-related lock (see [here](https://github.com/tox-dev/py-filelock/blob/d8c6eb077b9695b0886ff9c10945a2687d245f09/src/filelock/_unix.py#L33-L41) for UNIX). That also means that the lock file persists after successful runs (but since this is a cache directory I don't see a problem with that).

valeriupredoi: hmm I thought the `lock` file needs to go poof when the locked operations finished successfully - maybe I'm too oldschool and things changed forwards in Filelock since I've last used it

schlunma: Yeah I though so too, but as you can see in the source code of `filelock` above, acquiring/releasing the lock does not create/delete the lock file, but uses [`fcntl.flock`](https://docs.python.org/3/library/fcntl.html#fcntl.flock) (if the file is not present it of course creates it first).

remi-kazeroni:
Thanks @schlunma for your contribution! And thanks @Paulinebonnet111 and @valeriupredoi for the reviews!

esmvalcore/cmor/_fixes/icon/_base_fixes.py
valeriupredoi: I'm pressing the disengage button before I start concerning myself with flocks and glocks :rofl:

PR #1935, opened 2023-02-17
Combined `offline` and `always_search_esgf` into a single option `search_esgf`

### LINKED ISSUE #1931: Combine `offline` and `always_search_esgf` into a single option `search_esgf`
The @ESMValGroup/technical-lead-development-team would like to combine the config-user.yml options `always_search_esgf` and `offline` into a new option `search_esgf` with possible values `never` (corresponds to the current `offline: true`), `default` (corresponds to the current `offline: false`), and `always` (corresponds with the current `offline: false` and `always_search_esgf: true`) and then retire the `offline` option at some point in the future.

_Originally posted by @bouweandela in https://github.com/ESMValGroup/ESMValCore/issues/1609#issuecomment-1424153588_

schlunma:
### Deprecation

This PR adds a new config option `search_esgf`, which replaces the old options `offline` and `always_search_esgf`. `offline` has been deprecated (scheduled for removal in v2.10.0), `always_search_esgf` is simply removed (has not been included in a release yet). The following table shows the usage of the new option:

| Old `offline` | Old `always_search_esgf` | New `search_esgf` | Description |
|---|---|---|---|
| `False` | `False` | `'default'` | Use local files when possible (regardless of version), otherwise download |
| `False` | `True` | `'always'` | Use local files only if latest version is available, otherwise download | 
| `True` | `False` | `'never'` | Only use local files, never download |
| `True` | `True` | `'never'` | Only use local files, never download |

In addition, this PR cleans and unifies the code to deprecate configuration options in order to allow changing one configuration option (e.g., `search_esgf`) based on another (deprecated) option (e.g., `offline`). In addition, now a deprecation warning is always raised when that option is use. Previously, only a warning was raised when the option was different to the default.

Closes #1931 

Link to documentation:
- https://esmvaltool--1935.org.readthedocs.build/projects/ESMValCore/en/1935/quickstart/configure.html#user-configuration-file
- https://esmvaltool--1935.org.readthedocs.build/projects/ESMValCore/en/1935/quickstart/find_data.html#enabling-automatic-downloads-from-the-esgf
- https://esmvaltool--1935.org.readthedocs.build/projects/ESMValCore/en/1935/quickstart/run.html#running

***

***


schlunma:
I just realized that it is a bit weird that we have an `search_esgf` option `default` now, but this is actually not the default setting (which is `never`). We could either

1. Make `default` the default setting (i.e., enable downloads by default). In the past we decided against that, but since there have been major improvements in the ESGF download feature maybe we should do that now?
2. Rename `default` to something different (I don't have a better suggestion at the moment).

@ESMValGroup/technical-lead-development-team @ESMValGroup/scientific-lead-development-team any opinions?

remi-kazeroni (commented):
Thanks very much for tackling this @schlunma and putting that together quickly! 🍻 The overview table in the description is quite handy to understand the changes. 👍 

I'd only be able to review this next Wednesday if that PR is not yet merged. But I had a general comment. Up to now, the default was `offline: True` which is now replaced by `search_esgf: never`. However, the new "default" behaviour is now `search_esgf: default`, i.e. our current `offline: False`. Back in the days, a majority of us was not in favour of allowing downloads by default. But I think many of our users have were happy with the auto-download option and would not mind enabling that by default. I would not mind allowing auto-download by default, I think it helps to improve the user-friendliness of the Tool. New users may have less trouble setting up (CMIP) data pools when working on systems with no data. But if other developers would prefer to keep the old "default" (`offline: True`), then we would need to find another name for the value in `search_esgf: default`... Maybe `missing` but I'm not sure...

bouweandela:
Thanks for taking this on @schlunma! Would it make sense to split this into two separate pull requests? One for adding the `search_esgf` option and one for displaying the warnings at the end? I'm a bit concerned that showing the warnings at the end is not trivial so might be best bumped to the next release.

> I tried to use [warnings.catch_warnings](https://docs.python.org/3/library/warnings.html#warnings.catch_warnings) for this, but it turns out that this can only record all warnings (or no warning at all), which is not what we want.

The warnings can be filtered afterwards though, each warning has a `.category` attribute that contains the class of the warning. You could use that to select only the `ESMValCoreDeprecationWarning`s. See [Testing warnings](https://docs.python.org/3/library/warnings.html#testing-warnings) for an example of how this can be done. I did try this out myself too, but I also ran into an issue: somehow not all warnings were caught. Maybe that could be avoided by not using a context manager though..

> As an alternative, I added a global [set](https://github.com/ESMValGroup/ESMValCore/blob/bf66348276771c6c3de897c8ff6254f642d9982c/esmvalcore/exceptions.py#L58-L63) that stores these warnings and raise them at the very end of the run.

If you run the tool with `max_parallel_tasks` other than `1`, it will use multiple processes. Because global variables are process specific, any warning coming from the preprocessors will be silently ignored with this approach.

schlunma:
> Thanks for taking this on @schlunma! Would it make sense to split this into two separate pull requests? One for adding the `search_esgf` option and one for displaying the warnings at the end? I'm a bit concerned that showing the warnings at the end is not trivial so might be best bumped to the next release.

Yeah, that makes sense (I though it's much easier to do that). We should make sure that this goes into v2.9 though, so that users have at least one release where all the warnings are shown at the bottom.

I opened an issue here: https://github.com/ESMValGroup/ESMValCore/issues/1938, let's discuss the implementation details there.

valeriupredoi:
cheers @schlunma :beer: And thanks for agreeing with @bouweandela about splitting (PR, not bananas :banana: ). `search_esgf:` as the only key config parameter sounds good to me, with values - what are the values - `default, always, never`? I'd change `default` to `optimized` or `conditional` ie downloads only if it's missing local data; also I am still uneasy to have the default value set to anything but `never` - we need to find a better way to tell users to have the downloads on only if they need to, otherwise we're gonna start filling up disks through users' unawareness in no time

schlunma:
@ESMValGroup/esmvaltool-coreteam @ESMValGroup/technical-lead-development-team @ESMValGroup/scientific-lead-development-team @ESMValGroup/userengagementteam

Hi everyone, could I ask you all briefly on your input on this?

**Question: What should be the default value for the new `search_esgf` option?**

- :+1: Use local files when possible (regardless of version), otherwise download (currently called `search_esgf=default`, the old `offline=False`).
- :-1: Only use local files, never download (currently called `search_esgf=never`, the old `offline=True`). This is the default option in the current release.

Some arguments for the first option can be found [here](https://github.com/ESMValGroup/ESMValCore/pull/1935#pullrequestreview-1303749140) (more user-friendly, ESGF download has been improved massively over the last months), some arguments for the second one can be found [here](https://github.com/ESMValGroup/ESMValCore/pull/1935#issuecomment-1436992344) (users might unintentionally fill up their disks without being aware of it).

Please react on this comment with a ":+1:" or ":-1: " to vote on this. If you have any other arguments for/against either of the options or any other comment on this, add a comment below! Thanks!!

valeriupredoi:
thanks @schlunma - my apologies, but I had to change your phrasing to **Question: What should be the DEFAULT VALUE (easily changeable in the user config file) for the new `search_esgf` option?** because there was a slight bias there in your initial statement, leaning towards a " :+1: " :grin:

valeriupredoi:
Currently score is 3-1 I really hope I don't end up like Brazil in the 2014 World Cup :rofl:

remi-kazeroni:
A few suggestions to make the 👍 option (`search_esgf=default`) more acceptable:

- Add some information for the `download_dir` of the config-user.yml file [here](https://github.com/ESMValGroup/ESMValCore/blob/main/esmvalcore/config-user.yml#L29-L30) to tell our users that this directory should be created in a file system allowing them to store tens of GB of data.
- Group settings related to auto-download settings (`download_dir`, `search_esgf`) one after the other to ease readability.
- Currently, when the auto-download is run for the first time, the default directory `~/climate_data` (from `download_dir: ~/climate_data`) is automatically created. We could leave it to the users to create their own `climate_data` directory where they want. If the auto-download is triggered but not `download_dir` exists, the Tool could return a meaningful error like "Please create your download_dir in a partition suited to store tens of GB of data" instead of creating it _by default_ in their home directory.
- Remind the amount of data downloaded at the very end of the log file, together with the runtime and maximum memory used. Currently, this information is only displayed at the end of the data finding step, e.g. `INFO    Downloaded 748.04 MB in 1 minute and 59.47 seconds (6.26 MB/s)` but that is hidden by lots of messages coming later on (even when using `log_level: info`).

valeriupredoi:
> A few suggestions to make the +1 option (`search_esgf=default`) more acceptable:
> 
>     * Add some information for the `download_dir` of the config-user.yml file [here](https://github.com/ESMValGroup/ESMValCore/blob/main/esmvalcore/config-user.yml#L29-L30) to tell our users that this directory should be created in a file system allowing them to store tens of GB of data.
> 
>     * Group settings related to auto-download settings (`download_dir`, `search_esgf`) one after the other to ease readability.
> 
>     * Currently, when the auto-download is run for the first time, the default directory `~/climate_data` (from `download_dir: ~/climate_data`) is automatically created. We could leave it to the users to create their own `climate_data` directory where they want. If the auto-download is triggered but not `download_dir` exists, the Tool could return a meaningful error like "Please create your download_dir in a partition suited to store tens of GB of data" instead of creating it _by default_ in their home directory.
> 
>     * Remind the amount of data downloaded at the very end of the log file, together with the runtime and maximum memory used. Currently, this information is only displayed at the end of the data finding step, e.g. `INFO    Downloaded 748.04 MB in 1 minute and 59.47 seconds (6.26 MB/s)` but that is hidden by lots of messages coming later on (even when using `log_level: info`).

this is gold - fully agree with @remi-kazeroni on this :beer:

remi-kazeroni (commented):
Thanks a lot for your contribution @schlunma! I took a look at the code and did various tests with the old options and the new one, using all possible values. All looks good to me. The new `search_esgf` option seems to have the same exact behaviour as the previous two options in all cases 👍  I only have a few minor comments, please feel free to include those.

I would have press the Approve button if the [debate](https://github.com/ESMValGroup/ESMValCore/pull/1935#issuecomment-1441498690) was settled but it seems the match is going to end with the penalties again (3-3)...

Also, I would like to have #1609 merged first. I'm afraid this will result in a not so easy merge conflict here...

valeriupredoi:
4-3 - tiiiight :grin: @schlunma did you just merge 1609 piecemeal?

schlunma:
Yeah, since there (hopefully) won't be many changes to 1609 anymore I thought I'd already solve the merge conflicts (were only about 30 of them :joy:)

valeriupredoi:
> Yeah, since there (hopefully) won't be many changes to 1609 anymore I thought I'd already solve the merge conflicts (were only about 30 of them joy)

good man! you are the conflicts slayer :axe:

valeriupredoi:
well, bad news for you buddy, you need to solve them back again :rofl:

schlunma:
Not all of them though :sweat_smile:

valeriupredoi:
> Not all of them though sweat_smile

famous last words :rofl:

schlunma:
`git merge -s ours` is my friend :sweat_smile:

schlunma:
I will continue to work on this on Monday, by then we can hopefully decide who the winner of our poll is :+1:

bouweandela:
> I will continue to work on this on Monday, by then we can hopefully decide who the winner of https://github.com/ESMValGroup/ESMValCore/pull/1935#issuecomment-1441498690 is :+1: 

With three opposing votes, I don't think this is a change we can make even if there is a majority. Our policy is that even a single opposing view can block a change. Maybe we could try to implement the changes proposed by @remi-kazeroni in https://github.com/ESMValGroup/ESMValCore/pull/1935#issuecomment-1441738717 (for the next release?) and then revisit the decision to see if there is still a lot of opposition.

remi-kazeroni:
> > I will continue to work on this on Monday, by then we can hopefully decide who the winner of [#1935 (comment)](https://github.com/ESMValGroup/ESMValCore/pull/1935#issuecomment-1441498690) is 👍
> 
> With three opposing votes, I don't think this is a change we can make even if there is a majority. Our policy is that even a single opposing view can block a change. 

Agreed, there are enough opposing votes not to make `default` the "default". But we would need the rename the `default` option to something else. So far, suggestions above were: `optimized`, `conditional`, and `missing`. Not sure I have a favourite one here...

> Maybe we could try to implement the changes proposed by @remi-kazeroni in [#1935 (comment)](https://github.com/ESMValGroup/ESMValCore/pull/1935#issuecomment-1441738717) (for the next release?) and then revisit the decision to see if there is still a lot of opposition.

Yes, maybe that can be done for the next release. This way we keep the old behaviour: no download by default but once the automatic download is turned on, data is downloaded by default in the user's home directory (`~/climate_data`). Maybe by then, we could switch to not having a default `download_dir` (especially in the user's home) instead of `search_esgf: never`.

bouweandela:
> So far, suggestions above were: `optimized`, `conditional`, and `missing`. Not sure I have a favourite one here...

`missing` would be my favorite out of those three, or we could do `when_missing` to capture the conditional part?

bettina-gier:
I'd add `supplement` or `fill`, but out of those above I agree with Bouwe's `when_missing`, `when_not_locally_available` seems too long.

valeriupredoi:
missing_2003_teaser_original_film_art

valeriupredoi:
or we can do `are_you_sure_the_data_is_not_there_please_check_again_fella` :rofl:

schlunma:
All right, I

- renamed the `default` option to `when_missing`.
- added some more text regarding the `download_dir`.
- regrouped the options in the `config-user.yml` file so that all download-related options are grouped together.
- cleaned and unified the code to deprecate configuration options to allow changing one configuration option (e.g., `search_esgf`) based on another (deprecated) option (e.g., `offline`). In addition, now a deprecation warning is always raised when that option is used (previously, only a warning was raised when the option was different to the default).

I'd like to ignore the remaining Codacy options since IMHO non-used function arguments are clearer than using a bunch of `_`, `__`, etc.

This is ready for review now :+1:

valeriupredoi:
super trooper @schlunma - I'll go through the code tomorrow, before our TLT call :+1:

esmvalcore/config/_config_validators.py
```diff
@@ -4,10 +4,12 @@
...
 import os.path
 import warnings
-from collections.abc import Iterable
+from collections.abc import Callable, Iterable
 from functools import lru_cache, partial
 from pathlib import Path
-from typing import Any, Optional, Union
+from typing import TYPE_CHECKING, Any, Optional, Union
+
+from packaging import version
```
bouweandela, line 12: Don't forget to add the new dependency to setup.py/environment.yml

esmvalcore/dataset.py
```diff
@@ -598,35 +598,41 @@ def _find_files(self) -> None:
...
+        # is used: check ESGF
+        local_files = {f.name: f for f in self.files}
+        search_result = esgf.find_files(**self.facets)
+        for file in search_result:
+            if file.name not in local_files:
+                # Use ESGF files that are not available locally.
+                self.files.append(file)
+            else:
+                # Use ESGF files that are newer than the locally available
+                # files (can only happen in 'always' mode)
```
bouweandela, line 630: This also happens in `when_missing` mode when files covering part of the the requested timerange are missing but not the entire range

notebooks/loading-and-processing-data.ipynb
```diff
@@ -275,11 +275,11 @@
       "WARNING:esmvalcore.cmor.check:There were warnings in variable tas:\n",
       " Coordinate lon units degrees converted to degrees_east\n",
       " Coordinate lat units degrees converted to degrees_north\n",
-      "loaded from file /home/bandela/climate_data/cmip5/output1/CCCma/CanESM2/historical/mon/atmos/Amon/r1i1p1/v20120718/tas_Amon_CanESM2_historical_r1i1p1_185001-200512.nc\n",
+      "loaded from file ~/climate_data/cmip5/output1/MPI-M/MPI-ESM-MR/historical/mon/atmos/Amon/r1i1p1/v20120503/tas_Amon_MPI-ESM-MR_historical_r1i1p1_185001-200512.nc\n",
```
bouweandela, line 278: These warnings do not appear whith this model, they're specific to CanESM2

bouweandela:
Thanks for picking this up @schlunma! When I run the tool with `offline: false` in config-user.yml, I get the deprecation warning twice:
```
/home/bandela/src/esmvalgroup/esmvalcore/esmvalcore/config/_config_validators.py:340: ESMValCoreDeprecationWarning: The configuration option or command line argument `offline` has been deprecated in ESMValCore version 2.8.0 and is scheduled for removal in version 2.10.0. Please use the options `search_esgf=never` (for `offline=True`) or `search_esgf=when_missing` (for `offline=False`) instead. These are exact replacements.
  warnings.warn(deprecation_msg, ESMValCoreDeprecationWarning)
/home/bandela/src/esmvalgroup/esmvalcore/esmvalcore/_config/__init__.py:13: ESMValCoreDeprecationWarning: The private module `esmvalcore._config` has been deprecated in ESMValCore version 2.8.0 and is scheduled for removal in version 2.9.0. Please use the public module `esmvalcore.config` instead.
  warnings.warn(
/home/bandela/src/esmvalgroup/esmvalcore/esmvalcore/config/_config_validators.py:340: ESMValCoreDeprecationWarning: The configuration option or command line argument `offline` has been deprecated in ESMValCore version 2.8.0 and is scheduled for removal in version 2.10.0. Please use the options `search_esgf=never` (for `offline=True`) or `search_esgf=when_missing` (for `offline=False`) instead. These are exact replacements.
  warnings.warn(deprecation_msg, ESMValCoreDeprecationWarning)
```

esmvalcore/config/_config_validators.py
schlunma: Good find, I thought it's already there. I also noticed that it's missing in ESMValTool, issue: https://github.com/ESMValGroup/ESMValTool/issues/3059.

notebooks/loading-and-processing-data.ipynb
schlunma: I re-ran the entire notebook after clearing all outputs, so I guess these warnings also appear for the other model. Nevertheless, I will wait until https://github.com/ESMValGroup/ESMValCore/pull/1945 is merged before further changes to the notebooks here.

schlunma:
> Thanks for picking this up @schlunma! When I run the tool with `offline: false` in config-user.yml, I get the deprecation warning twice:
> 
> ```
> /home/bandela/src/esmvalgroup/esmvalcore/esmvalcore/config/_config_validators.py:340: ESMValCoreDeprecationWarning: The configuration option or command line argument `offline` has been deprecated in ESMValCore version 2.8.0 and is scheduled for removal in version 2.10.0. Please use the options `search_esgf=never` (for `offline=True`) or `search_esgf=when_missing` (for `offline=False`) instead. These are exact replacements.
>   warnings.warn(deprecation_msg, ESMValCoreDeprecationWarning)
> /home/bandela/src/esmvalgroup/esmvalcore/esmvalcore/_config/__init__.py:13: ESMValCoreDeprecationWarning: The private module `esmvalcore._config` has been deprecated in ESMValCore version 2.8.0 and is scheduled for removal in version 2.9.0. Please use the public module `esmvalcore.config` instead.
>   warnings.warn(
> /home/bandela/src/esmvalgroup/esmvalcore/esmvalcore/config/_config_validators.py:340: ESMValCoreDeprecationWarning: The configuration option or command line argument `offline` has been deprecated in ESMValCore version 2.8.0 and is scheduled for removal in version 2.10.0. Please use the options `search_esgf=never` (for `offline=True`) or `search_esgf=when_missing` (for `offline=False`) instead. These are exact replacements.
>   warnings.warn(deprecation_msg, ESMValCoreDeprecationWarning)
> ```

This is weird, I am only getting it once (test with every combination of `offline` as config file option and command line argument. Due to the way sessions and configs are created the code that issues the warnings runs more than once, but the [default setting](https://docs.python.org/3/library/warnings.html#the-warnings-filter) of the `warnings` module is to show warnings just once for each location, so I would only expect this warning once.

However, I also noticed to other problems:

- The deprecation warnings shown at the very beginning on screen are missing in the log files (regular + debug log). I guess the reasons is that logging is not configured that that stage, but this is really a problem for user who don't use the tool interactively. They will never see this deprecation warning.
- I get the deprecation warning about `use_legacy_supplementaries` even though I am not using this in my config file/command line and the recipe does not include `fx_variables:`. I will investigate this.

schlunma:
> Thanks for picking this up @schlunma! When I run the tool with `offline: false` in config-user.yml, I get the deprecation warning twice:
> 
> ```
> /home/bandela/src/esmvalgroup/esmvalcore/esmvalcore/config/_config_validators.py:340: ESMValCoreDeprecationWarning: The configuration option or command line argument `offline` has been deprecated in ESMValCore version 2.8.0 and is scheduled for removal in version 2.10.0. Please use the options `search_esgf=never` (for `offline=True`) or `search_esgf=when_missing` (for `offline=False`) instead. These are exact replacements.
>   warnings.warn(deprecation_msg, ESMValCoreDeprecationWarning)
> /home/bandela/src/esmvalgroup/esmvalcore/esmvalcore/_config/__init__.py:13: ESMValCoreDeprecationWarning: The private module `esmvalcore._config` has been deprecated in ESMValCore version 2.8.0 and is scheduled for removal in version 2.9.0. Please use the public module `esmvalcore.config` instead.
>   warnings.warn(
> /home/bandela/src/esmvalgroup/esmvalcore/esmvalcore/config/_config_validators.py:340: ESMValCoreDeprecationWarning: The configuration option or command line argument `offline` has been deprecated in ESMValCore version 2.8.0 and is scheduled for removal in version 2.10.0. Please use the options `search_esgf=never` (for `offline=True`) or `search_esgf=when_missing` (for `offline=False`) instead. These are exact replacements.
>   warnings.warn(deprecation_msg, ESMValCoreDeprecationWarning)
> ```

@remi-kazeroni and I found the reason for this! I don't use a default user config file in `.esmvaltool`, Remi does. I only get this error once, Remi twice. I guess you have a default config as well @bouweandela?

This has not been introduced with this PR. If you use the `main` branch you'll also get two deprecation warnings if you use `use_legacy_supplementaries` in the default config in `.esmvaltool`.

I have no idea what's going on there. Since the warning is raised twice from the exact same location, it should only be printed once. If no one has an idea about this, I suggest we simply ignore this. Two warnings is better than no warning.

notebooks/loading-and-processing-data.ipynb
bouweandela: Strange, I don't see them when running the notebook for #1945.

bouweandela:
I found a really simple way to print the warnings issued from loading the configuration file again and added it in [a05c7d1](https://github.com/ESMValGroup/ESMValCore/pull/1935/commits/a05c7d1c1807fff2387a488de4f26158bff5fd72). Please feel free to not use that if you find it doesn't do a good job.

doc/quickstart/configure.rst
```diff
@@ -53,38 +53,40 @@ omitted in the file.
...
   # Used by some recipes to look for additional datasets.
   auxiliary_data_dir: ~/auxiliary_data
 
+  # Automatic data download from ESGF --- [never]/when_missing/always
+  # Use automatic download of missing CMIP3, CMIP5, CMIP6, CORDEX, and obs4MIPs
+  # data from ESGF. ``never`` disables this feature, which is useful if you are
+  # working on a computer without an internet connection. ``when_missing``
+  # enables the automatic download for files that are not available locally.
+  # ``always`` will always check ESGF for the latest version of a file, and
+  # only uses local files if they correspond to that latest version.
```
valeriupredoi, line 66:
```suggestion
  # will only use local files if they correspond to that latest version.
```

doc/quickstart/configure.rst
```diff
@@ -53,38 +53,40 @@ omitted in the file.
...
+  search_esgf: never
+
+  # Directory for storing downloaded climate data
+  # Make sure to use a directory where you can store multiple GBs of data. Your
+  # home directory on a HPC is usually not suited for that purpose, so please
```
valeriupredoi, line 71:
```suggestion
  # home directory on a HPC is usually not suited for this purpose, so please
```

doc/quickstart/configure.rst
```diff
@@ -53,38 +53,40 @@ omitted in the file.
...
+  # adapt the default value in this case!
```
valeriupredoi, line 72:
```suggestion
  # change the default value in this case!
```

schlunma:
> The warning was issued twice because it is first issued from the `esmvalcore.config` module when it the config-user.yml file is loaded when the module is loaded, and then from the `emvalcore._main` module when the config-user.yml file gets loaded there. The [default filter](https://docs.python.org/3/library/warnings.html#the-warnings-filter) prints a warning once for every module that issues it. I just pushed [5eeed56](https://github.com/ESMValGroup/ESMValCore/pull/1935/commits/5eeed56ea65705ae1d2a2e9fcd4f57fd774118b4) that should solve the issue.

Ahh, that makes sense. I missed the part about "once per module" :facepalm: 

Unfortunately, your change didn't work for me. I am still getting two warnings with `.esmvaltool/config-user.yml`, and now even three warnings when I use a command line argument AND a configuration option.

Maybe we should simply use the filter option `"once"`?

> I found a really simple way to print the warnings issued from loading the configuration file again and added it in [a05c7d1](https://github.com/ESMValGroup/ESMValCore/pull/1935/commits/a05c7d1c1807fff2387a488de4f26158bff5fd72). Please feel free to not use that if you find it doesn't do a good job.

That worked great for me! Maybe trying to solve this for all deprecation warnings was overly enthusiastic :smiley:

doc/quickstart/configure.rst
```diff
@@ -136,11 +138,17 @@ omitted in the file.
...
+The ``search_esgf`` setting can be used to disable or enable automatic
+downloads from ESGF.
+If ``search_esgf`` is set to ``never``, the tool does not download any data
+from the ESGF.
+If ``search_esgf`` is set to ``when_missing``, the tool will download any CMIP3,
+CMIP5, CMIP6, CORDEX, and obs4MIPs data that is required to run a recipe but
+not available locally and store it in ``download_dir`` using the ``ESGF``
 directory structure defined in the :ref:`config-developer`.
+If ``search_esgf`` is set to ``always``, the tool will check for the data
+mentioned above on ESGF regardless of the data availability on the local
```
valeriupredoi, line 150:
```suggestion
mentioned above, on ESGF, regardless of the data availability on the local
```

doc/quickstart/configure.rst
```diff
@@ -136,11 +138,17 @@ omitted in the file.
...
+filesystem, and only use local files if they correspond to the latest version.
```
valeriupredoi, line 151: this is a bit twisty - what's the workflow doing: first, go to ESGF, find the data there, then come back to local, see if the latest versioned data is there, and user didn't request an (other, particular) version. If local doesn't have it or user specified a specific version, download from ESGF, else, use local data?

bouweandela:
Will have another look..

doc/quickstart/configure.rst
schlunma: Yes, that's basically what happens. I tried to summarize this as best as possible, but I am very open to better suggestions :+1:

valeriupredoi (changes requested):
couple more minor comments from me :beer:

doc/quickstart/configure.rst
valeriupredoi: you did well, but I have a bit of a Hemingway gene myself so how about this:

If ``search_esgf`` is set to ``always``, the tool will first check the ESGF for the needed data, regardless of any local data availability; if the data found on ESGF is newer than the local data (if any) or the user specifies a version of the data that is available only from the ESGF, then that data will be downloaded; otherwise, local data will be used.

esmvalcore/dataset.py
```diff
@@ -598,35 +598,41 @@ def _find_files(self) -> None:
...
+                # Use ESGF files that are not available locally.
+                self.files.append(file)
+            else:
+                # Use ESGF files that are newer than the locally available
+                # files
+                local_file = local_files[file.name]
+                if 'version' in local_file.facets:
+                    if file.facets['version'] > local_file.facets['version']:
+                        idx = self.files.index(local_file)
+                        self.files[idx] = file
```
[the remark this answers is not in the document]
schlunma: The `else` here is that the local file is up-to-date, i.e., we should use this (notice that `self.files` already contains that). If local is missing that data it does not end up in that part of the `if` (the one right after the `for` loop).

setup.py
```diff
@@ -47,6 +47,7 @@
         'nested-lookup',
         'netCDF4',
         'numpy',
+        'packaging',
```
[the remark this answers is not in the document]
valeriupredoi: note to self then: explain ot to myself why we need it :laughing:

esmvalcore/dataset.py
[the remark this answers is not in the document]
valeriupredoi: aha gotcha, makes sense then - thanks, man!

valeriupredoi (approved):
you're a :star: Manu! Cheers for all this wonderful work - together with @bouweandela I nominate you guys for the Contributors of the Version Award for 2.8 :trophy:

bouweandela:
> Unfortunately, your change didn't work for me. I am still getting two warnings with .esmvaltool/config-user.yml, and now even three warnings when I use a command line argument AND a configuration option.
>
> Maybe we should simply use the filter option "once"?

It only works when you specify no configuration file, i.e. with the command `esmvaltool run examples/recipe_python.yml`. Unfortunately, I ran out of time to fix this today and I'm not working tomorrow, so please feel free to undo the code changes I made if you think the previous behavior was better.

schlunma:
I suggest we just live with multiple warnings then, I also tried to solve this with `action=once` but was not succesful.

If there are no further comments I suggest to merge this by noon tomorrow. Thanks everyone for reviewing!!

valeriupredoi:
aren't the warnings being generated by different, multiple, tasks? I couldn't care less for multiple warning TBF - if at all, show them more often, maybe somebody decides to pay attention to them :laughing:

schlunma:
I think they are all issues from the same process (all the reading of the config file happens at the very beginning of the run where we only have a single process as far as I am aware).

valeriupredoi:
awesome :partying_face:

Issue #1948, opened 2023-03-01
Recipe runs sometimes fail due to multiple processes attempting to cleanup the same directory


schlunma:
**Describe the bug**

#1609 and related PRs introduced a race condition: running the same recipe over and over sometimes leads to successful runs and sometimes to errors like:

```
FileNotFoundError: [Errno 2] No such file or directory: '/home/b/b309141/scratch/esmvaltool_output/recipe_anav13jclim_20230228_201248/preproc/CMIP5_BNU-ESM_fx_historical_r0i0p0_sftlf_fixed'
```

**Reasons**

Supplementary variables do not get a `diagnostic` or `variable_group` facet when read from the recipe:

https://github.com/ESMValGroup/ESMValCore/blob/6d285d283095b55a5e6d89e01e24536e94f256e5/esmvalcore/_recipe/to_datasets.py#L360-L367

Thus, when creating the output file, which is eventually used to determine the fixed file path (used by the `fix_file` preprocessor), these directories end up directly in the `preproc` directory:

https://github.com/ESMValGroup/ESMValCore/blob/6d285d283095b55a5e6d89e01e24536e94f256e5/esmvalcore/local.py#L485-L490

An example is shown in the error message above: `.../preproc/CMIP5_BNU-ESM_fx_historical_r0i0p0_sftlf_fixed`. Finally, if multiple diagnostics use the same supplementary variable and the tool is run with multiple processes, this might lead to the aforementioned error in the [cleanup](https://github.com/ESMValGroup/ESMValCore/blob/main/esmvalcore/preprocessor/_io.py#L349) preprocessor.

**Solution**

I think adding `diagnostic` and `variable_group` to the supplementary variables would be solution to that, but I am not sure if these have been omitted on purpose (especially since the `_get_output_file` function specifically supports variables without these facets). @bouweandela probably has an answer to that. A problem with this might be that this might lead to conflicts if the diagnostic uses the supplementary variable as regular variable. So maybe we attach some kind of suffix to these fixed dirs in addition?

**Please attach**
[...]

PR #1949, opened 2023-03-01
Fixed race condition that may result in errors in `cleanup` and deprecate `cleanup`

schlunma:
## Description

This PR removes the race condition that may lead to errors in the `cleanup` preprocessor by using temporary directories to store fixed files.

Closes https://github.com/ESMValGroup/ESMValCore/issues/1948

## Deprecation

This PR deprecates the [`cleanup`](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/api/esmvalcore.preprocessor.html?highlight=cleanup#esmvalcore.preprocessor.cleanup) preprocessor, which is not needed anymore. Users should not use that anymore.

***

***


schlunma:
Makes sense. I guess this just hasn't surfaced yet since not many datasets actually use `fix_file`.

Just to be sure I understand this correctly: create this temporary file for **each** Dataset, also the non-supplementaries?

bouweandela:
> Just to be sure I understand this correctly: create this temporary file for each Dataset, also the non-supplementaries?

Yes, but a directory, since each dataset can consist of multiple files

schlunma:
Done!

schlunma:
Did you look at the code before approving? :stuck_out_tongue_closed_eyes: 

https://github.com/ESMValGroup/ESMValCore/blob/2fcdf1faabaa864b8ecd0cd73d38f60b969e302b/esmvalcore/_main.py#L459-L470

valeriupredoi:
> Did you look at the code before approving? stuck_out_tongue_closed_eyes
> 
> https://github.com/ESMValGroup/ESMValCore/blob/2fcdf1faabaa864b8ecd0cd73d38f60b969e302b/esmvalcore/_main.py#L459-L470

`shutil.rmtree(Manu)` pshhh :laughing: Yeah I missed a big chunk indeed :grin:

esmvalcore/cmor/_fixes/fix.py
```diff
@@ -197,21 +198,25 @@ def get_fixes(project, dataset, mip, short_name, extra_facets=None):
         return fixes
 
     @staticmethod
-    def get_fixed_filepath(output_dir, filepath):
+    def get_fixed_filepath(
+        output_dir: str | Path,
+        filepath: str | Path,
+    ) -> str:
```
bouweandela, line 204: Would it be possible to make the return value a `Path`? I know this is not strictly backward compatible, but it would be nice to try and switch over to using `Path`s instead of `str` when we mean a path everywhere in the tool.

esmvalcore/dataset.py
```diff
@@ -848,3 +849,21 @@ def _update_timerange(self):
         check.valid_time_selection(timerange)
 
         self.set_facet('timerange', timerange)
+
+    def get_temporary_fixed_file_dir(self) -> Path:
```
bouweandela, line 853: Could you make this method private (i.e. start the name with an `_`)? Removing things from the public API is difficult and I would like to avoid exposing this functionality. As soon as conversion from an `xarray.DataArray` to an `iris.cube.Cube` works properly (and I know the iris team has plans for this), we can get rid of fix file altogether by loading the file as an uninterpreted `xarray.DataArray`, applying fixes, and then converting to a cube. That will make computations much faster for datasets that require file fixes.

esmvalcore/dataset.py
```diff
@@ -848,3 +849,21 @@ def _update_timerange(self):
...
+        -------
+        Path
+            Path to new temporary directory.
+
+        """
+        fixed_file_dir = self.session.fixed_file_dir
+        fixed_file_dir.mkdir(parents=True, exist_ok=True)
+        facets_for_prefix = ('project', 'dataset', 'mip', 'short_name')
+        prefix = '_'.join(
+            [str(self.facets.get(facet, '')) for facet in facets_for_prefix] +
```
bouweandela, line 866:
```suggestion
            ["-".join(str(elem) for elem in value) if isinstance(value, (list, tuple)) else str(value) for facet, value in self.facets.items() if facets in facets_for_prefix] +
```
would be nicer since it avoids sequences of underscores and the characeters `[](),` in the directory name (the latter only when we have the full list of facets from the `summary` method).

bouweandela (changes requested):
Since the `cleanup` preprocessor function is no longer usable since #1609, we might as well remove it completely. Is that something that we would like to do in this pull request or in a different pull request?

esmvalcore/cmor/_fixes/fix.py
schlunma: I also thought about this, but decided against it because of backwards-compatibility. I will change it, but without a deprecation warning.

esmvalcore/dataset.py
schlunma: I made this public on purpose. `fix_file` (which takes `output_dir` as an argument) is a public function, so I thought it's nice to provide a public method to easily get an output dir. I guess it will take some time until this proper solution via `xarray.DataArray` is available (iris needs to implement it, then we need to implement it). But yeah, I will make it private.

schlunma:
Removing `cleanup` would introduce yet another deprecation, but let me see what I can do.

schlunma:
Thanks @bouweandela for your review, I implemented all requested changes.

I am currently re-running a couple of recipes, please do not merge before these runs have finished.

schlunma:
(I won't address the missing test in the IPSL CMORizer)

bouweandela:
Sorry to come up with this suggestion so late, but thinking about this a bit more: we're now creating a huge number of output directories for fix_file (one for every dataset load), while very few datasets need this. It would be better only to create the directory when we actually need it. Shall we move the temporary directory creation to the function `Fix.get_fixed_filepath` instead?

remi-kazeroni:
> Removing `cleanup` would introduce yet another deprecation, but let me see what I can do.

Thanks a lot for your work @schlunma! Please remember to use the "deprecated feature" label if that's needed and add a note about the deprecation in the PR header 🍻

esmvalcore/dataset.py
```diff
@@ -848,3 +867,19 @@ def _update_timerange(self):
...
+        Returns
+        -------
+        Path
+            Path to new temporary directory.
+
+        """
+        fixed_file_dir = self.session._fixed_file_dir
+        fixed_file_dir.mkdir(parents=True, exist_ok=True)
+        prefix = self._get_joined_summary_facets('_', join_lists=True)
+        if prefix != '':
```
bouweandela, line 883: This never happens, it's impossible to load a dataset without specifying at the very least project, mip, variable, dataset.

esmvalcore/dataset.py
```diff
@@ -848,3 +867,19 @@ def _update_timerange(self):
...
+            prefix += '_'
+        return Path(tempfile.mkdtemp(prefix=prefix, dir=fixed_file_dir))
```
bouweandela, line 885:
```suggestion
        return self.session._fixed_file_dir / prefix
```

esmvalcore/dataset.py
```diff
@@ -848,3 +867,19 @@ def _update_timerange(self):
...
+        """Create and return new temporary directory for storing fixed files.
+
+        Returns
+        -------
+        Path
+            Path to new temporary directory.
+
+        """
+        fixed_file_dir = self.session._fixed_file_dir
+        fixed_file_dir.mkdir(parents=True, exist_ok=True)
```
bouweandela, line 881:
```suggestion
```

esmvalcore/cmor/_fixes/fix.py
```diff
@@ -197,21 +198,24 @@ def get_fixes(project, dataset, mip, short_name, extra_facets=None):
...
         -------
-        str
+        Path
             Path to the fixed file.
         """
-        if not os.path.isdir(output_dir):
-            os.makedirs(output_dir)
-        return os.path.join(output_dir, os.path.basename(filepath))
+        output_dir = Path(output_dir)
+        output_dir.mkdir(parents=True, exist_ok=True)
```
bouweandela, line 220:
```suggestion
        output_dir = Path(tempfile.mkdtemp(prefix=output_dir))
```

esmvalcore/preprocessor/_io.py
```diff
@@ -347,7 +347,34 @@ def _get_debug_filename(filename, step):
...
 
 def cleanup(files, remove=None):
-    """Clean up after running the preprocessor."""
+    """Clean up after running the preprocessor.
+
+    Warning
+    -------
+    .. deprecated:: 2.8.0
+        This function has been deprecated in ESMValCore version 2.8.0 and is
+        scheduled for removal in version 2.10.0.
```
bouweandela, line 356:
```suggestion
        This function is no longer used and has been deprecated since
        ESMValCore version 2.8.0. It is scheduled for removal in version 2.10.0.
```

esmvalcore/dataset.py
```diff
@@ -96,6 +97,22 @@ class Dataset:
         Facets describing the dataset.
     """
 
+    _SUMMARY_FACETS = (
```
bouweandela, line 100:
```suggestion
    _summary_facets = (
```
Maybe use lowercase here to be consistent with other places where we have class attributes?

schlunma:
> Sorry to come up with this suggestion so late, but thinking about this a bit more: we're now creating a huge number of output directories for fix_file (one for every dataset load), while very few datasets need this. It would be better only to create the directory when we actually need it. Shall we move the temporary directory creation to the function `Fix.get_fixed_filepath` instead? I added some code suggestions below to do this.

This used to be the behavior pre-#1609, but has been changed here:

https://github.com/ESMValGroup/ESMValCore/blob/f87e25e566d681d3eb3d9fcf346b5ef0abbd289f/esmvalcore/cmor/fix.py#L58-L59

I thought this was on purpose, that's why I did not change that here...I will change it.

bouweandela:
Great, thanks!

esmvalcore/dataset.py
schlunma: PEP8 [recommends](https://peps.python.org/pep-0008/#constants) to write constants in all-capital letter. We also have examples in our code where class attributes are capitals, so I'd like to leave it like this.

schlunma:
Done. The fixed file directory is now only created if necessary. Note that I added an optional argument to `fix_file` to ensure backwards-compatibility.

esmvalcore/cmor/_fixes/fix.py
```diff
@@ -197,21 +218,38 @@ def get_fixes(project, dataset, mip, short_name, extra_facets=None):
         return fixes
 
     @staticmethod
-    def get_fixed_filepath(output_dir, filepath):
+    def get_fixed_filepath(
+        output_dir: str | Path,
+        filepath: str | Path,
+        create_temporary_dir: bool = False,
```
bouweandela, line 224: Good that you're preserving backward compatibility here. How about calling the argument `add_suffix` or `add_unique_suffix` or something similar? And then explain this in the help text with e.g. 'adds a unique suffix to output dir for thread safety'? To me, that describes the feature better than `create_temporary_dir`, because the directory isn't really temporary in the sense that it is not stored in the location for temporary files nor is it removed automatically afterward.

bouweandela (approved):

bouweandela:
Thanks for fixing this @schlunma!

I made a small final change myself in [50dfc8e](https://github.com/ESMValGroup/ESMValCore/pull/1949/commits/50dfc8e3c6613d019585b272c7f4e583745b3cca) instead of putting in another review comment, to not hold up the release process any longer. I hope that's OK with you.

remi-kazeroni:
Thanks very much @schlunma for tackling this last minute issue before the feature freeze! And also many thanks to @bouweandela and @valeriupredoi for the reviews! I'm going to merge now. This means: feature freeze is now and we're finalizing v2.8.0rc1 🍺

Issue #1981, opened 2023-03-21
Provenance tracking in `mask_multimodel` is slow and needs a lot of memory


schlunma:
EDIT: The reason for the issue below is the provenance tracking (see this [comment](https://github.com/ESMValGroup/ESMValCore/issues/1981#issuecomment-1478107436)).

While developing a new recipe, @bettina-gier and I found something very weird. A recipe that previously finished in 6 min did not finish in 8 hours after adding the `mask_multimodel` preprocessor.

This can be reproduced with the following simple recipe:

<details>
<summary>Example recipe</summary>

```yml
# ESMValTool
---
documentation:
  title: test
  description: test
  authors:
    - schlund_manuel

preprocessors:

  test:
    mask_landsea:
      mask_out: sea
    regrid:
      target_grid: 2x2
      scheme: nearest
    mask_multimodel:

diagnostics:

  diag_test:
    variables:
      tas:
        preprocessor: test
        project: CMIP5
        ensemble: r1i1p1
        mip: Amon
        exp: historical
        start_year: 2000
        end_year: 2004
        additional_datasets:
          - {dataset: ACCESS1-0}
          - {dataset: BNU-ESM}
          - {dataset: CanESM2}
          - {dataset: CCSM4}
          - {dataset: CNRM-CM5}
          - {dataset: CSIRO-Mk3-6-0}
          - {dataset: GFDL-ESM2M}
          - {dataset: HadGEM2-ES}
          - {dataset: GISS-E2-R}
          - {dataset: inmcm4}
          - {dataset: IPSL-CM5B-LR}
          - {dataset: MIROC-ESM}
          - {dataset: MPI-ESM-LR}
          - {dataset: MRI-CGCM3}
          # - {dataset: NorESM1-M}
    scripts:
      null
```
</details>

When varying the number of datasets, I get the following behavior of the maximum memory usage and run time (note that all runs were performed on the same interactive Levante node):

![mem](https://user-images.githubusercontent.com/32543114/226639599-63ba508f-f32a-4bc9-b058-785b271529dc.png)

![time](https://user-images.githubusercontent.com/32543114/226639627-b5aa4938-b0d8-4831-8d89-4f23ba05b837.png)

These are clearly exponential relationships! No wonder this results in a timeout with 20+ models...Good news is that the exponential behavior is already present in v2.7.0 (so no need to investigate that right now), but v2.8.0rc2 has made it worse by a factor of ~1.5.

[...]

PR #1984, opened 2023-03-22
Make sure to only use provenance copies when using `product.wasderivedfrom()`

schlunma:
## Description

This PR makes sure to only use provenance copies (`product.copy_provenance()`) as input for `product.wasderivedfrom()`. This prevents recursively adding large amounts of ancestors which blows up memory usage and run time.

Closes https://github.com/ESMValGroup/ESMValCore/issues/1981

***

***


schlunma:
> This could be implemented more efficiently by making fewer copies. For example, this implementation seems to copy all input file provenance for all output files, leading to n x (n-1) copies for the multi dataset mask (assuming all datasets are masked), where n is the number of datasets. However, only the input file provenance needs to be copied, which would be just n copies.

Makes sense, though the performance gain is probably not noticeable for common recipes. Implemented in https://github.com/ESMValGroup/ESMValCore/pull/1984/commits/341c650e28721bdaa69b543e9ee9ad54f19bc70b.

valeriupredoi:
good work, gents! @bouweandela could you maybe pls check the new commits, review, and merge this then :beer:

valeriupredoi:
@bouweandela frenly yodeling ping
BouweYodel

valeriupredoi:
@bouweandela one for you here, bud - quick one, get you back to work after the Easter break :grin:

esmvalcore/preprocessor/_bias.py
```diff
@@ -106,7 +106,7 @@ def bias(products, bias_type='absolute', denominator_mask_threshold=1e-3,
         cube.metadata = cube_metadata
         cube.units = new_units
         product.attributes['units'] = new_units
-        product.wasderivedfrom(reference_product)
+        product.wasderivedfrom(reference_product.copy_provenance())
```
bouweandela, line 109: This copy is not needed, since `reference_product` is not updated in this loop: https://github.com/ESMValGroup/ESMValCore/blob/2ecbb5c3cf183b3298b840b9187a5039529f6b39/esmvalcore/preprocessor/_bias.py#L88-L89

esmvalcore/preprocessor/_multimodel.py
```diff
@@ -618,11 +618,16 @@ def _multiproduct_statistics(products,
         ignore_scalar_coords=ignore_scalar_coords,
     )
     statistics_products = set()
+
+    # Create copies of products to avoid recursively adding ancestors
+    copied_products = {p.copy_provenance() for p in products}
```
bouweandela, line 623: These copies are not needed, since `statistics_products` are never derived from other `statistics_products`, only from the original input `products` to the function.

esmvalcore/preprocessor/_bias.py
schlunma: I am aware of that, but thought this might be safer this way. Is it really a good idea to attach the original product as ancestor? Any potential changes (e.g., addition of further ancestors) to this ancestor down the road will also affect the original product, and that just feels wrong. I can already see us having a similar problem in a year when this has been forgotten...

How expensive is this copying anyway? We are not doing this a thousand times, so I don't think it has a big influence on performance.

bouweandela: > Is it really a good idea to attach the original product as ancestor? Any potential changes (e.g., addition of further ancestors) to this ancestor down the road will also affect the original product, and that just feels wrong.

Indeed that would be wrong and that's why the current implementation does not do that, `wasderivedfrom` already creates a copy: https://github.com/ESMValGroup/ESMValCore/blob/a4b5bc67d97f76b120403fc51a69536857740695/esmvalcore/_provenance.py#L241.

bouweandela (approved):
Looking good now, thanks @schlunma!

PR #1988, opened 2023-03-24
Introduce `Fix.session` to access `Session` in fixes

### LINKED ISSUE #1986: Use `auxiliary_data_dir` in fixes
It would be very helpful if the [`Fix`](https://github.com/ESMValGroup/ESMValCore/blob/d36db3b0a6aed86706800386074ff32b9aac5ad7/esmvalcore/cmor/_fixes/fix.py#L12) base class had access to the `auxiliary_data_dir` specified in the user-config file. An example use case would be the ICON fixes, which need a (vertical + horizontal) grid file to work properly.

I will implement this as an additional attribute `Fix.session` which includes `Fix.session['auxiliary_data_dir']`

schlunma:
## Description

To make fixes even more flexible I added a `session` attribute which includes the current `Session`. Example use case is loading additional data in the fix from the `auxiliary_data_dir`, which can now be accessed via `Fix.session['auxiliary_data_dir']`.

Closes #1986

Link to documentation: https://esmvaltool--1988.org.readthedocs.build/projects/ESMValCore/en/1988/develop/fixing_data.html#fix-structure

***

- [ ] [🛠][1] ~All [checks below this pull request](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-checks) were successful~ (Codacy issues cannot be fixed)

***


esmvalcore/cmor/_fixes/fix.py
```diff
@@ -5,29 +5,43 @@
 import inspect
 import tempfile
 from pathlib import Path
+from typing import TYPE_CHECKING, Optional
 
 from ..table import CMOR_TABLES
 
+if TYPE_CHECKING:
+    from ...config import Session
+    from ..table import VariableInfo
```
valeriupredoi, line 14: I'd really make this imports use the `esmvalcore` package than do the up, up and up dir

esmvalcore/cmor/_fixes/fix.py
```diff
@@ -202,11 +226,15 @@ def get_fixes(project, dataset, mip, short_name, extra_facets=None):
...
+                (name.lower(), value) for (name, value) in
+                inspect.getmembers(fixes_module, inspect.isclass)
+            )
             for fix_name in (short_name, mip.lower(), 'allvars'):
                 try:
-                    fixes.append(classes[fix_name](vardef, extra_facets))
+                    fixes.append(classes[fix_name](
+                        vardef, extra_facets=extra_facets, session=session
+                    ))
                 except KeyError:
```
valeriupredoi, line 238: note that `session` may fail with a key error here too but the exception is supposed to catch the `fix_name` key error only and not other stuff, maybe call session once before if not None to catch anything bad from it?

valeriupredoi (approved):
two small comments that don't warrant a rejection :grin: Cheers, bud :beer:

esmvalcore/cmor/_fixes/fix.py
schlunma: `session` is already initialized here, but I agree that it's probably better to convert the `try/except` to a `if fix_name in classes`. Got some weird error messages lately because of a similar problem.

esmvalcore/cmor/_fixes/fix.py
valeriupredoi: it shouldn't since circ imports issues happen when one module imports another module that imports the first module, but OK - I'd have changed all those .. and . imports with package-level imports but we'd prob have to go elsewhere do that too, outside the scope of this PR :+1:

bouweandela: Indeed we may want to consider using more absolute imports, though I like relative imports if they're importing from a module at the same level and are deeply nested. From [PEP8](https://peps.python.org/pep-0008/#imports):

Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path):
```python
import mypkg.sibling
from mypkg import sibling
from mypkg.sibling import example
```
However, explicit relative imports are an acceptable alternative to absolute imports, especially when dealing with complex package layouts where using absolute imports would be unnecessarily verbose:
```python
from . import sibling
from .sibling import example
```

PR #1989, opened 2023-03-27
Merge v2.8.x into main

valeriupredoi:
## Description

This is usual bookkeeping to merge the release branch into the main branch after the release. Do we still do that or we just keep the old release branches to die on their own, forgotten and marginalized? :grin: @ESMValGroup/technical-lead-development-team 

***


valeriupredoi:
Gents, this is to keep the git history have a record of all the stuff we did in the release branch, but may not get to the main branch. It's bookkeeping only, not mandatory

schlunma:
If I understand [`setuptools-scm`](https://github.com/pypa/setuptools_scm/#default-versioning-scheme) correctly, we need the [commit](https://github.com/ESMValGroup/ESMValCore/commit/31ff2566084f69ca7bd7d25ef8e00f7bf73493b5) associated with the `v2.8.0` tag in the main branch to get a `v2.9.0dev***` for the development installation. This is currently **not** the case:

```bash
> git branch -r --contains 31ff2566084f69ca7bd7d25ef8e00f7bf73493b5
  origin/v2.8.x
```

Thus, I think merging (or some other way to get this commit into main) **is** mandatory. Not sure if a PR with a squashed commit works though.

valeriupredoi:
@schlunma very good point there! Indeed I was *suspecting* (but never bothered checking yesterday, was planning on doing that today) that we need the tagged branch in so we up ourselves from 2.8dev to 2.9dev = also very good point, this I knew and should have yapped about it here - we should not squash the commits when we merge this branch - just a simple merge :beer:

valeriupredoi:
yes that `31ff` commit is being picked up by Readthedocs too as the `stable` tag

remi-kazeroni:
I don't have a strong opinion on this but I would recommend to decide rather soon here. Otherwise, there's a risk that the 2 branches diverge in a way that making bugfix releases becomes tricky... I think the main point here is to get the commit containing the tag back into main and get the version number of dev installation correct. I guess one drawback of this approach is that our commit history may not look so clean...

valeriupredoi:
foremost the commit history needs to be complete, Remi - if it's not clean, it's not a problem - we just need to keep track of what we're doing via the history - that was my first take on this merge, Manu's point about the tag commit is a bonus, and a well needed one too - so I would merge without squashing the sooner the better :+1:

remi-kazeroni:
> Thanks! I let someone else merge without squashing 👍

which seems not enabled for this repository...

valeriupredoi:
@ESMValGroup/technical-lead-development-team pls one of yous merge this **without squashing** please :beer: We need an issue open asap to document these steps in our release dox, as Remi pointed out in one of his comments

valeriupredoi:
> > Thanks! I let someone else merge without squashing +1
> 
> which seems not enabled for this repository...

indeed! I remember I merged without squashing for 2.7 - did we change the merge options in the meantime?

valeriupredoi:
OK I've just allowed rebasing - I think we need that always, side by side with squashing, but I'll let @ESMValGroup/technical-lead-development-team decide

zklaus:
Hm. I would think we want to

- Keep the `v2.8.x` branch around
- Merge it into `main`
- Not have `main` merged into `v2.8.x`. Why did we do that?

PR #2046, opened 2023-05-19
Respect `ignore_warnings` settings in `Dataset.load`

### LINKED ISSUE #2045: `ignore_warnings` in cfg-developer.yml is not considered anymore
The `ignore_warnings` key ([example](https://github.com/ESMValGroup/ESMValCore/blob/8b6be3c3e994cae1609d8acc1cc0a7007a4d697d/esmvalcore/config-developer.yml#L138-L143)) in the `config-developer.yml` file is not considered anymore, which leads to a cluttered log when processing EMAC files.

The reason for this is that the new `Dataset.load` method does not pass this keyword argument to the `load` function:

https://github.com/ESMValGroup/ESMValCore/blob/8b6be3c3e994cae1609d8acc1cc0a7007a4d697d/esmvalcore/dataset.py#L731

schlunma:
## Description

This PR fixes an issue where ignored warnings defined in the config-developer.yml file are ignored by `Dataset.load`.

Closes #2045

Link to documentation: https://esmvaltool--2046.org.readthedocs.build/projects/ESMValCore/en/2046/api/esmvalcore.dataset.html#esmvalcore.dataset.Dataset.load

***

- [x] ~~[🛠][1] All [checks below this pull request](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-checks) were successful~~ Codacy issue cannot be fixed

***


esmvalcore/_recipe/recipe.py
```diff
@@ -212,9 +212,6 @@ def _get_default_settings(dataset):
 
     settings = {}
 
-    # Configure (deprecated, remove for v2.10.0) load callback
```
schlunma, line 215: I *think* this can be safely removed, since the `load` step is handled in `Dataset.load`. For example, `fix_metadata` settings are also **not** considered here.

schlunma: NVM, it is used...will revert

esmvalcore/dataset.py
```diff
@@ -663,9 +663,15 @@ def files(self) -> Sequence[File]:
     def files(self, value):
         self._files = value
 
-    def load(self) -> Cube:
+    def load(self, **kwargs) -> Cube:
```
bouweandela, line 666: This method is intended to be an easy-to-use method that 'just' loads the data. It calls many preprocessor functions, not just `esmvalcore.preprocessor.load`. Therefore I foresee problems later on when someone decides they also need to pass on keyword arguments to other functions and then there may be a need to disentangle those. Rather than getting into that, it would be easier to use and maintain if the `Datset.load` method just understands how the tool has been configured in config-developer (it has access to that information through the `session` object), figures out what the correct parameters to `esmvalcore.preprocessor.load` are and passes those in.

bouweandela (changes requested):
Thanks for looking into this issue @schlunma!

esmvalcore/dataset.py
schlunma: I had that in the beginning, but I thought it's cleaner that way since `callback` is also passed via `settings`. Now we will get two different ways of specifying kwargs for `load`. Anyway, I will change it back (`callback` will be removed in v2.10.0 anyway).

(As far as I understand, `session` does not include information about the config-developer file, only the config-user file. But it's still super easy to access that information).

schlunma: Done

bouweandela (approved):

Issue #2042, opened 2023-05-19
Add support for delayed saves


bouweandela:
Since [iris 3.6](https://scitools-iris.readthedocs.io/en/stable/whatsnew/3.6.html), iris supports delayed saves. This is a great new feature that allows:
- checking that a preprocessor run will succeed before doing any data crunching, saving (compute) time if there is something wrong with the data
- faster preprocessor runs because
  - intermediate results can be reused to write multiple output files (e.g. regrid only once for multimodel statistics)
  - I suspect this will allow to better overlap computations with writing/reading data, but I did not measure it so far.

See #1714 for an example implementation.

bouweandela:
Dropping this from the v2.10 milestone because there is no more time. Some funding has been allocated in ESiWACE3, so there may be time to work on this for v2.11 or v2.12.

Issue #2041, opened 2023-05-19
Add support for running tasks on a Dask cluster


bouweandela:
Since [iris 3.6](https://scitools-iris.readthedocs.io/en/stable/whatsnew/3.6.html), it is possible to use [Dask distributed](https://distributed.dask.org/en/stable/) with iris. This is a great new feature that makes it possible to implement distributed computing (e.g. running a single recipe on multiple cluster nodes) in ESMValCore and ESMValTool. See #1714 for an example implementation. I would like to add this feature.

bouweandela:
After gaining some more experience with Dask I'm not sure if this is a good idea because
- diagnostic tasks are likely to use more memory than workers have
- for preprocessing tasks, metadata handling may also exceed the amount of memory that is available on a Dask worker, though this could probably be fixed
- for preprocessing tasks a more fine-grained approach would parallelize much better

Issue #1193, opened 2021-06-24
Memory issues despite lazy operations


schlunma:
Right now, I'm trying to analyze 25 years of monthly ERA5 data (variable `ta`). Since this dataset is 4D (time, height, lat, lon), the file sizes are pretty large. After concantenation, the cube's shape is `(300, 37, 721, 1440)`, which corresponds to about 86 GiB assuming `float64`.

When using the following recipe to extract the zonal means

<details>

 ```yaml
# ESMValTool
---
documentation:
  description: Test ERA5.

  authors:
    - schlund_manuel

  references:
    - acknow_project


preprocessors:

  mean:
    zonal_statistics:
      operator: mean


diagnostics:

  ta_obs:
    variables:
      ta:
        preprocessor: mean
        mip: Amon
        additional_datasets:
          - {dataset: ERA5, project: native6, type: reanaly, version: v1, tier: 3, start_year: 1990, end_year: 2014}
    scripts:
      null
 ```

</details>

on a node with 64 GiB memory, the process hangs indefinitely after reaching

```
2021-06-24 11:49:23,750 UTC [24261] INFO    Starting task ta_obs/ta in process [24261]
2021-06-24 11:49:23,843 UTC [24232] INFO    Progress: 1 tasks running, 0 tasks waiting for ancestors, 0/1 done
```

I'm fairly sure this is due to memory issues. As the following memory samples (for every second) show, the memory slowly ramps up to maximum memory and then suddenly drops. I guess somehow the OS shuts down the process after filling up the entire RAM.

<details>

 ```
             total       used       free     shared    buffers     cached
Mem:           62G       2.7G        59G        80K       6.5M       653M
Mem:           62G       9.2G        53G        80K       6.5M       1.7G
Mem:           62G       9.4G        53G        80K       6.5M       1.8G
Mem:           62G        16G        46G        80K       6.5M       3.0G
Mem:           62G        15G        47G        80K       6.5M       3.0G
Mem:           62G        22G        39G        80K       6.5M       4.2G
Mem:           62G        22G        40G        80K       6.5M       4.2G
Mem:           62G        29G        32G        80K       6.5M       5.4G
Mem:           62G        28G        34G        80K       6.5M       5.5G
Mem:           62G        36G        26G        80K       6.5M       6.6G
Mem:           62G        35G        27G        80K       6.5M       6.8G
Mem:           62G        42G        20G        80K       6.5M       7.9G
[...]

Issue #47, opened 2019-02-26
Collapsing cubes with weights kwarg very inefficient (full realization of data, twice)


valeriupredoi:
two used cases from running a recipe and profiling it via debug and resource file:

```
First instance example
-----------------------
2019-02-26 14:03:40.811397      7250.1  7576.5  108     2.7     0       0.606   0.006
2019-02-26 14:03:41.897243      7251.2  7577.7  108     3.8     0       0.606   0.006
2019-02-26 14:03:42.984943      7252.3  7578.9  108     5.8     0       0.606   0.006
2019-02-26 14:03:44.071390      7253.4  7580.0  108     6.4     0       0.606   0.006
2019-02-26 14:03:45.156149      7254.5  7581.2  109     7.9     0       0.606   0.006
2019-02-26 14:03:46.243115      7255.5  7582.4  108     9.4     0       0.606   0.006
2019-02-26 14:03:47.341183      7256.6  7583.6  109     11.1    1       0.606   0.006
2019-02-26 14:03:48.465922      7257.8  7584.8  111     11.7    1       0.606   0.006
2019-02-26 14:03:49.567116      7258.9  7586.0  110     13.2    1       0.606   0.006
2019-02-26 14:03:50.656928      7260.0  7587.1  96      0.6     0       0.606   0.007

2019-02-26 14:03:41,277 UTC [16489] DEBUG   esmvaltool.preprocessor:197 Running preprocessor step average_region
2019-02-26 14:03:41,280 UTC [16489] DEBUG   esmvaltool.preprocessor:187 Running area_average(precipitation_flux / (kg m-2 s-1)   (time: 1740; latitude: 360; longitude: 720)
2019-02-26 14:03:41,284 UTC [16489] INFO    esmvaltool.preprocessor._area_pp:179 Calculated grid area:(1740, 360, 720)
2019-02-26 14:03:50,494 UTC [16489] DEBUG   esmvaltool.preprocessor:197 Running preprocessor step cmor_check_data


Second instance example
-----------------------
2019-02-26 14:36:54.558816      9243.9  9778.3  108     12.4    1       0.708   0.007
2019-02-26 14:36:55.667920      9245.0  9779.5  111     13.5    1       0.708   0.007
2019-02-26 14:36:58.053553      9247.4  9781.9  100     0.5     0       0.708   0.007

[...]

PR #2049, opened 2023-05-19
Add support for configuring Dask distributed

### LINKED ISSUE #2040: Add support for configuring Dask distributed
Since [iris 3.6](https://scitools-iris.readthedocs.io/en/stable/whatsnew/3.6.html), it is possible to use [Dask distributed](https://distributed.dask.org/en/stable/) with iris. This is a great new feature that will allow for better memory handling and distributed computing. See #1714 for an example implementation. However, it does require some extra configuration. 

My proposal would be to allow users to specify the arguments to [distributed.Client](https://distributed.dask.org/en/stable/api.html#client) and to the associated cluster, e.g. [distributed.LocalCluster](https://distributed.dask.org/en/stable/api.html#distributed.LocalCluster) or [dask_jobqueue.SLURMCluster](https://jobqueue.dask.org/en/latest/generated/dask_jobqueue.SLURMCluster.html#dask_jobqueue.SLURMCluster) in this configuration. This could either be added under a new key in config-user.yml or in a new configuration file in the `~/.esmvaltool` directory:

Add to the user configuration file
----------------------------------

We could add these new options to `config-user.yml` under a new key `dask`, e.g.

Example config-user.yml settings for running locally using a LocalCluster:
```yaml
dask:
  cluster:
    type: distributed.LocalCluster
```
Example settings for using an externally managed cluster (e.g. set it up from a Jupyter notebook)
```yaml
dask:
  client:
    address: tcp://127.0.0.1:45695
```
Example settings for running on Levante:
```yaml
dask:
  client: {}
  cluster:
    type: dask_jobqueue.SLURMCluster
    queue: interactive
    account: bk1088
    cores: 8
    memory: 16GiB
    local_directory: "/work/bd0854/b381141/dask-tmp"
    n_workers: 2
```

New configuration file
----------------------

Or, we could add the new configuration in a separate file, e.g. called `~/.esmvaltool/dask.yml` or `~/.esmvaltool/dask-distributed.yml`.

Example config-user.yml settings for running locally using a LocalCluster:
```yaml
cluster:
  type: distributed.LocalCluster
```
Example settings for using an externally managed cluster (e.g. set it up from a Jupyter notebook)
```yaml
client:
  address: tcp://127.0.0.1:45695
```
Example settings for running on Levante:
```yaml
client: {}
cluster:
  type: dask_jobqueue.SLURMCluster
  queue: interactive
  account: bk1088
  cores: 8
  memory: 16GiB
  local_directory: "/work/bd0854/b381141/dask-tmp"
  n_workers: 2
```

@ESMValGroup/esmvaltool-coreteam Does anyone have an opinion on what the best approach is here? A new file or add to config-user.yml?

bouweandela:
## Description

Closes #2040

Link to documentation:

***

***

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -10,6 +10,7 @@ dependencies:
   - cftime
   - compilers
   - dask
+  - distributed
   - esgf-pyclient>=0.3.1
   - esmpy!=8.1.0
   - filelock
@@ -18,7 +19,7 @@ dependencies:
   - geopy
   - humanfriendly
   - importlib_resources
-  - iris>=3.4.0
+  - iris>=3.6.0
   - iris-esmf-regrid >=0.6.0  # to work with latest esmpy
   - isodate
   - jinja2
@@ -72,3 +73,6 @@ dependencies:
   # Not on conda forge - vprof
   - yamllint
   - yapf
+  - pip
+  - pip:
+      - ESMValTool_sample_data==0.0.3
--- a/esmvalcore/config/_dask.py
+++ b/esmvalcore/config/_dask.py
@@ -1,0 +1,64 @@
+import contextlib
+import importlib
+import logging
+from pathlib import Path
+
+import yaml
+from distributed import Client
+
+logger = logging.getLogger(__name__)
+
+CONFIG_FILE = Path.home() / '.esmvaltool' / 'dask.yml'
+
+
+@contextlib.contextmanager
+def get_distributed_client():
+    """Get a Dask distributed client."""
+    dask_args = {}
+    if CONFIG_FILE.exists():
+        config = yaml.safe_load(CONFIG_FILE.read_text(encoding='utf-8'))
+        if config is not None:
+            dask_args = config
+
+    client_args = dask_args.get('client') or {}
+    cluster_args = dask_args.get('cluster') or {}
+
+    # Start a cluster, if requested
+    if 'address' in client_args:
+        # Use an externally managed cluster.
+        cluster = None
+        if cluster_args:
+            logger.warning(
+                "Not using Dask 'cluster' settings from %s because a cluster "
+                "'address' is already provided in 'client'.", CONFIG_FILE)
+    elif cluster_args:
+        # Start cluster.
+        cluster_type = cluster_args.pop(
+            'type',
+            'distributed.LocalCluster',
+        )
+        cluster_module_name, cluster_cls_name = cluster_type.rsplit('.', 1)
+        cluster_module = importlib.import_module(cluster_module_name)
+        cluster_cls = getattr(cluster_module, cluster_cls_name)
+        cluster = cluster_cls(**cluster_args)
+        client_args['address'] = cluster.scheduler_address
+    else:
+        # No cluster configured, use Dask basic scheduler, or a LocalCluster
+        # managed through Client.
+        cluster = None
+
+    # Start a client, if requested
+    if dask_args:
+        client = Client(**client_args)
+        logger.info(f"Dask dashboard: {client.dashboard_link}")
+    else:
+        logger.info("Using the Dask basic scheduler.")
+        client = None
+
+    try:
+        yield client
+    finally:
+        if client is not None:
+            client.close()
+        if cluster is not None:
+            cluster.close()
--- a/setup.py
+++ b/setup.py
@@ -28,9 +28,8 @@ REQUIREMENTS = {
     # Use with pip install . to install from source
     'install': [
         'cartopy',
-        # see https://github.com/SciTools/cf-units/issues/218
         'cf-units',
-        'dask[array]',
+        'dask[array,distributed]',
         'esgf-pyclient>=0.3.1',
         'esmf-regrid',
         'esmpy!=8.1.0',
@@ -56,8 +55,8 @@ REQUIREMENTS = {
         'pyyaml',
         'requests',
         'scipy>=1.6',
-        'scitools-iris>=3.4.0',
-        'shapely[vectorized]',
+        'scitools-iris>=3.6.0',
+        'shapely',
         'stratify',
         'yamale',
     ],
--- a/esmvalcore/_task.py
+++ b/esmvalcore/_task.py
@@ -19,9 +19,11 @@ from typing import Optional
 
 import psutil
 import yaml
+from distributed import Client
 
 from ._citation import _write_citation_files
 from ._provenance import TrackedFile, get_task_provenance
+from .config._dask import get_distributed_client
 from .config._diagnostics import DIAGNOSTICS, TAGS
 
 
@@ -191,7 +193,9 @@ def write_ncl_settings(settings, filename, mode='wt'):
     lines = []
 
     # ignore some settings for NCL diagnostic
-    ignore_settings = ['profile_diagnostic', ]
+    ignore_settings = [
+        'profile_diagnostic',
+    ]
     for sett in ignore_settings:
         settings_copy = dict(settings)
         if 'diag_script_info' not in settings_copy:
@@ -414,7 +418,9 @@ class DiagnosticTask(BaseTask):
         run_dir.mkdir(parents=True, exist_ok=True)
 
         # ignore some settings for diagnostic
-        ignore_settings = ['profile_diagnostic', ]
+        ignore_settings = [
+            'profile_diagnostic',
+        ]
         for sett in ignore_settings:
             settings_copy = dict(self.settings)
             settings_copy.pop(sett, None)
@@ -718,10 +724,22 @@ class TaskSet(set):
         max_parallel_tasks : int
             Number of processes to run. If `1`, run the tasks sequentially.
         """
-        if max_parallel_tasks == 1:
-            self._run_sequential()
-        else:
-            self._run_parallel(max_parallel_tasks)
+        with get_distributed_client() as client:
+            if client is None:
+                address = None
+            else:
+                address = client.scheduler.address
+                for task in self.flatten():
+                    if (isinstance(task, DiagnosticTask)
+                            and Path(task.script).suffix.lower() == '.py'):
+                        # Only insert the scheduler address if running a
+                        # Python script.
+                        task.settings['scheduler_address'] = address
+
+            if max_parallel_tasks == 1:
+                self._run_sequential()
+            else:
+                self._run_parallel(address, max_parallel_tasks)
 
     def _run_sequential(self) -> None:
         """Run tasks sequentially."""
@@ -732,7 +750,7 @@ class TaskSet(set):
         for task in sorted(tasks, key=lambda t: t.priority):
             task.run()
 
-    def _run_parallel(self, max_parallel_tasks=None):
+    def _run_parallel(self, scheduler_address, max_parallel_tasks):
         """Run tasks in parallel."""
         scheduled = self.flatten()
         running = {}
@@ -757,7 +775,8 @@ class TaskSet(set):
                     if len(running) >= max_parallel_tasks:
                         break
                     if all(done(t) for t in task.ancestors):
-                        future = pool.apply_async(_run_task, [task])
+                        future = pool.apply_async(_run_task,
+                                                  [task, scheduler_address])
                         running[task] = future
                         scheduled.remove(task)
 
@@ -790,7 +809,14 @@ def _copy_results(task, future):
     task.output_files, task.products = future.get()
 
 
-def _run_task(task):
+def _run_task(task, scheduler_address):
     """Run task and return the result."""
-    output_files = task.run()
+    if scheduler_address is None:
+        client = contextlib.nullcontext()
+    else:
+        client = Client(scheduler_address)
+
+    with client:
+        output_files = task.run()
+
     return output_files, task.products


doc/quickstart/configure.rst
```diff
@@ -199,6 +199,99 @@ the user.
...
+Lazy data is represented by `dask arrays <https://docs.dask.org/en/stable/array.html>`_.
+Dask arrays consist of many small
+`numpy arrays <https://numpy.org/doc/stable/user/absolute_beginners.html#what-is-an-array>`_
+(called chunks) and if possible, computations are run on those small arrays in
+parallel.
+In order to figure out what needs to be computed when, Dask makes use of a
+'`scheduler <https://docs.dask.org/en/stable/scheduling.html>`_'.
+The default scheduler in Dask is rather basic, so it can only run on a single
+computer and it may not always find the optimal task scheduling solution,
+resulting in excessive memory use when using e.g. the
```
valeriupredoi, line 220:
```suggestion
resulting in excessive memory use when running an already memory-intensive task like e.g. the
```

doc/quickstart/configure.rst
```diff
@@ -199,6 +199,99 @@ the user.
...
+:func:`esmvalcore.preprocessor.multi_model_statistics` preprocessor function.
+Therefore it is recommended that you take a moment to configure the
+`Dask distributed <https://distributed.dask.org>`_ scheduler.
+A Dask scheduler and the 'workers' running the actual computations, are
+together called a Dask 'cluster'.
```
valeriupredoi, line 225:
```suggestion
collectively called a 'Dask cluster'.
```

doc/quickstart/configure.rst
```diff
@@ -199,6 +199,99 @@ the user.
...
+
+Create a Dask distruted cluster on the computer running ESMValCore using
+all available resources:
+
+.. code:: yaml
+
+  cluster:
+    type: distributed.LocalCluster
+
+this should work well for most personal computers.
```
valeriupredoi, line 245:
```suggestion
this should work well for most personal computers. Note that, if running this configuration on a shared node of an HPC cluster, Dask will try and use as many resources it can find available, and this may lead to overcrowding the node by a single user (you)!
```

valeriupredoi: or, maybe encase this in a .. note::

environment.yml
```diff
@@ -18,7 +19,7 @@ dependencies:
   - geopy
   - humanfriendly
   - importlib_resources
-  - iris>=3.4.0
+  - iris>=3.6.0
```
valeriupredoi, line 22: this may conflict with Tool's current environment (the Py311 env that @remi-kazeroni is using now) - let's see, I had issues with such a pin yesterday

environment.yml
```diff
@@ -72,3 +73,6 @@ dependencies:
   # Not on conda forge - vprof
   - yamllint
   - yapf
+  - pip
+  - pip:
+      - ESMValTool_sample_data==0.0.3
```
valeriupredoi, line 78: aghh I really don't like this - can we not make it a conda forge package?

valeriupredoi (changes requested):
small bits and bobs before I go into the Python scripts, bud :beer:

environment.yml
bouweandela: It's only for testing purposes, I'm not sure if we should bother the busy conda-forge peeps with it.

doc/quickstart/configure.rst
bouweandela: The problem isn't so much that it's memory-intensive, but that the task graph becomes too complicated for the built-in scheduler.

environment.yml
valeriupredoi: hope so - of course, this pin is essential here :grin:

environment.yml
valeriupredoi: yeah but this type of mixed conda-pip conda environment file is a bit of a no-no, can we not have it in `setup.py` and post-install it with `pip`? This will not get into the conda package anyway since you're not allowed mixed conda-pip stuff in the conda recipe

bouweandela: Why is that better?

valeriupredoi: plus, it falls in the same category with vprof, conda renegade

bouweandela: Yeah, `vprof`, that's something else to think about (is it still up to date and the best choice?). Anyway, I'll remove those lines again if they bother you, they're quite irrelevant anyway.

esmvalcore/config/_dask.py
```diff
@@ -0,0 +1,65 @@
...
+import importlib
+import logging
+from pathlib import Path
+
+import yaml
+from distributed import Client
+
+logger = logging.getLogger(__name__)
+
+CONFIG_FILE = Path.home() / '.esmvaltool' / 'dask.yml'
```
valeriupredoi, line 12: could you add support for central installation where this file would sit elsewhere and would have a certain configuration, prob site-specific? Not here, but keep it in mind (issue open etc), so we can decide how to do it

esmvalcore/config/_dask.py
```diff
@@ -0,0 +1,65 @@
...
+                "Not using Dask 'cluster' settings from %s because a cluster "
+                "'address' is already provided in 'client'.", CONFIG_FILE)
+    elif cluster_args:
+        # Start cluster.
+        cluster_type = cluster_args.pop(
+            'type',
+            'distributed.LocalCluster',
+        )
+        cluster_module_name, cluster_cls_name = cluster_type.rsplit('.', 1)
+        cluster_module = importlib.import_module(cluster_module_name)
```
valeriupredoi, line 42: what could be `cluster_module_name` (sorry, not familiar with this myself) - would it be worth encasing it in a try/except and reverting to a default at except (if any default exists? - again, me being a n00b here)

setup.py
```diff
@@ -56,8 +55,8 @@
         'pyyaml',
         'requests',
         'scipy>=1.6',
-        'scitools-iris>=3.4.0',
-        'shapely[vectorized]',
+        'scitools-iris>=3.6.0',
+        'shapely',
         'stratify',
         'yamale',
     ],
```
valeriupredoi, line 62: did you forget to add the pip-installable ESMValTool sample data?

valeriupredoi (changes requested):
couple more comments, bud. test_dask.py to follow, then we done here :grin:

esmvalcore/config/_dask.py
bouweandela: The default is defined on line 38 above, does that help?

setup.py
bouweandela: I don't quite understand. Could you explain in more detail what you mean with this comment?

esmvalcore/config/_dask.py
valeriupredoi: ah Local - cool, sorry, missed that

setup.py
valeriupredoi: the sample data PyPI package that you removed from environmen.yml, man

doc/quickstart/configure.rst
```diff
@@ -199,6 +199,106 @@ the user.
...
+``~/.esmvaltool/dask.yml``, where ``~`` is short for your home directory.
+In this file, under the ``client`` keyword, the arguments to
+:obj:`distributed.Client` can be provided.
+Under the ``cluster`` keyword, the type of cluster (e.g.
+:obj:`distributed.LocalCluster`), as well as any arguments required to start
+the cluster can be provided.
+
+Below are some example configurations:
+
+Create a Dask distruted cluster on the computer running ESMValCore using
```
remi-kazeroni, line 237:
```suggestion
Create a Dask distributed cluster on the computer running ESMValCore using
```

doc/quickstart/configure.rst
```diff
@@ -199,6 +199,106 @@ the user.
...
+this should work well for most personal computers.
+
+.. note::
+
+   Note that, if running this configuration on a shared node of an HPC cluster,
+   Dask will try and use as many resources it can find available, and this may
+   lead to overcrowding the node by a single user (you)!
+
+
+Create a Dask distruted cluster on the computer running ESMValCore, with
```
remi-kazeroni, line 254:
```suggestion
Create a Dask distributed cluster on the computer running ESMValCore, with
```

remi-kazeroni (commented):
Thanks a lot @bouweandela! Just taking a look at the well written Docs to get a better idea of these new developments.

doc/quickstart/configure.rst
valeriupredoi: "distruted" is a perfectly fine word, look it up in Oxford Dictionary :rofl:

schlunma:
I'd like to test this with a couple of recipes, please do not merge just yet :grin:

valeriupredoi:
> I'd like to test this with a couple of recipes, please do not merge just yet grin

good call, Manu! :beer:

doc/quickstart/configure.rst
```diff
@@ -199,6 +199,106 @@ the user.
...
+
+  cluster:
+    type: distributed.LocalCluster
+    n_workers: 4
+    memory_limit: 2 GiB
+
+Create a Dask distributed cluster on the
+`Levante <https://docs.dkrz.de/doc/levante/running-jobs/index.html>`_
+supercomputer using the `Dask-Jobqueue <https://jobqueue.dask.org/en/latest/>`_
+package:
```
schlunma, line 267: It would be nice to mention that this needs to be installed by the user (e.g., `mamba install dask-jobqueue`) because it's not part of our environment.

schlunma (commented):
Hi @bouweandela, I just tested this thoroughly - this is really amazing stuff and works super well! :rocket: For the first time ever I was able to run ESMValTool on multiple nodes! In addition, monitoring an ESMValTool run with a dask dashboard is also super convenient. See here a nice visualization of our example recipe:

esmvaltool_dask

I have some general comments/questions:

- How does this interact with our current `multiprocessing`-based parallelization? I tested this with different `--max-parallel-tasks` options and all seemed to work well, but I do not quite understand how. Does every process use it's own cluster? Or do they share it? Is this safe?
- As far as I understand the only way to use multiple nodes in a `SLURMCluster` is to use the `cluster.scale(jobs=n_jobs)` method, see [here](https://jobqueue.dask.org/en/latest/generated/dask_jobqueue.SLURMCluster.html). Would it make sense to add a `scale` option to the configuration file where one could add keyword arguments for `scale`?
- When running this, I always get `UserWarning: Sending large graph of size 56.83 MiB. This may cause some slowdown. Consider scattering data ahead of time and using futures.`, even for the example recipe. Not sure if this is/will be a problem. Not something we need to tackle here though; just wanted to document it.

Thanks!! :+1:

valeriupredoi:
That's wonderful to see, Manu, awesome you tested, and of course, awesome work by Bouwe! I'll let Bouwe answer your q's (although I know the answer for the first q from code review hehe), but about the jobqueue dep - we should probably just add it in our env, I'd reckon whatever's Dask-related stuff should be tweaked at a minimum by the user. About data scattering - I've done it before, and have seen marginal performance improvement, but I used fairly small data sizes, so it could be better for busloads of data, as we have 😁

bouweandela:
Thanks for testing @schlunma!

> How does this interact with our current multiprocessing-based parallelization? I tested this with different --max-parallel-tasks options and all seemed to work well, but I do not quite understand how. Does every process use it's own cluster? Or do they share it? Is this safe?

When no cluster is configured, every process will use it's own. This is what is causing memory errors at the moment. With the pull request, this is fixed (provided that a user sets up a cluster), by passing the scheduler address to the process running the task and making it connect to the cluster. You can see that here in the code:
https://github.com/ESMValGroup/ESMValCore/blob/98edcb1fd24b8871cceaa30f0dfb6cf9efb927ae/esmvalcore/_task.py#L774-L775
https://github.com/ESMValGroup/ESMValCore/blob/98edcb1fd24b8871cceaa30f0dfb6cf9efb927ae/esmvalcore/_task.py#L808-L818
The multiprocessing configuration is used to parallelize the metadata crunching (i.e. all the stuff with coordinates etc). In a future pull request, I intend to make it possible to also run that on a Dask cluster, see #2041.

> As far as I understand the only way to use multiple nodes in a SLURMCluster is to use the cluster.scale(jobs=n_jobs) method, see [here](https://jobqueue.dask.org/en/latest/generated/dask_jobqueue.SLURMCluster.html). Would it make sense to add a scale option to the configuration file where one could add keyword arguments for scale?

This can also be configured with the parameter `n_workers`:
https://jobqueue.dask.org/en/latest/generated/dask_jobqueue.SLURMCluster.html#dask-jobqueue-slurmcluster

> When running this, I always get UserWarning: Sending large graph of size 56.83 MiB. This may cause some slowdown. Consider scattering data ahead of time and using futures., even for the example recipe. Not sure if this is/will be a problem. Not something we need to tackle here though; just wanted to document it.

I suspect this warning occurs because some of the preprocessor functions are not lazy, e.g. `multi_model_statistics`.

doc/quickstart/configure.rst
bouweandela: I like @valeriupredoi's suggestion of just adding it to the dependencies. It doesn't have any dependencies that we do not already have and it's a very small Python package.

bouweandela:
@schlunma Did you also have a chance to test this with some recipes using a large amount of data? I would be interested in your results.

doc/quickstart/configure.rst
schlunma: Sounds good, that's even better!

sloosvel:
Hi, do I have time to try to test this in our infrastructure?

schlunma:
> When no cluster is configured, every process will use it's own. This is what is causing memory errors at the moment.

Could you elaborate on that? What do you mean by "at the moment"? The current `main`? Do you have an example for such a memory error?

>With the pull request, this is fixed (provided that a user sets up a cluster), by passing the scheduler address to the process running the task and making it connect to the cluster. The multiprocessing configuration is used to parallelize the metadata crunching (i.e. all the stuff with coordinates etc).

True, could have guessed that from the code :grin: I suppose the cluster is smart enough to handle input from different processes? Are there any recommendations for the ratio `max_parallel_tasks/n_workers`? E.g., does it make sense to use 16 parallel tasks with just 1 worker?

> In a future pull request, I intend to make it possible to also run that on a Dask cluster, see #2041.

Cool!

> This can also be configured with the parameter `n_workers`: https://jobqueue.dask.org/en/latest/generated/dask_jobqueue.SLURMCluster.html#dask-jobqueue-slurmcluster

I tried that with `n_workers=4`, but just got 1 node. I now used `n_workers=16` and got 4 nodes. So I guess the number of nodes is a combination of the different parameters.

> @schlunma Did you also have a chance to test this with some recipes using a large amount of data? I would be interested in your results.

Not yet, but I can try to run a heavy recipe with this :+1:

doc/quickstart/configure.rst
```diff
@@ -199,6 +199,106 @@ the user.
...
+
+.. code:: yaml
+
+  cluster:
+    type: dask_jobqueue.SLURMCluster
+    queue: interactive
+    account: bk1088
+    cores: 8
+    memory: 16GiB
```
sloosvel, line 276: Do you know if there is any way to avoid having to specify the memory requirements? 

I am not allowed to submit jobs with memory requirements in our infrastructure, because they are set by default depending on the type of node you are running with. So my job submissions fail because of that. 

However, the configuration seems to require `memory` to be specified. I am just wondering if there is any way to set it to null or something.

valeriupredoi:
hellow cool stuff! Very cool indeed @schlunma :woman_dancing: Here's some prelim opinions:

- lazy preprocs are the only ones that work as expected because this whole machinery is optimized for lazy stuff
- settings depend on the amount of data and computation needed - and resources used really do vary based on the cluster settings - afraid there is no best set of settings parameters (eg try half the data for your lazy regrid runs and you'll see that performance changes and other settings return the best balance)
- the concurrent Futures issue seems o be coming from very small scattered data packets that get created, then deleted then recreated, see eg https://github.com/dask/distributed/issues/4612 - it'd be good to first scatter the data (since it's not lazy) then do computations, since once scattered the data is smaller, and default scattering by the cluster may not happen anymore (that is bad, since the cluster doesn't really know how to do it properly)

Very cool, still :beer:

bouweandela:
> I suppose the cluster is smart enough to handle input from different processes? Are there any recommendations for the ratio max_parallel_tasks/n_workers?

Provided that all preprocessor functions are lazy, I would recommend setting `max_parallel_tasks` equal to the number of CPU cores in the machine, since coordinates are mostly realized, meaning this will use on average 1 CPU core per task. I suspect that in most cases it does not make sense to have `n_workers` larger than `max_parallel_tasks` until we have delayed saving mentioned in #2042, since a task can only save one file at a time and I would expect a worker to be able to process all data required for a single output file. Depending on what you're actually trying to compute of course.

valeriupredoi:
this [post](https://stackoverflow.com/questions/46837416/what-causes-dask-job-failure-with-cancellederror-exception) says the Cancelled error comes from some node/worker being idle for too long - interesting - so it's either too short a computation that deletes the key on the worker, but the Client still needs it/recreates it, or the thing's too long - it'd be worth examining the "timeout" options see https://docs.dask.org/en/stable/configuration.html - beats me which one though, used to be a `scheduler_timeout`

doc/quickstart/configure.rst
bouweandela: Done in [25dc5ce](https://github.com/ESMValGroup/ESMValCore/pull/2049/commits/25dc5ce6f7e39e8a9b02f0777bcdcd70b4f2328e)

esmvalcore/config/_dask.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@contextlib.contextmanager
+def get_distributed_client():
+    """Get a Dask distributed client."""
+    dask_args = {}
+    if CONFIG_FILE.exists():
+        config = yaml.safe_load(CONFIG_FILE.read_text(encoding='utf-8'))
+        if config is not None:
+            dask_args = config
```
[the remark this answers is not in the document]
bouweandela: Added in [25dc5ce](https://github.com/ESMValGroup/ESMValCore/pull/2049/commits/25dc5ce6f7e39e8a9b02f0777bcdcd70b4f2328e)

schlunma:
I now also tested the (fully lazy) recipe of https://github.com/ESMValGroup/ESMValCore/issues/1193 now (which used to **not** run back then):

<details>

 ```yaml
# ESMValTool
---
documentation:
  description: Test ERA5.

  authors:
    - schlund_manuel

  references:
    - acknow_project


preprocessors:

  mean:
    zonal_statistics:
      operator: mean


diagnostics:

  ta_obs:
    variables:
      ta:
        preprocessor: mean
        mip: Amon
        additional_datasets:
          - {dataset: ERA5, project: native6, type: reanaly, version: v1, tier: 3, start_year: 1990, end_year: 2014}
    scripts:
      null
 ```

</details>

The good new is: it is now running without any dask configuration, and I could also get a significant speed up (5:41 min -> 1:56 min) by using a `LocalCluster`.

Replacing the lazy `zonal_statistics` with the non-lazy `area_statistics` lead to the same error as already reported.

esmvalcore/config/_dask.py
```diff
@@ -0,0 +1,77 @@
...
+def check_distributed_config():
+    """Check the Dask distributed configuration."""
+    if not CONFIG_FILE.exists():
+        logger.warning(
+            "Using the Dask basic scheduler. This may lead to slow "
+            "computations and out-of-memory errors. See https://docs."
+            "esmvaltool.org/projects/ESMValCore/en/latest/quickstart/"
+            "configure.html#dask-distributed-configuration for information "
+            "on how to configure the Dask distributed scheduler."
+        )
```
schlunma, line 24: I am not too happy with this being a warning at the current stage: as shown by my tests, using a distributed scheduler can actually lead to recipes not running anymore. Thus, I would be very careful recommending this to users at the moment.

We should either phrase this more carefully (maybe add that this is an "experimental feature") or convert it to an info message. Once we are more confident with this we can change it back.

schlunma: Also: this does not raise a warning if the dask config file exists but is empty. Should we also consider this case?

bouweandela: I can change it back to INFO if @valeriupredoi agrees because he asked for a WARNING in https://github.com/ESMValGroup/ESMValCore/pull/2049#discussion_r1202438286. Note that not configuring the scheduler can also lead to recipes not running, so it rather depends on what you're trying to do what the best setting is, as also noted in the documentation.

valeriupredoi: yeah that was me before Manu#s actual testing. But alas, Bouwe is right - I'd keep it as a warning but add what Manu suggests - experimental feature with twitchy settings that depend on the actual run

valeriupredoi: > Also: this does not raise a warning if the dask config file exists but is empty. Should we also consider this case?

Won't that default to a LocalCluster etc?

schlunma: > I'd keep it as a warning but add what Manu suggests - experimental feature with twitchy settings that depend on the actual run

Fine for me!

> Won't that default to a LocalCluster etc?

No, this also results in the basic scheduler: https://github.com/ESMValGroup/ESMValCore/pull/2049/files#diff-b046a48e3366bf6517887e3c39fe7ba6f46c0833ac02fbbb9062fb23654b06bdR64-R69

schlunma (approved):
Thanks for all the changes @bouweandela! I love the changes to the documentation, with this I could get a recipe to run in 30s which previously took 6 mins!

Just ouf of curiosity: where did you get that `n_jobs = n_workers / processes` from? I searched for that yesterday but couldn't find anything.

Please make sure to adapt the warning as discussed in the corresponding comment. Will approve now since I will be traveling next week.

Thanks again, this is really cool! :rocket:

esmvalcore/config/_dask.py
valeriupredoi: aw. Nicht gut. That needs be communicated to the user methinks

valeriupredoi:
> where did you get that n_jobs = n_workers / processes from?

I think that's just Bouwe being sensible, as he usually is :grin: ie you'd ideally want one process per worker

valeriupredoi:
awesome testing @schlunma - cheers for that! I'll dig deeper into those errors you've been seeing once I start testing myself, sometime next week :+1:

sloosvel (approved):
Thanks @bouweandela ! I managed to create some SLURM clusters just fine with these changes. Regarding the configuration, is it possible to have multiple configurations in `dask.yml`? Not every recipe will require the same type of resources.

However,  I am running in a machine that has quite modest technical specifications, to put it nicely, and even running in a distributed way does not seem to help much in the performance and memory usage. 

Some issues that were found for instance:
-  The concatenation of multiple files seems to slow down
- The heaviest jobs all seem to die during the saving of the files.
- Even for jobs that finish successfully, I get errors in the middle of the debug log about workers dying, not being responsive and communications dying. Not sure where these are coming from.

It could very well be that I am not configuring the jobs properly though. 

It is true though that everything that is already lazy performs even faster. The lazy regridder went from taking 4 minutes to run in a high resolution experiment, to taking less than 2 minutes.

doc/quickstart/configure.rst
```diff
@@ -199,6 +199,145 @@ the user.
...
+
+  cluster:
+    type: dask_jobqueue.SLURMCluster
+    queue: shared
+    account: bk1088
+    cores: 8
+    memory: 7680MiB
+    processes: 2
+    interface: ib0
+    local_directory: "/work/bd0854/b381141/dask-tmp"
```
fnattino, line 294: The Dask config should accept environment variables, maybe useful to show this for the following?
```suggestion
    local_directory: "$SCRATCH"
```

fnattino (approved):
Very nice @bouweandela, looks great to me! Nice that you have both the cluster and the client config sections, so that the following also works as a 'shortcut' to create a local cluster:
```
client:
  n_workers: 2
  threads_per_worker: 4
```

bouweandela:
> Just ouf of curiosity: where did you get that `n_jobs = n_workers / processes` from? I searched for that yesterday but couldn't find anything.

@schlunma From the documentation of the `processes` argument:

> Cut the job up into this many processes.

i.e. `processes` is the number of workers launched by a single job.

bouweandela:
>  Regarding the configuration, is it possible to have multiple configurations in dask.yml? Not every recipe will require the same type of resources.

Good idea. Shall I add a warning that the current configuration file format is experimental and may change in the next release? We can then discuss how to best add this to the configuration at the SMHI workshop https://github.com/ESMValGroup/Community/discussions/98.

doc/quickstart/configure.rst
bouweandela: Thanks for reviewing @fnattino! I will add a note to the text, but not change the configuration example here because this [environmental variable is not provided on Levante](https://docs.dkrz.de/doc/levante/file-systems.html) so then the example wouldn't work.

bouweandela:
@schlunma and @sloosvel Thank you for running the tests! Really good to learn from your experiences. Regarding the [dying workers](https://distributed.dask.org/en/stable/killed.html): I suspect they are using more than the configured amount of memory and then killed, so giving them more memory may solve the problem, though in case the preprocessor functions you're using are not lazy that may not help.

Regarding the `area_statistics` preprocessor function: this is actually lazy (except for `median`), but there seems to be something odd going on. If I run the example recipe provided by @schlunma in https://github.com/ESMValGroup/ESMValCore/pull/2049#issuecomment-1563062806 with just one model and one year of data, I already get the warning that 1.5 GB of data is sent to the workers (i.e. the size of the `weights` array), even though the weights array is lazy. Maybe https://github.com/ESMValGroup/ESMValCore/issues/47 is applicable here.

esmvalcore/config/_dask.py
bouweandela: Done in https://github.com/ESMValGroup/ESMValCore/pull/2049/commits/80f5c62e9f58240789e43c3912c1d444de6ffd4a

bouweandela:
> The concatenation of multiple files seems to slow down

@sloosvel If this is a problem for your work, could you please open an issue and add an example recipe so we can look into it?

bouweandela:
@ESMValGroup/technical-lead-development-team I think this is ready to be merged. Could someone please do a final check and merge?

remi-kazeroni:
Thanks a lot @bouweandela for your work on this great extension! It should definitely one of the highlights of the next Core release! And thanks to everyone who reviewed this PR and tested the new Dask capabilities! Merging now 👍

PR #2061, opened 2023-05-24
Fix tas/tos CMIP6: FIO, KACE, MIROC, IITM

### LINKED ISSUE #2060: Dataset problem: tas/tos FIO-ESM-2-0, IITM-ESM, KACE-1-0, MIROC6
There are some models that require fixes in "mon"  tas and tos related to lat/lon rounding discrepancies between historical and ssp scenarios, and/or bound issues in lat/lon/time coordinates.

The models are: FIO-ESM-2-0, IITM-ESM, KACE-1-0, MIROC6

The issues have been identified in Levante's pool of CMIP6 data.

I'll open a PR with the fixes and tests

Cheers

Pep

pepcos:
There are some fixes needed for tas and/or tos in CMIP6 datasets: FIO-ESM-2-0, IITM-ESM, KACE-1-0 and MIROC6

FIO-ESM-2-0:
- OceanFixGrid
- Amon non-contiguous lat/lon bounds
- Omon rounding error in the 6th decimal

IITM-ESM:
- OceanFixGrid
- non-contiguous time bounds for "mon"

KACE-1-0:
- OceanFixGrid
- non-contiguous time bounds for "mon"

MIROC6:
- Some files have coordinates in float64 and some in float32

Closes #2060 

***

***


esmvalcore/cmor/_fixes/cmip6/fio_esm_2_0.py
```diff
@@ -0,0 +1,64 @@
...
+        return cubes
+
+
+class Amon(Fix):
+    """Fixes for Amon vars."""
+
+    def fix_metadata(self, cubes):
+        """Fix metadata.
+
+        FIO-ESM-2-0 ta data contains error in coordinate bounds.
```
sloosvel, line 38:
```suggestion
        FIO-ESM-2-0 Amon data contains error in coordinate bounds.
```
I guess this is for the whole realm?

esmvalcore/cmor/_fixes/cmip6/fio_esm_2_0.py
```diff
@@ -0,0 +1,64 @@
...
+            # Check both lat and lon coords and replace bounds if necessary
+            latitude = cube.coord("latitude")
+            if latitude.has_bounds():
+                if latitude.bounds[1:, 0] != latitude.bounds[:-1, 1]:
+                    latitude.bounds = None
+                    latitude.guess_bounds()
+                    iris.util.promote_aux_coord_to_dim_coord(cube, "latitude")
+
+            longitude = cube.coord("longitude")
+            if longitude.has_bounds():
```
sloosvel, line 59: I guess that if the coordinates do not have bounds they will be guessed anyway in the checks? In any case, I don't follow why you need to promote the coordinate afterwards, what is preventing it being a dimensional coordinate when the data is being loaded?

esmvalcore/cmor/_fixes/cmip6/miroc6.py
```diff
@@ -1,11 +1,43 @@
...
 
-
 Cli = ClFixHybridPressureCoord
 
-
 Clw = ClFixHybridPressureCoord
+
+
+class Tos(Fix):
+    """Fixes for ** all vars."""
```
sloosvel, line 15:
```suggestion
    """Fixes for tos."""
```

esmvalcore/cmor/_fixes/cmip6/fio_esm_2_0.py
pepcos: It was so that bounds that are badly described get fixed in the specific model fix. It is true that they would be fixed anyway later on. Should I remove the bound guessing from here and expect no bounds if they are badly defined?

Regarding the promotion, it has no reason really so I'll remove it

sloosvel: Nono, I was just wondering why you were only guessing the bounds in the case they were present but wrong, and doing nothing if they were not present. But since this is a fix in Amon, if some variable does not have bounds, they will be guessed anyway in the checks. So it's fine as it is.

sloosvel:
Thanks for addressing all the changes @pepcos ! I tested with the following recipe and it finished successfully:

```yaml
documentation:
  title: |
    bla
  authors: 
    - loosveldt-tomas_saskia
    - cos_josep
  description: |
    bla

datasets:
  - dataset: FIO-ESM-2-0
    ensemble: 'r1i1p1f1'
    exp: historical
    project: CMIP6
    timerange: '2014/2014'
  - dataset: IITM-ESM
    ensemble: 'r1i1p1f1'
    exp: historical
    project: CMIP6
    timerange: '2014/2014'

preprocessors:
    prep:
      regrid:
         target_grid: 1.5x1.5
         scheme: area_weighted 
      extract_season:
        season: "JJA"
      seasonal_statistics:
        operator: mean
        seasons: ["JJA"]

diagnostics:
  test:
    variables:
      tos:
        preprocessor: prep
        mip: Omon
        grid: gn
        additional_datasets:
          - dataset: KACE-1-0-G
            ensemble: 'r1i1p1f1'
            exp: historical
            project: CMIP6
            timerange: '2014/2014'
            grid: gr
      tas:
        preprocessor: prep
        mip: Amon
        grid: '*'
        additional_datasets:
          - dataset: MIROC6
            ensemble: 'r10i1p1f1'
            exp: [historical, ssp585]
            project: CMIP6
            timerange: '2014/2015'
    scripts: null
```

I'll approve all changes when all tests are green

sloosvel (approved):

esmvalcore/cmor/_fixes/cmip6/miroc6.py
```diff
@@ -1,11 +1,41 @@
...
+            Input cubes.
+
+        Returns
+        -------
+        iris.cube.CubeList
+        """
+        for cube in cubes:
+            latitude = cube.coord('latitude')
+            if latitude.bounds is not None:
+                latitude.bounds = latitude.bounds.astype(np.float32)
```
bouweandela, line 33: Is this large data? The convention is to use `float64` (https://github.com/ESMValGroup/ESMValCore/blob/2e7e3848b0a7ab1bd5485605eb0d151d3dbb8bab/esmvalcore/cmor/tables/cmip6/Tables/CMIP6_coordinate.json#L837) for coordinates.

sloosvel: Yes, we looked into that. The problem is that scenario data for this dataset comes with coordinates in float32 whereas the historical one is in float64, so the concatenation of the two experiments failed. 

We tried converting the scenario to float64 and the concatenation was impossible due to rounding errors. The only way we found to be able to work with both experiments was by setting less precision in the coordinates.

bouweandela: Yes, I can see that you implemented the rounding on line 34 and I think that's fine. But why are you choosing to convert the datatype fo float32 instead of float64?

pepcos: If it is converted to float64 the coordinates that are in 32 get some nonsense values at the last decimal places and it is not possible to round it afterwards so that they match with float64 values. Not sure why, though

bouweandela: It looks like you're only rounding the bounds and not the points, at least in the code I can see here. How about rounding the points to the same precision?

pepcos: Yes, we are aware, but there is no rounding needed for the points if we convert to float32.

Nevertheless, if we convert to float64, even if we round we get differences in the historical and ssp coordinates. Here is a debug, note the max and min values of the differences. (the top difference is between the bounds and the lower is the points)
Screenshot from 2023-05-30 16-38-45

In this case the fix is as follows (note that I added the points rounding just in case):
```python
class Tos(Fix):
    """Fixes for ** all vars."""

    def fix_metadata(self, cubes):
        """
        ***Fix latitude_bounds and longitude_bounds data type and round to 4 d.p.

        Parameters
        ----------
        cubes : iris.cube.CubeList
            Input cubes.

        Returns
        -------
        iris.cube.CubeList

        """
        for cube in cubes:
            latitude = cube.coord('latitude')
            if latitude.bounds is None:
                latitude.guess_bounds()
            latitude.bounds = latitude.bounds.astype(np.float64)
            latitude.bounds = np.round(latitude.bounds, 4)
            latitude.points = latitude.points.astype(np.float64)
            latitude.points = np.round(latitude.points, 4)
            longitude = cube.coord('longitude')
            if longitude.bounds is None:
                longitude.guess_bounds()
            longitude.bounds = longitude.bounds.astype(np.float64)
            longitude.bounds = np.round(longitude.bounds, 4)
            longitude.points = longitude.points.astype(np.float64)
            longitude.points = np.round(longitude.points, 4)
        return cubes
```

bouweandela: If just converting to `float32` already does the trick, how about doing something like:
```
latitude.points = latitude.points.astype(np.float32).astype(np.float64)
```
?

pepcos: I agree, this is a very sexy move.

Nevertheless, it invents the data from the 6th decimal in the points and bounds. it will probably not influence the results, but do you think we should allow this?

example input lat points:  [23.0211550, 26.3700972]
example outout lat points: [23.021154403686523, 26.370098114013672]

its in the meters scale so it shouldn't be a big problem

bouweandela: It's no worse then inventing zeros and what else can you do if the modellers don't provide the data? The more precise alternative would be add the correct values from a file that has them.

bouweandela:
@pepcos Thanks for fixing the coordinate data type!
@sloosvel There have been some changes since you previously reviewed this pull request, are you still happy with it? If yes, I will merge it.

sloosvel:
Sure, looks good to me!

PR #2070, opened 2023-05-29
Set `use_src_mask` parameter in regridding of datasets with grid discontinuities

### LINKED ISSUE #2014: Dataset problem: Nemo grid datasets cannot be regridded using iris-esmf
**Describe the dataset issue**
When trying to regrid ocean variables that are in a dataset that use the Nemo grid (CNRM, CMCC, EC-Earth, IPSL), the regridder in iris-esmf  fails due to the coordinates not being contiguous. In those particular datasets, removing the first grid point (which is always masked in ocean variables) seems to be enough in order to make the the coordinates contiguous and have the regridder work properly.

It would be good to have a fix that removes this point, even though it would mean that the data after fixing would have a different shape than the original data. I already worked on the fix, but it would be good to know if anyone has further comments.

**Assign the right people**
If you are already a member of the ESMValTool GitHub project, please assign Valeriu Predoi (valeriupredoi) and
Klaus Zimmermann (zklaus) to the issue. They will then check the issue raised and propagate it further to the
data model developers.

**Please attach**
  - Attach a text file with the output of `ncdump -h /path/to/file.nc`

Cheers :beer:

sloosvel:
## Description

This PR allows the regridding of some datasets that have grid discontiguities using the iris-esmf regridder. Those discontiguities tend to happen in lat and lon values where the data is masked, so setting `use_src_mask` to True is enough to be able to regrid them. 

It does not solve the problem that some regridded datasets come out of the area weighted regridder with a vertical blank line. 

But at least, many other datasets that could not use the iris regridder due to assertion erros in coord.is_contiguous() (see https://github.com/ESMValGroup/ESMValTool/issues/2661) , now pass the check and come out of the regridder looking nicely.

Closes #2014 

Link to documentation:

***

***


esmvalcore/preprocessor/_regrid.py
```diff
@@ -383,6 +383,18 @@ def extract_location(cube, location, scheme):
                          scheme)
 
 
+def _check_grid_discontiguities(cube, scheme):
+    """Check if there are grid discontiguities and set use_src_mask to True."""
+    try:
+        discontiguities = iris.util.find_discontiguities(cube)
+    except NotImplementedError:
```
valeriupredoi, line 390: can the `NotImplementedError` be tossed by iris based on its version (if so, can you pls add a comment stating starting what version of iris that function is implemented)?

valeriupredoi (approved):
nice and easy to rewvie, many thanks @sloosvel :beer: Please see a couple comments from me - minor. More importantly, do you need a certain version of iris-esmf-regrid for this to work or the >=0.6.0 current pin will do for you? (which will become 0.7.0 in 2.9 courtesy of @schlunma developments). Speaking of the :smiling_imp: - @schlunma would you mind have a look and merging this pls bud? :beer:

tests/unit/preprocessor/_regrid/test_regrid.py
```diff
@@ -262,5 +263,57 @@ def test_regrid_is_skipped_if_grids_are_the_same():
...
+    cube = iris.cube.Cube(
+        data,
+        aux_coords_and_dims=[(lat_coord, (0, 1)), (lon_coord, (0, 1))],
+    )
+
+    scheme = {}
+    scheme = _check_grid_discontiguities(cube, scheme)
+    assert scheme == {'use_src_mask': True}
+
+
```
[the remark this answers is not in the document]
sloosvel: I thought it was being tested in the test above, if there are no discontinuities the NotImplementedError just returns a scheme where the parameter is not added.

esmvalcore/preprocessor/_regrid.py
sloosvel: Okey, I will browse through iris to see when that was added

tests/unit/preprocessor/_regrid/test_regrid.py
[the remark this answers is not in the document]
valeriupredoi: ah think I missed that then :smiley:

esmvalcore/preprocessor/_regrid.py
valeriupredoi: thinking about it even more (yeh Monday mornings) it'd be nice to have a debug warning about it too, if you find the iris version then that'd be good be included in the warning message :+1:

bouweandela: We have pinned iris to the latest version, so this should be needed.

esmvalcore/preprocessor/_regrid.py
```diff
@@ -383,6 +383,18 @@ def extract_location(cube, location, scheme):
...
+        pass
+    else:
+        if np.isin(True, discontiguities):
```
bouweandela, line 393:
```suggestion
        if discontiguities.any():
```
would look a bit more descriptive to me, but the current implementation is also fine.

sloosvel:
> Would it make sense to add some documentation for this feature?

Good point but I cannot make it happen for today, it's a holiday here and I am having trouble connecting with my laptop. In any case `use_src_mask` can always be called from the recipe.

valeriupredoi:
and you're also having lunch :rofl: :pizza:

esmvalcore/preprocessor/_regrid.py
```diff
@@ -384,6 +384,18 @@ def extract_location(cube, location, scheme):
                          scheme)
 
 
+def _check_grid_discontiguities(cube, scheme):
+    """Check if there are grid discontiguities and set use_src_mask to True."""
```
bouweandela, line 388:
```suggestion
    """Check if there are grid discontiguities and set use_src_mask to True."""
    scheme = dict(scheme)
```
to avoid updating the input `scheme` dict? It looks like the intention of the function is to return a new `scheme` dict, but currently it modifies the input dict.

esmvalcore/preprocessor/_regrid.py
```diff
@@ -384,6 +384,18 @@ def extract_location(cube, location, scheme):
...
+    try:
+        discontiguities = iris.util.find_discontiguities(cube)
+    except NotImplementedError:
+        pass
+    else:
+        if discontiguities.any():
+            scheme['use_src_mask'] = True
```
bouweandela, line 395: Is it possible for users to provide the value of `use_src_mask` in the recipe? If yes, it would be good to at least log a message that we're changing their setting here.

bouweandela:
Thanks for adding the documentation. Would it make sense to add it to the `esmvalcore.preprocessor.regrid` as well, so it shows up in the API docs?

PR #2075, opened 2023-05-31
Unpin shapely

valeriupredoi:
## Description

We discussed starting from @zklaus [comment](https://github.com/ESMValGroup/ESMValTool/pull/3187#issuecomment-1566602182) and agreed that we should let shapely be modern again; we'd have to fix recipe_shapeselst.yml, as Klaus points out, but we better fix a recipe and have modern dependencies, than the other way round.

***

***


bouweandela:
Why do we need versions larger than 2?

valeriupredoi:
It's not a massively important pin, but 2.0+ has introduced a lot of changes, it's not that <2.0.0 would break stuff AFAIK though. @zklaus suggested it, and in Klaus I trust when it comes to these things 😁

bouweandela:
I just ran some tests by creating an environment with shapely <2 (I got 1.8.5) and all our unit tests pass, so I would really like to have some better explanation of why we need this before we consider merging this.

bouweandela:
The feature freeze for the v2.9 release was yesterday, so I'm removing this from the milestone.

valeriupredoi:
no biggie, man - I think we should be perfectly safe without pinning it for now (remember you made me plop that pytest-json dep in for when an older esmpy gets installed due to an older shapely); so we can always pin if needed; let's just keep an eye on the package build for ESMValTool, there may be an issue, if at all

zklaus:
Shapely 2 comes with a partly new API that facilitates much better performance due to consistent vectorization. This API is already available in version 1.8, which is a transitional version, i.e. API that is being removed in version 2 is merely deprecated in 1.8 and throws deprecation warnings.

Imho, the current situation is slightly dangerous (only slightly because, after all, shapely is not used a lot in our code base). This is because people will likely get version 1.8 and might still use the old API, ignoring the deprecation warnings. When the new version 2 comes in those parts will break. If we pin to >2 now, we will see the breakage and be able to fix it; all further use of shapely will be forced to use only the new API.

valeriupredoi:
@zklaus thanks a lot, man! I wasn't aware of the API evolution in the 2+ versions, but I had a hunch we should go for the modern ones, hence this PR - @bouweandela I'd say we should resurrect this :skull:

.github/workflows/run-tests.yml
```diff
@@ -21,6 +21,7 @@ on:
   push:
     branches:
       - main
+      - unpin_shapely
```
bouweandela, line 24:
```suggestion
```

bouweandela (changes requested):
Thanks for explaining! Could you please clean up @valeriupredoi?

bouweandela (approved):

valeriupredoi:
Cheers bud 🍺

valeriupredoi:
Ha! Merged 2min before getting off the bus, very shapely of me 😁

PR #2108, opened 2023-06-21
remove `with_mypy` or `with-mypy` extension/optional tool for prospector

valeriupredoi:
## Description

~There is an issue with prospector=1.10.2 at the moment and installing it with `with_mypy` makes it not work at all, so I am removing this extra tool for now, so we have some functionality out of it, rather than a complete shutdown;~

~Nevermind, `prospectpr=1.10.2` needs `with-mypy` instead of `with_mypy` due to `poetry` changing version, and @carlio (the prospector main dev) tells me this may change in the future again, depending how poetry wakes up in the morning (what version it is one)~

@bouweandela suggested we remove the extra tool option altogether since we run the mypy tests separately anyway :+1: 

See https://github.com/landscapeio/prospector/issues/624

***

***

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,7 +94,7 @@ REQUIREMENTS = {
         'docformatter',
         'isort',
         'pre-commit',
-        'prospector[with_pyroma]>=1.9.0',
+        'prospector[with_pyroma, with-mypy]>=1.9.0',
         'vprof',
         'yamllint',
         'yapf',
--- a/setup.py
+++ b/setup.py
@@ -94,7 +94,7 @@ REQUIREMENTS = {
         'docformatter',
         'isort',
         'pre-commit',
-        'prospector[with_pyroma,with_mypy]>=1.9.0',
+        'prospector[with_pyroma]>=1.9.0',
         'vprof',
         'yamllint',
         'yapf',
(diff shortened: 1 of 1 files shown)


bouweandela:
We can just drop the `with_mypy`/`with-mypy` extra because we're installing mypy separately already.

valeriupredoi:
aha! Let me do that now - even better! :beer:

bouweandela:
> @bouweandela suggested we remove the extra tool option altogether since we run the mypy tests separately anyway +1

What I actually meant is that we're installing `mypy` as a separate tool as well, so there is no need to pull it in through the prospector `with_mypy` extra. Because it is installed, it works fine with `prospector` too.

bouweandela (approved):

valeriupredoi:
> > @bouweandela suggested we remove the extra tool option altogether since we run the mypy tests separately anyway +1
> 
> What I actually meant is that we're installing `mypy` as a separate tool as well, so there is no need to pull it in through the prospector `with_mypy` extra. Because it is installed, it works fine with `prospector` too.

indeed - mypy is a core dep for us here - may want to deouble check in Tool though

PR #2160, opened 2023-08-01
Updated doc about fixes and added type hints to fix functions

### LINKED ISSUE #2129: Rethinking `fix_file`
Currently, `fx_file` has the call signature `(filename) -> (filename)`. Thus, files need to be copied if they need to be modified (overwriting input files is a very bad idea), which is very expensive. Usually, the `netCDF4` library is used to perform these kind of modifications.

A possible alternative that has often come up is xarray. At the moment, there is a method [`DataArray.to_iris()` ](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.to_iris.html#xarray.DataArray.to_iris) which allows converting a `DataArray` into a `Cube` (no idea how efficient this is, though). In addition, the iris devs [are working on an improved xarray-iris bridge](https://github.com/SciTools/iris/issues/4994).

Thus, it would be very nice to allow an efficient usage of xarray in `fix_file`. My question is now: how would that look like in practice? Possible solutions I could think of are:

- Allow `fx_file` to return xarray objects (currently only `DataArray`?) and `load` to read these kind of objects.
- Allow `fx_file` to return cubes and `load` to read cubes.
- Rewrite `fx_file` in such a way that it is a callback in `load` (with one of the call signatures above).

The reason I am asking this now is that I want to get started with reading the ERA5 GRIB files that are available on Levante (see https://github.com/ESMValGroup/ESMValCore/issues/1991). However, [iris-grib](https://github.com/SciTools/iris-grib) cannot open them. On the other hand, xarray (in combination with ECMWF's [cfgrib](https://github.com/ecmwf/cfgrib)) works perfectly fine. Thus, an option to include xarray into our pipeline would be super helpful for me.

@ESMValGroup/technical-lead-development-team opinions?

schlunma:
## Description

This PR updates the outdated doc about fixes (add the necessary function `get_cube_from_list` to `fix_metadata` and replaced the old `extract_strict` with `extract_cube`) and adds type hints to many fix-related functions.

There is NO change in functionality.

Links to documentation:
- https://esmvaltool--2160.org.readthedocs.build/projects/ESMValCore/en/2160/develop/fixing_data.html#create-the-fix
- https://esmvaltool--2160.org.readthedocs.build/projects/ESMValCore/en/2160/recipe/preprocessor.html#dataset-specific-fixes

***

***


esmvalcore/cmor/_fixes/fix.py
```diff
@@ -57,24 +59,24 @@ def fix_file(
...
         Parameters
         ----------
-        filepath: Path
+        filepath:
             File to fix.
-        output_dir: Path
+        output_dir:
             Output directory for fixed files.
-        add_unique_suffix: bool, optional (default: False)
+        add_unique_suffix: optional
```
valeriupredoi, line 66: why are you removing the types from the docstring? It's still nice to have them in the API docs, I don't think they get parsed from the functional definition, do they?

esmvalcore/preprocessor/_io.py
```diff
@@ -114,19 +120,23 @@ def _delete_attributes(iris_object, atts):
...
 
 
-def load(file, callback=None, ignore_warnings=None):
-    """Load iris cubes from files.
+def load(
+    file: str | Path,
+    callback: Optional[Callable] = None,
+    ignore_warnings: Optional[list[dict]] = None,
+) -> CubeList:
+    """Load iris cubes.
```
valeriupredoi, line 128:
```suggestion
    """Load iris cubes from string or Path objects.
```

esmvalcore/preprocessor/_io.py
```diff
@@ -114,19 +120,23 @@ def _delete_attributes(iris_object, atts):
...
 
     Parameters
     ----------
-    file: str
+    file:
         File to be loaded.
```
valeriupredoi, line 133:
```suggestion
        File to be loaded. Could be string or POSIX Path object.
```

esmvalcore/preprocessor/_io.py
```diff
@@ -180,10 +193,13 @@ def load(file, callback=None, ignore_warnings=None):
...
             raw_cubes = iris.load_raw(file, callback=callback)
     logger.debug("Done with loading %s", file)
+
     if not raw_cubes:
         raise ValueError(f'Can not load cubes from {file}')
+
     for cube in raw_cubes:
-        cube.attributes['source_file'] = file
+        cube.attributes['source_file'] = str(file)
+
```
valeriupredoi, line 202: I would put a check on `file`'s type in here (str is a bit too general, so we are exposing us to fails if load is called with `load("s3://cow")`) - a placeholder so we maybe get support for eg object storage "files" in the future

valeriupredoi (approved):
thanks, Manu! A couple minor points from me :beer:

esmvalcore/preprocessor/_io.py
schlunma: Sorry, I don't fully get your comment - at the moment, only valid `Path` and `str` objects can be passed to `load`. Since a couple of version iris can digest `Path` objects, that's why I am calling `Path(file)` at the beginning of the function now. Here, we needs `str(file)` for the attribute to be properly set.

Since now other types are supported, I don't think we need to add additional checks here. If the path is invalid, we will either get an iris error during `iris.load_raw` or a `ValueError` from our code.

esmvalcore/cmor/_fixes/fix.py
schlunma: Very good point. I added `autodoc_typehints = 'both'` to our RTD config - now they are properly shown in the function description, see https://esmvaltool--2160.org.readthedocs.build/projects/ESMValCore/en/2160/api/esmvalcore.preprocessor.html#esmvalcore.preprocessor.fix_file.

Before that, they have only been shown in the function signature at the very beginning.

valeriupredoi: Danke, Manu!

esmvalcore/preprocessor/_io.py
valeriupredoi: indeed, and my second phrasing of the comment was as vague as a Tory declaring their taxes - what I really meant was we should probably think of examining the `str` path a bit more in the context when it is not a valid string of a POSIX path, and throw up the error, but a little bit more tailored towards cases such eg the str path starts with "s3" or "https" - the first steps towards supporting remote/object store files. Clearly, not a priority now, so totally up to you if you want to go start going down this rabbit hole :rabbit:

valeriupredoi:
thanks a lot for addressing the review comments @schlunma - if you don't (and I expect you not to) want do anything about https://github.com/ESMValGroup/ESMValCore/pull/2160#discussion_r1283046039 then this is good to be merged, you wanna poke the TLT for a merge? :beer:

schlunma:
Thanks for reviewing V! You're right, I don't want to address this here :sweat_smile:

schlunma:
In order to proceed with https://github.com/ESMValGroup/ESMValCore/pull/2157, it would be nice to have this merged today (there will be a lot of merge conflicts which I would like to solve before finalizing https://github.com/ESMValGroup/ESMValCore/pull/2157). Since this is a pure maitenance PR, I don't think this be a problem. Any objections @ESMValGroup/technical-lead-development-team?

Issue #1991, opened 2023-03-30
Adding DKRZ drs for `native6` to enable usage of local vast collection of ERA5 data (hourly, grb)


remi-kazeroni:
From the reviewers of our IS-ENES3 deliverable D9.5:

_Angelika has just (almost*) completed to retrieve the 1940–2022 hourly time series from ECMWF Mars Tape Archive to Levante /pool/data/ERA5.
This comprises 
	• surface level analysis (49 parameters)
	• surface level forecasts (55 parameters)
	• model level analysis (16 parameters) (*retrieval 1940–1958 ongoing)
	• pressure level analysis (16 parameters) (*retrieval 1940–1950 ongoing). 
The data are not the 0.25° regridded ERA5 data that users can download from Copernicus CDS, but the native resolution ERA5 data (T639/N320) that can be retrieved from Mars, only.
The now around 1550 Tb of data are stored directly on Levante’s disk storage, globally accessible via /pool/data/ERA5/E5._

We could think of adding a `drs` in our [config file](https://github.com/ESMValGroup/ESMValCore/blob/main/esmvalcore/config-developer.yml#L95) to enable ESMValTool to access this vast collection of ERA5 data. This would be interesting if DKRZ users would like to benefit from having access to that data pool. It could be an interesting test to check how well grib files are handled by ESMValTool.

I'm not sure this could completely replace our own local collection of RAWOBS ERA5 data (downloaded from CDS with `era5cli`). The reason is such retrieval of huge amount of data cannot be done by our users on their own. These would need to continue relying on tools like `era5cli` or `cdsapi` to create their own local pools of ERA5 data on their own machines/clusters. Thus, it might be good to continue testing ESMValTool as done now with ERA5 in our own RAWOBS to better reproduce what's done by a majority of our users.

This idea is similar to that of #1246 for Jasmin. See also [DKRZ docs](https://docs.dkrz.de/doc/dataservices/finding_and_accessing_data/era_data/index.html) on ECMWF reanalysis products available locally.

schlunma:
I started looking into this some days ago. Here are some insights:

#### Reading the data

[...]

Issue #1907, opened 2023-01-27
CMOR checks ignore wrong `var_name`


schlunma:
## Problem

If the `var_name` of a netCDF file is different to the one expected by the CMOR tables (but everything else matches), our CMOR checker simply ignores this and happily writes an output file with the wrong `var_name`. Example:

Asking for the variable

```yaml
...
      gpp:
        project: CMIP5
        mip: Lmon
        exp: esmrcp85
        ensemble: r1i1p1
        start_year: 2010
        end_year: 2010
        additional_datasets:
          - {dataset: CESM1-BGC}
...
```

when a file with the correct DRS is present (e.g., `CMIP5/NSF-DOE-NCAR/CESM1-BGC/esmrcp85/mon/land/Lmon/r1i1p1/v1/gpp/gpp_Lmon_CESM1-BGC_esmrcp85_r1i1p1_200601-210012.nc`) but with wrong `var_name`:

```
...
variables:                                                                                                                                                                                                                                                                    
        float not_gpp(time, lat, lon) ;                                                                                                                                                                                                                                       
                not_gpp:standard_name = "gross_primary_productivity_of_carbon" ;                                                                                                                                                                                              
                not_gpp:long_name = "Carbon Mass Flux out of Atmosphere due to Gross Primary Production on Land" ;                                                                                                                                                            
                not_gpp:units = "kg m-2 s-1" ;                                                                                                                                                                                                                                
                not_gpp:cell_methods = "time: mean (interval: 30 days) area: mean where land" ;
...
```

will lead to funny warnings like `WARNING There were warnings in variable not_gpp: not_gpp: attribute positive not present` but also leads to a `preproc` file with the very same wrong name, which eventually will lead to problems in our diagnostics (which expect correct var names). In addition, we often select cubes based on their `var_name` in ESMValCore (example [here](https://github.com/ESMValGroup/ESMValCore/blob/main/esmvalcore/cmor/_fixes/fix.py#L92-L93) or [here](https://github.com/ESMValGroup/ESMValCore/blob/d832c8c26110b1e878074d043b7ccf6273c34170/esmvalcore/preprocessor/_derive/__init__.py#L87-L88)), which is also very problematic with wrong var names.

## Solutions

I see 2 possible solutions here:

[...]

PR #1125, opened 2021-05-12
Add era5 variable mappings

zklaus:
## Description

Closes #issue_number

Link to documentation:

Table with ERA5 variable names: https://confluence.ecmwf.int/pages/viewpage.action?pageId=82870405#ERA5:datadocumentation-Table1
***

## [Before you get started](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#getting-started)

- [ ] [☝ Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do

## [Checklist](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#checklist-for-pull-requests)

It is the responsibility of the author to make sure the pull request is ready to review. The icons indicate whether the item will be subject to the [🛠 Technical][1] or [🧪 Scientific][2] review.

[1]: https://docs.esmvaltool.org/en/latest/community/review.html#technical-review
[2]: https://docs.esmvaltool.org/en/latest/community/review.html#scientific-review

- [ ] [🧪][2] The new functionality is [relevant and scientifically sound](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#scientific-relevance)
- [ ] [🛠][1] This pull request has a [descriptive title and labels](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-title-and-label)
- [ ] [🛠][1] Code is written according to the [code quality guidelines](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#code-quality)
- [ ] [🧪][2] and [🛠][1] [Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available
- [ ] [🛠][1] [Unit tests](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#tests) have been added
- [ ] [🛠][1] Changes are [backward compatible](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#backward-compatibility)
- [ ] [🛠][1] Any changed [dependencies have been added or removed](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#dependencies) correctly
- [ ] [🛠][1] The [list of authors](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#list-of-authors) is up to date
[...]

PR #2178, opened 2023-08-23
Add support for native ERA5 data in GRIB format

schlunma:
## Description

This PR allows ESMValCore to process native ERA5 data in GRIB format, which is for example [available on Levante](https://docs.dkrz.de/doc/dataservices/finding_and_accessing_data/era_data/index.html#era-data) in the `/pool/data/ERA5` directory.

#### Reading the data

The following settings are necessary in the user configuration file:

```yaml
rootpath:
  ...
  native6:
    /pool/data/ERA5: DKRZ-ERA5-GRIB
  ...
```

I added an extra facets file which includes reasonable default for all supported variables. You can check it out [here](https://github.com/ESMValGroup/ESMValCore/blob/911ed28199ea3c14991c344cdcd7d65ba5e4220b/esmvalcore/config/extra_facets/native6-mappings.yml).

Thus, reading this data is as easy as

```yaml
datasets:
  - {project: native6, dataset: ERA5, timerange: '2000/2001', short_name: tas, mip: Amon}
  - {project: native6, dataset: ERA5, timerange: '2000/2001', short_name: cl, mip: Amon, tres: 1H, frequency: 1hr}
  - {project: native6, dataset: ERA5, timerange: '2000/2001', short_name: ta, mip: Amon, type: fc, typeid: '12'}
```

#### Regridding

Native ERA5 data in GRIB format is on a reduced Gaussian grid (i.e., an unstructured grid). Thus, in 99% of the use cases, it is necessary to regrid this data, especially since no cell areas are available for the data (thus, we cannot even calculate global/regional statistics over the native data). This is done automatically by the CMORizer (as [recommended by the ECMWF](https://confluence.ecmwf.int/display/CKB/ERA5%3A+What+is+the+spatial+reference#heading-Interpolation)), but can be disabled in the recipe:

```yaml
datasets:
  - {project: native6, dataset: ERA5, timerange: '2000/2001', short_name: tas, mip: Amon, automatic_regrid: false}
```

This PR depends on the following other PRs:

- [x] https://github.com/SciTools/iris-grib/pull/357
- [x] https://github.com/ESMValGroup/ESMValCore/pull/2231
- [x] https://github.com/ESMValGroup/ESMValCore/pull/2344
- [x] https://github.com/ESMValGroup/ESMValCore/pull/2350
- [x] https://github.com/ESMValGroup/ESMValCore/pull/1894 (optional, but very helpful)

***

Closes https://github.com/ESMValGroup/ESMValCore/issues/1991
Closes https://github.com/ESMValGroup/ESMValTool/issues/3238

Link to documentation: https://esmvaltool--2178.org.readthedocs.build/projects/ESMValCore/en/2178/quickstart/find_data.html#supported-native-reanalysis-observational-datasets

***

***


schlunma:
This is ready from my side, but there's two issues that need to be resolved before I mark this ready for review:

- [ ] https://github.com/ESMValGroup/ESMValCore/pull/2181
- [ ] For some reason I can't get `iris-grib` to run on CircleCI; locally it works well. This is probably closely related to the problem reported by @remi-kazeroni [here](https://github.com/ESMValGroup/ESMValTool/pull/3141#issuecomment-1686550089).

I tested this thoroughly with the following recipe: [recipe_000.yml.txt](https://github.com/ESMValGroup/ESMValCore/files/12437112/recipe_000.yml.txt)

An example run is available on Levante here: `/home/b/b309141/scratch/esmvaltool_output/recipe_000_20230825_080240`

Note that with the default dask scheduler, this recipe ran into a timeout after 8 hours with 67/76 tasks finished. With the following dask configuration, I could run the same recipe on the same node (regular Levante compute node with 256 GiB of memory) in **5:27 min (!!)**. This is amazing 🚀 

@ESMValGroup/technical-lead-development-team

schlunma:
Regarding the installation: Findings by Rémi:

> I could get the dependency tests pass locally by adding python-eccodes<1.6.0 to environment*yml and setup.py which is needed [to install iris-grid](https://iris-grib.readthedocs.io/en/stable/#getting-started). The tests would still fail locally with python-eccodes not pinned (i.e. 1.6.0).

> For the latest version of python-eccodes, I still get ImportError: /opt/conda/envs/esmvaltool/lib/python3.11/site-packages/gribapi/_bindings.cpython-311-x86_64-linux-gnu.so: undefined symbol: codes_bufr_key_is_coordinate which seems to be an issue upstream.

> If I understand correctly, the python-eccodes would need to be rebuilt on [conda-forge](https://github.com/conda-forge/python-eccodes-feedstock) to be compatible with the latest eccodes (2.30.2, but still at 2.21.2 on conda-forge). But perhaps our in-house conda-forge specialists, @zklaus and @valeriupredoi, would know better how to best handle eccodes and the Python bindings for the Tool 🍻

Answer by @zklaus:

> Thanks for bringing this up! I dug a bit deeper and it looks like this python-eccodes upstream (at ECMWF) declares a min_recommended_version for Eccodes which currently stands at 2.21.0. However, with release 1.6.0 they added linking to a function that was only added to Eccodes in 2.31.0. Conda-forge relies on the declaration in python-eccodes to do the pinning, so I have opened an issue there (https://github.com/ecmwf/eccodes-python/issues/82) to update that. In the meantime, we can already fix the conda-forge package, which is tracked in https://github.com/conda-forge/python-eccodes-feedstock/issues/86.

> Could you possibly test if things work with python-eccodes-1.6.0 and eccodes-2.31.0?

schlunma:
Since both PRs linked by @zklaus are merged now, I simply tested the installation with the environment provided in this PR (`iris-grib` added to `environment.yml` and `setup.py`; no other pins) again.

On CircleCI, the tests fail again with a `RuntimeError: Could not load the ecCodes library!` (see [here](https://app.circleci.com/pipelines/github/ESMValGroup/ESMValCore/9853/workflows/020a70fa-bab1-4fff-9a07-ccc528a0f117/jobs/42657)).

Locally, the tests ran fine. To install ESMValCore, I ran

```bash
cd ESMValCore
mamba env create -n esm -f environment.yml
mamba activate esm
pip install -e '.[develop]'
```

The conda environment creation gives me

```bash
eccodes                   2.32.0               h35c6de3_0    conda-forge
python-eccodes            1.6.1           py311h1f0f07a_1    conda-forge
iris-grib                 0.18.0             pyhd8ed1ab_0    conda-forge
```

After the `pip install` step, I get the deprecated eccodes-python package again, so there is still a problem there I suppose:

```bash
eccodes                   2.32.0               h35c6de3_0    conda-forge
eccodes-python            0.9.9                    pypi_0    pypi
python-eccodes            1.6.1           py311h1f0f07a_1    conda-forge
iris-grib                 0.18.0             pyhd8ed1ab_0    conda-forge
```

No idea though why it works locally but fails on CircleCI.

schlunma:
Thanks V for turning on the GA tests! I will open an issue at iris-grib about that dependency.

Here is a quick summary from our Tech Lead Team meeting call (I thought I posted this yesterday, but apparantly I forgot to click "send" 😄):

- The tests fails on CircleCI are to be expected since it uses an outdated Docker container. @bouweandela how do we update the image? Is this done automatically for new releases?
- I will include the regridding into our public `linear` scheme in a separate PR.

schlunma:
Corresponding iris-grib issue: https://github.com/SciTools/iris-grib/issues/356

bouweandela:
> The tests fails on CircleCI are to be expected since it uses an outdated Docker container. @bouweandela how do we update the image? Is this done automatically for new releases?

For a general description of the docker setup you can read [this](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#check-the-docker-images). The easiest way to update the docker image with a new dependency (that cannot be installed from PyPI) is to make a pull request that only adds that dependency. Once that is merged into `main`, a new `esmvalgroup/esmvalcore:development` container image will be built by dockerhub and when that is finished it will be used to run all the tests.

schlunma:
> The easiest way to update the docker image with a new dependency (that cannot be installed from PyPI) is to make a pull request that only adds that dependency. Once that is merged into `main`, a new `esmvalgroup/esmvalcore:development` container image will be built by dockerhub and when that is finished it will be used to run all the tests.

Done in https://github.com/ESMValGroup/ESMValCore/pull/2453

schlunma:
This is now ready for review.

Here is a recipe that can be used to test this: [recipe_era5_grib.yml.txt](https://github.com/user-attachments/files/15787632/recipe_era5_grib.yml.txt). Please not that this needs the `iris-grib` package and an updated config-user.yml file as described [here](https://github.com/ESMValGroup/ESMValCore/pull/2178#issue-1862691857).

Here is the output of that recipe: https://esmvaltool.dkrz.de/shared/esmvaltool/era5_grib_tests/recipe_era5_grib_20240611_114507/

axel-lauer (approved):
Great work @schlunma ! I looked at the output of your test recipe and everything looks fine. I also ran some tests for specific humidity (hus) at 100 hPa. The results look good, so https://github.com/ESMValGroup/ESMValTool/issues/3238 can be closed once this PR is merged. Yay!
The only thing I find slightly annoying is that you have to use the regrid preprocessor to obtain ERA5 output that is usable with all diagnostics. If I want to use the same preprocessor also for model datasets (which is what I do often), I cannot analyze the model data on their native grid.
An idea (please feel free to ignore) might be to introduce an additional facet for native6 datasets, e.g. "default_regridding: true", which would apply the "default" regridding to the data, i.e. 0.25°x0.25° with a linear scheme as recommended by ECMWF.

schlunma:
Great, thanks for reviewing @axel-lauer!

> The only thing I find slightly annoying is that you have to use the regrid preprocessor to obtain ERA5 output that is usable with all diagnostics. If I want to use the same preprocessor also for model datasets (which is what I do often), I cannot analyze the model data on their native grid. An idea (please feel free to ignore) might be to introduce an additional facet for native6 datasets, e.g. "default_regridding: true", which would apply the "default" regridding to the data, i.e. 0.25°x0.25° with a linear scheme as recommended by ECMWF.

Good point. Initially I decided not to do that because the fix (more specifically `fix_metadata`) is executed for each input file separately (and for hourly data there are lots of files!), so this would be a huge performance bottleneck. However, I just realized that we also have `fix_file` which is executed after cube concatenation and time range clipping, so this would be the perfect place for the regridding.

The only issue I see here is that there is a risk of regridding twice if users use a custom regrid preprocessor and the "default" regridding. Do you think this is a problem? Should we enable the "default" regridding by default (I guess yes?)?

> P.S.: I think iris-grib needs to be added to the environment so it does not need to be installed manually.

This has already been added in https://github.com/ESMValGroup/ESMValCore/pull/2453 to let the tests pass 👍

schlunma:
> An idea (please feel free to ignore) might be to introduce an additional facet for native6 datasets, e.g. "default_regridding: true", which would apply the "default" regridding to the data, i.e. 0.25°x0.25° with a linear scheme as recommended by ECMWF.

This is implemented in https://github.com/ESMValGroup/ESMValCore/pull/2178/commits/cc632dc4aed304d8751f200e729bf08c59da414b now.

@ESMValGroup/technical-lead-development-team could any of you have a brief look on this and perform a technical review? That would be awesome, thanks!

esmvalcore/config/extra_facets/native6-mappings.yml
```diff
@@ -0,0 +1,192 @@
...
+# A complete list of supported keys is given in the documentation (see
+# ESMValCore/doc/quickstart/find_data.rst).
+---
+
+ERA5:
+
+  # Settings for all variables of all MIPs
+  '*':
+    '*':
+      family: E5
```
bouweandela, line 16: Is there a place where the meaning of these facets is explained? If yes, it would be nice to add a link here to make it easier to understand where they come from and how to update them.

esmvalcore/cmor/_fixes/native6/era5.py
```diff
@@ -433,3 +569,16 @@ def fix_metadata(self, cubes):
...
+        """Fix data."""
+        regridding_enabled = self.extra_facets.get('regrid', True)
+        if regridding_enabled and has_unstructured_grid(cube):
+            logger.debug(
+                "Automatically regrid ERA5 data (variable %s) to %s° "
+                "grid using bilinear regridding",
+                self.vardef.short_name,
+                DEFAULT_ERA5_GRID,
+            )
+            cube = regrid(cube, DEFAULT_ERA5_GRID, 'linear')
```
bouweandela, line 583: While I can see that the feature is useful, in particular to make seamlessly switching between NetCDF and GRIB ERA5 data possible, I'm not not sure if it should be enabled by default: I just checked a few recipes what the effect of this would be:
```
clouds/recipe_lauer22jclim_fig5_lifrac.yml
clouds/recipe_lauer22jclim_fig1_clim.yml
recipe_climwip_test_performance_sigma.yml
bock20jgr/recipe_bock20jgr_fig_1-4.yml
ipccwg1ar6ch3/recipe_ipccwg1ar6ch3_fig_3_19.yml
model_evaluation/recipe_model_evaluation_basics.yml
```
and in all cases it leads to double regridding. Therefore I think that disabled by default may be a better option.

Another concern is that a fix seems conceptually the wrong place to implement regridding, as this starts mixing preprocessing with data loading. But maybe there is no way around it.

bouweandela: It would also be possible to automatically add a regrid preprocessor in `esmvalcore/_recipe/recipe.py` for ERA5 data, but I'm not sure if that would be any cleaner.

esmvalcore/config/extra_facets/native6-mappings.yml
```diff
@@ -0,0 +1,192 @@
+# Extra facets for native6 data
```
bouweandela, line 1: It would be nice to have `era5` in the filename, to make it clear that this is about ERA5 data. How about `native6-era5.yml`?

esmvalcore/config/extra_facets/native6-mappings.yml
```diff
@@ -0,0 +1,192 @@
...
+ERA5:
+
+  # Settings for all variables of all MIPs
+  '*':
+    '*':
+      family: E5
+      regrid: true
+      type: an
+      typeid: '00'
+      version: ''  # necessary to get a nice output file name
```
bouweandela, line 20:
```suggestion
      version: 'v1'  # necessary to get a nice output file name
```
may be nicer? This is also the version we use for NetCDF ERA5 data.

bouweandela: Actually, I think this is already specified in the recipe

esmvalcore/config/extra_facets/native6-mappings.yml
```diff
@@ -0,0 +1,192 @@
...
+
+    # Variable-specific settings
+    albsn:
```
bouweandela, line 23: Would it be possible to add `var_name` or a similar name (e.g. standard_name, long_name) to this mapping and add a simple check in `AllVars.fix_metadata` that the input data actually contains the right variable? I've always been concerned that people may put the wrong file in the wrong directory and get completely wrong results out of these fixes.

esmvalcore/config/extra_facets/native6-mappings.yml
schlunma: Added in https://github.com/ESMValGroup/ESMValCore/pull/2178/commits/4c3f1beb0e55fa70d0e7c9bdfcc1096231ecce12

esmvalcore/config/extra_facets/native6-mappings.yml
schlunma: Done in https://github.com/ESMValGroup/ESMValCore/pull/2178/commits/4c3f1beb0e55fa70d0e7c9bdfcc1096231ecce12

esmvalcore/config/extra_facets/native6-mappings.yml
schlunma: Hmm, I am not sure about this. Does it really make sense to invent an arbitrary version number here? The only reason we need this is to derive the [output file name](https://github.com/ESMValGroup/ESMValCore/blob/4c3f1beb0e55fa70d0e7c9bdfcc1096231ecce12/esmvalcore/config-developer.yml#L106); it's not used at all to find the data.

esmvalcore/config/extra_facets/native6-mappings.yml
schlunma: In principle I like that idea, but I don't know if that works in practice. For example, are we sure that the names of the netCDF input files are the exact same as for the GRIB input files? If this is not the case, the fix will start failing for one or the other.

I would argue that for the GRIB files the risk of using wrong files is rather low, as the GRIB ID (= the variable name) is [part of the file name](https://github.com/ESMValGroup/ESMValCore/blob/cc632dc4aed304d8751f200e729bf08c59da414b/esmvalcore/config-developer.yml#L105). So this is more of a problem for the netCDF input files, for which we simple use `*.nc` as input file pattern.

esmvalcore/preprocessor/_io.py
```diff
@@ -130,7 +130,14 @@ def load(
         # warnings.filterwarnings
         # (see https://github.com/SciTools/cf-units/issues/240)
         with suppress_errors():
-            raw_cubes = iris.load_raw(file, callback=_load_callback)
+            # GRIB files need to be loaded with iris.load, otherwise we will
+            # get separate (lat, lon) slices for each time step, pressure
+            # level, etc.
+            grib_formats = ('.grib2', '.grib', '.grb2', '.grb', '.gb2', '.gb')
+            if file.suffix in grib_formats:
+                raw_cubes = iris.load(file, callback=_load_callback)
```
valeriupredoi, line 138: Manu, you should tell @trexfeathers and Iris folk to tell ECMWF to list Iris as a viable/recommended GRIB loader, see https://confluence.ecmwf.int/display/CKB/What+are+GRIB+files+and+how+can+I+read+them - one q with regards to Iris stability wrt GRIB files - is this a gods-given one now? If not, I'd argue we should perform a few consistency tests on the loaded cube(s)

schlunma: Since the recent update(s) of iris-grib I didn't have any difficulties reading those ERA5 GRIB files. Since I would not expect that these files will change any time soon, I won't expect any problems with that. We also perform the same extensive CMOR checks on those files as we do for the netCDF files, so I am pretty confident we are fine here.

About reading GRIB files in general - I am not 100% sure if we (or better, iris-grib) support all possible GRIB files now, but I wouldn't know what to do about this here. If we encounter problems at some point in the future, we can fix that there. Again, given that we have extensive CMOR checks in our pipeline, I am fairly sure we will know if there are problems, even if we won't get very obvious errors.

esmvalcore/config/extra_facets/native6-mappings.yml
bouweandela: Yes, we invent version numbers for all observational and reanalysis datasets.

doc/quickstart/find_data.rst
```diff
@@ -131,6 +140,74 @@ ERA5
   of both liquid and solid phases to vapor (from underlying surface and vegetation)."
   Therefore, the ERA5 (and ERA5-Land) CMORizer switches the signs of ``evspsbl`` and ``evspsblpot`` to be compatible with the CMOR standard used e.g. by the CMIP models.
 
+.. _read_native_era5_grib:
+
+ERA5 (in GRIB format available on DKRZ's Levante)
```
bouweandela, line 145: Mention that ERA5 data in grib format can also be downloaded from ECMWF?

esmvalcore/cmor/_fixes/native6/era5.py
bouweandela: There is the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) and to me at least, automatic regridding for one dataset and not for any other dataset is rather surprising, even if it may make sense in this case.

esmvalcore/config/extra_facets/native6-mappings.yml
bouweandela: I would expect (but have not checked), that ECMWF uses the same names for their variables, regardless of the file format. e.g. are these variable names present in both file formats? https://confluence.ecmwf.int/pages/viewpage.action?pageId=82870405#ERA5:datadocumentation-Table4

esmvalcore/config/extra_facets/native6-mappings.yml
schlunma: This is certainly not true for all observational and reanalysis datasets. I will change it in the extra facets file..

schlunma: https://github.com/ESMValGroup/ESMValCore/pull/2178/commits/0fea4abd583b8f5fb5c1a6d1a2d300d6090a23ff

esmvalcore/config/extra_facets/native6-mappings.yml
schlunma: While this is probably true, I don't think this check is relevant for the GRIB files. As I mentioned, they have their GRIB ID (= variable name) in the file name. We also don't check the variable name for CMIP model output.

Thus, IMHO, it would make more sense to implement this in a different PR (this one here as already >1000 lines).

doc/quickstart/find_data.rst
schlunma: I am not sure how useful that is in practice to be honest. All the facets and paths here are tailored towards the files on Levante. I *think* you'll get different paths if you download them manually since the names of the facets seem to be [different](https://cds.climate.copernicus.eu/cdsapp#!/dataset/reanalysis-era5-complete?tab=form) in the docs.

To me, it's easier for the user if we just recommend manual download of the files in netCDF format via era5cli.

esmvalcore/cmor/_fixes/native6/era5.py
schlunma: One could also argue that the least surprise would be to get files on a regular grid (like the ones you get when you download the netCDF data).

How about checking the extension of the files in `esmvalcore/_recipe/recipe.py` and applying the regridding based on this? This would avoid double regridding.

esmvalcore/preprocessor/_io.py
trexfeathers: Hi folks, I've raised SciTools/iris-grib#515

Iris-grib definitely cannot load all possible GRIB files. As you may have noticed, the GRIB specification, and how [eccodes](https://confluence.ecmwf.int/display/ECC) interprets this, is not as specific as NetCDF. This means it can be hard to know for certain if Iris-grib is doing the right thing, and we typically therefore rely on finding expert users who can tell us.

So the list of loadable templates is limited to those where we could find a user to check our work. And this also means that we might later discover problems with our existing work, but so far that appears to be rare.

schlunma:
There is currently a problem with iris-grib: https://github.com/SciTools/iris-grib/issues/520

schlunma:
@bouweandela would you be able to have another look on my answers to your review comments? It would be great to get this into main very soon. We want to use the features from this PR to add support for further datasets in GRIB format (e.g., CAMS reanalysis). Thanks 🙏

trexfeathers:
> There is currently a problem with iris-grib: [SciTools/iris-grib#520](https://github.com/SciTools/iris-grib/issues/520)

Note that a new version of Iris-grib is now available and that should fix this problem?

schlunma:
> Note that a new version of Iris-grib is now available and that should fix this problem?

Looks like this is indeed the case, the tests run fine now 🎉

bouweandela:
> @bouweandela would you be able to have another look on my answers to your review comments?

Not this week, but hopefully next week.

doc/quickstart/find_data.rst
```diff
@@ -131,6 +140,74 @@ ERA5
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+ERA5 data in monthly, daily, and hourly resolution is `available on Levante
+<https://docs.dkrz.de/doc/dataservices/finding_and_accessing_data/era_data/index.html#era-data>`__
+in its native GRIB format.
+To read these data with ESMValCore, use the root path ``/pool/data/ERA5`` with
+DRS ``DKRZ-ERA5-GRIB`` in your :ref:`user configuration file`, for example:
```
bouweandela, line 152: The documentation build appears to be broken because this link was removed in #2448

esmvalcore/cmor/_fixes/native6/era5.py
```diff
@@ -24,17 +27,23 @@ def get_frequency(cube):
...
             )
         return "fx"
 
     interval = time.points[1] - time.points[0]
+
     if interval - 1 / 24 < 1e-4:
         return "hourly"
-
+    if interval - 1.0 < 1e-4:
+        return "daily"
```
bouweandela, line 46: Why is this code needed? ERA5 data is always either hourly or monthly means as far as I'm aware

esmvalcore/cmor/_fixes/native6/era5.py
```diff
@@ -149,6 +204,20 @@ def fix_metadata(self, cubes):
         return cubes
 
 
+class O3(Fix):
+    """Fixes for o3."""
+
+    def fix_metadata(self, cubes):
+        """Convert mass mixing ratios to mole fractions."""
+        for cube in cubes:
+            cube.units = "kg kg-1"
```
bouweandela, line 213: This change of units does not seem to have any effect since different units are assigned on line 217

esmvalcore/cmor/_fixes/native6/era5.py
```diff
@@ -337,6 +457,19 @@ def fix_metadata(self, cubes):
         return cubes
 
 
+class Toz(Fix):
+    """Fixes for toz."""
+
+    def fix_metadata(self, cubes):
+        """Convert 'kg m-2' to 'm'."""
+        for cube in cubes:
+            cube.units = "kg m-2"
```
bouweandela, line 466: Same comment here as in https://github.com/ESMValGroup/ESMValCore/pull/2178/files#r1790469038

esmvalcore/cmor/_fixes/native6/era5.py
```diff
@@ -390,10 +515,25 @@ def _fix_coordinates(self, cube):
...
+                    and has_unstructured_grid(cube)
+                ):
+                    coord.guess_bounds()
 
         self._fix_monthly_time_coord(cube)
 
+        # Fix coordinate increasing direction
+        if cube.coords("latitude") and not has_unstructured_grid(cube):
+            lat = cube.coord("latitude")
+            if lat.points[0] > lat.points[-1]:
```
bouweandela, line 530: Why is this additional check needed? It does not seem to be applied to grib data because that is on an unstructured grid and for NetCDF data it should just be applied

esmvalcore/cmor/_fixes/native6/era5.py
schlunma: Fixed in https://github.com/ESMValGroup/ESMValCore/pull/2178/commits/a031ea9516ce1db269396d46e380fad89d1191bd

doc/quickstart/find_data.rst
schlunma: Fixed in https://github.com/ESMValGroup/ESMValCore/pull/2178/commits/a031ea9516ce1db269396d46e380fad89d1191bd

esmvalcore/cmor/_fixes/native6/era5.py
schlunma: The ERA5 GRIB data on Levante is also available in daily format (`1D`), see also https://docs.dkrz.de/doc/dataservices/finding_and_accessing_data/era_data/index.html#era5-data-via-pool-data-era5.

esmvalcore/cmor/_fixes/native6/era5.py
schlunma: This was more intended as a comment. Fixed in https://github.com/ESMValGroup/ESMValCore/pull/2178/commits/44040c37027c245e4f42f7f451c132439519812f

esmvalcore/cmor/_fixes/native6/era5.py
schlunma: Klaus once used a smart word to describe that our fixes should only fix wrong data, but do nothing for correct data (I can't remember what it was). This is what I'm trying to do here. If someone applies this fix to data with a correctly ordered lat, this fix should not reverse it.

schlunma: Okay, ChatGPT tells me it's "idempotent" 😁

esmvalcore/cmor/_fixes/native6/era5.py
bouweandela: > How about checking the extension of the files in esmvalcore/_recipe/recipe.py and applying the regridding based on this? This would avoid double regridding.

Sounds good, could you give that a try?

esmvalcore/config/extra_facets/native6-mappings.yml
bouweandela: > We also don't check the variable name for CMIP model output.

That is exactly what the CMOR checks do, right? In this case, we're overwriting the variable name as part of the fix, regardless of what was there. As long as people stick with using the variables that you have pre-defined for them in this pull request, that seems quite safe, but what if they would start adding others? Or did you add support for all the ERA5 data available on DRKZ?

doc/quickstart/find_data.rst
bouweandela: I'm not sure if we should recommend using the NetCDF data because 1) we have encountered issues with it in the past and 2) the NetCDF option is still listed as experimental and not suitable for production use on the CDS download page: https://cds.climate.copernicus.eu/datasets/reanalysis-era5-pressure-levels?tab=download.

esmvalcore/cmor/_fixes/native6/era5.py
bouweandela: I can see that is a nice feature, but overly defensive code increases the maintenance burden. E.g. what if someone tries to use the data of only a single latitude point? (as we've had with the time dimension #2511)

bouweandela: Anyway, I'm fine with leaving this as is.

esmvalcore/config/extra_facets/native6-mappings.yml
schlunma: > That is exactly what the CMOR checks do, right?

No, the CMOR checks don't do that: https://github.com/ESMValGroup/ESMValCore/issues/1907. They will simply overwrite wrong long names and standard names (regardless of what was there), but wrong variables names are simply ignored.

Sorry, I don't get your point. What would checking the variable name help if people started to use (unsupported) variables? We could of course add a list of supported variable names and raise errors for others, but just checking against a predefined value given in the extra facets won't help at all (as you said, users could simply overwrite that in the recipe).

For GRIB, I really don't see the point in adding this check. For all supported variables, a GRIB ID is predefined, and for all others, users need to give that in the recipe. And if users **explicitly** give that in the recipe, I think we should be fine with that and assume they know what they're doing.

esmvalcore/cmor/_fixes/native6/era5.py
schlunma: > E.g. what if someone tries to use the data of only a single latitude point? (as we've had with the time dimension #2511)

Nothing would happen since using the index `-1` on a list with one element is totally fine, and the if statement would simply evaluate to `False`.

valeriupredoi: not quite: idempotence is when you apply an operator, then no matter however many times you apply the same operator, the result doesn't change anymore, past the first result

doc/quickstart/find_data.rst
schlunma: Ok, I am happy to include this in the documentation, but please be aware that I didn't test this and it would probably need some effort from the user side to make it work (either by putting the files in the correct directories or by setting up a new DRS).

schlunma: Added in https://github.com/ESMValGroup/ESMValCore/pull/2178/commits/e4d642a465417cf17fca7d329150389f37b0f616

bouweandela: Thanks!

esmvalcore/config/extra_facets/native6-mappings.yml
bouweandela: My main concern is that if users download data, either as NetCDF or as GRIB, there is a risk that they make a mistake and put a file containing the wrong variable in the directory with the variable name or GRIB ID that we use to look up the files corresponding to a particular variable. Since we are overwriting all variable names in the fix, this may go unnoticed. That is why it would make sense to me to check that the input variable name matches with what we expect before overwriting.

A secondary concern is that it may not be clear to users which variables are supported and which are not. This is probably a bigger issue with NetCDF input than with GRIB files though, since you added a nice list of supported in GRIB [here](https://github.com/ESMValGroup/ESMValCore/blob/read_era5_grib/esmvalcore/config/extra_facets/native6-era5.yml).

> No, the CMOR checks don't do that: https://github.com/ESMValGroup/ESMValCore/issues/1907. They will simply overwrite wrong long names and standard names (regardless of what was there), but wrong variables names are simply ignored.

You're right, that does not look safe at all.

schlunma: Ok, just to make sure I understand you correctly: you propose to add a new facet (something like `raw_name`) to the extra facets table and check the variables name of the raw file against that, right?

I fully agree that this would make sense for the netcdf files. They currently use `*.nc` as input file pattern, so basically every netcdf file in the input directory is read. This will lead to various problems if users put wrong files into the input directories.

However, for ERA5 GRIB files, I think the situation is a lot different:

1. For the data pool available on DKRZ, the input file convention is `'{family}{level}{typeid}_{tres}_*_{grib_id}.grb'`. Since the GRIB ID is basically the variable name, there's no way of selecting a wrong file. Also, users won't need to copy around files themselves. If they do that, they're on their own (we have no way of making sure everything works fine if users start modifying and renaming files; but this can happen for any data, not only ERA5 GRIB).
2. For self-downloaded data, we specifically mention that this is *experimental* and users need additional setup. I don't know what the naming conventions for self-downloaded data are (I don't even know if they are consistent across the [various ERA5 products](https://cds.climate.copernicus.eu/datasets) you can download), so I am really hesitant to enforce any additional restrictions on the data without more tests and an in-depth analysis.

In summary: I think checking variables names is a good idea for the netcdf files and maybe for self-downloaded GRIB files, but this should go into a separate PR. The main point of this PR is to add support for the ERA5 data pool that is available on Levante, and I am convinced that there's no risk of using wrong files if users don't start modifying/renaming/copying files themselves.

esmvalcore/cmor/_fixes/native6/era5.py
schlunma: Implemented in https://github.com/ESMValGroup/ESMValCore/pull/2178/commits/e123244d16b99564c29a62e6d84f15568fcf6030

esmvalcore/config/extra_facets/native6-mappings.yml
bouweandela: I expect users will be quick to rebuild the DKRZ data structure on their own computers if they don't have access to DKRZ. But you're right that having the grib_id in the filename provides some safety, so I'm happy to leave this feature for a future pull request. See also https://github.com/ESMValGroup/ESMValCore/pull/1125 for some previous discussion on the topic.

bouweandela (approved):

esmvalcore/_recipe/recipe.py
```diff
@@ -227,6 +228,34 @@ def _get_default_settings(dataset):
...
+        # Settings recommended by ECMWF
+        # (https://confluence.ecmwf.int/display/CKB/ERA5%3A+What+is+the+spatial+reference#heading-Interpolation)
+        settings["regrid"] = {"target_grid": "0.25x0.25", "scheme": "linear"}
+        logger.debug(
+            "Automatically regrid native6 ERA5 data in GRIB format with the "
+            "settings %s",
+            settings["regrid"],
+        )
+
+
```
bettina-gier, line 258: If we want to add other GRIB data sources later on, wouldn't it be better to not hardcode ERA5 here, but instead have it as an overwriteable default? Or do you only want to regrid ERA5 on default because of the ECMWF recommendations?

bettina-gier (approved):
Found a few typos and had a comment for futureproofing the automatic regridding in case we are adding more grib datasets (CAMS soon TM). Runs fine though and input looks reasonable!
Think we can merge this with 3 pull request reviews and all issues and dependencies solved?

esmvalcore/_recipe/recipe.py
schlunma: ERA5 is hardcoded here on purpose. We really only want the automatic regridding in this specific case. If we want to extend it later, we can simply add another `elif` 👍

schlunma:
While testing this again, @bettina-gier and myself got the following error `ZeroDivisionError: integer division or modulo by zero` as already reported [here](https://github.com/ESMValGroup/ESMValCore/issues/2607#issuecomment-2520419014).

<details>
<summary>Full traceback:</summary>

```
2024-12-06 09:38:18,767 UTC [4154330] ERROR   Program terminated abnormally, see stack trace below for more information:
"""
  File "/work/bd0854/b309141/micromamba/envs/test_np1/lib/python3.12/multiprocessing/pool.py", line 125, in worker
                    ^^^^^^^^^^^^^^^^^^^
  File "/home/b/b309141/tmp/ESMValCore/esmvalcore/_task.py", line 895, in _run_task
    output_files = task.run()
                   ^^^^^^^^^^
  File "/home/b/b309141/tmp/ESMValCore/esmvalcore/_task.py", line 290, in run
                        ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/b/b309141/tmp/ESMValCore/esmvalcore/preprocessor/__init__.py", line 730, in _run
  File "/home/b/b309141/tmp/ESMValCore/esmvalcore/preprocessor/__init__.py", line 527, in apply
    self.cubes = preprocess(
                 ^^^^^^^^^^^
  File "/home/b/b309141/tmp/ESMValCore/esmvalcore/preprocessor/__init__.py", line 430, in preprocess
    _run_preproc_function(
  File "/home/b/b309141/tmp/ESMValCore/esmvalcore/preprocessor/__init__.py", line 364, in _run_preproc_function
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/b/b309141/tmp/ESMValCore/esmvalcore/preprocessor/_shared.py", line 237, in wrapper
    result = func(data, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/b/b309141/tmp/ESMValCore/esmvalcore/preprocessor/_time.py", line 826, in climate_statistics
    agg_kwargs = update_weights_kwargs(
                 ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/b/b309141/tmp/ESMValCore/esmvalcore/preprocessor/_shared.py", line 162, in update_weights_kwargs
    callback(cube, **callback_kwargs)
  File "/home/b/b309141/tmp/ESMValCore/esmvalcore/preprocessor/_time.py", line 870, in _add_time_weights_coord
    get_time_weights(cube),
    ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/b/b309141/tmp/ESMValCore/esmvalcore/preprocessor/_shared.py", line 377, in get_time_weights
    time_weights = time_weights.rechunk(time_chunks)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/bd0854/b309141/micromamba/envs/test_np1/lib/python3.12/site-packages/dask/array/core.py", line 2779, in rechunk
    return rechunk(self, chunks, threshold, block_size_limit, balance, method)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/bd0854/b309141/micromamba/envs/test_np1/lib/python3.12/site-packages/dask/array/rechunk.py", line 349, in rechunk
    chunks = normalize_chunks(
             ^^^^^^^^^^^^^^^^^
  File "/work/bd0854/b309141/micromamba/envs/test_np1/lib/python3.12/site-packages/dask/array/core.py", line 3151, in normalize_chunks
    chunks = _convert_int_chunk_to_tuple(shape, chunks)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/work/bd0854/b309141/micromamba/envs/test_np1/lib/python3.12/site-packages/dask/array/core.py", line 3177, in _convert_int_chunk_to_tuple
    return sum(
           ^^^^
  File "/work/bd0854/b309141/micromamba/envs/test_np1/lib/python3.12/site-packages/dask/array/core.py", line 3180, in <genexpr>
    blockdims_from_blockshape((s,), (c,))
  File "/work/bd0854/b309141/micromamba/envs/test_np1/lib/python3.12/site-packages/dask/array/core.py", line 1279, in blockdims_from_blockshape
    return tuple(
           ^^^^^^
  File "/work/bd0854/b309141/micromamba/envs/test_np1/lib/python3.12/site-packages/dask/array/core.py", line 1280, in <genexpr>
    ((bd,) * (d // bd) + ((d % bd,) if d % bd else ()) if d else (0,))
              ~~^^~~~
ZeroDivisionError: integer division or modulo by zero
```

</details>

The reason for this is that the intermediate [`rechunk`](https://github.com/ESMValGroup/ESMValCore/blob/4c36a0c7be2a48b32e1394e6e3212c9f7913a2f7/esmvalcore/preprocessor/__init__.py#L416) between the preprocessor steps create the following chunks after the regrid step / before the climate_statistics steps:

```
chunks before rechunk
((36, 13), (720,), (1440,))
chunks after rechunk
((0, 36, 13), (720,), (1440,))
```

The only non-default preprocessors used here are `regrid` and `climate_statistics`. The leading zero here then causes the error during the calculation of the time weights. The very funny thing here [again](https://github.com/ESMValGroup/ESMValCore/issues/2607) is that this depends on the timerange. The error appears for 3 days, but not 10 days (see recipe)...Also, as reported by Tina, this does not appear in older Dask versions (I am using 2024.12.0 for this test).

@bouweandela do you think this is related this PR here or another bug in Dask. I would really like to merge this here.

Recipe:                                 
[recipe_era5_grib.yml.txt](https://github.com/user-attachments/files/18036066/recipe_era5_grib.yml.txt)

bouweandela:
> do you think this is related this PR here or another bug in Dask

If this PR works fine with the older version of Dask, it's probably a Dask bug. Especially since we just call `rechunk` on a Dask array without any arguments, this looks very Dask internal. It would be really good to open issue(s) about our recent problems with Dask (I believe the size 0 chunks are the third this month?) on the Dask repository to make sure that things get fixed and we can keep using recent versions of Dask.

schlunma:
> It would be really good to open issue(s) about our recent problems with Dask (I believe the size 0 chunks are the third problem you've found in the past month?) on the Dask repository to make sure that things get fixed and we can keep using recent versions of Dask.

The big problem here really is that it's super hard to isolate this. During the last weeks I spent days trying to come up with a minimal example example of a Dask bug which does not use any external packages and couldn't do it. I really hope to finalize this at some point.

So if anyone has the resources to look into this one here, I would be super grateful.

schlunma:
The size 0 chunk problem [will be fixed](https://github.com/dask/dask/pull/11622) in the next Dask release 🎉

PR #2189, opened 2023-08-31
Add normalization with statistics to many statistics preprocessors

### LINKED ISSUE #2185: New preprocessor: subtract global mean from field
A common operation in climate science is to subtract global means from a field. I will implement that.

schlunma:
## Description

This PR adds the argument `normalize` to all statistical preprocessors that are based on `Cube.collapsed` (exceptions: multi-model statistics and ensemble statistics [their call signature is different], and climate statistics [we already have the preprocessor `anomaly`, which does the very same as this PR]). This PR skips statistical preprocessors that use `Cube.aggregated_by` since these need extra care and cannot simply make use of iris' rich broadcasting capabilities.

For example, this is useful to subtract the global mean from a field or divide by the global standard deviation:

```yaml
area_statistics:
  operator: mean
  normalize: subtract
```

Closes #2185

Link to documentation:
- https://esmvaltool--2189.org.readthedocs.build/projects/ESMValCore/en/2189/recipe/preprocessor.html#zonal-statistics
- https://esmvaltool--2189.org.readthedocs.build/projects/ESMValCore/en/2189/recipe/preprocessor.html#meridional-statistics
- https://esmvaltool--2189.org.readthedocs.build/projects/ESMValCore/en/2189/recipe/preprocessor.html#area-statistics
- https://esmvaltool--2189.org.readthedocs.build/projects/ESMValCore/en/2189/recipe/preprocessor.html#volume-statistics
- https://esmvaltool--2189.org.readthedocs.build/projects/ESMValCore/en/2189/recipe/preprocessor.html#axis-statistics

***

***


esmvalcore/preprocessor/_area.py
```diff
@@ -913,3 +913,72 @@ def _mask_cube(cube: Cube, masks: dict[str, np.ndarray]) -> Cube:
...
+    area_stats_cube = area_statistics(cube, **area_statistics_kwargs)
+
+    # Normalize cube
+    # Note: the area statistics cube is always broadcastable to the input cube
+    # due to iris' broadcasting rules enabled by the named dimensions
+    metadata = cube.metadata
+    if operation == 'subtract':
+        cube = cube - area_stats_cube
+    elif operation == 'divide':
+        cube = cube / area_stats_cube
```
valeriupredoi, line 972: do you want to catch a zero-division warning here and perform a check on the `area_stats_cube` if that pops up?

valeriupredoi (approved):
looks spiffy, bud :beer: One tiny comment from me

esmvalcore/preprocessor/_area.py
schlunma: Good point. I just tested this; with the current version of the code this already raises an `RuntimeWarning: invalid value encountered in divide` warning and gives `nan` values in the output. I can add a nicer message here. Would it also make sense to mask the cube instead of the `nan`s? I am not sure, maybe it also makes sense to keep the `nan`s to properly distinguish between the real mask and the zero-divisions.

schlunma: I found a weird problem with iris: https://github.com/SciTools/iris/issues/5523. Sometimes the output of zero divisions is masked, sometimes not. Thus, I chose to mask them consistently here.

valeriupredoi: so in light of that issue, let's wait for the linked iris issue to be fixed before we do anything ourselves - in the lower stacks - it's fine to fix here (I reckon) :beers:

esmvalcore/_recipe/check.py
```diff
@@ -456,3 +460,33 @@ def reference_for_bias_preproc(products):
...
+
+    # Check that area_statistics_kwargs are valid
+    area_stats_kwargs = step_settings.get(
+        'area_statistics_kwargs', {'operator': 'mean'}
+    )
+    area_stats_settings_to_check = {'area_statistics': area_stats_kwargs}
+    try:
+        check_preprocessor_settings(area_stats_settings_to_check)
+    except (TypeError, ValueError) as exc:
+        raise RecipeError(
```
valeriupredoi, line 489: agh - Codacy may be right this time - I am a bit uneasy to mask something that's not TypeError nor ValueError and let it through just as a user warning - that may trigger a fall off the cliff anyway but harder to catch where from

esmvalcore/preprocessor/_area.py
```diff
@@ -913,3 +913,96 @@ def _mask_cube(cube: Cube, masks: dict[str, np.ndarray]) -> Cube:
...
+
+    # Normalize cube
+    # Note: the area statistics cube is always broadcastable to the input cube
+    # due to iris' broadcasting rules enabled by the named dimensions
+    if operation == 'subtract':
+        cube = cube - area_stats_cube
+
+    elif operation == 'divide':
+        # Raise proper zero-division warning if necessary
+        if (cube.core_data() == 0.0).any():
```
valeriupredoi, line 976:
```suggestion
        if (area_stats_cube.core_data() == 0.0).any():
```

valeriupredoi (changes requested):
nice warning! But area stats is the zero troublesome cube :smiley:

esmvalcore/_recipe/check.py
valeriupredoi: it won't terminate it here if it doesn't catch one of the (TypeError, ValueError) - I mean it will, but down the line when things will get stuck somewhere else - if, say, `check_preprocessor_settings(area_stats_settings_to_check)` fails with OSError (for some bizarre reason), the OSError will show up as a warning in the stdout ie "Caught OSError etc etc" but the thing will keep on chugging. Stuff will hit the fan later though

esmvalcore/preprocessor/_area.py
schlunma: Haha, true that!

valeriupredoi (approved):
cheers, Manu! Don't sweat too much about me comments about the RecipeError - all fine, and if the user doesn't know what went wrong, too bad, not my problem :rofl:

esmvalcore/_recipe/check.py
valeriupredoi: yes of course it will! I am a doofus, this is not in another try/except block of its own :man_facepalming:

valeriupredoi:
@bouweandela would you be a darling a merge this, please? (I've been merging your stuff all morning :rofl: )

bouweandela:
@schlunma Would it make sense to apply the same changes here as in #2191? I.e. remove the nesting from the function signature and it to:
```python
def normalize_with_area_statistics(
    cube: Cube,
    operation: str = 'subtract',
    operator: str,
    **kwargs: Any,
) -> Cube:
```
? (Or, if you think having bot operation and operator is confusing, rename one of the two?)

schlunma:
I honestly think it is better the way it is. The keywords here are only relevant for the `area_statistics` part of this preprocessor, whereas in #2191 they are relevant for the actual preprocessors.

I have seen this kind of API in other packages as well, e.g., the `line_kws` in [`seaborn.regplot`](https://seaborn.pydata.org/generated/seaborn.regplot.html) which are only relevant for a part of the function.

schlunma:
I would be nice to merge this PR after #2191; I want to add a proper check for the `area_statistics` keywords here introduced with #2191.

bouweandela:
Thinking ahead a bit, this kind of feature could be implemented by adding just one extra keyword argument to all of the current `*_statistics` preprocessor functions. Maybe people would also be interested in knowing the spread of an ensemble for the `ensemble_statistics` preprocessor function, or normalizing by `volume_statistics`, etc. If we add new preprocessor functions for all of those we end up with many preprocessor functions. Are you really keen on getting this included into v2.10 (today is the deadline for the feature freeze) or would you be willing to think about it a bit more?

schlunma:
I like the idea of generalizing this, but I don't think we should simply add a keyword to the existing `statistics` preprocessors. I think it would be weird if `area_statistics` somehow returned a cube that still has `latitude/longitude` dimensions.

It would probably be clearer to add a `normalize_with_statistics` preprocessor that takes the statistic as argument. Probably easy to implement for most preprocessors except for the multi-model ones since their call signature is different.  I can try to implement this for all preprocessors except for the multi-model ones today. Would that work for you?

bouweandela:
That would probably be nicer than this approach, right?

schlunma:
One issue with this generalized approach is that `normalize_with_statistics` requires different supplementary variables dependent on the statistic used (this is given as an argument to the function). Currently, the type of supplementary variable that a preperocessor uses is given with a [decorator](https://github.com/ESMValGroup/ESMValCore/blob/87e3f9fe95d2496222fb4029ce1e13feed1752b1/esmvalcore/preprocessor/_normalize.py#L18-L21), which knows nothing about the preprocessor arguments. In addition, if this `normalize_with_statistics` is used multiple times in the recipe, it might require different supplementary variables. This is also a problem for the current implementation, since the necessary supplementaries are [stored in a dictionary with the name of the preprocessor as key](https://github.com/ESMValGroup/ESMValCore/blob/87e3f9fe95d2496222fb4029ce1e13feed1752b1/esmvalcore/preprocessor/_supplementary_vars.py#L36).

@bouweandela any ideas how to solve this smartly without changing our entire code base?

bouweandela:
In that case, adding the normalize option as an extra keyword argument to existing preprocessor functions may be a more attractive solution because it avoids all these technical issues. I agree that it may feel a bit counter-intuitive that the returned cube can have different dimensions based on the arguments though. Maybe we could ask the community for their opinion at the next monthly community meeting?

bouweandela:
I just came across discussion https://github.com/ESMValGroup/ESMValTool/discussions/3353 where this seems relevant.

schlunma:
> In that case, adding the normalize option as an extra keyword argument to existing preprocessor functions may be a more attractive solution because it avoids all these technical issues.

Agreed, that would be much simpler. I'll try to implement something like this for v2.11 because I need this new feature.

> I just came across discussion [ESMValGroup/ESMValTool#3353](https://github.com/ESMValGroup/ESMValTool/discussions/3353) where this seems relevant.

I don't think that's relevant if I understand the problem correctly (weighting != normalization).

doc/recipe/preprocessor.rst
```diff
@@ -1921,6 +1925,10 @@ argument:
 Parameters:
     * `operator`: Operation to apply.
       See :ref:`stat_preprocs` for more details on supported statistics.
+    * `normalize_with_stats`: If given, do not return the statistics cube
+      itself, but rather the input cube normalized with the statistics cube.
```
valeriupredoi, line 1929:
```suggestion
      itself, but rather, the input cube, normalized with the statistics cube.
```

valeriupredoi (approved):
I believe I approved this PR once, in ancient times, not sure what happened in the meantime - still looks good to me, but I *think* we should dd a test for the volume normalization case too. At any rate, looking good, bar missing a few commas (yes, I am that kind of reviewer haha) :grin: :beer:

valeriupredoi:
oh sorry, Manu - I got alerted I am up for providing a review - just saw it's still a draft - carry on, then

schlunma:
This is ready for review now

valeriupredoi (approved):
this could be the fourth time I approve, but hey better more than never :grin:

valeriupredoi:
not willing to approve this yet another time, I'll pass the ball to @bouweandela - if you are happy with all the stuff here, please merge, bud :beer:

valeriupredoi:
frenly ping @bouweandela :baby_chick:

esmvalcore/preprocessor/_shared.py
```diff
@@ -181,3 +182,73 @@ def update_weights_kwargs(
...
+    Cube
+        Input cube normalized with statistics cube.
+
+    """
+    if normalize_with_stats == 'subtract':
+        normalized_cube = cube - statistics_cube
+
+    elif normalize_with_stats == 'divide':
+        # Raise proper zero-division warning if necessary
+        if (statistics_cube.core_data() == 0.0).any():
```
bouweandela, line 221: This kind of check can be computationally expensive if the cube is large. Instead of checking and warning at runtime, I would recommend just documenting the behaviour in the docstring (i.e. write that normalization with 'divide' will mask invalid values) and not do any checks at runtime because you're going to do this anyway and chances are real that users will not see the warning anyway in between the hundreds of warnings coming from iris.

doc/recipe/preprocessor.rst
```diff
@@ -1954,11 +1954,15 @@ along the longitude coordinate.
 Parameters:
     * `operator`: Operation to apply.
       See :ref:`stat_preprocs` for more details on supported statistics.
+    * `normalize_with_stats`: If given, do not return the statistics cube
```
bouweandela, line 1957: Would you be ok with calling this argument just `normalize` or `normalize_operator` instead? The abbreviation `stats` is for `statistics` is not immediately clear to me and writing `normalize_with_statistics` seems a bit long.

esmvalcore/preprocessor/_area.py
```diff
@@ -221,14 +233,19 @@ def zonal_statistics(
...
         )
     (agg, agg_kwargs) = get_iris_aggregator(operator, **operator_kwargs)
-    cube = cube.collapsed('longitude', agg, **agg_kwargs)
-    cube.data = cube.core_data().astype(np.float32, casting='same_kind')
-    return cube
+    stat_cube = cube.collapsed('longitude', agg, **agg_kwargs)
+    if normalize_with_stats is not None:
+        result = get_normalized_cube(cube, stat_cube, normalize_with_stats)
+    else:
+        result = stat_cube
```
bouweandela, line 240: A bit of a nitpick, but I would find the following easier to read
```suggestion
    result = cube.collapsed('longitude', agg, **agg_kwargs)
    if normalize_with_stats is not None:
        result = get_normalized_cube(cube, result, normalize_with_stats)
```

bouweandela (commented):
Thanks for making all the changes @schlunma! Looks great already, just a few more suggestions for making it even better. Feel free to merge as-is if you feel they're not relevant.

doc/recipe/preprocessor.rst
schlunma: Still feels weird to me to call that simply `normalize`, but the current one is also certainly not optimal. Will change to `normalize`.

bouweandela: Thanks!

PR #2217, opened 2023-10-04
Check type of argument passed to `esmvalcore.cmor.table.read_cmor_tables`

### LINKED ISSUE #2216: `esmvalcore.cmor.table.read_cmor_tables()` does not restrict type of argument, consequently gets confused and kicks bucket
`read_cmor_tables(cfg_developer: Optional[Path] = None) -> None:` needs to restrict type of argument or check if argument is of type `Path` before doing anything - instead, it happily accepts `cfg_developer` of type `Dict` and crashes when getting `stat()` method, coz la-dee-dah `Dict` obejcts don't have `stat()` methods, see bkwards incompatible failed [tests](https://github.com/ESMValGroup/ESMValTool/actions/runs/6345829794/job/17238451504) in ESMValTool

valeriupredoi:
## Description

`read_cmor_tables` is happy to accept any type of argument, but is only happy if the arg is of type Path - no checks anywhere up in the stack make it fail with something the user would be a bit baffled about, so here's a check.

Closes #2216 

Link to documentation:

***

***


zklaus:
Is this about the nightly CI fails? Please note #2214. I think that is addressing the problem.

valeriupredoi:
no @zklaus - slightly different, but involving changes from the same PR by Bouwe - it's a loose typing case in `cmor.table.read_cmor_tables` that lets `cfg_developer` be whatever the Pope wants and no check is performed on its type - it's public API so this should be a bit more tight IMHO - note that it's not a strictly needed check but it'd be *nice* to have it

esmvalcore/cmor/table.py
```diff
@@ -124,9 +124,17 @@ def read_cmor_tables(cfg_developer: Optional[Path] = None) -> None:
...
+    ------
+    TypeError
+        If `cfg_developer` is not a Path-like object
     """
     if cfg_developer is None:
         cfg_developer = Path(__file__).parents[1] / 'config-developer.yml'
+    elif not isinstance(cfg_developer, Path):
+        raise TypeError("cfg_developer is not a Path-like object, got ",
+                        cfg_developer)
     mtime = cfg_developer.stat().st_mtime
```
bouweandela, line 138: maybe just do
```suggestion
    mtime = Path(cfg_developer).stat().st_mtime
```
instead of the above? Or do you prefer the specific error message?

valeriupredoi: I like the specific catch - raise that I put in - a bit more friendly for users since it prints out what they are passing in, also we  may want to tailor that in the future and accept other types of config developer?

bouweandela: > we may want to tailor that in the future and accept other types of config developer?

Let's not do that, that would be awful ;-)

bouweandela (approved):

bouweandela:
I removed the `bug` label because this is just adding a nicer error message, it doesn't really solve a bug.

PR #2219, opened 2023-10-05
Add file encoding (and some read modes) at open file step

### LINKED ISSUE #1585: Specify encoding when reading YAML and other text files
When reading a YAML or other text file, it would be best if we specify the encoding, for example:
```python
with open('/path/to/file.yml', encoding='utf-8') as file:
    content = yaml.safe_load(file)
```

this prevents problems such as https://github.com/ESMValGroup/ESMValTool/issues/973 and https://github.com/ESMValGroup/ESMValTool/issues/2652.

valeriupredoi:
## Description

As discussed in https://github.com/ESMValGroup/ESMValCore/issues/1585 we should specify the type of encoding (bogstandard UTF-8 here) when opening file objects; binary open does NOT require that (in fact enforces it to be binary only), so I added encodings to all the YAML files we open, and some read mode 'r' when needed.

Closes #1585 

***

***

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -62,7 +62,8 @@ jobs:
       - run: flake8
       - run: |
           locale
-          sudo update-locale LC_ALL=C
+          echo "export LC_ALL=C" >> $GITHUB_ENV
+      - run: |
           locale
           pytest -n 2 -m "not installation" 2>&1 | tee test_linux_artifacts_python_${{ matrix.python-version }}/test_report.txt
       - name: Upload artifacts
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -21,6 +21,7 @@ on:
   push:
     branches:
       - main
+      - add_file_encoding
   # run the test only if the PR is to main
   # turn it on if required
   #pull_request:
@@ -59,7 +60,10 @@ jobs:
       - run: python -V 2>&1 | tee test_linux_artifacts_python_${{ matrix.python-version }}/python_version.txt
       - run: pip install -e .[develop] 2>&1 | tee test_linux_artifacts_python_${{ matrix.python-version }}/install.txt
       - run: flake8
-      - run: pytest -n 2 -m "not installation" 2>&1 | tee test_linux_artifacts_python_${{ matrix.python-version }}/test_report.txt
+      - run: |
+          sudo update-locale LC_ALL=C
+          locale
+          pytest -n 2 -m "not installation" 2>&1 | tee test_linux_artifacts_python_${{ matrix.python-version }}/test_report.txt
       - name: Upload artifacts
         if: ${{ always() }}  # upload artifacts even if fail
         uses: actions/upload-artifact@v2
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -21,7 +21,6 @@ on:
   push:
     branches:
       - main
-      - add_file_encoding
   # run the test only if the PR is to main
   # turn it on if required
   #pull_request:
@@ -60,12 +59,7 @@ jobs:
       - run: python -V 2>&1 | tee test_linux_artifacts_python_${{ matrix.python-version }}/python_version.txt
       - run: pip install -e .[develop] 2>&1 | tee test_linux_artifacts_python_${{ matrix.python-version }}/install.txt
       - run: flake8
-      - run: |
-          locale
-          echo "export LC_ALL=C" >> $GITHUB_ENV
-      - run: |
-          locale
-          pytest -n 2 -m "not installation" 2>&1 | tee test_linux_artifacts_python_${{ matrix.python-version }}/test_report.txt
+      - run: pytest -n 2 -m "not installation" 2>&1 | tee test_linux_artifacts_python_${{ matrix.python-version }}/test_report.txt
       - name: Upload artifacts
         if: ${{ always() }}  # upload artifacts even if fail
         uses: actions/upload-artifact@v2
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -61,6 +61,7 @@ jobs:
       - run: pip install -e .[develop] 2>&1 | tee test_linux_artifacts_python_${{ matrix.python-version }}/install.txt
       - run: flake8
       - run: |
+          locale
           sudo update-locale LC_ALL=C
           locale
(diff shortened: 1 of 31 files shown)


valeriupredoi:
what's Codecov's deal? It's very confused

bouweandela:
Great to see some action on this topic! `grep -RIn 'open(' $(find -name '*.py')` still finds quite a few file opens is missing even with the changes here. Could you have another look?

valeriupredoi:
@bouweandela cheers! yes, but:
- whatever is open for writing there is no need to specify the encoding since the Python writer knows exactly what to use
- JSON and image opening I'd rather stay away from forcing the encoding and let Python choose it
- bytes mode opening should not have encoding specified

This leaves a couple open file calls which I'll look into now :+1:

bouweandela:
We could check if the problem is solved by doing `export LC_ALL=C` and running the tests, right?

valeriupredoi:
you do that, I am not sure if I can revert the locale vars on my ancient OS :rofl: :rofl: JK, can you try pls - while I am looking at the other calls that need encoding

valeriupredoi:
eugh I can't change my `locale`:
```
(esmvaltool) valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ sudo update-locale LC_CTYPE="C"
(esmvaltool) valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ locale
LANG=en_GB.UTF-8
LANGUAGE=en_GB:en
LC_CTYPE="en_GB.UTF-8"
LC_NUMERIC="en_GB.UTF-8"
LC_TIME="en_GB.UTF-8"
LC_COLLATE="en_GB.UTF-8"
LC_MONETARY="en_GB.UTF-8"
LC_MESSAGES="en_GB.UTF-8"
LC_PAPER="en_GB.UTF-8"
LC_NAME="en_GB.UTF-8"
LC_ADDRESS="en_GB.UTF-8"
LC_TELEPHONE="en_GB.UTF-8"
LC_MEASUREMENT="en_GB.UTF-8"
LC_IDENTIFICATION="en_GB.UTF-8"
LC_ALL=
```
very possible since OS is *a little* old :grin:

valeriupredoi:
OK @bouweandela - I think we should be OK - I got GA to do that, locale to `C.UTF-8` and things look good test-wise

.github/workflows/run-tests.yml
```diff
@@ -59,7 +60,10 @@ jobs:
       - run: python -V 2>&1 | tee test_linux_artifacts_python_${{ matrix.python-version }}/python_version.txt
       - run: pip install -e .[develop] 2>&1 | tee test_linux_artifacts_python_${{ matrix.python-version }}/install.txt
       - run: flake8
-      - run: pytest -n 2 -m "not installation" 2>&1 | tee test_linux_artifacts_python_${{ matrix.python-version }}/test_report.txt
+      - run: |
+          sudo update-locale LC_ALL=C
```
zklaus, line 65: What is this `update-locale` thing? You shouldn't need to try and change the default locale for the whole machine. Just do an `export LC_ALL=C`?

valeriupredoi:
Aye, am gonna add encoding to everything then - just to be on the safe side

.github/workflows/run-tests.yml
valeriupredoi: Meh. Gonna add encoding to all open file calls anyway

zklaus:
> Aye, am gonna add encoding to everything then - just to be on the safe side

Hm. I know this is something that @bouweandela also brought up, but I feel a bit more cautious about it. For Yaml, the case is clear cut; the Yaml specs dictate UTF(-8). For all other text files, it is perfectly reasonable to use different encodings, isn't it?

Also, a side question: Why are we ever opening Yaml files in binary mode?

valeriupredoi:
@zklaus indeed that was me concern too in https://github.com/ESMValGroup/ESMValCore/pull/2219#issuecomment-1748985891 - but me not being in the know of file encodings I assumed @bouweandela knows a tad me, so I was thinking of following his advice - how about I do that and test with export LC_ALL=C on GAs?

valeriupredoi:
OK @bouweandela this should now cover all instances of open file, except shapefiles, bytes, and images - I really don't want to faff around those (wb and rb strictly don't accept encoding anyway) :beer:

.github/workflows/run-tests.yml
bouweandela: The thing we want to test is that it works fine with a default character set other than utf-8, as that is what we keep running into (see  linked issues).

bouweandela:
> Hm. I know this is something that @bouweandela also brought up, but I feel a bit more cautious about it. For Yaml, the case is clear cut; the Yaml specs dictate UTF(-8). For all other text files, it is perfectly reasonable to use different encodings, isn't it?

The problem is that Python uses the default encoding setting of the machine that it is running on instead of the encoding of the file, as there is no way to specify that, and therefore this causes problems when the system encoding is set to something ancient. To avoid this issue, we need to specify the encoding in all text files we are reading and writing so they work across platforms and users.

> Also, a side question: Why are we ever opening Yaml files in binary mode?

It looks like `yaml` has some code to automatically try to figure out the correct encoding in case you do this, so that should work fine: https://github.com/yaml/pyyaml/blob/155ec463f6a854ac14ccd5e2dda8017ce42a508a/lib/yaml/reader.py#L122

bouweandela (approved):

valeriupredoi:
Cheers for chipping in @bouweandela 🍺 Thing is, don't know how to change my UNIX default encoding without having to use a Klingon translator to manage commands in the terminal 😁

bouweandela:
>  Thing is, don't know how to change my UNIX default encoding

@valeriupredoi I think you could do this (haven't tried it myself): https://askubuntu.com/a/120068. So basically you 1) make sure that the locale has been generated with the requested character set and then 2) export the `LANG` and `LC_ALL environmental variables to use it. After closing the terminal the changes should be gone because your environmental variables have been restored to normal, so no need to worry about messing up your system.

valeriupredoi:
thanks @bouweandela but you assume I can still run `dpkg-*` which is a false assumption on my abandonware OS :rofl:

valeriupredoi:
many thanks for approving and merging, gents - let's wait see if some people with bizarre encodings complain, then we can get back to it, by then I should be running a modern OS too, so I can test :grin:

bouweandela:
I just checked and the command is available in Ubuntu 14.04. I believe that is your favorite OS, isn't it? :rofl:

valeriupredoi:
no, I "evolved" - I am now a big fan of 16.04 :rofl:

PR #2231, opened 2023-10-12
Refactor regridding

schlunma:
## Description

This PR refactors our regridding. The main purpose is to simplify the code so that new schemes can be added more easily, which will be necessary for https://github.com/ESMValGroup/ESMValCore/pull/2178.

### Features

1. At the moment, the schemes `linear`, `area_weighted` and `nearest` work for regular and irregular grids. However, for unstructured grids, we have to use `unstructured_nearest`. This PR unifies this behavior so that `nearest` can also be used for unstructured grids (just like for irregular grids, I added a check for unstructured grids [here](https://github.com/ESMValGroup/ESMValCore/pull/2231/files#diff-940f3c2344a86b95804366fef409006557427f90affe7313b31af9c7d6ba6655R542) that automatically selects the correct regridding scheme for unstructured grids).

2. In addition, this PR unifies all regridding schemes so they can be used the (iris-intended) way within [`Cube.regrid`](https://scitools-iris.readthedocs.io/en/latest/generated/api/iris.cube.html#iris.cube.Cube.regrid), just like the regridding schemes provided by iris. These are now located in the new [`esmvalcore.preprocessor.regridding_schemes`](https://esmvaltool--2231.org.readthedocs.build/projects/ESMValCore/en/2231/api/esmvalcore.regridding_schemes.html#) module. With this, regridding is as easy as

```py
loaded_scheme = _load_scheme(cube, scheme)
cube = cube.regrid(target_grid_cube, loaded_scheme)
```

3. Finally, this PR adds a check of the validity of the specified regridding schemes early in an ESMValTool run, so that the tool won't fail because of a bad scheme halfway through the recipe (potentially after hours of running).

### Deprecations

This PR deprecated two regridding schemes, which will be removed with ESMValCore v2.13.0:

- `unstructured_nearest`: Please use the scheme `nearest` instead. This is an exact replacement for data on unstructured grids. ESMValCore is now able to determine the most suitable regridding scheme based on the input data.
- `linear_extrapolate`: Please use a generic scheme with `reference: iris.analysis:Linear` and `extrapolation_mode: extrapolate` instead. This is an exact replacement, e.g.:

```yaml
preprocessors:
      regrid_preprocessor:
        regrid:
          target_grid: 2.5x2.5
          scheme:
            reference: iris.analysis:Linear
            extrapolation_mode: extrapolate
```

Related to https://github.com/ESMValGroup/ESMValCore/pull/2178

Link to documentation:
- https://esmvaltool--2231.org.readthedocs.build/projects/ESMValCore/en/2231/recipe/preprocessor.html#built-in-regridding-schemes
- https://esmvaltool--2231.org.readthedocs.build/projects/ESMValCore/en/2231/api/esmvalcore.regridding_schemes.html#

***

- [ ] [🛠][1] ~Changes are [backward compatible](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#backward-compatibility)~ no: see deprecations above
- [ ] [🛠][1] ~All [checks below this pull request](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-checks) were successful~ Remaining Codacy checks about `Too few public methods (1/2) (too-few-public-methods)` cannot be fixed; it is necessary to use classes here due to the way iris regridding schemes are defined.

***


valeriupredoi (changes requested):
hey Manu, no actual changes needed, just a couple questions for now - the review block is more for me since I'm not done going through the code yet (just got to `_regrid_esmpy`

esmvalcore/preprocessor/_regrid_esmpy.py
```diff
@@ -39,6 +40,130 @@
...
+        _ESMPyRegridder
+            Regridder instance.
+
+        """
+        return _ESMPyRegridder(
+            src_cube,
+            tgt_cube,
+            method=self._METHOD,
+            mask_threshold=self.mask_threshold,
+        )
```
valeriupredoi, line 146: OK call me old fashioned but to me this is a bit wrong: public method in a private class is something I don't agree with, especially since the `ESMPYScheme` name doesn't reflect the fact that an actual regridder instance is built within its bellows. What's wrong with the old generic and public `regrid` wrapper that the user is exposed to?

esmvalcore/preprocessor/regrid_schemes.py
```diff
@@ -0,0 +1,118 @@
...
+__all__ = [
+    'ESMPyAreaWeighted',
+    'ESMPyLinear',
+    'ESMPyNearest',
+    'GenericFuncScheme',
+    'UnstructuredNearest',
+]
+
+
+class _GenericRegridder:
```
valeriupredoi, line 28: make this public

esmvalcore/preprocessor/regrid_schemes.py
```diff
@@ -0,0 +1,118 @@
...
+        tgt_cube:
+            Cube defining the target grid.
+
+        Returns
+        -------
+        _GenericRegridder
+            Regridder instance.
+
+        """
+        return _GenericRegridder(src_cube, tgt_cube, self.func, **self.kwargs)
```
valeriupredoi, line 118: ah I see why the one above is private! OK - I'd argue you can coalesce that class into this one, and get rid of `_GenericRegridder`, and this is already public and does all the bits one needs

valeriupredoi (changes requested):
this is a very nice unification PR, Manu, many thanks! I am very happy to see this happening, but I have a few concerns related to the way how things are structured ie private/public access to various classes and methods, and a bit of extra complexity that IMHO we can do without. Nothing major tho :beer:

esmvalcore/preprocessor/_regrid_esmpy.py
schlunma: I agree with you that this looks too complicated, but this is how iris expects regridding schemes to look like! The scheme class needs a public method `regridder(self, src_cube, target_grid)` that returns an object of a regridder class. This regridder class needs to have a method `def __call__(self, cube)` that actually performs the regridding. [Here](https://github.com/SciTools/iris/blob/8694479607d5967561affa05b85a6b8fb49477df/lib/iris/analysis/__init__.py#L3083-L3125) is an example for a regridder in iris.

The advantage of this new syntax is that it's much more generic. All of our regridding schemes are used the same way (within `iris.cube.Cube.regrid()`. In the old version, some schemes used the iris syntax, some use this `regrid` function that you mentiond, etc.

esmvalcore/preprocessor/regrid_schemes.py
schlunma: See comment above about how iris implements regridders. This needs to stay like this.

esmvalcore/preprocessor/_regrid.py
```diff
@@ -470,59 +486,223 @@ def is_dataset(dataset):
...
+    """Check if irregular regridding with ESMF should be used."""
+    if isinstance(scheme, str) and scheme in HORIZONTAL_SCHEMES_IRREGULAR:
+        try:
+            lat_dim = cube.coord('latitude').ndim
+            lon_dim = cube.coord('longitude').ndim
+            if lat_dim == lon_dim == 2:
+                return True
+        except iris.exceptions.CoordinateNotFoundError:
+            pass
+    return False
```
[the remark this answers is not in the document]
schlunma: I actually like this idea: implemented a public `has_irregular_grid` in https://github.com/ESMValGroup/ESMValCore/pull/2231/commits/13f7499013c6fc8eafdc83c514e89eb57c13ab1d, now both "special" grid types are treated similarly :+1:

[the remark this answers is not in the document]
valeriupredoi: yay! So my rant wasn't that useless after all :grin:

esmvalcore/preprocessor/_regrid_esmpy.py
valeriupredoi: indeed - I caught that, and liked it - but here's the thing - aligning ourselves to iris is not something we should aim for. If we find a nice, generic, and robust way of writing code and functionality we should go that way, regardless iris do other things; in this case, it's prob useful since we are in essence wrapping our regridding around iris regridding, so I'll grab the bait :fish:

esmvalcore/preprocessor/regrid_schemes.py
valeriupredoi: mhmm, :fish:

esmvalcore/iris_helpers.py
```diff
@@ -157,3 +157,29 @@ def merge_cube_attributes(
...
+    -------
+    bool
+        ``True`` if input cube has an unstructured grid, else ``False``.
+
+    """
+    try:
+        lat = cube.coord('latitude')
+        lon = cube.coord('longitude')
+    except CoordinateNotFoundError:
+        return False
```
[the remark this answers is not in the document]
valeriupredoi: OK, second thought - nevermind, it's better this way :beer:

valeriupredoi (approved):
approving this with very many thanks to Manu for the (rather heavy lifting) work, but I'd like a second pair of reviewing eyes, specifically @bouweandela or @zklaus please - Manu could you hold fire (merge, that is) until one of those two fellas have looked at it too, if you please, bud :beer:

esmvalcore/preprocessor/regrid_schemes.py
```diff
@@ -0,0 +1,118 @@
...
+        Returns
+        -------
+        Cube
+            Regridded cube.
+
+        """
+        return self.func(cube, self.tgt_cube, **self.kwargs)
+
+
+class GenericFuncScheme:
```
bouweandela, line 77: Is this scheme useful to users of the Python API? My understanding is that generic schemes are useful to define the scheme in the recipe, but what does it add for the Python user? If that is correct, we could make this scheme private.

esmvalcore/preprocessor/_regrid_esmpy.py
```diff
@@ -39,6 +40,130 @@
...
+
+
+class ESMPyAreaWeighted(_ESMPyScheme):
```
bouweandela, line 149: The `regridder` method of this class (and also for the other schemes in this file) returns an `_ESMPyRegridder` according to the documentation. It would be nice to avoid leaking private classes into the documentation, would it be possible to return something public instead? And hide the base class because it is also private? I believe there is a keyword that you can add for that.

esmvalcore/preprocessor/_regrid.py
```diff
@@ -563,105 +742,30 @@ def regrid(cube, target_grid, scheme, lat_offset=True, lon_offset=True):
...
 
     # Horizontal grids from source and target (almost) match
     # -> Return source cube with target coordinates
-    if _horizontal_grid_is_close(cube, target_grid):
+    if _horizontal_grid_is_close(cube, target_grid_cube):
         for coord in ['latitude', 'longitude']:
-            cube.coord(coord).points = target_grid.coord(coord).points
-            cube.coord(coord).bounds = target_grid.coord(coord).bounds
+            cube.coord(coord).points = target_grid_cube.coord(coord).points
+            cube.coord(coord).bounds = target_grid_cube.coord(coord).bounds
```
bouweandela, line 755: Not new in the pull request, but maybe it would be nice not to realize the coordinates here?
```suggestion
            cube.coord(coord).points = target_grid_cube.coord(coord).core_points()
            cube.coord(coord).bounds = target_grid_cube.coord(coord).core_bounds()
```
and what abound the bounds, why are they not updated?

esmvalcore/preprocessor/_regrid.py
```diff
@@ -49,21 +68,31 @@
 _LON_RANGE = _LON_MAX - _LON_MIN
 
 # A cached stock of standard horizontal target grids.
-_CACHE: Dict[str, iris.cube.Cube] = {}
+_CACHE: Dict[str, Cube] = {}
```
bouweandela, line 71: Use @functools.lru_cache instead?

esmvalcore/preprocessor/regrid_schemes.py
schlunma: I don't have a strong opinion on this. This class is only used if a regridding function `f(src_cube, grid_cube, **kws) -> Cube` is available (which is not the case for every generic regridding scheme). This can then be turned into an iris-compatible regridder, which might be helpful for users. Your call.

esmvalcore/preprocessor/_regrid.py
schlunma: But from the code it looks like the bounds are update, aren't they?

esmvalcore/preprocessor/_regrid.py
schlunma: Done in https://github.com/ESMValGroup/ESMValCore/pull/2231/commits/969084d759437fb27ffc0a89816bb07034c4d574

esmvalcore/preprocessor/_regrid.py
schlunma: Done in https://github.com/ESMValGroup/ESMValCore/pull/2231/commits/cdb1f47e5296893c97003043f403e7dcb2a1e3a8

esmvalcore/preprocessor/_regrid_esmpy.py
```diff
@@ -39,6 +40,148 @@
...
+    """General ESMPy regridder.
+
+    Parameters
+    ----------
+    src_cube:
+        Cube defining the source grid.
+    tgt_cube:
+        Cube defining the target grid.
+    method:
+        Regridding algorithm.
```
bouweandela, line 53: It would be nice to mention the valid method names here

bouweandela (approved):
Thanks for these improvements @schlunma!

schlunma:
Thanks for the reviews @bouweandela and @valeriupredoi ! Merging this now since two people approved.

valeriupredoi:
yay :partying_face:

Issue #2234, opened 2023-10-17
Uas and vas cmor time fixes in ERA5


malininae:
Hi folks, while working on the #1895 I've noticed that `uas` and `vas` time fixes for the hourly data in ERA are missing, I'm adding them. PR to follow.

malininae:
The fix is not needed, the `uas` and `vas` are instantaneous variables so the time shift is not required. See [this comment](https://github.com/ESMValGroup/ESMValCore/pull/2242#discussion_r1372759053)

Issue #1895, opened 2023-01-17
Wind speed derivation from 'uas' and 'vas'


malininae:
Hi, 

I might have missed something, but I was wondering if there's interest in a `sfcwind` variable derivation from `uas` and `vas`? It's may be not super important for the CMIP data, but it is for ERA5 (and may be other reanalysis or observation). In couple of weeks, I might come to that, if others are supportive.

valeriupredoi:
cheers, Liza! I support anything that scientists may need - I can assist with the PR :+1: How do you think @ESMValGroup/scientific-lead-development-team ?

malininae:
@valeriupredoi the other option is to create `sfcWind` directly in ERA5 cmorizer, but I'm not sure it's such a good idea. As said, I'll come to that hopefully next week. I hope to join this month's meeting, can briefly discuss that issue there.

katjaweigel:
Thanks for the issue @malininae. I would be also interested (thought about it before but never calculated it so far). I think one issue we must be aware of, is the nonlinearity of the calculation. It should be fine to calculate hourly sfcWind from hourly uas and vas, but I'm not sure, how the daily/monthly sfcWind is defined, i.e. if it needs to be calculated as mean of the hourly sfcWind or if we can use the monthly mean uas and vas directly?

malininae:
> Thanks for the issue @malininae. I would be also interested (thought about it before but never calculated it so far). I think one issue we must be aware of, is the nonlinearity of the calculation. It should be fine to calculate hourly sfcWind from hourly uas and vas, but I'm not sure, how the daily/monthly sfcWind is defined, i.e. if it needs to be calculated as mean of the hourly sfcWind or if we can use the monthly mean uas and vas directly?

@katjaweigel, a very good point! I did a brief analysis of the daily `sfcWind` for one extratropical storm comparing ERA5 and HighResMIP models, but what I did there, I processed ERA5 in the diagnostic (again, naughty me) and calculated hourly `sfcWind` and then took a daily mean from there. It was a quick analysis, just to get a rough estimate, but I wanted to look into it further.  I'll need to double check the script, but isn't ERA5 cmorized as daily `uas` and `vas`? So may be the cmorizer should be done for the hourly tables, or indeed we might want to add the `sfcWind` calculation into the ERA5 cmorizer.

katjaweigel:
[...]

PR #2242, opened 2023-10-23
Adding sfcWind derivation from uas and vas 

malininae:
Following #1895 and #2234 the `sfcWind` derivation from `uas` and `vas` has been added and the wind components fixes for `uas` and `vas` from ERA5 has been added. NB: it was impossible to add a `height10m` to the custom CMOR table.


valeriupredoi:
@malininae many thanks for this! Looks good to me, but the time coordinate of the fixed cube is different from the time coordinate in the cmor cube:
```
cmor_cube coords:
DimCoord :  time / (days since 1850-1-1 00:00:00, standard calendar)
    points: [1990-01-01 00:00:00, 1990-01-01 01:00:00, 1990-01-01 02:00:00]
    shape: (3,)
    dtype: float64
    standard_name: 'time'
    long_name: 'time'
    var_name: 'time'
 ```
 vs
 ```
 fixed_cube coords:
DimCoord :  time / (days since 1850-1-1 00:00:00, standard calendar)
    points: [1989-12-31 23:30:00, 1990-01-01 00:30:00, 1990-01-01 01:30:00]
    shape: (3,)
    dtype: float64
    standard_name: 'time'
    long_name: 'time'
    var_name: 'time'
 ```

**Review by valeriupredoi — CHANGES_REQUESTED** [review]
time coord of fixed cube needs looking into, apart from that all fine :beer:

File: esmvalcore/preprocessor/_derive/__init__.py [hunk 1]
```diff
@@ -52,11 +52,11 @@ def get_required(short_name, project):
         List of dictionaries (including at least the key `short_name`).
 
     """
-    if short_name not in ALL_DERIVED_VARIABLES:
+    if short_name.lower() not in ALL_DERIVED_VARIABLES:
```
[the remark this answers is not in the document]
malininae: The wind variable in a recipe would be spelled `sfcWind`, so the `.lower()` part would be required to marry the input from the recipe to the preprocessor.

malininae:
> time coord of fixed cube needs looking into, apart from that all fine 🍺

Is this the output from ERA? If yes, I've added the 30 mins offset for the hourly `uas` and `vas`

esmvalcore/preprocessor/_derive/__init__.py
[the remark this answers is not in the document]
valeriupredoi: `short_name` comes from the CMOR table, so that will always be lower case - as I recall, it doesn't come from the recipe

valeriupredoi:
> > time coord of fixed cube needs looking into, apart from that all fine 🍺
> 
> Is this the output from ERA? If yes, I've added the 30 mins offset for the hourly `uas` and `vas`

indeed! See https://github.com/ESMValGroup/ESMValCore/pull/2242#issuecomment-1777503052 - exactly half hour early :grin: So I guess this is the fix motivated by #2234 - so please adjust the test cube as well, so that the test reflects the correct time points. In the future it'd be good if you opened a separate PR per issue or fix needing solving, things are a bit more clear that way :+1:

esmvalcore/preprocessor/_derive/__init__.py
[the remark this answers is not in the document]
malininae: Interesting, from the quick look, the most variables indeed are spelled in lowercase, so wind seems to be an exception:  (see e.g.  https://github.com/ESMValGroup/ESMValCore/blob/bdad237ded5e42e915ce05ca7d284b10eef30eb4/esmvalcore/cmor/tables/cmip6/Tables/CMIP6_day.json#L431C18-L431C18). That being said, as I mentioned in the discussion, I am happy to rename it as a `der_sfcwind` highlighting that this is a wind derived from `uas` and `vas` and not the native `sfcWind` :-)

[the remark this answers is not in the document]
valeriupredoi: sorry to be pedantic, Liza, but as you can see from the table `"standard_name": "wind_speed"` - I am fairly sure all standard names need be lower case in CMOR format, but @zklaus (our resident CMOR/standards specialist) can correct me if I is wrong :smiley:

valeriupredoi:
> > > > time coord of fixed cube needs looking into, apart from that all fine 🍺
> > > 
> > > 
> > > Is this the output from ERA? If yes, I've added the 30 mins offset for the hourly `uas` and `vas`
> > 
> > 
> > indeed! See [#2242 (comment)](https://github.com/ESMValGroup/ESMValCore/pull/2242#issuecomment-1777503052) - exactly half hour early 😁 So I guess this is the fix motivated by #2234 - so please adjust the test cube as well, so that the test reflects the correct time points. In the future it'd be good if you opened a separate PR per issue or fix needing solving, things are a bit more clear that way 👍
> 
> Sorry!!! I thought since they are related better to handle them in one pull request, will do better the next time 😊 By adjusting the test cube you mean the derived one? If yes, I'm currently applying the function to the models too.

not a problem at all - just for future reference :beer: 

As for the failing tests, the `uas` and `vas` CMOR times don't match the fixed times, please have a look at how the time coordinate gets created for the CMOR cubes via `time = _cmor_time('E1hr')` in the `vas` and `uas` cases - I, for one, am not sure which is the correct one, it seems to me that 30min offset should not break the world, but you are the one to decide, in any case, either the function that assembles the CMOR time or the fix has to change :+1:

valeriupredoi:
fgfhrtsrwter

esmvalcore/preprocessor/_derive/__init__.py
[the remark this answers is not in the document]
valeriupredoi: and a massive :man_facepalming: to me since all this time I kept yapping about `standard_name` when in fact Liza is fixing a problem with `short_name` - sorry, Liza, my bad, and a slip of attention!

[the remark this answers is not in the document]
valeriupredoi: I even wrote `short_name` once in a comment, was still thinking of `standard_name` :rofl:

esmvalcore/cmor/_fixes/native6/era5.py
```diff
@@ -334,6 +334,26 @@ def fix_metadata(self, cubes):
...
 
 
+class Uas(Fix):
+    """Fixes for uas."""
+
+    def fix_metadata(self, cubes):
+        """Fix metadata."""
+        for cube in cubes:
+            fix_hourly_time_coordinate(cube)
+        return cubes
```
bouweandela, line 344: Shifting the hourly time points back by half an hour is done for variables that are accumulations because for ERA5 data those represent the past hour and in CMOR the convention is to put the time point at the center of the interval instead of at the end. However, for instantaneous variables like the 10 metre U/V wind component this does not apply. The time point represents the moment at which the variable had this value, so this fix should not be used. You can read more about it here: https://confluence.ecmwf.int/display/CKB/ERA5%3A+data+documentation

esmvalcore/cmor/tables/custom/CMOR_sfcWind.dat
```diff
@@ -0,0 +1,23 @@
+SOURCE: CMIP5
```
bouweandela, line 1: Is the custom CMOR table entry really needed? If I remember correctly, the definition of derived variables will automatically be taken from another mip table (from the same project) if they are not available in the mip table that is specified in the recipe.

bouweandela (changes requested):

esmvalcore/cmor/_fixes/native6/era5.py
valeriupredoi: cheers @bouweandela - I didn't know this! :beer:

esmvalcore/cmor/tables/custom/CMOR_sfcWind.dat
valeriupredoi: no, indeed it's not needed - but it shows up in pretty much all other custom tables, so we shouldn't fix this here, but in a dedicated PR

esmvalcore/cmor/_fixes/native6/era5.py
malininae: @bouweandela Oh, good to know, thanks! A question though, why do we do a 30 mins down for `tasmax`, `tasmin`? By the looks of it "reanalysis: the minimum or maximum values are in the hour (the processing period) ending at the validity date/time", so as far as I understand, the validity is at the end of timestep and it's similar what you are saying for CMOR.

esmvalcore/cmor/tables/custom/CMOR_sfcWind.dat
malininae: So the problem was, that when I ran the derivation, it wasn't working without the custom table see [here](https://github.com/ESMValGroup/ESMValCore/issues/1895#issuecomment-1741596407l). Based on the error message, I attributed it to the fact that `uas` and `vas` are in `3hr` table, while `sfcWind` is in `CF3hr` or `E3hr`, but I'm not sure that is the reason now.

esmvalcore/cmor/_fixes/native6/era5.py
bouweandela: `tasmin` and `tasmax` are always related to some interval in time. Where the time point is located does not really matter, what matters is the bounds. The developers of ERA5 chose to put the time point at the end of the interval, while for CMIP data it is customary to put the time point in the middle of the interval. For easier comparison of ERA5 with CMIP data, we shift the time point to the middle of the interval when that is possible without sacrificing the validity of the data.

malininae: Thanks! Removed the fix for `Uas` and `Vas` :smile: Sorry, I caused that issue!

esmvalcore/cmor/tables/custom/CMOR_sfcWind.dat
valeriupredoi: let's tackle that in a separate PR - good points from both Bouwe and Liza, but it's just not the place for it here

valeriupredoi (approved):
cheers muchly for the very useful work, Liza! @bouweandela maybe you got time give this one last glance, and merge, pls bud? :beer:

esmvalcore/cmor/tables/custom/CMOR_sfcWind.dat
bouweandela: It seems to work without the custom CMOR table for me. I tried this recipe:
```yaml
documentation:
  description: |
    Example recipe that plots a map and timeseries of temperature.

  title: Recipe that runs an example diagnostic written in Python.

  authors:
    - andela_bouwe
    - righi_mattia

  maintainer:
    - schlund_manuel

  references:
    - acknow_project

  projects:
    - esmval
    - c3s-magic

datasets:
  - {dataset: ERA5, project: native6, tier: 3, type: reanaly}

diagnostics:

  map:
    description: Global map of temperature in January 2000.
    themes:
      - phys
    realms:
      - atmos
    variables:
      sfcWind:
        mip: E1hr
        timerange: 1990/PT1M
        derive: true
        caption: |
          Global map of {long_name} in January 2000 according to {dataset}.
    scripts:
      script1:
        quickplot:
          plot_type: pcolormesh
          cmap: Reds
```
@malininae Could you try again and share the error message?

bouweandela: You're right, the 'derived' argument is only used for looking things up in the custom CMOR table. I've opened #2256 to improve that.

malininae:
> Hey @malininae thanks for the work, that might be useful for me too. I'm interested in sfcWind from ERA5, but on a monthly frequency. However, I just wanted to point out that the calculation is not linear, which means deriving sfcWind from monthly wind components is not the same as monthly mean of sfcWind calculated from higher frequencies. I'm not sure how monthly windspeed at 10m is calculated in the [published ERA5 monthly dataset](https://cds.climate.copernicus.eu/cdsapp#!/dataset/reanalysis-era5-single-levels-monthly-means?tab=form), but probably from higher frequencies.
> 
> Do you think it makes sense to limit the required input data to hourly and/or 3hourly data?

@lukruh thanks for the comment! It's indeed not an obvious question. I am writing a paper on extreme winds now and I looked through the 'stopping  point' when `sqrt(uas^2+vas^2)` is not giving the same answer for different time aggregations. For my region of interest and my application, the stopping will be at 6 hours, however for some regions 6 hours is too much already and I would stop at 3. Daily give too much of an uncertainty due to non-linearity, as you mention. At the same time I was suggested by a model developer to use daily data. I agree, that I should add the constrain on the frequency, but which one to use for the, honestly, I don't know. I would suggest a day maybe and raise a warning if the frequency is above 3 hours?

bouweandela:
To move this pull request forward, I would be keen to get #2256 merged first, then take out the custom variable from this pull request and then merge this.

bouweandela:
https://github.com/ESMValGroup/ESMValCore/pull/2256 has new been merged.

bouweandela (approved):

PR #2243, opened 2023-10-24
revisit the comment-triggered Github Actions test

valeriupredoi:
## Description

This revists the run GA tests triggered by a comment on a PR issue; this time round this should be running only on the branch that's used for the PR and not on all other branches; needs be merged in main first, so it can trigger :man_facepalming: 

***

PR #2244, opened 2023-10-25
Remove workflow that runs Github Actions tests from PR comment

valeriupredoi:
Testing the implementation from #2243 (now that the workflow in on `main` branch); if things don't pan out well, this PR will remove that workflow.

**Conclusion**: this stuff still doesn't work properly

**Pros**

- it actually runs triggered via a comment containing "runGAtests" (as you'll see the flow triggered by this very comment)

**Cons**

- because the workflow is in `main` (and it can't work from any other branch), every blithering comment on every pull request will trigger the flow; it will be skipped everywhere but in the PR that actually needs it to run, but it'll still show up as a Skipped workflow in the Actions UI list, that'll clog our Actions list terribly in no time; Github has no known way to hide/remove automatically skipped workflow runs (pish!)
- it is unclear if the branch identifier actually chooses the correct branch since I am seeing the tests fail when they shouldn't fail

**TL;DR**

Too much faff to get something done, that Github should offer in a working, optimized manner :man_facepalming:


valeriupredoi:
runGAtests

valeriupredoi:
whoopidy-whoppydoo it darn works! :partying_face:

valeriupredoi:
fgfhrtsrwter

valeriupredoi:
runGAtests

valeriupredoi:
runGAtests

bouweandela (approved):

valeriupredoi:
cheers @bouweandela - sorry to have wasted your time on these PRs, but I think that option would be very useful, if only it worked well :beer:

PR #2292, opened 2024-01-02
Add height for sfcWindmax in MPI HighRes models

malininae:
Following #2291 here's the fix for the missing coordinates for the sfcWindmax variable in MPI-ESM1-2-HR and MPI-ESM1-2-XR.


Karen-A-Garcia:
Looks good from what I see :)

valeriupredoi (approved):
hey Liza, happy new year! Thanks a bunch for this - am approving, but I'd be very grateful if you added a couple tests for those fixes, please :beer:

malininae:
> hey Liza, happy new year! Thanks a bunch for this - am approving, but I'd be very grateful if you added a couple tests for those fixes, please 🍺

Oops, sorry! Based on the other examples, I thought the ones which are re-using a class don't need tests. Will make tests in the next weeks!

valeriupredoi:
ow I missed that those were re-used classes, my bad - no need then, those are tested already :+1: :beer:

Issue #2293, opened 2024-01-03
Missing height 2m coordinates in GFDL-CM4


Karen-A-Garcia:
There is a height issue with the GFDL-CM4 historical model with variables requiring height. An example is the hurs:

```
esmvalcore.cmor.check.CMORCheckError: There were errors in variable hurs:
 height2m: does not exist
in cube:
relative_humidity / (%)             (time: 7300; latitude: 180; longitude: 288)
    Dimension coordinates:
        time                             x               -               -
        latitude                         -               x               -
        longitude                        -               -               x
    Cell methods:
        0                           area: time: mean
    Attributes:
        Conventions                 'CF-1.7 CMIP-6.0 UGRID-1.0'
        activity_id                 'CMIP'
        branch_method               'standard'
        branch_time_in_child        0.0
        branch_time_in_parent       -638385
        contact                     'gfdl.climate.model.info@noaa.gov'
        data_specs_version          '01.00.27'
        experiment                  'historical'
        experiment_id               'historical'
        external_variables          'areacella'
        forcing_index               1
        frequency                   'day'
        further_info_url            'https://furtherinfo.es-doc.org/CMIP6.NOAA-GFDL.GFDL-CM4.historical.non ...'
        grid                        'atmos data regridded from Cubed-sphere (c96) to 180,288; interpolation ...'
        grid_label                  'gr1'
        initialization_index        1
        institution                 'National Oceanic and Atmospheric Administration, Geophysical Fluid Dynamics ...'
        institution_id              'NOAA-GFDL'
        interp_method               'conserve_order2'
        license                     'CMIP6 model data produced by NOAA-GFDL is licensed under a Creative Commons ...'
        mip_era                     'CMIP6'
        nominal_resolution          '100 km'
        original_name               'hurs'
        parent_activity_id          'CMIP'
        parent_experiment_id        'piControl'
        parent_mip_era              'CMIP6'
        parent_source_id            'GFDL-CM4'
        parent_time_units           'days since 1850-1-1 00:00:00'
        parent_variant_label        'r1i1p1f1'
        physics_index               1
        product                     'model-output'
        realization_index           1
        realm                       'atmos'
        references                  'see further_info_url attribute'
        source                      'GFDL-CM4 (2018): \naerosol: interactive\natmos: GFDL-AM4.0.1 (Cubed-sphere ...'
[...]

PR #2263, opened 2023-11-27
Fixing missing height coordinates in MPI-ESM1-2-HR and MPI-ESM1-2-XR

Karen-A-Garcia:
This is a follow up pull request for the issue #2262. This is all using daily resolution CMIP6 data. In MPI_ESM1_2_HR in HighResMIP, tasmax was missing the 2m height coordinate and uas and vas were missing the 10m height coordinate. In MPI_ESM1_2_XR in HighResMIP, tasmax and tasmin were missing the 2m height coordinate and uas and vas were missing the 10m height coordinate. I just added quick fixes for those.

PR #2294, opened 2024-01-03
Fixing missing height 2m coordinates in GFDL-CM4 and KIOST-ESM

Karen-A-Garcia:
This is a follow up to issues #2293 and #2295. This is all using daily resolution CMIP6 data. In GFDL-CM4 in historical experiment tasmax, tasmin, sfcWind, and hurs are missing the 2m height coordinate. In Kiost-esm tasmin and tasmax has height issues. I just added quick fixes for those and tests for them. This is similar to what I did in pull request #2263.


valeriupredoi:
hi @Karen-A-Gracia happy new year and many thanks for the fixes and PR! I fixed the conflict in the test module, and will merge after them tests pass. A couple small pointers: when you import classes or functions from module, it's best practice to import them all in one line, rather than eg `from moduleA import class1 / from moduleA import class2` do `from moduleA import (class1, class2, )`. Also, once you see merge conflicts, it's always good to, given you got time, fix them, or if the conflicts are something you don't really know how to fix, ask for help :beer:

PR #2309, opened 2024-01-25
Do not read `~/.esmvaltool/config-user.yml` if `--config-file` is used

### LINKED ISSUE #2280: ~/.esmvaltool/config-user.yml is always read even when --config-file is specified
**Describe the bug**

If a user specified a custom user configuration file via `--config-file`, the default user configuration file at `~/.esmvaltool/config-user.yml` is [still read](https://github.com/ESMValGroup/ESMValCore/blob/main/esmvalcore/config/_config_object.py#L242). This can lead to very confusing results, e.g., if an invalid option is present in `~/.esmvaltool/config-user.yml` but not in the user-defined file, there will be an error which is very hard to solve for users without a deep knowledge of the tool.

It would be really nice if only the user-defined configuration file is read.

I guess this is a side effect of how the configuration is provided for the (experimental) API. @ESMValGroup/technical-lead-development-team does anyone have an idea how to solve this in a backwards-compatible way without too many changes? I can't think of an easy solution to this at the moment, so this might be something to consider when we rewrite our configuration file/code (see https://github.com/ESMValGroup/ESMValCore/issues/795).

schlunma:
## Description

Closes #2280

Link to documentation: https://esmvaltool--2309.org.readthedocs.build/projects/ESMValCore/en/2309/api/esmvalcore.config.html#esmvalcore.config.CFG

***

- [x] [🛠][1] ~All [checks below this pull request](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-checks) were successful~ Codacy issues cannot be avoided (also, they have been there before)

***


schlunma:
Here is my though process regarding this PR:

1. The first time in the code where we can properly access the `config-file` from the command line is in [`esmvalcore._main.ESMValTool.run`](https://github.com/ESMValGroup/ESMValCore/blob/989839f596a6b4433731c0efcb6b5cb4f3ca7045/esmvalcore/_main.py#L333). However, at that time, the `CFG` is already populated by an indirect import of `esmvalcore.config` earlier in [`esmvalcore._main.ESMValTool.__init__`](https://github.com/ESMValGroup/ESMValCore/blob/989839f596a6b4433731c0efcb6b5cb4f3ca7045/esmvalcore/_main.py#L303) (which actually happens when populating the entry points with the CMORizer CLI from ESMValTool).
2. Thus, I chose to read the CLI arguments directly via `sys.argv` in [`esmvalcore.config`](https://github.com/ESMValGroup/ESMValCore/blob/7d20d6e9cb64f2eedf16d0a5346088b5606109c6/esmvalcore/config/_config_object.py#L243). This feels very hacky, but ensures that the correct config file is used, regardless of when this module is imported (which might change in the future). This works well for the entire preprocessor.
3. However, when [running the diagnostics](https://github.com/ESMValGroup/ESMValCore/blob/7d20d6e9cb64f2eedf16d0a5346088b5606109c6/esmvalcore/_task.py#L523), no `--config-file` CLI argument is present, so this is not working. My thought was now to set an environment variable the first time `sys.argv` is parsed right at the beginning. That way, also the diagnostics have access to the correct config file.

All in all, these feels very brittle and not at all robust, but it's the best I can think of that the moment :sweat_smile: :

schlunma:
All right, this is working now. Need to add doc, tests, and clean a bit, then we should be ready to go.

esmvalcore/config/_config_object.py
```diff
@@ -93,29 +112,140 @@ def _load_default_config(cls, filename: Union[os.PathLike, str]):
...
+                '_ESMVALTOOL_USER_CONFIG_FILE_' in os.environ
+        ):
+            config_user = os.environ['_ESMVALTOOL_USER_CONFIG_FILE_']
+
+        # (3) Try to get user configuration file from CLI arguments
+        if config_user is None:
+            config_user = Config._get_config_path_from_cli()
+
+        # (4) Default location
+        if config_user is None:
```
valeriupredoi, line 180: will `Config._get_config_path_from_cli()` return a None if that's not specified via cmd line? Or an empty string?

esmvalcore/config/_config_object.py
```diff
@@ -93,29 +112,140 @@ def _load_default_config(cls, filename: Union[os.PathLike, str]):
...
+            return None
+
+        for arg in sys.argv:
+            for opt in ('--config-file', '--config_file'):
+                if opt in arg:
+                    # Parse '--config-file=/file.yml' or
+                    # '--config_file=/file.yml'
+                    partition = arg.partition('=')
+                    if partition[2]:
+                        return partition[2]
```
valeriupredoi, line 223: yeah - see - it could return an empty string if `--config-file` has no value - remember that we can run whatever we want with Fire eg `esmvaltool --config-file` - this will raise a Fire error (not a bool operator etc, line 518 in _main.py):
```
    try:
        fire.Fire(ESMValTool())
    except fire.core.FireExit:
        raise
 ```
 but I think that's way down the stack for this case

**Review by valeriupredoi — CHANGES_REQUESTED** [review]
this is a beautiful implementation - got one single low level worry (see me comment about using the --config-file with no value), but also got a more higher level question: what happens when simply does `import esmvalcore` - that does not, on its own, go through `_main.py` and does not dynamically import the environment set up via esmvalcore's installation = so my feeling is that an environ variable exposed to the user at a mere import call needs be put in an eg activation script like they did with ESMF_MKFILE

File: esmvalcore/config/_config_object.py [hunk 3]
```diff
@@ -38,49 +43,63 @@ class Config(ValidatedConfig):
...
         """
         new = cls()
+        new.update(CFG_DEFAULT)
+
+        config_user_path = cls._get_config_user_path(filename)
 
         try:
-            mapping = _read_config_file(filename)
-            mapping['config_file'] = filename
-        except IOError:
```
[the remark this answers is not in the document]
schlunma: It was raised, I just changed it to `FileNotFoundError` since this fits much better :grin:

esmvalcore/config/_config_object.py
schlunma: It will return `None`, so we are fine :+1:

esmvalcore/config/_config_object.py
```diff
@@ -38,49 +43,63 @@ class Config(ValidatedConfig):
...
         return new
 
     @classmethod
-    def _load_default_config(cls, filename: Union[os.PathLike, str]):
+    def _load_default_config(cls):
         """Load the default configuration."""
         new = cls()
 
-        mapping = _read_config_file(filename)
+        config_user_path = Path(esmvalcore.__file__).parent / 'config-user.yml'
```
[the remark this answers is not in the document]
schlunma: Done in https://github.com/ESMValGroup/ESMValCore/pull/2309/commits/5ae85aba49fae35e8e516095a96bad44ef64d6cd

esmvalcore/config/_config_object.py
schlunma: Very good point! I added some code to simply return `None` now here in https://github.com/ESMValGroup/ESMValCore/pull/2309/commits/2ed5a78d63b792bad27363551cf11da7ff39c74e and let fire handle this further down the line. Before that, a very ugly `IndexError` has been raised.

schlunma:
> what happens when simply does `import esmvalcore` - that does not, on its own, go through `_main.py` and does not dynamically import the environment set up via esmvalcore's installation = so my feeling is that an environ variable exposed to the user at a mere import call needs be put in an eg activation script like they did with ESMF_MKFILE

If you just `import esmvalcore` (or more specifically `from esmvalcore.config import CFG`) in a script or a notebook the code tries to read the default user config at `~/.esmvaltool/config-user.yml`, and if that does not exist uses the default config. This is exactly the behavior we have right now (and the one we want).

We don't want anyone apart from the ESMValTool main process to set the `_ESMVALTOOL_USER_CONFIG_FILE_` manually. The only purpose of this is that child processes of the main process use the same configuration file as the main process (without that, child processes would again default to `~/.esmvaltool/config-user.yml`, which we don't want, see #2280). That's also the reason why there is no public documentation of this environment variable.

schlunma:
This is now ready to merge from my side

esmvalcore/config/_config_object.py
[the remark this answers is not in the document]
valeriupredoi: aha gotcha - good call!

esmvalcore/config/_config_object.py
valeriupredoi: good man!

valeriupredoi:
good man Manu! Let me have a testy-test done, then I'll approve then merge :beers:

valeriupredoi:
`esmvalcore`-only (dev installed on its own) tests pass beautifully! I am now running a set of tests with a dev installed `esmvaltool` and dev installed `esmvaltool` off this branch - if things go according to plan I will be rejoicing and praising Lord Manu :grin:

valeriupredoi:
hang tight, got distracted with S3 stuffs - on to it in a minute

valeriupredoi (approved):
outstanding work, Manu! Tested proper, and all was peachy, thank you, sir :beer:

PR #2327, opened 2024-02-07
change 'mean' to 'percentile' in doc strings of preprocessor statistics

lukruh:
For me this looks like a mistake that got copy pasted to some doc strings. This should rather be 'percentile' than 'mean'.
I don't have time to go through the complete checklist for a PR.


schlunma (approved):
Thanks for fixing this @lukruh, this is indeed a typo 🚀

PR #2332, opened 2024-02-12
cmip6 gfdl_cm4: fix tas height fix to work for concatenated scenarios

mwjury:
In its current form the fix for correcting the height in TAS of GFDL-CM4 fails when concatenating the historical with ssp experiments with:
```
ValueError: Can not concatenate cubes: failed to concatenate into a single cube.
  Scalar coordinates values or metadata differ: height != height
```

This is due to:
- the height coord being missing in the historical experiment, and being added by the current fix:
```
    	double height ;
            height:units = "m" ;
            height:standard_name = "height" ;
            height:long_name = "height" ;
            height:positive = "up" ;
```

- the height coord in the scenarios being differently defined than the one added by the current fix

```
        double height ;
            height:long_name = "height" ;
            height:units = "m" ;
            height:missing_value = 1.e+20 ;
            height:_FillValue = 1.e+20 ;
            height:cell_methods = "time: point" ;
            height:axis = "Z" ;
            height:positive = "up" ;
            height:standard_name = "height" ;
            height:description = "~2 m standard surface air temperature and surface humidity  height" ;
```

However, the error is raised solely due to:

`            height:description = "~2 m standard surface air temperature and surface humidity  height" ;
`


valeriupredoi (approved):
many thanks @mwjury :beer:

valeriupredoi:
though a slightly less destructive approach would be to check if that `description` attr exists, and add it with the said value - but am not 100% sure 1. how useful that is, and 2. if that will change in future data iterations and that'll break `iris` again :+1:

Issue #2036, opened 2023-05-17
Dataset problem: GERICS-REMO2015 differences between original grid and standardised grid too large.


ljoakim:
**Describe the dataset issue**
Running a recipe with the GERICS-REMO2015 dataset causes an error when the grid for the dataset is recomputed in `cordex_fixes.py`. The grid in the original data contains `lon` values with a 'seam' where values goes from close to 360, to zero at the seam. This causes the difference between the original and the recomputed standardized grid to be too large where original values are close to 360 since the standardized grid uses negative coordinate values for the same locations.

- Data file has been changed by you in any way: No
- Project: CORDEX
- Full dataset description:
  - dataset: REMO2015
  - domain: EUR-11
  - experiment: historical
  - ensemble: r3i1p1
  - mip: mon
  - institute: GERICS
  - rcm_version: v1
  - driver: MPI-M-MPI-ESM-LR
  - variable used: tas
  - version: v20190925

- Problems encountered: differences between original grid and standardised grid to large.
- Pointer to existing copy of data on ESGF node: e.g. http://esgf1.dkrz.de/thredds/fileServer/cordex/cordex/output/EUR-11/GERICS/MPI-M-MPI-ESM-LR/historical/r3i1p1/GERICS-REMO2015/v1/mon/tas/v20190925/tas_EUR-11_MPI-M-MPI-ESM-LR_historical_r3i1p1_GERICS-REMO2015_v1_mon_199101-200012.nc

**Please attach**
  - The recipe that you are trying to run, you can find a copy in the `run` directory in the output directory
[recipe_remo.txt](https://github.com/ESMValGroup/ESMValCore/files/11497439/recipe_remo.txt)

  - The `main_log_debug.txt` file, this can also be found in the `run` directory in the output directory
[main_log_debug.txt](https://github.com/ESMValGroup/ESMValCore/files/11497440/main_log_debug.txt)

  - Attach a text file with the output of `ncdump -h /path/to/file.nc`
[ncdump.txt](https://github.com/ESMValGroup/ESMValCore/files/11497441/ncdump.txt)

  - ... and also a cutout from `ncdump -v lon [file]` to show the 'seam' in the `lon` coordinate:
[ncdump_lon.txt](https://github.com/ESMValGroup/ESMValCore/files/11497443/ncdump_lon.txt)

Cheers :beer:

mwjury:
[...]

PR #2334, opened 2024-02-13
cordex GERICS REMO2015  lon differences above 10e-4

mwjury:
Several of the EURO-CORDEX GERICS REMO2015 realizations fail during cmor checkes due to:
`ERROR   Differences between the original grid and the standardised grid are above 10e-4 degrees_east.`
, or rather differences being above 360°:
```
2024-02-09 11:08:05,322 UTC [2442951] DEBUG   esmvalcore.cmor._fixes.cordex.cordex_fixes:115 Maximum difference between original rlatpoints and standard EUR-11 domain points  for dataset remo2015 and driver NCC-NorESM1-M is: 3.552713678800501e-15.
2024-02-09 11:08:05,323 UTC [2442951] DEBUG   esmvalcore.cmor._fixes.cordex.cordex_fixes:115 Maximum difference between original rlonpoints and standard EUR-11 domain points  for dataset remo2015 and driver NCC-NorESM1-M is: 3.552713678800501e-15.
2024-02-09 11:08:05,370 UTC [2442951] DEBUG   esmvalcore.cmor._fixes.cordex.cordex_fixes:115 Maximum difference between original latpoints and standard EUR-11 domain points  for dataset remo2015 and driver NCC-NorESM1-M is: 0.00028290596755198294.
2024-02-09 11:08:05,387 UTC [2442951] DEBUG   esmvalcore.cmor._fixes.cordex.cordex_fixes:115 Maximum difference between original lonpoints and standard EUR-11 domain points  for dataset remo2015 and driver NCC-NorESM1-M is: **360.00001502170517.**
```

This fix simply checks for these before running the _check_grid_differences.


mwjury:
Provides a fix for part of #2036

valeriupredoi (changes requested):
many thanks @mwjury - unfortunately I myself am not well familiar with CORDEX data, but by the looks of it, I *think* you ought to also roll the data, if you are switching from -180,+180 to a 0,360 grid. Rolling is easy eg apply a dedicated function we have in the CMORizer utils, like so:

```python
def roll_cube_data(cube, shift, axis):
    """Roll a cube data on specified axis."""
    cube.data = da.roll(cube.core_data(), shift, axis=axis)
    return cube
```

with a use like so:

```python
                if cube_coord.points[0] < 0. and \
                        cube_coord.points[-1] < 181.:
                    cube_coord.points = \
                        cube_coord.points + 180.
                    cube.attributes['geospatial_lon_min'] = 0.
                    cube.attributes['geospatial_lon_max'] = 360.
                    nlon = len(cube_coord.points)
                    roll_cube_data(cube, nlon // 2, -1)
```

I'm going to put this on hold for now and ask @schlunma and/or @bouweandela who are a lot more familiar with CORDEX data :beers:

valeriupredoi:
BTW unfortunately @zklaus has left for greener pastures :sheep:

mwjury:
> many thanks @mwjury - unfortunately I myself am not well familiar with CORDEX data, but by the looks of it, I _think_ you ought to also roll the data, if you are switching from -180,+180 to a 0,360 grid. Rolling is easy eg apply a dedicated function we have in the CMORizer utils, like so:

Thanks @valeriupredoi! Yeah CORDEX fun :hammer:. In case of cordex, lats/lons are both 2d matrices. The issue here is that in the raw coord system itself the lons range from [~315 - 60], while the fix/coord-check uses [-45 - 60]. So the data matrix is fine and does not need to be rolled. Due to the 2d nature of longitudes, I currently see no other way than to using indices, but of course I'm happy to learn :beers: 

> BTW unfortunately @zklaus (our former resident CORDEX specialist) has left for greener pastures 🐑 - maybe @sloosvel (if you are back) or anybody else?
That's a pity. @sloosvel is not back unfortunately.

schlunma:
Hi @mwjury! Unfortunately, I am not familiar with CORDEX data at all. However, adding/subtracting 360° to the coordinates sounds reasonable to me in that context (and should not require rolling). In our generic fix we have a slightly more sophisticated method of making sure that longitude is in [0, 360], feel free to check that out:

https://github.com/ESMValGroup/ESMValCore/blob/d936e4e65436c08c8d8cd02decbce3a3a6908b6a/esmvalcore/cmor/_fixes/fix.py#L407-L410

I think your method is 100% fine as well.

valeriupredoi (approved):
excellent - very many thanks both you guys! I am happy to let this through, and get it merged then, unless @mwjury wants to implement the slightly fancier method pointed out by @schlunma :beer: x2

PR #2344, opened 2024-02-23
Cache regridding weights if possible

### LINKED ISSUE #2341: Use iris' regridder caching for faster regridding?
Some regridding schemes of iris allow reusing the regridding weights (see [here ](https://scitools-iris.readthedocs.io/en/latest/further_topics/which_regridder_to_use.html)for an overview of them), which might significantly speed up our `regrid` preprocessor if multiple variables of the same dataset are regridded. This is described [here](https://scitools-iris.readthedocs.io/en/latest/userguide/interpolation_and_regridding.html#caching-a-regridder) and basically boils down to using the following pattern

```py
regridder = iris.analysis.Nearest().regridder(source_cube, target_cube)
for cube in list_of_cubes_on_source_grid:
    result = regridder(cube)
```

instead of

```py
for cube in list_of_cubes_on_source_grid:
    result = cube.regrid(target_cube, iris.analysis.Nearest())
```

This should be fairly easy to implement with a `dict` cache that uses the source and target latitude and longitude dimensions as keys (we cannot use `lru_cache` since the `hash` of a coordinate is simply its `id`, therefore this would never use cached results for different cubes).

One drawback here is that we have an additional comparison of the coordinates for each `regrid` call, which makes it slightly slower if just a single variable for many data sets is analyzed.

@ESMValGroup/esmvaltool-coreteam is this something we want? I can open a draft PR so that the discussion is easier.

schlunma:
## Description

This implements [regridder weights caching](https://scitools-iris.readthedocs.io/en/latest/userguide/interpolation_and_regridding.html#caching-a-regridder), which may reduce regridding time dramatically (if many variables of the same data set are analyzed).

I also modernized existing regridding tests so that they use pytest now instead of unittest, and actually test the regridding instead of using mocks.

Closes #2341

Link to documentation: https://esmvaltool--2344.org.readthedocs.build/projects/ESMValCore/en/2344/recipe/preprocessor.html#horizontal-regridding

***

***


doc/recipe/preprocessor.rst
```diff
@@ -780,10 +780,6 @@ regridding is based on the horizontal grid of another cube (the reference
 grid). If the horizontal grids of a cube and its reference grid are sufficiently
 the same, regridding is automatically and silently skipped for performance reasons.
 
-The underlying regridding mechanism in ESMValCore uses
-:obj:`iris.cube.Cube.regrid`
-from Iris.
-
```
valeriupredoi, line 786: any reason why you removing this, bud? Quite too general a statement indeed, but it's not wrong

esmvalcore/preprocessor/_regrid.py
```diff
@@ -639,6 +639,62 @@ def _load_generic_scheme(scheme: dict):
...
+    ----
+    If possible, this uses an existing regridder to reduce runtime (see also
+    https://scitools-iris.readthedocs.io/en/latest/userguide/
+    interpolation_and_regridding.html#caching-a-regridder.)
+
+    """
+    # Use existing regridder if possible. This first checks the regridding
+    # scheme name and shapes of source and target coordinates, and only if
+    # these match check coordinates themselves (this is much more expensive).
+    coord_key = _get_coord_key(src_cube, tgt_cube)
```
valeriupredoi, line 658: have completely dropped the option that regrid preprocessor be optionally called with an "MxN" target grid rather than a target grid cube? If we did, carry on - if we didn't, then this will break it

valeriupredoi: or this one gets used only when target grid is a cube?

valeriupredoi (changes requested):
looking good, bud :beer: A couple q's from me please - also, do you have a feel how big those caches can get to? ie would memory clogging get severe enough to forgo caching and just go for CPU time use instead?

doc/recipe/preprocessor.rst
schlunma: After this PR it will be wrong 😬 The new syntax is `regridded_cube = regridder(cube)`.

esmvalcore/preprocessor/_regrid.py
schlunma: At this point in the code, `tgt_cube` will always be a cube, since it has been create from `_get_target_grid_cube`. It's still possible to use `MxN` specs of course.

schlunma:
Thanks for reviewing V! Memory usage should be minimal: as mentioned [here](https://github.com/ESMValGroup/ESMValCore/issues/2341#issuecomment-1961400007), the weights should only be around 10 MiB for very high resolution grids (1000x1000), and much smaller for "normal" resolutions.

esmvalcore/preprocessor/_regrid.py
valeriupredoi: gotcha, thanks, Manu (I was suspecting that's the case, but better confirm, like in aviation :airplane: )

doc/recipe/preprocessor.rst
valeriupredoi: aha - they changed the API indeed - maybe just drop a mention about the new func from iris? Up to you

valeriupredoi (approved):
thanks for answering, Manu! Glad @bouweandela asked the same q about memory :grin:

doc/recipe/preprocessor.rst
schlunma: The old API is still there, but we are using a different one now that allows us to re-use the weights 😁 See https://scitools-iris.readthedocs.io/en/stable/userguide/interpolation_and_regridding.html#caching-a-regridder.

valeriupredoi: yes, I forgot that from 2 hours ago when I reviewed - remember, am an old man :rofl:

valeriupredoi:
@bouweandela maybe you could have a look too, given that I took you are not 100% convinced about the need of caching (reading from the issue) - and pls merge if all's good by ye too :beer:

bouweandela:
Yes, a 10 percent increase in runtime in the best case doesn't seem like a huge gain, but it's nice to have of course. I am concerned about the size of the cache though, in the issue you talked about 1GB https://github.com/ESMValGroup/ESMValCore/issues/2341#issuecomment-1961400007 and I saw that there is some discussion planned at the workshop about re-using weights:
> High resolution model data: often the weights to interpolate grids can be reused, not only for different variables but even across different experiments, time periods, and simulations of the same model. Could ESMValTool support the reading and use of precalculated weights because their calculation is very time consuming? (@katjaweigel?)

I image they would be rather large arrays too if they are so expensive to compute.

To avoid this turning into a memory leak, would it be an option to:
1) add a method to the `regrid` preprocessor function to clear the cache, similar to how it's done if you're using lru_cache, so Python users can clear the cache when they are done with it?
1) make weights caching optional by adding a `cache_weights` or similar argument to the preprocessor function?

schlunma:
> Yes, a 10 percent increase in runtime in the best case doesn't seem like a huge gain, but it's nice to have of course. I am concerned about the size of the cache though, in the issue you talked about 1GB [#2341 (comment)](https://github.com/ESMValGroup/ESMValCore/issues/2341#issuecomment-1961400007)

The 10% reduction in run time were only for this specific example, it's definitely not an upper bound. Also, the 1GiB is a really extreme example (e.g., a 0.1x0.1° grid would lead to a weights array of ~25 MiB).

> To avoid this turning into a memory leak, would it be an option to:
> 
> 1. add a method to the `regrid` preprocessor function to clear the cache, similar to how it's done if you're using lru_cache, so Python users can clear the cache when they are done with it?
> 
> 2. make weights caching optional by adding a `cache_weights` or similar argument to the preprocessor function, so it can be enabled from the recipe if it is relevant?

That sounds very reasonable, I will do that! 👍

valeriupredoi:
from me own experience it's always better to read from file/memory if possible, than it is to compute (yes OK call me Dr Obvious :rofl: ) - and given Manu's reassuring info on upper limits for mem intake, I think it's a go - but I do like Bouwe's suggestions too. My only concern, that I just thought of, is maybe this is better it sat in iris?

valeriupredoi:
Also (sorry, I sat down did MO crap today, so now my brain is free at last) - Manu, maybe it'd be worth ploppoing a worst case scenario test in the tests ie build a super-high res netCDF file on the fly, and do the weights caching dance on it - we'd be able to monitor the memory using the test performance tool we have in Github Actions, and of course, decorate it with eg `@highmem` so we don't run it usually :beer:

schlunma:
> My only concern, that I just thought of, is maybe this is better it sat in iris?

Actually this entire functionality [is already in iris](https://scitools-iris.readthedocs.io/en/stable/userguide/interpolation_and_regridding.html#caching-a-regridder) in the form of `Regridder` classes that compute the weights and store them. With this PR, we are using those `Regridder` classes instead of the regridding *schemes* which do not offer weights caching.

> Also (sorry, I sat down did MO crap today, so now my brain is free at last) - Manu, maybe it'd be worth ploppoing a worst case scenario test in the tests ie build a super-high res netCDF file on the fly, and do the weights caching dance on it - we'd be able to monitor the memory using the test performance tool we have in Github Actions, and of course, decorate it with eg @highmem so we don't run it usually 🍺

I am honestly not sure if that's worth it, especially if we implement the option to turn off weights caching 😬 What do we want to achieve with such a test? Kill the CI machines? 😄 🤖

schlunma:
Added option to enable/disable (default: disabled) caching weights: https://github.com/ESMValGroup/ESMValCore/pull/2344/commits/b6eaade2d39c237136d8aa2da68333b1494da913

Added function to clear cache similar to `lru.cache_clear()`: `regrid.cache_clear()`, see https://github.com/ESMValGroup/ESMValCore/pull/2344/commits/52a536ccd0e14f56820ba745816c607e22015420

bouweandela:
Thanks for making the changes @schlunma! @valeriupredoi Could you please do a final review to review the new code, since you reviewed the original PR, and then we can merge.

valeriupredoi:
On it, bud 🍺

valeriupredoi:
> What do we want to achieve with such a test? Kill the CI machines?

Not a smart thing to do, especially since today Skynet's closer than ever :laughing: OK, leave the poor machines alone then

valeriupredoi (approved):
bew stuff looks good! Cheers gents @schlunma and @bouweandela :beer:

Issue #2357, opened 2024-03-01
[CI/CD/GA] Tests failing with one test fail


valeriupredoi:
See [fail](https://github.com/ESMValGroup/ESMValCore/actions/runs/8104817928/job/22152114274) - unfotunately it is 5min before Friday pub o'clock, so I'll look into it on Monday morning :beers:

schlunma:
I am pretty sure this is related to the new release of iris ([3.8.1](https://github.com/SciTools/iris/releases/tag/v3.8.1))

valeriupredoi:
you are absolutely correct @schlunma - see https://github.com/SciTools/iris/issues/5806

valeriupredoi:
OK that xfailed test we xfail in #2363 needs be un-xfailed, and we need to fix the data instead, see https://github.com/SciTools/iris/issues/5806#issuecomment-2077658271

valeriupredoi:
closed via Manu's https://github.com/ESMValGroup/ESMValCore/pull/2403 - Manu, remember that invoking "proper fix" don't close an issue :grin:

PR #2363, opened 2024-03-06
xfail `tests/integration/cmor/_fixes/test_common.py::test_cl_hybrid_height_coord_fix_metadata` while Iris folk fix behaviour

valeriupredoi:
## Description

Issue described in https://github.com/SciTools/iris/issues/5806 and reported internally in #2357 

Remember to plug intended behaviour after Iris folk fix that, or, un-xfail the test

@schlunma gives a nice overview of the test: This test tests a fix that converts a non-CF-compliant netCDF file into a CF-compliant file. One issue of this netCDF file is that the units of the bounds differ to those of the points.

Since 3.8, iris now [ignores](https://github.com/SciTools/iris/pull/5746) bounds with different (incompatible) units than the points. I think we need to wait for a reply of the iris people before we can properly fix this. It would be especially helpful to know if we can somehow retrieve the bounds after loading the cube.

***

***

### DIFF
--- a/tests/integration/cmor/_fixes/test_common.py
+++ b/tests/integration/cmor/_fixes/test_common.py
@@ -141,6 +141,7 @@ def hybrid_height_coord_fix_metadata(nc_path, short_name, fix):
     assert air_pressure_coord.units == 'Pa'
 
 
+@pytest.mark.xfail(reason="github.com/SciTools/iris/issues/5806")
 def test_cl_hybrid_height_coord_fix_metadata(test_data_path):
     """Test ``fix_metadata`` for ``cl``."""
     vardef = get_var_info('CMIP6', 'Amon', 'cl')


schlunma (approved):
I am fairly sure we need to adapt our test somehow, but for now this is good.

bouweandela:
Could over of you please give a bit more detail about why this test is failing?

schlunma:
This test tests a fix that converts a non-CF-compliant netCDF file into a CF-compliant file. One issue of this netCDF file is that the units of the bounds differ to those of the points.

Since 3.8, iris now [ignores](https://github.com/SciTools/iris/pull/5746) bounds with different (incompatible) units than the points. I think we need to wait for a reply of the iris people before we can properly fix this. It would be especially helpful to know if we can somehow retrieve the bounds after loading the cube.

valeriupredoi:
thanks, Manu! And yes, valid question, Bouwe :beer:

bouweandela:
Just to make it clear that this is broken with iris 3.8, could you please open an issue in our own repository too with this description and which datasets can now now longer be fixed by ESMValCore? I'm not sure if the iris folks have plans to support all kinds of non-compliant data, so the fix may need an update if we want to use iris 3.8.

PR #2393, opened 2024-04-10
Unify dtype handling of preprocessors

### LINKED ISSUE #2370: `annual_statistics` may change dtype
**Describe the bug**

Very similar to https://github.com/ESMValGroup/ESMValCore/issues/1236 to and https://github.com/ESMValGroup/ESMValCore/issues/1235: The `annual_statistics` my change the dtype of the data.

Since we have lots of custom code in our preprocessors ([example](https://github.com/ESMValGroup/ESMValCore/blob/9728667e6b11298c7ec05bea8683ead15abc9e04/esmvalcore/preprocessor/_area.py#L238)), I think it would make sense to somehow sense to generalize this.

I am thinking about adding a decorator that does that. An alternative could be to add it to our `preprocess` function just like the [rechunking](https://github.com/ESMValGroup/ESMValCore/blob/main/esmvalcore/preprocessor/__init__.py#L368). However, this won't be useful for people who use the preprocessor outside of ESMValTool, so I would prefer a decorator.

@ESMValGroup/technical-lead-development-team thoughts?

schlunma:
## Description

Many preprocessors have use some kind of custom code to make sure that dtypes are preserved. This PR adds a decorator `@preserve_float_dtype` unifies this task. Only `float` dtypes will be preserved, other dtypes are allowed to change. This ensures that `int` objects may be transformed to `float` objects when taking the mean for example (similar to what numpy does).

Closes #2370

***

***

### DIFF
--- a/esmvalcore/preprocessor/_area.py
+++ b/esmvalcore/preprocessor/_area.py
@@ -24,6 +24,7 @@ from esmvalcore.preprocessor._shared import (
     get_iris_aggregator,
     get_normalized_cube,
     guess_bounds,
+    preserve_float_dtype,
     update_weights_kwargs,
 )
 from esmvalcore.preprocessor._supplementary_vars import (
@@ -189,6 +190,7 @@ def _extract_irregular_region(cube, start_longitude, end_longitude,
     return cube
 
 
+@preserve_float_dtype
 def zonal_statistics(
     cube: Cube,
     operator: str,
@@ -235,10 +237,10 @@ def zonal_statistics(
     result = cube.collapsed('longitude', agg, **agg_kwargs)
     if normalize is not None:
         result = get_normalized_cube(cube, result, normalize)
-    result.data = result.core_data().astype(np.float32, casting='same_kind')
     return result
 
 
+@preserve_float_dtype
 def meridional_statistics(
     cube: Cube,
     operator: str,
@@ -284,7 +286,6 @@ def meridional_statistics(
     result = cube.collapsed('latitude', agg, **agg_kwargs)
     if normalize is not None:
         result = get_normalized_cube(cube, result, normalize)
-    result.data = result.core_data().astype(np.float32, casting='same_kind')
     return result
 
 
@@ -367,6 +368,7 @@ def _try_adding_calculated_cell_area(cube: Cube) -> None:
     variables=['areacella', 'areacello'],
     required='prefer_at_least_one',
 )
+@preserve_float_dtype
 def area_statistics(
     cube: Cube,
     operator: str,
@@ -414,7 +416,6 @@ def area_statistics(
         `cell_area` is not available.
 
     """
-    original_dtype = cube.dtype
     has_cell_measure = bool(cube.cell_measures('cell_area'))
 
     # Get aggregator and correct kwargs (incl. weights)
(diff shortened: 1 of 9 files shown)


valeriupredoi:
Manu, I'm sorry, man - I promised I'd look into this, and it fell off my radar screen like a Russian Sukhoi over the Black Sea. I'll at least help out here :+1:

schlunma:
Since `inspect.getfullargspec` apparently [does not work with decorated functions](https://stackoverflow.com/questions/147816/preserving-signatures-of-decorated-functions#comment68410839_147878), I changed this to `inspect.signature` instead, which provides very similar functionality and handles decorated functions properly.

bouweandela:
> Since inspect.getfullargspec apparently [does not work with decorated functions](https://stackoverflow.com/questions/147816/preserving-signatures-of-decorated-functions#comment68410839_147878)

I'm a bit puzzled by this comment. Why does the decorator that we use for indicating that a preprocessor function requires ancillary variables or cell measures not cause this issue?
https://github.com/ESMValGroup/ESMValCore/blob/bbd307dc2202f28065807dda4e0f63ad122336c6/esmvalcore/preprocessor/_area.py#L366

schlunma:
The `register_supplementaries` decorator does not need to consider the arguments or return value of the decorated function, so the wrapper can simply return the decorated function itself (with the original signature):

https://github.com/ESMValGroup/ESMValCore/blob/bbd307dc2202f28065807dda4e0f63ad122336c6/esmvalcore/preprocessor/_supplementary_vars.py#L35-L37

On the other hand, the `preserve_float_dtype` decorator needs to access the first argument and change the return value:

https://github.com/ESMValGroup/ESMValCore/blob/5b470db800a0137908ef1a920a8a6606aee204ad/esmvalcore/preprocessor/_shared.py#L257-L266

Thus, the signature of the decorated function changes to that of the wrapper. In theory, this can be solved with the [`functools.wraps`](https://docs.python.org/3/library/functools.html#functools.wraps) decorator. However, as mentioned [here](https://stackoverflow.com/questions/147816/preserving-signatures-of-decorated-functions#comment68410839_147878), this is only considered by `inspect.signature`, not by `inspect.getfullargspec`.

valeriupredoi:
this shall get reviewed today by yours truly - sorry, Manu, was busy all week with non-ESMValX suffs

valeriupredoi (approved):
many thanks, Manu! And sorry I didn't get around to revieweing it on Fruday - after work pub o'clock happened :beer: The PR and its implementation looks very good; got one design-related question though: would you rather have this done as it is done now - using a decorator - and thus, the dtype change happens somewhat under the hood, but still, visible enough for future developers to make use of it for future preprocs, or:
- via a function that *must* be used whenever dtype needs preserved
- via some magic in the `__init__` where this is done automatically everytime an input dtype needs be preserved (probably by altering `globals`, can't think of a concerete way just now)

I think this way is the best way: the functional approach is bulky, and the completely under the hood approach is sneaky, and may create problems down the way. But what do @ESMValGroup/technical-lead-development-team think?

Either way, it'd be nice to add a phrase or two in the docs - not for every decorated func, but at top level - say dtype is preserved via a decorator etc :beer:

schlunma:
Thanks for reviewing V! I also think the decorator is the preferable way (that's why I implemented it 😄)

valeriupredoi:
> Thanks for reviewing V! I also think the decorator is the preferable way (that's why I implemented it 😄)

About docs: I'd think that we should mention that preproc X or Y preserves the dtype - like Numpy docs say - would you be OK to add that to each preproc's API docs? I can do that, I owe you a bit of help on this since I promised I will help :grin: Alternatively, we can provide a list of preproc functions that preserve dtypes, like irirs do with their lazy/realized data funcs

schlunma:
I think it would make more sense to document those that do **not** preserve float dtypes (IMHO, our default should be to preserve them). It would be even better to fix them 😄

So maybe we can mention in the beginning that (hopefully) all of our preprocessors preserve float dtypes (ints may be changed)? How does that sound?

valeriupredoi:
that too! I was debating this amongst my few working brain cells (given my new old age) just a bit earlier - a good point, Manu!

schlunma:
Updated the doc: https://esmvaltool--2393.org.readthedocs.build/projects/ESMValCore/en/2393/api/esmvalcore.preprocessor.html#preprocessor-functions

valeriupredoi:
@bouweandela would you be OK to give this a onesie and merge, plss, bud? :beer:

PR #2305, opened 2024-01-22
Pin pandas to avoid broken `round` function

schlunma:
## Description

This PR fixes [recent test failures](https://app.circleci.com/pipelines/github/ESMValGroup/ESMValCore/10270/workflows/774a0b7d-be4a-436a-89b2-2b1429d1b77b/jobs/44121) that are caused by a [bug in pandas](https://github.com/pandas-dev/pandas/issues/57002). This bug breaks the `round` function that we use to round to the nearest second:

https://github.com/ESMValGroup/ESMValCore/blob/d6008f3d010e3ad6d2708d47ed8d48aa396af739/esmvalcore/cmor/_fixes/icon/icon.py#L495-L497

In addition, this removes the use of two deprecated features:
1. Replace 'S' by 's' (see https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases)
2. Surround `DatetimeProperties.to_pydatetime` by `np.array` as recommended in the deprecation message. Unfortunately, this does not remove the warning, but hopefully this will avoid problems in the future.

***

## [Checklist](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#checklist-for-pull-requests)

It is the responsibility of the author to make sure the pull request is ready to review. The icons indicate whether the item will be subject to the [🛠 Technical][1] or [🧪 Scientific][2] review.

[1]: https://docs.esmvaltool.org/en/latest/community/review.html#technical-review
[2]: https://docs.esmvaltool.org/en/latest/community/review.html#scientific-review

- [x] [🧪][2] The new functionality is [relevant and scientifically sound](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#scientific-relevance)
- [x] [🛠][1] This pull request has a [descriptive title and labels](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-title-and-label)
- [x] [🛠][1] Code is written according to the [code quality guidelines](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#code-quality)
- [x] [🧪][2] and [🛠][1] [Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available
- [x] [🛠][1] [Unit tests](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#tests) have been added
[...]

PR #2395, opened 2024-04-12
[Numpy2] Support for `numpy==2.0.0` (and pin `iris >=3.11`)

### LINKED ISSUE #2460: [Numpy2] Preprocessor `distance_metric` returns value 0 instead of `--` (masked element) with `numpy==2.0.0` for metrics that use `np.sqrt` or `da.sqrt`
- test: `tests/unit/preprocessor/_compare_with_refs/test_compare_with_refs.py`
- test result: when an all-masked array is used as data input, `distance_metric` returns data element 0 instead of `masked` or `--`, with numpy=2 and certain (not all!) metrics
- see https://github.com/ESMValGroup/ESMValCore/actions/runs/9584890389/job/26429529577?pr=2395
- Output of interest:
```
FAILED tests/unit/preprocessor/_compare_with_refs/test_compare_with_refs.py::test_distance_metric[weighted_pearsonr-nan-1.0-Pearson's r-pearsonr_tas-1] - AssertionError: 
FAILED tests/unit/preprocessor/_compare_with_refs/test_compare_with_refs.py::test_distance_metric_masked_data[emd-1.98625-0.0-EMD-emd_tas-K-True] - AssertionError: 
FAILED tests/unit/preprocessor/_compare_with_refs/test_compare_with_refs.py::test_distance_metric_fully_masked_data[rmse-2.34520788-0.0-RMSE-rmse_tas-K-True] - AssertionError: 
FAILED tests/unit/preprocessor/_compare_with_refs/test_compare_with_refs.py::test_distance_metric_fully_masked_data[rmse-2.34520788-0.0-RMSE-rmse_tas-K-False] - AssertionError: 
FAILED tests/unit/preprocessor/_compare_with_refs/test_compare_with_refs.py::test_distance_metric_fully_masked_data[weighted_rmse-2.0-0.0-RMSE-rmse_tas-K-True] - AssertionError: 
FAILED tests/unit/preprocessor/_compare_with_refs/test_compare_with_refs.py::test_distance_metric_fully_masked_data[weighted_rmse-2.0-0.0-RMSE-rmse_tas-K-False] - AssertionError:
```

could be Scipy could be something else, this needs a bit of due dilligence :beer:

valeriupredoi:
## Description

This PR introduces the changes we need to support Numpy 2.0 (these changes were collected from running the tests, so they may not be 100% exhaustive, but given that we have 95% test coverage, I'd say they are 95% exhaustive; we can deal with missed cases as they come along, if any; I **would not** use the ruff plugin test Numpy recommends to migrate, since that will introduce a lot of style changes that we may want to make when we include ruff as our linter). Most changes are in our testing suite, but there are a couple in the source code; these are **backwards compatible** changes, ie [all tests pass](https://github.com/ESMValGroup/ESMValCore/actions/runs/9616129161/job/26524937053?pr=2395) when running with numpy=1.26.4

The only test that fails (both for Numpy2 and numpy legacy) is related to the pins on pandas, that need to go since we need to use the latest pandas for environment compatibility, see https://github.com/ESMValGroup/ESMValCore/issues/2459 ; I'll try fix that issue here, since it's related to (but not a direct consequence of) Numpy2.

Closes #2460
Closes https://github.com/ESMValGroup/ESMValCore/issues/2459

~iris issue to monitor https://github.com/SciTools/iris/issues/6043~

-> which that is already in iris==3.11, which leads to

~**TODO** pin `iris>=3.11`~

-> pinned in commits dddc0831b98305f7b55a7b0eee4ec18480ed3792 and 858ae1179c7f69f8cbe8f8b7d75c32c5bed9dbb9

The second issue closure is more of a brush it under the carpet via a pytest xfail, the original issue that kept creeping up is https://github.com/ESMValGroup/ESMValCore/pull/2305 and the actual issue is with Pandas, and listed in https://github.com/pandas-dev/pandas/issues/57002 - we can wait on its fix (and have us test quietness via the test `xfail`) or @schlunma can propose a more elegant solution in a separate PR?

**Historical guff**

OK @ESMValGroup/technical-lead-development-team we finally have a correct set of tests with the latest Numpy=2.0.0 (stable) - don't mind the ones related to `pandas`, I had to unpin it to gte pandas=2.2.2 built with Numpy 2.0 (otherwise the environment wouldn't solve) -> we have some work to do :+1: https://github.com/ESMValGroup/ESMValCore/actions/runs/9582595825/job/26421917358?pr=2395

~Numpy 2.0 is impending~: it is now fully out (see https://anaconda.org/conda-forge/numpy ); resources https://anaconda.org/conda-forge/numpy/files and https://github.com/numpy/numpy/releases - they recommend a test with the RC and list the changes roadmap in https://numpy.org/devdocs/numpy_2_0_migration_guide.html so we can test ourselves. Note: this branch should not be merged! If everything works out fine, we'll just pick up numpy 2.0 as per normal. ATTN: @ESMValGroup/technical-lead-development-team 
- seems the 2.0.0rc1 is a bit of a ninja - in https://github.com/numpy/numpy/issues/26271 it is reported that the channel is `conda-forge/label/numpy_dev`
- moving to said conda channel results in conflicts for `python-stratify`, `shapely`, and `pandas`, since most prob they are pinning the upper bound of numpy to <2.0, see [GA tests](https://github.com/ESMValGroup/ESMValCore/actions/runs/8688337876/job/23823752190?pr=2395)
- a manual install is then needed; that results in [mayhem](https://github.com/ESMValGroup/ESMValCore/actions/runs/8688527436/job/23824349407?pr=2395) - some of the fails are due to `cftime`, as @bouweandela points out in comments, see https://github.com/Unidata/cftime/issues/318 (and linked PR), others are due to actual package compilation against a Numpy=1.x that barfs when it sees 2.x (those should be solved at recompile/redeploy time), and others are due to netCDF4 complaining about dtypes - that needs a closer look


valeriupredoi:
UGH! This doesn't look good at all https://github.com/ESMValGroup/ESMValCore/actions/runs/8664961338/job/23762564320

bouweandela:
xref https://github.com/Unidata/cftime/issues/318

valeriupredoi:
Spot on @bouweandela 🍻 Fortunately there is a mature PR for it https://github.com/Unidata/cftime/pull/319

valeriupredoi:
OK environment barfs due to python-stratify, that has an old autobot PR to update to numpy 2.0 https://github.com/conda-forge/python-stratify-feedstock/pull/26 - will go ping there

valeriupredoi:
OK @ESMValGroup/technical-lead-development-team we finally have a correct set of tests with the latest Numpy=2.0.0 (stable) - don't mind the ones related to `pandas`, I had to unpin it to gte pandas=2.2.2 built with Numpy 2.0 (otherwise the environment wouldn't solve) -> we have some work to do :+1: https://github.com/ESMValGroup/ESMValCore/actions/runs/9582595825/job/26421917358?pr=2395

esmvalcore/preprocessor/_regrid_esmpy.py
```diff
@@ -243,7 +243,10 @@ def get_grid(esmpy_lat, esmpy_lon,
         num_peri_dims = 1
     else:
         num_peri_dims = 0
-    grid = esmpy.Grid(np.array(esmpy_lat.shape),
+
+    # previous to numpy 2.0, a np.array(esmpy_lat.shape) was used
+    # numpy>=2.0 throws ValueError: matrix transpose with ndim<2 is undefined
+    grid = esmpy.Grid(np.vstack(esmpy_lat.shape),
```
schlunma, line 249: Is this backwards-compatible? If yes, do we need this comment?

esmvalcore/preprocessor/_regrid_unstructured.py
```diff
@@ -169,8 +169,11 @@ def _get_weights_and_idx(
         src_points_with_convex_hull = self._add_convex_hull_twice(
             src_points, hull.vertices
         )
-        src_points_with_convex_hull[-2 * n_hull:-n_hull, 1] -= 360
-        src_points_with_convex_hull[-n_hull:, 1] += 360
+        # explicitly casting to int32 since without it, in Numpy 2.0
+        # one gets OverflowError: Python integer 360 out of bounds for int8
+        # see notes https://numpy.org/devdocs/numpy_2_0_migration_guide.html
+        src_points_with_convex_hull[-2 * n_hull:-n_hull, 1] -= np.int32(360)
+        src_points_with_convex_hull[-n_hull:, 1] += np.int32(360)
```
schlunma, line 176: Would this work with `360.0` instead of `np.int32(360)`? The points should be floats, not integers, so to me this makes much more sense. If this works, you can also remove the comment

esmvalcore/preprocessor/_derive/co2s.py
```diff
@@ -15,6 +15,8 @@ def _get_first_unmasked_data(array, axis):
     indices = da.meshgrid(
         *[da.arange(array.shape[i]) for i in range(array.ndim) if i != axis],
         indexing='ij')
+    # numpy>=2.0 indices becomes a tuple
```
schlunma, line 18: Remove comment? I don't think this is useful

esmvalcore/preprocessor/_time.py
```diff
@@ -1559,7 +1559,9 @@ def _transform_to_lst_eager(
     """
     # Apart from the time index, all other dimensions will stay the same; this
     # is ensured with np.ogrid
-    idx = np.ogrid[tuple(slice(0, d) for d in data.shape)]
+    # NOTE: in numpy 2.0 np.ogrid returns ndarray or tuple of ndarrays
```
schlunma, line 1562: Remove comment? This should be fully backwards-compatible, so I don't think it's useful.

tests/integration/cmor/_fixes/icon/test_icon.py
```diff
@@ -1607,6 +1607,9 @@ def test_invalid_time_units(cubes_2d):
 # Test fix with (sub-)hourly data
 
 
+# remove xfail when https://github.com/pandas-dev/pandas/issues/57002
+# gets fixed; pinning pandas not a viable solution due deps issues
+@pytest.mark.xfail(reason='Bug in pandas needs be fixed.')
```
schlunma, line 1612: As a short-term solution this is fine (it shouldn't break any recipes), but a proper fix would be much nicer.

tests/unit/preprocessor/_time/test_time.py
```diff
@@ -920,9 +920,12 @@ def test_season_not_available(self):
             name='clim_season',
             seasons=['JFMAMJ', 'JASOND'],
         )
+
+        # numpy>=2.0 these need to be explicitly cast to numpy strings
+        two_seasons = [np.str_('JASOND'), np.str_('JFMAMJ')]
```
schlunma, line 925: I guess this is not backwards-compatible? How does the error look like? It might make sense to adapt this in the actual code so the error message still looks nice

esmvalcore/preprocessor/_compare_with_refs.py
```diff
@@ -451,7 +451,21 @@ def _calculate_rmse(
...
+    # and https://docs.dask.org/en/stable/_modules/dask/array/reductions.html
+    if npx.__name__ == "dask.array":
+        da_squared_error = npx.ma.average(squared_error,
+                                          axis=axis,
+                                          weights=weights)
+        rmse = npx.reductions.safe_sqrt(da_squared_error)
+    else:
+        rmse = npx.ma.sqrt(
+            npx.ma.average(squared_error, axis=axis, weights=weights)
+        )
```
schlunma, line 468:
```suggestion
    mse = npx.ma.average(squared_error, axis=axis, weights=weights)
    if isinstance(mse, da.Array):
        rmse = da.reductions.safe_sqrt(mse)
    else:
        rmse = np.ma.sqrt(mse)
```

schlunma (changes requested):
Thanks V, looks really good already! I have a couple of comments, mostly about removing the comments if the change is fully backwards-compatible.

bouweandela:
Maybe we should wait for Iris to be numpy 2 compatible: https://github.com/SciTools/iris/pull/6035

bouweandela:
It looks like that will happen when Iris 3.11 is released at the earliest, planned in October: https://github.com/SciTools/iris/discussions/5571#discussioncomment-9792428.

valeriupredoi:
many thanks for the reviews and comments @schlunma and @bouweandela :beer: x2 I don't think we can wait until October TBF, I say we get this in and see what's the worst it can happen, and fairly sure not much, and if there are problems, we can alert iris and tell them to hurry up :grin:

esmvalcore/preprocessor/_regrid_esmpy.py
valeriupredoi: yes, it is, functionally and it returns the same sort of structure, only typed differently

esmvalcore/preprocessor/_regrid_unstructured.py
valeriupredoi: not sure, will test - but I agree about points' types indeed

bouweandela:
> I say we get this in and see what's the worst it can happen

I'm sorry, but I have to disagree. What would most likely happen is that users will get wrong results and/or crashes.

valeriupredoi:
> > I say we get this in and see what's the worst it can happen
> 
> I'm sorry, but I have to disagree. What would most likely happen is that users will get wrong results and/or crashes.

that is indeed the possibility; I guess there's one way to find out about the magnitude of that - run the iris tests with a numpy>=2 pin in an iris fork

esmvalcore/preprocessor/_regrid_unstructured.py
valeriupredoi: nopers, a float there throws this:
```
>       src_points_with_convex_hull[-2 * n_hull:-n_hull, 1] -= 360.
E       numpy._core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'subtract' output from dtype('float64') to dtype('int8') with casting rule 'same_kind'
```
so we either np.int32 the 360 or we np.float the left member

esmvalcore/preprocessor/_derive/co2s.py
valeriupredoi: done in 85b2aa4f6523a4ec1d2eb2e2d00db12d17f6b405

esmvalcore/preprocessor/_time.py
valeriupredoi: done in 05dea6644aee29f5da34d64215c71cbb3e6920a9

tests/unit/preprocessor/_time/test_time.py
valeriupredoi: yes it is, as it is now, ie the test does what it says on the lid for both numpy<2 and numpy>=2, here's the failed test for numpy>=2 if we don't change to forced typing:
```
>       with pytest.raises(ValueError, match=re.escape(msg)):
E       AssertionError: Regex pattern did not match.
E        Regex: "Seasons\\ \\('DJF',\\ 'MAM',\\ 'JJA',\\ 'SON'\\)\\ do\\ not\\ match\\ prior\\ season\\ extraction\\ \\['JASOND',\\ 'JFMAMJ'\\]\\."
E        Input: "Seasons ('DJF', 'MAM', 'JJA', 'SON') do not match prior season extraction [np.str_('JASOND'), np.str_('JFMAMJ')]."

tests/unit/preprocessor/_time/test_time.py:930: AssertionError
```

valeriupredoi:
Monitoring https://github.com/SciTools/iris/issues/6043

valeriupredoi:
OK folks, iris have implemented numpy2 support and closed https://github.com/SciTools/iris/issues/6043 - their 3.11 is fast approaching, see https://github.com/SciTools/iris/milestone/51 - so could I please ask you to get back to reviewing this so we can plop it in as soon as the new iris is out? :beer: The two review chevaliers: @schlunma @bouweandela :carousel_horse:

valeriupredoi:
@schlunma would you have a bit of time and come back to finishing the review on this, please, bud? Iris 3.11 is fully out and tested for here, and once we get this into `main`, then I can finish #2566 and with that we'll have support for the latest Python 3.13 too - we'll shipshape and completely up to date with our dependencies :beer:

esmvalcore/preprocessor/_regrid_esmpy.py
```diff
@@ -283,8 +283,11 @@ def get_grid(
         num_peri_dims = 1
     else:
         num_peri_dims = 0
+
+    # previous to numpy 2.0, a np.array(esmpy_lat.shape) was used
+    # numpy>=2.0 throws ValueError: matrix transpose with ndim<2 is undefined
```
schlunma, line 288:
```suggestion
```

Don't think we need that if it's backwards-compatible

pyproject.toml
```diff
@@ -61,7 +61,7 @@ dependencies = [
     "pyyaml",
     "requests",
     "scipy>=1.6",
-    "scitools-iris>=3.10.0",
+    "scitools-iris>=3.11",  # 3.11 first to support numpy2 and Python3.13
```
schlunma, line 64:
```suggestion
    "scitools-iris>=3.11",  # 3.11 first to support Numpy 2 and Python 3.13
```

environment.yml
```diff
@@ -18,7 +18,7 @@ dependencies:
   - fire
   - geopy
   - humanfriendly
-  - iris >=3.10.0
+  - iris >=3.11  # 3.11 first to support numpy2 and Python3.13
```
schlunma, line 21:
```suggestion
  - iris >=3.11  # 3.11 first to support Numpy 2 and Python 3.13
```

tests/unit/preprocessor/_compare_with_refs/test_compare_with_refs.py
```diff
@@ -740,6 +744,7 @@ def test_distance_metric_fully_masked_data(
     assert out_cube.dtype == np.float64
 
     expected_data = np.ma.masked_all(())
+    print("out/in/metric", out_cube.data, expected_data, metric)
```
schlunma, line 747:
```suggestion
```

esmvalcore/preprocessor/_regrid_unstructured.py
schlunma: Sorry, I still don't understand this. When I run the tests (numpy < 2), `src_points_with_convex_hull` is an array with dtype `float64` as expected by the CMOR standards. Numpy >= 2 should NOT change that. It would be wrong if this array is of dtype `int`.

I just tested this locally with dummy arrays and it works with `360` and `360.0` (the latter would be "more" correct). Maybe this was just a problem of an early numpy version. Would it be able to test this again?

tests/unit/preprocessor/_time/test_time.py
schlunma: The new error message in Numpy 2 will be

```
Seasons ('DJF', 'MAM', 'JJA', 'SON') do not match prior season extraction [np.str_('JASOND'), np.str_('JFMAMJ')].
```

This is not really nice. I think it would be much better to change the code [here](https://github.com/ESMValGroup/ESMValCore/blob/f74a2ada1609b5fa6fa657f6dcda2fee0a029f07/esmvalcore/preprocessor/_time.py#L635) to

```py
old_seasons = sorted({str(s) for s in cube.coord("clim_season").points})
```

to get the old nice-looking error message back.

schlunma (commented):
Thanks V, great that we will have support for numpy 2 soon! A couple of comments from me.

esmvalcore/preprocessor/_regrid_unstructured.py
valeriupredoi: nope, the test [still fails](https://app.circleci.com/pipelines/github/ESMValGroup/ESMValCore/12022/workflows/7aaea67a-5cf4-4978-93e9-890abbca7f1d/jobs/50488?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary) - it's looking at it as `int8` by default, hence my workaround to cast it to `int32`

valeriupredoi: @schlunma directly referencing you since I can't see my comment on the main PR page for some reason (and maybe you can't either, thanks, GitHub)

valeriupredoi:
OK for some reason my two comments are not showing up on the main page - was saying that that test still fails with numpy=2.1.3 (latest), see [test fail](https://app.circleci.com/pipelines/github/ESMValGroup/ESMValCore/12022/workflows/7aaea67a-5cf4-4978-93e9-890abbca7f1d/jobs/50488?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary) - numpy is looking at `360` as an `int8` and spitting an Overflow, hence my workaround to make it `int32` and avoid the Overflow - it could be an `float64` like any coord point in CMOR - just anything to avoid silly numpy thinking it's int8

esmvalcore/preprocessor/_regrid_unstructured.py
schlunma: Okay, I found the problem, was looking at the wrong test before. It looks like the default dtype for cubes and their coords is [`int8`](https://github.com/ESMValGroup/ESMValCore/blob/f74a2ada1609b5fa6fa657f6dcda2fee0a029f07/tests/unit/preprocessor/_regrid/__init__.py#L48)...I think it would make sense to change this to something more realistic like `int32` or `int` (`float` might break other tests). There are other tests to that use that, let me try to fix them too. Can I push to this branch?

valeriupredoi: absolutely, please do! :beer:

valeriupredoi:
green green green :green_circle: like at WEC races :checkered_flag: - cheers for fixing, Manu!

schlunma (approved):
All green now 🟢 Thanks V!!

valeriupredoi:
I'll merge now, if that's OK with you - will get @bouweandela to review the Python 3.13 PR #2566

Issue #2254, opened 2023-11-10
ERA5 hurs or huss derivation


malininae:
I've noticed that ERA5 single levels doesn't have surface humidity, so to obtain the `hurs` or `huss` one would need to use input from `tas`, `tdps` and `ps`. What's the best way to derive it operationally? The only solution I cam up with is to create a derivation function in the preprocessor. Would it be a good idea? (I am working with daily data, so hurs/huss derivation will make sense climatologically).  This issue is remotely related to #1806.

PR #2397, opened 2024-04-25
Adding hurs (realtive humidity) derivation script

malininae:
Hi folks,

Following the April 2024 monthly meeting, I'm adding the relative humidity derivation (`hurs`). Before being able to open a real PR, I would need to add filtering in case the `hurs_cb.data` >100%. Also need to write test. Will do it before the workshop. 

This PR is related to #2254 .


esmvalcore/preprocessor/_derive/hurs.py
```diff
@@ -0,0 +1,45 @@
...
+    def calculate(cubes):
+        """Compute relative humidity.
+
+        Relative humidity computed from dewpoint temperature and 
+        surface air temperature following Bohren and Albrecht 1998.
+        """
+        tdps_cube = cubes.extract_cube(NameConstraint(var_name='tdps'))
+        tas_cube = cubes.extract_cube(NameConstraint(var_name='tas'))
+
+        cubes_difference = iris.analysis.maths.subtract(tas_cube, tdps_cube)
```
bouweandela, line 35:
```suggestion
        cubes_difference = tas_cube - tdps_cube
```
Would this also work? Did you read the section on [cube maths](https://scitools-iris.readthedocs.io/en/stable/userguide/cube_maths.html) in the iris documentation?

bouweandela:
>  would need to add filtering in case the `hurs_cb.data` >100%

This could probably be done with code like this:
```python
import dask.array as da
hurs_cb.data = da.ma.where(hurs_cb.core_data() > 100., hurs_cb.core_data(), 100.)
```

bouweandela:
If you would like a quicker response in the future, it usually helps if you tag people so they get a notification, e.g. @bouweandela

esmvalcore/preprocessor/_derive/hurs.py
malininae: Thanks, so when I was trying it with simple `tas_cube-tdps_cube` the issue was that I was running out of memory, while it wasn't a case with `subtract`. I will test it again with the time diagnostic and will let you know.

bouweandela: That would be surprising because it should do the same thing, i.e. the minus operator on an Iris cube is implemented using the exact same function:
https://github.com/SciTools/iris/blob/c5e0353da11c263656c6c77e5e56eafefcede077/lib/iris/cube.py#L3968

malininae:
Finally finished the filtering and the test, hopefully, still in time for 2.12.

sloosvel (commented):
Hi @malininae ,

Code looks good to me, I think you can also use the divide operator / between cubes, as @bouweandela pointed out with the substract and multiply operations.

The release for 2.12 has been delayed, so since there is time, maybe it would be good to have a scientific review as well to test the results of the derive script.

malininae:
> Hi @malininae ,
> 
> Code looks good to me, I think you can also use the divide operator / between cubes, as @bouweandela pointed out with the substract and multiply operations.
> 
> The release for 2.12 has been delayed, so since there is time, maybe it would be good to have a scientific review as well to test the results of the derive script.

Thanks so much @sloosvel! I changed the division, and updated the tests to match [this paper](https://journals.ametsoc.org/view/journals/bams/86/2/bams-86-2-225.xml). The values with the rounding are consistent with the table, essentially what we expect. The code is essentially just programming the formula 12 in the paper. Please, let me know if scientific review still required or if there's someone specific I should tag.

sloosvel (approved):
Thanks @malininae ! I guess having such a comprehensive test is enough to validate the results

valeriupredoi (approved):
cheers muchly, folks :beer:

PR #2445, opened 2024-06-03
Updated esmf-related pins

schlunma:
## Description

With the new [release](https://github.com/SciTools-incubator/iris-esmf-regrid/releases/tag/v0.10.0) of iris-esmf-regrid we can update some pins in the environment.

See also https://github.com/SciTools-incubator/iris-esmf-regrid/pull/342#issuecomment-2092921514.

***

## [Checklist](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#checklist-for-pull-requests)

It is the responsibility of the author to make sure the pull request is ready to review. The icons indicate whether the item will be subject to the [🛠 Technical][1] or [🧪 Scientific][2] review.

[1]: https://docs.esmvaltool.org/en/latest/community/review.html#technical-review
[2]: https://docs.esmvaltool.org/en/latest/community/review.html#scientific-review

- [x] [🧪][2] The new functionality is [relevant and scientifically sound](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#scientific-relevance)
- [x] [🛠][1] This pull request has a [descriptive title and labels](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-title-and-label)
- [x] [🛠][1] Code is written according to the [code quality guidelines](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#code-quality)
- [x] [🧪][2] and [🛠][1] [Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available
- [x] [🛠][1] [Unit tests](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#tests) have been added
- [x] [🛠][1] Changes are [backward compatible](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#backward-compatibility)
- [x] [🛠][1] Any changed [dependencies have been added or removed](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#dependencies) correctly
- [x] [🛠][1] The [list of authors](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#list-of-authors) is up to date
- [x] [🛠][1] All [checks below this pull request](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-checks) were successful

***

To help with the number pull requests:

[...]

PR #2447, opened 2024-06-04
[Merge after v2.11.0 Release] Remove support for Python 3.9

### LINKED ISSUE #2406: Drop support for Python 3.9
Drop support for Python 3.9 as per [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule).

valeriupredoi:
## Description

Closes #2406 

AFAIK I removed all references for Python 3.9 - if you know of any others please holler back so I can replace them :beer: 

***

***


bouweandela:
I think we do not need `from __future__ import annotations` anymore with recent Python versions, would you like to give that a try? i.e. remove it from everywhere and then run the GitHub Actions matrix tests to see if it works?

doc/quickstart/install.rst
```diff
@@ -22,7 +22,7 @@ and install ESMValCore into it with a single command:
 
 .. code-block:: bash
 
-    conda create --name esmvalcore -c conda-forge esmvalcore 'python=3.10'
+    conda create --name esmvalcore -c conda-forge esmvalcore 'python=3.12'
```
bouweandela, line 25: How about removing the python installation altogether? It will be pulled in as a dependency anyway.

```suggestion
    conda create --name esmvalcore -c conda-forge esmvalcore
```

bouweandela:
This can also be removed:
https://github.com/ESMValGroup/ESMValCore/blob/07ba25c425724e61be1ce109ea98a2d791abb3fa/tests/unit/preprocessor/_regrid_esmpy/test_regrid_esmpy.py#L384-L386

bouweandela:
This needs an update: https://github.com/ESMValGroup/ESMValCore/blob/af5490f9c0e285f65ab8a213629ada7e903c1e36/setup.cfg#L37-L39

doc/quickstart/install.rst
valeriupredoi: totally agreed, bud!

valeriupredoi:
> tests/unit/preprocessor/_regrid_esmpy/test_regrid_esmpy.py

done in https://github.com/ESMValGroup/ESMValCore/pull/2447/commits/df2aa31b94ce10c0f0adc62d9036f2a567067b67

valeriupredoi:
> This needs an update:
> 
> https://github.com/ESMValGroup/ESMValCore/blob/af5490f9c0e285f65ab8a213629ada7e903c1e36/setup.cfg#L37-L39

done in https://github.com/ESMValGroup/ESMValCore/pull/2447/commits/fb2411b50f1b4a28110251db11c97de1cc58391d

valeriupredoi:
@bouweandela have a look at the flake8 barf now that we've removed from future... - F821 is indeed an [undefined name](https://flake8.pycqa.org/en/latest/user/error-codes.html) in flake8 error parlance, but it's typing assignemnt, not variable assignment - any suggestions?

bouweandela:
I tried to fix the flake8 error (it happens because the class used as a type hint is not yet defined at the point where it is used, so in some cases you can just re-order the code and then it works), but then when I tried to run the tests and found that we still need the `from __future__ import annotations` also where we use the bare class name (i.e. without quotes) when the used classes are imported only for type checking. In short, I think we cannot yet get rid of the `from __future__ import annotations`. Thanks for trying though!

valeriupredoi:
> I tried to fix the flake8 error (it happens because the class used as a type hint is not yet defined at the point where it is used as a type hint, so in some cases you can just re-order the code and then it works), but then I tried to run the tests and found that we still need the `from __future__ import annotations` also where we use the bare class name (i.e. without quotes) when the used classes are imported only for type checking. In short, I think we cannot yet get rid of the `from __future__ import annotations`. Thanks for trying though!

cheers for testing, bud! Lemme revert the commit then - easy-peasy :+1:

valeriupredoi:
OK buds, reverted those two last commites (running GA and removing from future) :beer:

bouweandela (approved):
Thanks! Maybe we should wait with merging until the v2.11 release is done?

valeriupredoi:
Yep let's do that, maybe we do a rebuild when ESMValTool has the same package specs? We sreing esmpy>=8.6.0 causing some issues with resolving the env with python 3.12 at the mo at Tool, so we can wait a bit

valeriupredoi:
I'd argue we should add a label a la "warrants/motivates a rebuild" just as we have "needs new ESMValCore release" in Tool - what do you think, bud?

valeriupredoi:
also, excuse my verbosity today, what do you think I just remove 3.9 from the GA tests ie https://github.com/ESMValGroup/ESMValCore/actions/runs/9393366881 in a separate PR just so that release folks don't get annoyed when it comes to checking tests for the release?

bouweandela:
They may want to run the GitHub actions on the v2.11.x branch instead of on `main`, as it is getting rather behind `main` and I would not expect the failing Python 3.9 tests on the v2.11.x branch as they were introduced in https://github.com/ESMValGroup/ESMValCore/pull/2445#issuecomment-2145281138, which is not included in v2.11.x.

valeriupredoi:
sounds like what I'd do too :grin:

valeriupredoi:
@bouweandela now that @ehogan has released 2.11, would you mind if I merged this, so we get us rid of that dinosaur once and for all? :sauropod:

valeriupredoi:
:sauropod: just about to be extinct :grin: Many thanks for the review @bouweandela :beer:

Issue #2371, opened 2024-03-14
New configuration file plan


bouweandela:
Here is the design for the new way of organizing our configuration files, that we decided on at the last workshop. We will use a simplified version of the [Dask configuration system](https://docs.dask.org/en/stable/configuration.html). That means that the configuration will consist of nested dictionaries that are organized per component of the tool. There is a default configuration shipped with the ESMValCore and ESMValTool, that will be updated with the configuration specified by the user through one or more YAML files in a user-specified directory.

For a smooth transition, we will keep supporting the existing configuration key: value pairs, but add new ones as well.

Example of what a future configuration file could look like:
```yaml




```

TODO: finish typing this

schlunma:
Thank you very much Bouwe for writing this up, this looks really great! I am really looking forward to these changes ❤️ 

I have three comments:

1. In addition to the default settings shipped with ESMValTool, I think it should also be possible to consider system-wide configurations, for example in `/etc/esmvaltool/`. This would allow us to add default configurations for commonly used HPC systems like Levante, and users would not have to mess around with configuration on these machines.
1. What would happen if a specific setting is used in multiple configuration files? It's clear that user-defined options should overwrite system-wide configuration and the default configuration, but what about the case when a user (accidentally) has two files with conflicting settings? Would we raise an error? Would we allow that and simply use the "last" file encountered (e.g., after sorting them alphabetically)?
1. I think we should avoid allowing arbitrary strings in YAML keys. In my experience, this is a major source of confusion for new(er) users of the tool. For example, in the definition of preprocessors in the recipe, it can be really confusing to find out which keys are arbitrary and which are not:

```yaml
preprocessors:  # NOT arbitrary
  regrid:  # arbitrary
    regrid:  # NOT arbitrary
      target_grid: 2x2  # NOT arbitrary
      scheme: linear  # NOT arbitrary
```

Same goes for the definition of diagnostic scripts, variables, etc. My suggestion would be to avoid patterns like

```yaml
projects:
  CMIP6:
    data:
      CMIP6-local:
        type: esmvalcore.local.LocalDataSource  # this could be omitted for local data
        path: /work/bd0854/DATA/ESMValTool2/CMIP6_DKRZ
        dirname: '{activity}/{institute}/{dataset}/{exp}/{ensemble}/{mip}/{short_name}/{grid}/{version}'
        filename: '{short_name}_{mip}_{dataset}_{exp}_{ensemble}_{grid}*.nc'
[...]

Issue #2369, opened 2024-03-13
Make `dask.yml` path configurable 


malininae:
@Karen-A-Garcia and I are operationalizing ESMValTool and running a few recipes sequentially. Although our dask expertise are quite limited, we've got a problem that [this recipe](https://github.com/malininae/ESMValTool/blob/rapid_event_attr/esmvaltool/recipes/cccma_rapid_attribution/recipe_observational_analysis_heatwave.yml) doesn't seem to work with any dask setup we've tried, while the other recipes do work. We'll try to figure out what the deal is, but, I can see the value of adding the optional configurable `dask.yml` path to the `config_user.yml`, sort of like `config_developer.yml` is set up. Any objections? If someone could volunteer themselves, great, if not I can do it in a dream land called 'after April 15th'.

bouweandela:
It looks like that recipe is using a preprocessor function that is not yet lazy (see https://github.com/ESMValGroup/ESMValCore/issues/674 and https://github.com/SciTools/iris/pull/5795), therefore your best bet is the default threaded scheduler. If you run out of memory, try reducing the number of workers (`num_workers`) in your Dask configuration. This can e.g. be done by creating a file `~/.config/dask/dask.yml` with the content `num_workers: 4` (see the [Dask docs](https://docs.dask.org/en/stable/scheduling.html?highlight=num_workers#configuration) for more information.

bouweandela:
> I can see the value of adding the optional configurable dask.yml path to the config_user.yml, sort of like config_developer.yml is set up. Any objections?

Improving how users can configure Dask is indeed something we would like to do, see https://github.com/ESMValGroup/ESMValCore/issues/2040 for previous discussion on the topic. However, the plan is to tackle this as part of a larger overhaul of the configuration, as users often find the current way of configuring things, where settings are spread out over multiple files, confusing. See https://github.com/ESMValGroup/ESMValCore/issues/2371 for the plan.

k-a-webb:
I'd like to encourage the option of having a configurable the dask configuration.
Ideally it would be easy to select a specific dask configuration (or no configuration) for individual recipes, just as you can choose a specific `config_user` file.
 
[...]

PR #2448, opened 2024-06-05
Merge configuration object from multiple files (instead of one single file)

### LINKED ISSUE #805: Look into generating command line documentation automatically
It would be great if we could have complete documentation about the command line options available on readthedocs. Maybe there exist libararies that can automatically generate the documentation from the command line interface that we can use for this?

### LINKED ISSUE #1433: `esmvaltool run --help` does not show all possible command line options
Currently, `esmvaltool run --help` returns

```
> esmvaltool run --help
...

FLAGS
    --config_file=CONFIG_FILE
        Type: Optional[]
        Default: None
        Configuration file to use. If not provided the file ${HOME}/.esmvaltool/config-user.yml will be used.
    --resume_from=RESUME_FROM
        Type: Optional[]
        Default: None
        Resume one or more previous runs by using preprocessor output files from these output directories.
    --max_datasets=MAX_DATASETS
        Type: Optional[]
        Default: None
        Maximum number of datasets to use.
    --max_years=MAX_YEARS
        Type: Optional[]
        Default: None
        Maximum number of years to use.
    --skip_nonexistent=SKIP_NONEXISTENT
        Default: False
        If True, the run will not fail if some datasets are not available.
    --offline=OFFLINE
        Type: Optional[]
        Default: None
        If True, the tool will not download missing data from ESGF.
    --diagnostics=DIAGNOSTICS
        Type: Optional[]
        Default: None
        Only run the selected diagnostics from the recipe. To provide more than one diagnostic to filter use the syntax 'diag1 diag2/script1' or '("diag1", "diag2/script1")' and pay attention to the quotes.
    --check_level=CHECK_LEVEL
        Default: 'default'
        Configure the sensitivity of the CMOR check. Possible values are: `ignore` (all errors will be reported as warnings), `relaxed` (only fail if there are critical errors), default (fail if there are any errors), strict (fail if there are any warnings).
    Additional flags are accepted.

```

These are clearly not all possible arguments, see for example here:

https://github.com/ESMValGroup/ESMValCore/blob/4b8bf66ce6ef51cc6814b5d18357ae91684eb0b7/esmvalcore/experimental/config/_config_validators.py#L254-L289

@ESMValGroup/esmvaltool-coreteam  Is there a certain strategy which options are shows in the help and which not? I think it would definitely make sense to show additional options (not all of them!), like `run_diagnostic` which even [appears in the documentation](https://docs.esmvaltool.org/projects/esmvalcore/en/latest/quickstart/run.html#running).

schlunma:
## Description

This PR is the first step towards our new configuration system (see https://github.com/ESMValGroup/ESMValCore/issues/2371). It allows reading the configuration from multiple files (and directories), similar to [Dask's configuration](https://docs.dask.org/en/stable/configuration.html). The different files are merged using [`dask.config.collect()`](https://docs.dask.org/en/stable/configuration.html#dask.config.collect).

The following directories are considered (descending priority):

1. The directory specified via the ``--config_dir`` command line argument
2. User configuration directory (by default ``~/.config/esmvaltool``, but this can be changed with
the ``ESMVALTOOL_CONFIG_DIR`` environment variable)
3. Default configuration in repository (`esmvalcore/config/configurations/defaults`)

Please note that these directories may only contain those files that we used to call "user configuration files". If an old "developer configuration file" (or any other YAML file with a different format) is present in this directory, this will lead to errors about invalid configuration options.

Detailed instructions on how to update your existing setup can be found in the deprecation warnings that you get from running the tool, or in the section on deprecations below.

Related to #2371
Closes #805
Closes #1433

Link to documentation:
- https://esmvaltool--2448.org.readthedocs.build/projects/ESMValCore/en/2448/quickstart/configure.html
- https://esmvaltool--2448.org.readthedocs.build/projects/ESMValCore/en/2448/api/esmvalcore.config.html

### Usage of single configuration file `~/.esmvaltool/config-user.yml`

As mentioned above, ESMValTool will now read configuration *directories* (instead of one single file). By default, the directory `~/.config/esmvaltool` will be considered (can be changed with command line argument `--config_dir`; see below). To switch to the new format, run

```bash
mkdir -p ~/.config/esmvaltool && mv ~/.esmvaltool/config-user.yml ~/.config/esmvaltool
```

Then you can run ESMValTool with

```bash
esmvaltool run <other optional command line options> <recipe>
````

### Configuration option `config_file` / command line argument `--config_file`:

As mentioned above, ESMValTool will now read configuration *directories* (instead of one single file). Instead of specifying one file, specify a configuration directory with `--config_dir`. To switch to the new format, ensure that the directory that contains your old config file only contains this one YAML file (no other YAML files like old "developer" configuration files) and run ESMValTool with

```bash
esmvaltool run --config_dir=<path/to/your/old/config_dir> <other optional command line options> <recipe>
```

### `Config.load_from_file()`:

Please `Config.load_from_dirs()` instead, e.g.

```py
from pathlib import Path
config_file = Path('path/to/my/config_file.yml')
CFG.load_from_dirs([config_file.parent])
```

### `Session.config_dir`:

There is no replacement for this attribute.

### Specify `extra_facets_dir` as `tuple`:

Please use a `list` instead. For example,

```py
CFG['extra_facets_dir'] = ('path/1', 'path/2')  # deprecated, do not use anymore
CFG['extra_facets_dir'] = ['path/1', 'path/2']  # preferred syntax
```

## Backwards-incompatible change

In some cases, simply copying the old configuration file (e.g., from `~/.esmvaltool/config-user-yml`) to the new location might not be enough. In some cases, it might be necessary to adapt `drs` and `rootpath` settings. For example, on **Levante**, it might be necessary to change

```yaml
drs:  # deprecated, do not use anymore
  CMIP3: DKRZ
  CMIP5: DKRZ
  CMIP6: DKRZ
  CORDEX: BADC
```

to

```yaml
drs:  # preferred syntax
  CMIP3: DKRZ
  CMIP5: DKRZ
  CMIP6: DKRZ
  CORDEX: BADC
  obs4MIPS: default
```

<details>

Due to the new way settings are merged across configuration files, nested dictionaries are properly updated now.

Example:

The files

```yaml
drs:  # default config
  CMIP5: ESGF
  obs4MIPS: ESGF
```

and

```yaml
drs:  # user config
  CMIP5: DKRZ
  CMIP6: DKRZ
```

were previously merged to

```yaml
drs:  # final config
  CMIP5: DKRZ
  CMIP6: DKRZ
```

In the new syntax, they are merged to

```yaml
drs:  # final config
  CMIP5: DKRZ
  CMIP6: DKRZ
  obs4MIPS: ESGF
```

Thus, in the example above, it is necessary to explicitly specify `obs4MIPS: default`.

</details>

***

- [ ] [🛠][1] ~Changes are [backward compatible](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#backward-compatibility) see above~

***


esmvalcore/config/_config_object.py
```diff
@@ -233,23 +252,90 @@ def _get_config_path_from_cli() -> None | str:
...
+
+        1. If set, the directory specified with the ``ESMVALTOOL_CONFIG_DIR``
+           environment variable.
+
+        2. The user configuration directory (by default
+           ``~/.config/esmvaltool``, but this can be changed with the
+           ``--config_dir`` command line argument when running the
+           ``esmvaltool`` command line tool).
+
+        3. Default configuration, see :ref:`config_options`.
```
bouweandela, line 297: Would it be possible to change this to:

        1. The ``--config_dir`` command line argument when running the
           ``esmvaltool`` command line tool

        2. The user configuration directory: by default
           ``~/.config/esmvaltool``, but this can be changed with the
           ``ESMVALTOOL_CONFIG_DIR` environment variable.
        3. Default configuration, see :ref:`config_options`.

esmvalcore/config/_config_object.py
```diff
@@ -359,6 +474,142 @@ def _fixed_file_dir(self):
...
+
+    return None
+
+
+def _get_user_config(ignore_internal_env: bool = False) -> tuple[str, Path]:
+    """Get user configuration directory.
+
+    The following directories are considered (descending priority):
+
+    1. Internal `_ESMVALTOOL_USER_CONFIG_DIR_` environment variable (this
```
bouweandela, line 515: Would it be possible to use the `ESMVALTOOL_CONFIG_DIR` for this instead of adding an extra internal variable? That seems conceptually simpler.

esmvalcore/config/config_defaults/config-user.yml
```diff
@@ -1,5 +1,5 @@
 ###############################################################################
-# Example user configuration file for ESMValTool
```
bouweandela, line 2: How about calling the directory `esmvalcore/config/configurations`? That would allow storing all kinds of configurations in here that people could request from the command line, e.g. `esmvaltool config get dask_settings_levante_compute_partition` or `esmvaltool config get input_file_paths_jasmin` and that would then copy the files `dask_settings_levante_compute_partition.yml` or `input_file_paths_jasmin.yml` respectively from here.

bouweandela (commented):
Nice work here @schlunma!

bouweandela:
Would it be possible to do this in a backward compatible way? The deprecations are fine, but it would be nice if existing configurations kept working (at least for a while). For example, if  `ESMVALTOOL_CONFIG_DIR` is not set and `~/.config/esmvaltool` does not exist, we could just continue to read the values from `~/.esmvaltool/config-user.yml` or the command line `--config-file` argument, right? Then people would only need to switch when they actually need the new functionality.

doc/recipe/overview.rst
```diff
@@ -3,8 +3,8 @@
 Overview
 ********
 
-After ``config-user.yml``, the ``recipe.yml`` is the second file the user needs
-to pass to ``esmvaltool`` as command line option, at each run time point.
+The recipe is the main control fail of ESMValTool.
```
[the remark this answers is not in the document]
schlunma: Fixed in https://github.com/ESMValGroup/ESMValCore/pull/2448/commits/ee57cb067be67fb141c484b272d47c0cc14783b5 😅

esmvalcore/config/config_defaults/config-user.yml
schlunma: Would it really make sense to put all requestable configuration options into the default configuration directory that is always read? Wouldn't this create a big mess of configuration options? I think we should distinguish between the default options (these will be always read) and the options that can be requested and copied over to the user's config directory.

esmvalcore/config/_config_object.py
schlunma: The main reason we need the `_ESMVALTOOL_USER_CONFIG_DIR_` variable is that the path given via `--config_dir` is properly given to the diagnostics (these are run with `subprocess`, so `sys.argv` is different for those). Since the path for `--config_dir` and `ESMVALTOOL_CONFIG_DIR` can be different (in the current version but also in the change you suggest above), we need an additional environment variable.

esmvalcore/config/_config_object.py
schlunma: Yes, this is certainly possible and probably also more straightforward. I opted for the current version to stick as closely as possible to the old way of specifying the configuration (where the `--config-file` argument overwrote the default user config location `~/.esmvaltool/config-user-yml`).

schlunma:
> Would it be possible to do this in a backward compatible way? The deprecations are fine, but it would be nice if existing configurations kept working (at least for a while). For example, if `ESMVALTOOL_CONFIG_DIR` is not set and `~/.config/esmvaltool` does not exist, we could just continue to read the values from `~/.esmvaltool/config-user.yml` or the command line `--config-file` argument, right? Then people would only need to switch when they actually need the new functionality.

Yes, 100% agree. That's why I already implemented that in exactly this way. This will work for two more minor versions, and only then it will raise an error.

esmvalcore/config/_config_object.py
schlunma: Done in https://github.com/ESMValGroup/ESMValCore/pull/2448/commits/2decd64f0b43f58cb8145c1fd4b1cbd42decf9a6.

Turns out that changing this allowed me to get rid of all the complicated code that took care of parsing the CLI arguments (>200 lines!): if the environment variable `ESMVALTOOL_CONFIG_DIR` can be used to replace the default user dir instead of the CLI argument, no special parsing is needed.

Right now the tool will NOT raise an error if the user config dir (the one in home or given by the env variable) does not exist (there will be a message at the very beginning that the dir does not exist). This was much easier to implement this way and the behavior is identical to when the ESMValTool Python API is used. When a `--config_dir` is specified and it does not exist, the tool WILL raise an error.

esmvalcore/config/_config_object.py
schlunma: Removed in https://github.com/ESMValGroup/ESMValCore/commit/2decd64f0b43f58cb8145c1fd4b1cbd42decf9a6, see comment above

schlunma:
@bouweandela I think I addressed all your comments, could you have another look? Thanks!

esmvalcore/config/config_defaults/config-user.yml
bouweandela: Good point. How about putting the default in `esmvalcore/config/configurations/default` and then we can add others later? e.g. `esmvalcore/config/configurations/levante`, `esmvalcore/config/configurations/jasmin`, etc?

esmvalcore/config/config_defaults/config-user.yml
```diff
@@ -13,14 +13,6 @@
 # file.
 #
 ###############################################################################
-#
```
bouweandela, line 16: What is the plan for the ESMValTool copy, remove?

bouweandela (changes requested):
Thanks for the massive effort @schlunma! I have a few more suggestions for improving the documentation and then I think we can ask some people to test

esmvalcore/config/config_defaults/config-user.yml
schlunma: Yes, I think that's the easiest solution. We don't want users to simply copy-paste the file anymore but rather use `esmvaltool config ...`, so it doesn't need be located in the repo anymore.

doc/quickstart/configure.rst
```diff
@@ -1,203 +1,262 @@
...
-
-If the option ``--path`` is omitted, the file will be created in
-``${HOME}/.esmvaltool``
-
-The following shows the default settings from the ``config-user.yml`` file
-with explanations in a commented line above each option. If only certain values
-are allowed for an option, these are listed after ``---``. The option in square
-brackets is the default value, i.e., the one that is used if this option is
-omitted in the file.
+Example:
```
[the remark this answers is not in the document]
schlunma: Done in https://github.com/ESMValGroup/ESMValCore/pull/2448/commits/61445a022079c191ac604a17a3e1eb8b73773d25

doc/quickstart/configure.rst
```diff
@@ -1,203 +1,262 @@
...
 
-.. warning::
+These files can live in any of the following locations:
+
+1. The directory specified via the ``--config_dir`` command line argument (if
+   this points to an invalid directory, an error will be raised).
+
+2. The user configuration directory: by default ``~/.config/esmvaltool``, but
+   this can be changed with the ``ESMVALTOOL_CONFIG_DIR`` environment variable
+   (if this points to an invalid directory, this is silently ignored).
```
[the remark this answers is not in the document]
schlunma: Done in https://github.com/ESMValGroup/ESMValCore/pull/2448/commits/3829210074a604d258c64d81af5b9f909bd5d5e9

esmvalcore/config/config_defaults/config-user.yml
schlunma: Done in https://github.com/ESMValGroup/ESMValCore/pull/2448/commits/8b1fef80b706f5a3ca344ee34fe3deaa05730432

bouweandela (approved):

bouweandela:
@ESMValGroup/userengagementteam Could you please try this out and provide us with feedback on the user-friendliness?

k-a-webb:
Continuing from the thread in https://github.com/ESMValGroup/ESMValCore/issues/2369#issuecomment-2374176485

I've only started to try and use PR#2448, but quickly encountered an issue related to user-defined projects.
I have a custom configuration directory ([config-user.yml](https://github.com/user-attachments/files/17135383/config-user.yml.txt), [config-developer.yml](https://github.com/user-attachments/files/17135388/config-developer.yml.txt)):

```
esmvaltool_config:
  config-user.yml # slightly revised config-user.yml from ESMVal* v2.11.0
  config-developer.yml # contains custom projects, same as was used successfully in  ESMVal* v2.11.0
```

My issue is that the custom project raise an issue in `esmvalcore.config._validated_config` related to `esmvalcore.config/_config_validators.validate_drs` being set by the _default_ (i.e., not user specified) `esmvalcore/config/configurations/defaults/config-user.yml`. Since this file includes only:
```
drs: {'CMIP3': 'ESGF', 'CMIP5': 'ESGF', 'CMIP6': 'ESGF', 'CORDEX': 'ESGF', 'obs4MIPs': 'ESGF'}
```
Running ESMValTool with the custom config files:
`esmvaltool run recipes/test__Seasonal_Near-surface_air_temperature.yml --config_dir /path/to/esmvaltool_config/`
fail with the error message:
```
ERROR:esmvalcore._main:Program terminated abnormally, see stack trace below for more information:
  File "/fs/site5/eccc/crd/ccrn/users/rkw001/code/esmvalcore/ESMValCore_pr2448/esmvalcore/_main.py", line 405, in run
    CFG.load_from_dirs(config_dirs)
  File "/fs/site5/eccc/crd/ccrn/users/rkw001/code/esmvalcore/ESMValCore_pr2448/esmvalcore/config/_config_object.py", line 366, in load_from_dirs
    self.update(new_config_dict)
  File "<frozen _collections_abc>", line 982, in update
  File "/fs/site5/eccc/crd/ccrn/users/rkw001/code/esmvalcore/ESMValCore_pr2448/esmvalcore/config/_validated_config.py", line 65, in __setitem__
    raise InvalidConfigParameter(
esmvalcore.exceptions.InvalidConfigParameter: `CanESM5_dev` is not a valid config parameter.

  File "/fs/site5/eccc/crd/ccrn/users/rkw001/code/esmvalcore/ESMValCore_pr2448/esmvalcore/_main.py", line 591, in run
    fire.Fire(ESMValTool())
  File "/space/hall5/sitestore/eccc/crd/ccrn/users/rkw001/miniconda3/envs/a4d_env_pr2448/lib/python3.12/site-packages/fire/core.py", line 143, in Fire
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/space/hall5/sitestore/eccc/crd/ccrn/users/rkw001/miniconda3/envs/a4d_env_pr2448/lib/python3.12/site-packages/fire/core.py", line 477, in _Fire
                                ^^^^^^^^^^^^^^^^^^^^
  File "/space/hall5/sitestore/eccc/crd/ccrn/users/rkw001/miniconda3/envs/a4d_env_pr2448/lib/python3.12/site-packages/fire/core.py", line 693, in _CallAndUpdateTrace
                ^^^^^^^^^^^^^^^^^^^^^^
  File "/fs/site5/eccc/crd/ccrn/users/rkw001/code/esmvalcore/ESMValCore_pr2448/esmvalcore/_main.py", line 412, in run
    raise InvalidConfigParameter(
esmvalcore.exceptions.InvalidConfigParameter: Failed to parse configuration directory /home/rkw001/A4D_standard_diagnostics__v1.5_pr2448/work/v1.5pr_pr2448/esmvaltool_config (command line argument): `CanESM5_dev` is not a valid config parameter.
```

bouweandela:
Thanks for testing this! That is very useful feedback.

schlunma:
Thanks for the feedback @k-a-webb, that's great!

I think the issue here is that you put `config-user.yml` and `config-developer.yml` into the same directory. With the new way of reading configuration files, ESMValCore will interpret both as "user config files". Since these two are still very different, this fails. In the future, we'd like to get rid of this distinction, but for now, these need to be saved in separate directories.

I will add a note to the documentation to clarify this 👍

schlunma:
> I will add a note to the documentation to clarify this 👍

Done in https://github.com/ESMValGroup/ESMValCore/pull/2448/commits/3325b594d78a5b385f1c747323301c84a11ea884, please let me now if this is clear! Thanks!!

k-a-webb:
> I think the issue here is that you put config-user.yml and config-developer.yml into the same directory. With the new way of reading configuration files, ESMValCore will read both of them and interpret both as "user config files". Since we still distinguish between "user" and "developer" configuration files, and these are very different, this fails. 

Thanks! I was able to successfully run the recipe once I moved the `config-developer.yml` to a different directory than set by `--config_dir` (and updating the path in `config-user.yml`).

I only noticed after-the-fact that this was stated at the top of this PR thread, 

> Please note that these directories may only contain those files that we used to call "user configuration files". If an old "developer configuration file" (or any other YAML file with a different format) is present in this directory, this will lead to errors about invalid configuration options.

I would suggest adding a warning to this section of the documentation, [Developer configuration file](https://esmvaltool--2448.org.readthedocs.build/projects/ESMValCore/en/2448/quickstart/configure.html#developer-configuration-file), which states that the `config-developer.yml` file cannot be in the directory specified by `--config_dir`. I did not find anything in the documentation about the distinction between user/developer config files that you mentioned -- but perhaps I missed it.

I'm looking forward to the continued development of configurable custom configuration files! Thanks for all your hard work :)

Issue #2405, opened 2024-05-06
Automatically select a lazy regridding scheme in `regrid` preprocessor function for 2D lat/2D lon grids when possible


bouweandela:
Through iris-esmf-regrid and the [generic regridding schemes](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/recipe/preprocessor.html#horizontal-regridding), it is possible to regrid grids with 2D latitude and 2D longitude coordinates lazily. However, these regridding schemes are not automatically selected and instead the non-lazy ESMPyLinear/ESMPyNearest/ESMPyAreaWeighted built-in schemes are selected. This results in poor computational performance by default.

@schlunma @sloosvel Do you think it would be possible to improve the default so we prefer lazy schemes and automatically select those when possible?

sloosvel:
I'm not sure this is possible because the generic regridding schemes don't work for all models, you can get results such as this one using ESMFAreaWeighted:

![image](https://github.com/ESMValGroup/ESMValCore/assets/45196700/7dcb0c10-440e-43bc-9fae-deac30a5dcca)

bouweandela:
Could you provide some more details on that image, to help me understand the issue?
- What was the source grid or model?
- what was the target grid?
- Which regridding scheme was used?

sloosvel:
It happens to models that have OceanGridFix applied to them. The generated bounds work fine to regrid with the non-lazy regridders, but for some reason when you call the lazy AreaWeighted scheme, this happens. You can check with the recipe below for instance:

```yaml
documentation:
  title: test regrid
  description: |
    Test regrid
  authors:
     - loosveldt-tomas_saskia

datasets:
   - {dataset: FGOALS-g3, ensemble: r1i1p1f1, exp: historical, project: CMIP6, grid: gn, timerange: 1951/1951}


preprocessors:
  pp:
    regrid:
      target_grid: 1.5x1.5
      scheme: 
        reference: esmf_regrid.schemes:ESMFAreaWeighted

diagnostics:
  test_regrid:
    variables:
      tos:
        mip: Omon
        preprocessor: pp
    scripts: null


```

bouweandela:
An overview of regridding sea surface temperature for most models in CMIP6 to a 1x1 grid is available [here](https://esmvaltool.dkrz.de/shared/esmvaltool/regrid_tests/recipe_tos_regrid_20240518_212719/).

A first inspection reveals the following issues:
[...]

PR #2316, opened 2024-01-31
Proof of concept: running metadata and data computations on Dask

bouweandela:
## Description

This pull request splits the computation up in three stages:
1) Preprocessor functions are run in parallel using Dask without saving data
1) Preprocessor files are populated with data in parallel using Dask
1) Diagnostic scripts are run

Only works with `max_parallel_tasks: 1` at the moment.

Ideas for further improvements:
1) optimize multi-model functions, as these limit parallelism
1) use one delayed per group in multi-model/ensemble means to increase parallelism
1) try to make delayed operations 'pure', e.g. by copying the input cubes in `preprocess` before calling the preprocessor function
1) see if splitting `Dataset.load` prior to preprocessor step `concatenate` up in multiple delayeds improves parallelism

Blocking issues
---------------
These are things that block this from being used in practice.

1) ESMPy crashes if you try to from a different thread than the main one. Example script that produces the crash:
   ```python
   import threading

   import numpy as np


   def run():
       import esmpy
       m = esmpy.Manager(debug=True)
       esmpy.Grid(np.array((10, 20)),
                  num_peri_dims=1,
                  staggerloc=[esmpy.StaggerLoc.CENTER])


   def main():

       thread = threading.Thread(target=run)
       thread.start()
       thread.join()


   if __name__ == '__main__':
       main()
   ```
   results in `Segmentation fault (core dumped)` and a log file called `PET0.ESMF_LogFile` is written by ESMF with the following content:
   ```
   20240226 150217.785 INFO             PET0 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
   20240226 150217.785 INFO             PET0 !!! THE ESMF_LOG IS SET TO OUTPUT ALL LOG MESSAGES !!!
   20240226 150217.785 INFO             PET0 !!!     THIS MAY CAUSE SLOWDOWN IN PERFORMANCE     !!!
   20240226 150217.785 INFO             PET0 !!! FOR PRODUCTION RUNS, USE:                      !!!
   20240226 150217.785 INFO             PET0 !!!                   ESMF_LOGKIND_Multi_On_Error  !!!
   20240226 150217.785 INFO             PET0 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
   20240226 150217.785 INFO             PET0 Running with ESMF Version   : 8.4.2
[...]

PR #2457, opened 2024-06-14
Add an iris-esmf-regrid based regridding scheme

bouweandela:
## Description

Add an iris-esmf-regrid based regridding scheme

Part of #2405.

Link to documentation:

***

***


esmvalcore/preprocessor/_regrid.py
```diff
@@ -533,9 +535,13 @@ def _get_target_grid_cube(
...
 
-def _attempt_irregular_regridding(cube: Cube, scheme: str) -> bool:
+def _attempt_irregular_regridding(
+    src_cube: Cube,
+    tgt_cube: Cube,
+    scheme: str,
+) -> bool:
     """Check if irregular regridding with ESMF should be used."""
-    if not has_irregular_grid(cube):
+    if not (has_irregular_grid(src_cube) or has_irregular_grid(tgt_cube)):
```
bouweandela, line 544: @schlunma Does this also work for ugrid data?

doc/recipe/preprocessor.rst
```diff
@@ -997,9 +1001,12 @@ with the remaining entries of the ``scheme`` dictionary passed as keyword
 arguments.
 
 One package that aims to capitalize on the :ref:`support for unstructured grids
-introduced in Iris 3.2 <iris:ugrid>` is :doc:`iris-esmf-regrid:index`.
+introduced in Iris 3.2 <iris:ugrid>` is :doc:`esmf_regrid:index`.
 It aims to provide lazy regridding for structured regular and irregular grids,
```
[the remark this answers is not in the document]
schlunma: Are you referring to "our" definition of "irregular" (i.e., general curvilinear grids)? Yes, I think it does, since most likely some of the models [here](https://esmvaltool.dkrz.de/shared/esmvaltool/regrid_tests/recipe_tos_regrid_20240614_105846/) use that type of grid.

esmvalcore/preprocessor/_regrid.py
schlunma: Yes!

```py
>>> # cube is a UGRID-compliant cube
>>> ih.has_regular_grid(cube)
False
>>> ih.has_irregular_grid(cube)
False
>>> ih.has_unstructured_grid(cube)
True
```

Please note that in the UGRID word "irregular" and "unstructured" might mean the same, whereas for us "irregular" means "curvilinear"

doc/conf.py
```diff
@@ -438,20 +438,21 @@
...
 # Configuration for intersphinx
 intersphinx_mapping = {
-    'cf_units': ('https://cf-units.readthedocs.io/en/latest/', None),
+    'cf_units': ('https://cf-units.readthedocs.io/en/stable/', None),
     'cftime': ('https://unidata.github.io/cftime/', None),
     'esmvalcore':
     (f'https://docs.esmvaltool.org/projects/ESMValCore/en/{rtd_version}/',
      None),
     'esmvaltool': (f'https://docs.esmvaltool.org/en/{rtd_version}/', None),
+    "esmpy": ("https://earthsystemmodeling.org/esmpy_doc/release/latest/html/",
```
schlunma, line 447:
```suggestion
    'esmpy': ('https://earthsystemmodeling.org/esmpy_doc/release/latest/html/',
```

Just to be consistent with the other entries

esmvalcore/preprocessor/_regrid.py
```diff
@@ -39,10 +42,9 @@
     add_cell_measure,
 )
 from esmvalcore.preprocessor.regrid_schemes import (
-    ESMPyAreaWeighted,
-    ESMPyLinear,
-    ESMPyNearest,
     GenericFuncScheme,
+    ESMPyNearest,
+    IrisESMFRegrid,
```
schlunma, line 47: Run `isort`?

esmvalcore/preprocessor/_regrid.py
bouweandela: But for UGRID cubes we would like to use iris-esmf-regrid. What would be the best way to proceed? Add another function that checks if the source or target cube has a `iris.cube.Cube.mesh` property that is not `None` and that case use iris-esmf-regrid?

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def __init__(
+        self,
+        method: Literal["bilinear", "conservative", "nearest"],
+        mdtol: float | None = None,
+        use_src_mask: bool | np.ndarray = True,
+        use_tgt_mask: bool | np.ndarray = True,
+        src_resolution: int | None = None,
+        tgt_resolution: int | None = None,
+        tgt_location: Literal['face', 'node'] | None = None,
```
schlunma, line 102: Would it maybe make sense to pass all of these as `**kwargs`? It's nice to have them explicitly documented here, but if the regridders support new arguments in the future or change arguments we will need to adapt this code here to support this.

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
```diff
@@ -0,0 +1,208 @@
...
+            mask = da.ma.getmaskarray(data[slices])
+        else:
+            vertical_dim = cube.coord_dims(vertical_coord)[0]
+            slices = tuple(
+                slice(None) if i in horizontal_dims + (vertical_dim, ) else 0
+                for i in range(cube.ndim))
+            mask = da.ma.getmaskarray(data[slices])
+            mask_vertical_dim = sum(i < vertical_dim for i in horizontal_dims)
+            mask = mask.all(axis=mask_vertical_dim)
+        return mask
```
schlunma, line 168: Why do we need to explicitly calculate the mask here? From the [docs](https://iris-esmf-regrid.readthedocs.io/en/stable/_api_generated/esmf_regrid.schemes.html#module-esmf_regrid.schemes), it looks like `use_src_mask` and `use_tgt_mask` can only be booleans. So why don't we just use `True/False` here?

It's also dangerous to derive this only from the first time step, as the mask might be time-dependent (e.g. for satellite data).

doc/recipe/preprocessor.rst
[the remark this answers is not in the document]
bouweandela: Yes, I was referring to "our" definition of "irregular". I now understand that that overlaps with UGRID cubes, but our definition would also apply to cubes that do not follow the UGRID standard, so would not have a mesh, right? And then iris-esmf-regrid only supports the ones that have a `.mesh` attribute that is not `None`.

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
```diff
@@ -0,0 +1,208 @@
...
+        given to ESMF for calculation. If resolution is set, the source cube
+        must have strictly increasing bounds (bounds may be transposed
+        plus or minus 360 degrees to make the bounds strictly increasing).
+        Only available for method 'conservative'.
+    tgt_resolution:
+        If present, represents the amount of latitude slices per target cell
+        given to ESMF for calculation. If resolution is set, the target cube
+        must have strictly increasing bounds (bounds may be transposed
+        plus or minus 360 degrees to make the bounds strictly increasing).
+        Only available for method 'conservative'.
```
schlunma, line 83: I don't really understand this, but it looks like you just copied this from the regridder. In case you have a better understanding, could you rephrase it? For me, a "cell" contains one latitude, so multiple latitudes per cell sounds awkward...

tests/unit/preprocessor/_regrid/test_regrid.py
```diff
@@ -265,50 +267,126 @@ def make_test_cube(shape):
     return cube
 
 
-def test_rechunk_on_increased_grid():
-    """Test that an increase in grid size rechunks."""
-    with dask.config.set({'array.chunk-size': '128 M'}):
+def make_test_cube_unstructured(shape):
```
schlunma, line 270: I think this defines an irregular cube (using [our definitions](https://esmvaltool--2457.org.readthedocs.build/projects/ESMValCore/en/2457/recipe/preprocessor.html#grid-types)).

tests/unit/preprocessor/_regrid/test_regrid.py
```diff
@@ -265,50 +267,126 @@ def make_test_cube(shape):
...
+        iris.coords.AuxCoord(
+            np.broadcast_to(lon_points.reshape(1, -1), shape[-2:]),
+            standard_name='longitude',
+        ),
+        (-2, -1),
+    )
+    return cube
+
+
+def make_test_cube_irregular(shape):
```
schlunma, line 301: I think this defines an unstructured cube (using [our definitions](https://esmvaltool--2457.org.readthedocs.build/projects/ESMValCore/en/2457/recipe/preprocessor.html#grid-types)).

schlunma (changes requested):
Looks very good already Bouwe, it's great to see our custom ESMF code being replaced. Some additional comments:

1. In the PR description, please replace "unstructured" with "irregular" (you are replacing the schemes for irregular)
2. Would it make sense to deprecate our own `ESMFPy...` schemes? I don't think they are really useful anymore given that we have a much better solution for this now.
3. Would it maybe even make sense and go one step further by also replacing the schemes for regular regridding? This would make our code even easier. I think `esmf-regrid` also supports that kind of regridding with the same schemes. In principle, it would even work for unstructured grids, but in this case the data would need to be UGRID-compliant, and I am not sure we can enforce that (yet).

doc/recipe/preprocessor.rst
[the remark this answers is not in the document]
bouweandela: Oh wait, looks like I mixed up those definitions. In that case thuis text makes sense as it is.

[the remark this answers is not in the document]
schlunma: As far as I understand, our definition of "irregular" does NOT overlap with UGRID. UGRID only refers to unstructured data (1D lat and lon with common dimension), whereas our definition of irregular corresponds to structured curvilinear data (2D lat and lon with common dimensions).

> And then iris-esmf-regrid only supports the ones that have a .mesh attribute that is not None.

This is only true for unstructured data, not for (our) irregular data.

esmvalcore/preprocessor/_regrid.py
schlunma: Yeah, that sounds like a good suggestion to me!

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
bouweandela: It needs to be an array for data that contains a vertical dimension. This is allowed, as documented [here](https://iris-esmf-regrid.readthedocs.io/en/stable/_api_generated/esmf_regrid.schemes.html#esmf_regrid.schemes.ESMFAreaWeightedRegridder) and enforced by this code: https://github.com/SciTools-incubator/iris-esmf-regrid/blob/1772591d2ee1dbab06312e087ff694abe370cb72/esmf_regrid/schemes.py#L37-L76.

The mask seems mostly important to hide discontinuities in the grid. As long as the latitude and longitude coordinates are not dependent on time or vertical level, the mask computed here should be valid. It could mask out too much though, if fewer points are masked at a later timestep. However, computing the mask over all timesteps is computationally expensive as it requires loading the entire dataset from disk.

bouweandela: Are we currently supporting satellite datasets with time-varying masks?

tests/unit/preprocessor/_regrid/test_regrid.py
bouweandela: I mixed those two up

esmvalcore/preprocessor/_regrid.py
bouweandela: Done

doc/conf.py
bouweandela: Done

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
bouweandela: Yes, it's copy paste

tests/unit/preprocessor/_regrid/test_regrid.py
bouweandela: Fixed now

tests/unit/preprocessor/_regrid/test_regrid.py
bouweandela: Fixed now

bouweandela:
Thanks for reviewing @schlunma! 

> In the PR description, please replace "unstructured" with "irregular" (you are replacing the schemes for irregular)

I mixed those two up, it should be better now.

> Would it make sense to deprecate our own ESMFPy... schemes? I don't think they are really useful anymore given that we have a much better solution for this now.

Yes, but maybe with the next release of iris-esmf-regrid where the `nearest` scheme is also supported. @stephenworsley mentioned that he expects to sort out https://github.com/SciTools-incubator/iris-esmf-regrid/pull/276 and https://github.com/SciTools-incubator/iris-esmf-regrid/pull/369 and make a new release soon.

> Would it maybe even make sense and go one step further by also replacing the schemes for regular regridding? This would make our code even easier. I think esmf-regrid also supports that kind of regridding with the same schemes. In principle, it would even work for unstructured grids, but in this case the data would need to be UGRID-compliant, and I am not sure we can enforce that (yet).

I like this idea, but there is an issue with ESMF that prevents parallelisation using Dask (see #2316). I have reported it to the ESMF developers and they said they saw opportunities to solve it, but I'm not sure when this will be solved. I would be keen to improve our parallisation scheme for those cases where ESMF is not needed and then it is relevant to only use it when is really needed.

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
schlunma: > Are we currently supporting satellite datasets with time-varying masks?

Yes, we have some in our data pool. It's also quite common that a mask is height-dependent, i.e., for lower levels the regions with high altitude are masked.

I assume the mask you can pass to the functions can only be 2D? This is really not optimal...it would be much simpler if iris-esmf supported masks with the same shape as the input data.

schlunma:
Thanks for addressing my comments @bouweandela! Could you please add some documentation about the new schemes for meshes? It would also be great to use a common naming convention for this, e.g., at the moment, the documentation only talks about UGRID instead of iris meshes (in the [ICON doc](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/quickstart/find_data.html#icon)). Maybe you could move that part to regridding docs and only reference that in the ICON-specific section?

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
bouweandela: >  It's also quite common that a mask is height-dependent, i.e., for lower levels the regions with high altitude are masked.

The current code already takes care of that. I'll add something for the case where the mask is time varying.

> I assume the mask you can pass to the functions can only be 2D?

The mask is used to compute the regridding weights. Therefore the mask that is required describes the grid, so its dimensions need to match those of the grid.

bouweandela:
> Would it make sense to deprecate our own ESMFPy... schemes? I don't think they are really useful anymore given that we have a much better solution for this now.

Done in 9aaf1b2

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
bouweandela: I added the option to specify along which axes the mask varies using the `collapse_src_mask_along_axes` and `collapse_tgt_mask_along_axis` arguments, see [here](https://esmvaltool--2457.org.readthedocs.build/projects/ESMValCore/en/2457/api/esmvalcore.regridding_schemes.html#esmvalcore.preprocessor.regrid_schemes.IrisESMFRegrid) for the documentation. Does it look better now?

bouweandela:
@schlunma I've improved the documentation and testing and also added support for lazy nearest neighbour regridding now because this is at least partly supported with the new iris-esmf-regrid v0.11. Could you do another review, please?

esmvalcore/preprocessor/_regrid_esmpy.py
```diff
@@ -161,6 +161,11 @@ class ESMPyAreaWeighted(_ESMPyScheme):
 
     Does not support lazy regridding.
 
+    .. deprecated:: 2.12.0
```
schlunma, line 164: Would it make sense to also print a warning if these schemes are used in a recipe? E.g., add an `ESMValCoreDeprecationWarning` to the `_ESMPyRegridder` code?

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
```diff
@@ -0,0 +1,233 @@
...
+            coord = cube.coord(axis=axis, dim_coords=True)
+        except iris.exceptions.CoordinateNotFoundError:
+            try:
+                coord = cube.coord(axis=axis)
+            except iris.exceptions.CoordinateNotFoundError:
+                return tuple()
+        return cube.coord_dims(coord)
+
+    dims = {d for axis in axes for d in _get_dims_along_axis(cube, axis)}
+    return tuple(sorted(dims))
```
schlunma, line 42: I think this code could also added to the `preprocessor._shared` module. It's very general and could be useful for many preprocessors, not just regridding.

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
```diff
@@ -0,0 +1,233 @@
...
+    collapse_src_mask_along_axes:
+        When deriving the mask from the source cube data, collapse the mask
+        along the dimensions identified by these axes. Only points that are
+        masked at all time (``'T'``), vertical levels (``'Z'``), or both time
+        and vertical levels (``'TZ'``) will be considered masked.
+    collapse_tgt_mask_along_axes:
+        When deriving the mask from the target cube data, collapse the mask
+        along the dimensions identified by these axes. Only points that are
+        masked at all time (``'T'``), vertical levels (``'Z'``), or both time
+        and vertical levels (``'TZ'``) will be considered masked.
```
schlunma, line 95: If an axis exists, but is not mentioned here, the first slice along that axis will be used right? Would be good to mention that here.

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
```diff
@@ -0,0 +1,233 @@
...
+        which will tell :mod:`esmpy` which points to ignore. If an array is
+        provided, that will be used.
+        If set to :obj:`None`, it will be set to :obj:`True` for methods
+        ``'bilinear'`` and ``'conservative'`` and to :obj:`False` for method
+        ``'nearest'``. This default may be changed to :obj:`True` for all
+        schemes once `SciTools-incubator/iris-esmf-regrid#368
+        <https://github.com/SciTools-incubator/iris-esmf-regrid/issues/368>`_
+        has been resolved.
+    use_tgt_mask:
+        If True, derive a mask from (first time step) of the target cube,
```
schlunma, line 78:
```suggestion
        If True, derive a mask from of the target cube,
```
?

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
```diff
@@ -0,0 +1,233 @@
...
+        Only available for method 'conservative'.
+    tgt_resolution:
+        If present, represents the amount of latitude slices per target cell
+        given to ESMF for calculation. If resolution is set, the target cube
+        must have strictly increasing bounds (bounds may be transposed
+        plus or minus 360 degrees to make the bounds strictly increasing).
+        Only available for method 'conservative'.
+    tgt_location:
+        Either "face" or "node". Describes the location for data on the mesh
+        if the target is not a :class:`~iris.cube.Cube`.
```
schlunma, line 110: What happens if `tgt_location` is `None` for meshes?

schlunma (commented):
Great! I just tested this extensively with all different grid types, and everything works as excepted 🚀 

Just have a couple of minor comments on the doc. Cheers!

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
bouweandela: Yes. Do you think that is a safe option, also for ocean data? Or should we also allow e.g. coordinate names in addition to axes `'T'` and `'Z'`?

schlunma: It makes certainly sense in some use cases, as long as it's documented properly. And yes, allowing coordinate names in addition to axes sounds great!

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
bouweandela: Done in 42123c0

esmvalcore/preprocessor/_regrid_esmpy.py
bouweandela: Done in 286f43d

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
bouweandela: It looks like an exception will be raised: https://github.com/SciTools-incubator/iris-esmf-regrid/blob/9e2559af148b0e5ac057fcfeacac3a7ec3e18739/esmf_regrid/schemes.py#L498-L523

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
bouweandela: Done in cfc4330

bouweandela:
Thanks for reviewing @schlunma. I think I have addressed all comments now, could you have another look please?

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
schlunma: Ah, I was briefly confused why I didn't get an error in my tests when using the default scheme (which uses `tgt_location=None`), but since I used [unstructured->regular regridding](https://github.com/SciTools-incubator/iris-esmf-regrid/blob/9e2559af148b0e5ac057fcfeacac3a7ec3e18739/esmf_regrid/schemes.py#L1355), this keyword is ignored. I think it would be good to mention that this is required if target is a mesh, but ignored otherwise.

bouweandela: Improved in a848468

esmvalcore/preprocessor/_regrid_iris_esmf_regrid.py
```diff
@@ -0,0 +1,234 @@
...
+        plus or minus 360 degrees to make the bounds strictly increasing).
+        Only available for method 'conservative'.
+    tgt_resolution:
+        If present, represents the amount of latitude slices per target cell
+        given to ESMF for calculation. If resolution is set, the target cube
+        must have strictly increasing bounds (bounds may be transposed
+        plus or minus 360 degrees to make the bounds strictly increasing).
+        Only available for method 'conservative'.
+    tgt_location:
+        Only used if the target grid is an :class:`iris.mesh.MeshXY`. Describes
```
schlunma, line 102:
```suggestion
        Only used if the target grid is an :class:`iris.mesh.MeshXY` or an :class:`iris.cube.Cube` with an associated mesh. Describes
```

bouweandela: It is not used if the target is a cube (on a mesh), see the original docstring: https://iris-esmf-regrid.readthedocs.io/en/latest/_api_generated/esmf_regrid.schemes.html#esmf_regrid.schemes.ESMFBilinear

In that case, the `location` attribute of the cube is used: https://github.com/SciTools-incubator/iris-esmf-regrid/blob/9e2559af148b0e5ac057fcfeacac3a7ec3e18739/esmf_regrid/schemes.py#L504

schlunma: Good point, haven't checked the code in that detail 😄

schlunma (approved):
Awesome, thanks @bouweandela! This will facilitate regridding enormously 🚀

PR #2456, opened 2024-06-11
Disable collecting test coverage by default

bouweandela:
## Description

Collecting code coverage is time consuming. Disable it by default and enable it on request by running `pytest --cov`. This will save some time when developing locally and also reduces the runtime of the GitHub Actions tests by about 1 (Linux) to 2 (OSX) minutes.

Link to documentation: https://esmvaltool--2456.org.readthedocs.build/projects/ESMValCore/en/2456/contributing.html#test-coverage

***

## [Before you get started](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#getting-started)

- [ ] [☝ Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do

## [Checklist](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#checklist-for-pull-requests)

It is the responsibility of the author to make sure the pull request is ready to review. The icons indicate whether the item will be subject to the [🛠 Technical][1] or [🧪 Scientific][2] review.

[1]: https://docs.esmvaltool.org/en/latest/community/review.html#technical-review
[2]: https://docs.esmvaltool.org/en/latest/community/review.html#scientific-review

- [x] [🧪][2] The new functionality is [relevant and scientifically sound](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#scientific-relevance)
- [x] [🛠][1] This pull request has a [descriptive title and labels](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-title-and-label)
- [x] [🛠][1] Code is written according to the [code quality guidelines](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#code-quality)
- [x] [🧪][2] and [🛠][1] [Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available
- [x] [🛠][1] Changes are [backward compatible](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#backward-compatibility)
- [x] [🛠][1] The [list of authors](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#list-of-authors) is up to date
- [x] [🛠][1] All [checks below this pull request](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-checks) were successful

***

To help with the number pull requests:

[...]

PR #1819, opened 2022-11-21
No more sequential tests since SegFaults were not noticed anymore (knock on wood)

valeriupredoi:
## Description

Let's see if indeed SegFaults that we used to see are now gone (bar, of course, the ones - many of them - from netCDF>1.6.0 + iris). Don't merge this just yet (hence draft, I'd like to run a few of these test sets).

***

## [Checklist](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#checklist-for-pull-requests)

It is the responsibility of the author to make sure the pull request is ready to review. The icons indicate whether the item will be subject to the [🛠 Technical][1] or [🧪 Scientific][2] review.

[1]: https://docs.esmvaltool.org/en/latest/community/review.html#technical-review
[2]: https://docs.esmvaltool.org/en/latest/community/review.html#scientific-review

- [x] [🛠][1] This pull request has a [descriptive title and labels](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-title-and-label)
- [x] [🛠][1] All [checks below this pull request](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-checks) were successful

***

To help with the number pull requests:

- 🙏 We kindly ask you to [review](https://docs.esmvaltool.org/en/latest/community/review.html#review-of-pull-requests) two other [open pull requests](https://github.com/ESMValGroup/ESMValCore/pulls) in this repository

PR #2540, opened 2024-10-03
Use pyproject.toml instead of setup.py/setup.cfg

bouweandela:
## Description

Use pyproject.toml instead of setup.py/setup.cfg, as this is the modern way of doing things.

Also added some improvements suggested by https://learn.scientific-python.org/development/guides/repo-review/?repo=esmvalgroup%2Fesmvalcore&branch=pyproject-toml

Link to documentation: https://esmvaltool--2540.org.readthedocs.build/projects/ESMValCore/en/2540/contributing.html

Related to https://github.com/ESMValGroup/ESMValTool/issues/3584

***

- [x] [🛠][1] All [checks below this pull request](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-checks) were successful - except that a line that has been reformatted by ruff (probably because we now specify the required python version) is not covered by tests

***


setup.cfg
```diff
@@ -1,38 +1,6 @@
...
-
-[coverage:run]
-parallel = true
-[coverage:report]
-exclude_lines =
-    pragma: no cover
-    if __name__ == .__main__.:
-    if TYPE_CHECKING:
-
 [pycodestyle]
```
valeriupredoi, line 1: love the outcome of this issue, that was proposing support for `pyproject.toml` :rofl: https://github.com/PyCQA/pycodestyle/issues/813

valeriupredoi (commented):
absolutely YES for this :beer:

setup.cfg
bouweandela: We should be able to drop pycodestyle as soon as we're using ruff for linting too.

.circleci/install_triggers
```diff
@@ -1,5 +1,3 @@
 ^\.circleci/
 ^environment\.yml$
 ^pyproject.toml$
-^setup\.py$
-^setup\.cfg$
```
schlunma, line 5: `setup.cfg` still exists, did you remove that here on purpose?

tests/integration/cmor/_fixes/cmip6/test_cesm2.py
```diff
@@ -61,7 +61,6 @@ def test_get_cl_fix():
 )
 
 
-@pytest.mark.sequential
 @pytest.mark.skipif(
     sys.version_info < (3, 7, 6), reason="requires python3.7.6 or newer"
 )
```
schlunma, line 66: Why are you dropping the `sequential` mark? Also, would it make sense to remove the following `skipif` since we're not supporting Python 3.7 anymore?

schlunma (commented):
Great, thanks Bouwe, I really like this! 🚀 

There's one more suggestion at https://learn.scientific-python.org/development/guides/repo-review/?repo=esmvalgroup%2Fesmvalcore&branch=pyproject-toml about pyproject:

```
[PP309](https://learn.scientific-python.org/development/guides/pytest#PP309): Filter warnings specified

filterwarnings must be set (probably to at least ["error"]). Python will hide important warnings otherwise, like deprecations.

[tool.pytest.ini_options]
filterwarnings = ["error"]
```

Would it make sense to implement this as well?

bouweandela:
> Would it make sense to implement this as well?

I tried, but it causes the tests to fail because some of our dependencies are using deprecated features. It would be good to enable that though, but it will require extra work.

tests/integration/cmor/_fixes/cmip6/test_cesm2.py
bouweandela: It was removed in #1819, but these entries were forgotten. Using
https://github.com/ESMValGroup/ESMValCore/blob/edb1291944ded032fa2608a894c60ce70ca49304/pyproject.toml#L151
makes using undefined markers an error.

.circleci/install_triggers
bouweandela: Yes, it now only contains the configuration for pycodestyle, so it does not affect the (development) installation of the tool anymore.

schlunma: Makes sense. Would it make sense to remove all unused element in `setup.cfg` here? It looks like it contains more than the pycodestyle config:

https://github.com/ESMValGroup/ESMValCore/blob/a3557ecdde7d1b9e857610779105fe75049f7e04/setup.cfg#L1-L38

bouweandela: Looks like I messed up the last merge with `main`. Should be fixed now.

tests/integration/cmor/_fixes/cmip6/test_cesm2.py
bouweandela: Removed in e70ffd2

bouweandela:
>>    Would it make sense to implement this as well?

> I tried, but it causes the tests to fail because some of our dependencies are using deprecated features. It would be good to enable that though, but it will require extra work.

Created #2555 as a reminder.

.circleci/install_triggers
bouweandela: And then I forgot to move the changes from #2456 to pyproject.toml. Should be really fixed now :facepalm:

schlunma (approved):
Great, thanks Bouwe!!

esmvalcore/_task.py
```diff
@@ -609,9 +609,10 @@ def _run(self, input_files):
 
         returncode = None
 
-        with resource_usage_logger(process.pid, self.resource_log), open(
-            self.log, "ab"
-        ) as log:
+        with (
+            resource_usage_logger(process.pid, self.resource_log),
+            open(self.log, "ab") as log,
```
valeriupredoi, line 614: opening file as bytes object, but then adding to it? Not a negative remark, am just wondering why not as plaintext with encoding

pyproject.toml
```diff
@@ -1,7 +1,121 @@
 [build-system]
-requires = ["setuptools >= 40.6.0", "wheel", "setuptools_scm>=6.2"]
+requires = [
+    "setuptools >= 40.6.0",
+    "setuptools_scm>=6.2",
+]
```
valeriupredoi, line 5: just as a minor remark - why did we need wheel when we never built wheels for esmvalcore?

pyproject.toml
```diff
@@ -1,7 +1,121 @@
...
+description = "A community tool for pre-processing data from Earth system models in CMIP and running analysis scripts"
+dependencies = [
+    "cartopy",
+    "cf-units",
+    "dask[array,distributed]!=2024.8.0",  # ESMValCore/issues/2503
+    "dask-jobqueue",
+    "esgf-pyclient>=0.3.1",
+    "esmf-regrid>=0.11.0",
+    "esmpy!=8.1.0",  # not on PyPI
```
valeriupredoi, line 42: shall we remove this then?

valeriupredoi: at least the pin, I mean, not the entire dep

valeriupredoi (approved):
very cool! A few remoarks from me only, but I *think* we now need to add `importlib_metadata` to our deps if we use it directly

esmvalcore/_task.py
bouweandela: That saves decoding the text as it comes out of the process that runs the diagnostic script and then encoding it again to write it to file.

pyproject.toml
bouweandela: We do build wheels, check PyPI: https://pypi.org/project/ESMValCore/#files

However, it does not need to be listed as a dependency according to Repo-Review/[Scientific Python Library Development Guide](https://learn.scientific-python.org/development/guides/packaging-classic/#pep-517518-support-high-priority):

> Note that "wheel" is never required; it is injected automatically by setuptools only when needed.

pyproject.toml
bouweandela: Done in 953be1c

bouweandela:
> we now need to add `importlib_metadata` to our deps if we use it directly

Can you point me to a line of code where it is used? I couldn't find any.

pyproject.toml
bouweandela: https://github.com/ESMValGroup/ESMValCore/blob/12798741cd4422adc1b7a6ecbc987dbc6e488388/.github/workflows/build-and-deploy-on-pypi.yml#L35-L41

valeriupredoi: Thanks! Funny that we build wheels even if we don't require compilation, but maybe what I'm callin a "wheel" is the wrong thing here :wheel: All good!

pyproject.toml
valeriupredoi: Bedankt :beer:

valeriupredoi:
> > we now need to add `importlib_metadata` to our deps if we use it directly
> 
> Can you point me to a line of code where it is used? I couldn't find any.

nevermind, it was `from importlib.metadata` - dot is essential, V :grin:

bouweandela:
No thanks, I'm on it.

valeriupredoi:
superb @bouweandela :partying_face:

PR #2546, opened 2024-10-08
reformat datetime strings be in line with new `isodate==0.7.0` and actual ISO8601 and pin `isodate>=0.7.0`

### LINKED ISSUE #2545: Tests involving `isodate` failed
```bash
=========================== short test summary info ============================
FAILED tests/integration/recipe/test_check.py::test_valid_time_selection[19900101T12H00M00S/19920101T12H00M00] - esmvalcore.exceptions.RecipeError: Invalid value encountered for `timerange`. Valid value must follow ISO 8601 standard for dates and duration periods, or be set to '*' to load available years. Got ['19900101T12H00M00S', '19920101T12H00M00'] instead.
FAILED tests/integration/recipe/test_check.py::test_valid_time_selection[19900101TH00M00S/P2Y2M1DT12H00M00S] - esmvalcore.exceptions.RecipeError: Invalid value encountered for `timerange`. Valid value must follow ISO 8601 standard for dates and duration periods, or be set to '*' to load available years. Got ['19900101TH00M00S', 'P2Y2M1DT12H00M00S'] instead.
FAILED tests/integration/recipe/test_check.py::test_valid_time_selection[P2Y2M1D/19920101T12H00M00S] - esmvalcore.exceptions.RecipeError: Invalid value encountered for `timerange`. Valid value must follow ISO 8601 standard for dates and duration periods, or be set to '*' to load available years. Got ['P2Y2M1D', '19920101T12H00M00S'] instead.
FAILED tests/integration/recipe/test_recipe.py::test_recipe_iso_timerange[19900101TH00M00S/P2Y2M1DT12H00M00S-19900101TH00M00S-P2Y2M1DT12H00M00S] - isodate.isoerror.ISO8601Error: Unrecognised ISO 8601 date format: '19900101TH00M00S'
FAILED tests/integration/recipe/test_recipe.py::test_recipe_iso_timerange[P1Y2M1D/19920101T12H00M00S-P1Y2M1D-19920101T12H00M00S] - isodate.isoerror.ISO8601Error: Unrecognised ISO 8601 date format: '19920101T12H00M00S'
FAILED tests/integration/recipe/test_recipe.py::test_recipe_iso_timerange_as_dataset[19900101TH00M00S/P2Y2M1DT12H00M00S-19900101TH00M00S-P2Y2M1DT12H00M00S] - isodate.isoerror.ISO8601Error: Unrecognised ISO 8601 date format: '19900101TH00M00S'
FAILED tests/integration/recipe/test_recipe.py::test_recipe_iso_timerange_as_dataset[P1Y2M1D/19920101T12H00M00S-P1Y2M1D-19920101T12H00M00S] - isodate.isoerror.ISO8601Error: Unrecognised ISO 8601 date format: '19920101T12H00M00S'
= 7 failed, 3877 passed, 13 skipped, 1 xpassed, 1081 warnings in 112.33s (0:01:52) =
```

See https://app.circleci.com/pipelines/github/ESMValGroup/ESMValCore/11808/workflows/9fcf7ed9-5d01-4686-a504-6aca0941b2aa/jobs/49698.

valeriupredoi:
## Description

Closes #2545 

As of the new `isodate=0.7.0` released 11 hours ago, the package is a bit stricter what is actually iingesting as ISO8601 - in other words, @bouweandela was correct (like always :grin: )

OK so @schlunma raised a few good points about how we handle these `isodate` strings, in his comments below; as a response, here are some key elements belonging to ISO8601 from  https://www.digi.com/resources/documentation/digidocs/90001488-13/reference/r_iso_8601_duration_format.htm

- duration: always starts with P (so we need to check for that before calling `isodate.prase_duration()`)
- datetime from https://www.digi.com/resources/documentation/digidocs/90001488-13/reference/r_iso_8601_date_format.htm - T is the character denotic the start of the time bit of datetime, so if T in string then one needs to call `isodate.parse_datetime()`, otherwise if no T, then one needs to call `isodate.parse_date()` because those two functions barf if they get mixed inputs

***

***


valeriupredoi:
BRB

esmvalcore/_recipe/check.py
```diff
@@ -392,7 +392,10 @@ def _check_format_years(date):
...
 def _check_timerange_values(date, timerange):
     try:
-        isodate.parse_date(date)
+        if "T" in date:
+            isodate.parse_datetime(date)
+        else:
+            isodate.parse_date(date)
     except ValueError:
         try:
             isodate.parse_duration(date)
```
schlunma, line 401: In other parts of the code we use the following syntax:

https://github.com/ESMValGroup/ESMValCore/blob/99c6a7be1e995f8b4fcb61f3a44bd46aa828c397/esmvalcore/preprocessor/_time.py#L121-L130

I am honestly not sure which one is better. Will all datetimes contain a "T"? If not, then I would go for this other option.

.github/workflows/run-tests.yml
```diff
@@ -21,6 +21,7 @@ on:
   push:
     branches:
       - main
+      - fix_isodate_tests
```
schlunma, line 24:
```suggestion
```

Just so we don't forget that

schlunma (commented):
Great, thanks for tackling this V! One comment on the code, and one general one: is this backwards-compatible? I.e., will this still work with an older version of `isodate`? I guess yes, but would be nice to ensure that.

esmvalcore/_recipe/check.py
valeriupredoi: that's a very good call, Manu! I *think* we should change it there too with a check on T and call the correct isodate function. `parse_date()` will fail if there is a T in it, and `parse_datetime()` will fail if there is no T in it (ah Mr T :grin: ) - or, we can let it fail so we prompt the user to change their date(time) strings in the recipe, though this would be very annoying if those are actually correct - what do you think, you and @bouweandela ?

.github/workflows/run-tests.yml
valeriupredoi: lemme plug that in once we're done-done with this

valeriupredoi:
@schlunma cheers, Manu! Yes, it is backwards compatible bc the Circle tests still use the older `isodate=0.6.1`; if we want to, we can just pin `isodate>=0.7.0` in our env - the previous version is ancient anyhow; more pressing is you comment about the other bits where `isodate.parse_ ...` is invoked (well spotted!) - I reckon we ought to check, but am not fully sold on it

esmvalcore/_recipe/check.py
valeriupredoi: > Will all datetimes contain a "T"? 

Sorry, forgot to answer: no, some are just eg `19990605` ie just YYYYMMDD

valeriupredoi: damn am verbose today! Note that `except isodate.isoerror.ISO8601Error:` is actually a bit too general and may fail now if the string is eg `YYYYMMDDTxxHxxMxxS`

schlunma: Yeah, agreed, just like python's `datetime.datetime` also handles date with and without time. But if the existence of "T" is enough to distinguish the two, we should implement your solution. And maybe include a test for "P" to check if a duration is given?

schlunma: > damn am verbose today! Note that `except isodate.isoerror.ISO8601Error:` is actually a bit too general and may fail now if the string is eg `YYYYMMDDTxxHxxMxxS`

But that's not a valid format, is it? See https://github.com/ESMValGroup/ESMValCore/issues/2545. So I guess removing the try/except and just checking for the existence of "P" and "T" in the string is the best solution?

valeriupredoi: yeah that's exactly what I think too! Lemme massage that then :beer:

valeriupredoi: @schlunma @bouweandela have a look at cc94a26f0c02d03a74e42d78b9692c3ec1c020ff and tell me if you think it's a good framework - I'll make similar changes to `_time.py` and `local.py` (whatever that is) :beer:

schlunma: I like it! Maybe we could even put that if/elif/else construct in a little function and use that everywhere? 😁

valeriupredoi: sorry, kept editing the commit hash coz I keep forgetting I gotta Ruff it :grin: Well, we could, but that might be overkill coz we have to allow for one or two checked strings etc

valeriupredoi: ah crap! Both P and T could be in the same string and that's still duration, need to refactor. Off home though, in the morrow :pizza:

valeriupredoi: ah done that, now am off to :pizza:

esmvalcore/_recipe/check.py
```diff
@@ -391,20 +400,25 @@ def _check_format_years(date):
...
-                raise RecipeError(
-                    "Invalid value encountered for `timerange`. "
-                    "Valid value must follow ISO 8601 standard "
-                    "for dates and duration periods, or be "
-                    "set to '*' to load available years. "
-                    f"Got {timerange} instead."
-                ) from exc
+        else:
+            if "T" in date:
+                isodate.parse_datetime(date)
```
schlunma, line 410:
```py
if "P" in data:
    isodate.parse_duration(date)
elif "T" in date:
    isodate.parse_datetime(date)
else:
    isodate.parse_date(date)
```

should also work

valeriupredoi: ace, Manu! Plugged in in 7f631a14350cd94004cbf91d6c989f1bad09ff17

valeriupredoi:
OK I've now pretty much refactored all the handling of `isodate` strings (apart from the ones in `local.py` that are correctly handled) , added a couple tests, and pinned to `isodate>=0.7.0` - the old package, apart from being very old, has some funny ways of handling parsing exceptions, and those better in the new 0.7.0, hence some of our negative tests are not compatible anymore with the old one (in essence the same fail, but contorted exceptions for 0.6.1). May goal was no not catch any iso errors and filter them through our stack, but rather, let them immediately be raised from `isodate` - lot more straightforward that way :beer:

valeriupredoi:
the [upstream dev deps test](https://app.circleci.com/pipelines/github/ESMValGroup/ESMValCore/11827/workflows/c98ca245-eb85-415a-b672-6897bf4efabf/jobs/49778?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary) fails bc of esgf-pyclient and that means I have to go there and fix whatever broke there too - not in a rush though, we can get this in no problem in the meantime, just another bit for me to look into :confounded:

bouweandela:
It may have been a network issue, re-running tests seems to have solved it.

valeriupredoi:
cheers @bouweandela :beers: You here to review as well, or just just passing by :grin:

bouweandela:
Just passing by, it looks like @schlunma is doing an excellent job reviewing.

esmvalcore/_recipe/check.py
```diff
@@ -391,20 +400,24 @@ def _check_format_years(date):
 
 
 def _check_timerange_values(date, timerange):
+    # P must always be in a duration string
+    # if T in date, that is a datetime; otherwise it's date
```
schlunma, line 404:
```suggestion
    # Wildcards are fine
    if date == '*':
        return
    
    # P must always be in a duration string
    # if T in date, that is a datetime; otherwise it's date
```

feels more natural

esmvalcore/_recipe/check.py
```diff
@@ -391,20 +400,24 @@ def _check_format_years(date):
...
+            isodate.parse_date(date)
+    except isodate.isoerror.ISO8601Error as exc:
+        if date != "*":
+            raise RecipeError(
+                str(exc) + "\n" + "Invalid value encountered for `timerange`. "
+                "Valid value must follow ISO 8601 standard "
+                "for dates and duration periods, or be "
+                "set to '*' to load available years. "
+                f"Got {timerange} instead."
+            )
```
schlunma, line 420:
```suggestion
        raise RecipeError(
            str(exc) + "\n" + "Invalid value encountered for `timerange`. "
            "Valid value must follow ISO 8601 standard "
            "for dates and duration periods, or be "
            "set to '*' to load available years. "
            f"Got {timerange} instead."
        )
```

tests/unit/preprocessor/_time/test_time.py
```diff
@@ -486,6 +487,24 @@ def test_clip_timerange_single_year_4d(self):
                 coord_name
             )
 
+    def test_clip_timerange_start_date_invalid_isodate(self):
+        cube = self._create_cube(
+            [[[[0.0, 1.0]]]], [150.0], [[0.0, 365.0]], "standard"
+        )
+        with pytest.raises(isodate.isoerror.ISO8601Error) as exc:
```
schlunma, line 494: `pytest.raises` has a `match` keyword to check the exception string (see https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions), but feel free to ignore that comment

schlunma (commented):
Thanks V, looks great! Just have some minor comments (mostly suggestions, feel free to ignore). Cheers!

esmvalcore/_recipe/check.py
```diff
@@ -366,20 +366,29 @@ def _check_delimiter(timerange):
...
+        )
+
+    if timerange[0].startswith("P"):
+        try:
+            isodate.parse_duration(timerange[0])
+        except isodate.isoerror.ISO8601Error as exc:
+            raise RecipeError(
+                str(exc) + "\n" + "Invalid value encountered for `timerange`. "
+                f"{timerange[0]} is not valid duration according to ISO 8601."
+            )
```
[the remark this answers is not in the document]
valeriupredoi: that's a very good point - I usually do that myself, but I've noticed that the `isodate` exceptions are a bit cryptic, so I didn't want the user to be left with that as the last communication in the Traceback; if you think we sould really switch the order, then I'll do it, but otherwise I'd leave our (more friendly) exception as the last one :beer:

tests/unit/preprocessor/_time/test_time.py
valeriupredoi: ah I know the match= kwatg, cheers! meh, will leave it as is, if you don't mind, that'd be fine tuning a Trabant given how poorly those tests are (you see they using UnitTest classes :man_facepalming: )

esmvalcore/_recipe/check.py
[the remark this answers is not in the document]
schlunma: At least the one from the tests seems to be clearer if you switch the order:

Now:

```
"ISO 8601 time designator 'T' missing. Unable to parse datetime string '11X'
Invalid value encountered for `timerange`. P11X is not valid duration according to ISO 8601."
```

Switched order:

```
"Invalid value encountered for `timerange`. P11X is not valid duration according to ISO 8601:
ISO 8601 time designator 'T' missing. Unable to parse datetime string '11X'"
```

So I think switching makes sense 😅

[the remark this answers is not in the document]
valeriupredoi: Sounds good, buds! Lemme plug it in right after my neeting am in now 👍🍺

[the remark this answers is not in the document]
valeriupredoi: OK buds done them all now - isodate exc at the end (so we have continuity) in 5670abd57c924491f2432e2b7e501be84a3e42f6 and c8f59e4c291258a4cfb096dc71bff051a08a6235

schlunma (approved):
Thanks V!!

valeriupredoi:
Danke viel Manu :beer:

PR #2551, opened 2024-10-11
Allow grouping input cubes by date (instead of filename) for `fix_metadata`

### LINKED ISSUE #1806: Variable derivation for ERA5 on-the-fly CMORizer
I ran into the problem of not being able to figure out how to derive variables (using more than 1 input variable) with the ERA5 on-the-fly CMORizer. In this concrete case, I would like to calculate the CMOR variable rsut (TOA Outgoing Shortwave Radiation), which is not readily available from ERA5. rsut needs to be calculated from the available ERA5 variables as

rsut = era5_toa_incident_solar_radiation - era5_top_net_solar_radiation

The problem is (as it seems) that it is not possible to have 2 different variables at the time the functions from cmor/_fixes/native6/era5.py are called.

I guess one possible solution could be to have all variables needed for the variable derivation in the ERA5 "raw" files and then follow a similar approach as @schlunma implemented for the on-the-fly CMORizer for EMAC. But I am not sure this is supported by era5cli we use to download ERA5 data and would probably make obtaining the ERA5 data more complicated.

@ESMValGroup/esmvaltool-coreteam would you have an idea how to address the variable derivation problem for ERA5 data?

schlunma:
## Description

This PR allows grouping the input cubes for our `fix_metadata` functions by date, i.e., all files with the same data range are passed to the fix simultaneously). This can be enabled by setting the class variable `GROUP_CUBES_BY_DATE = True` in the corresponding fix class. This allows implementing fixes where variables from multiple input files are necessary (for example, to derive rsut for ERA5).

This solution only works for projects where the input files are located in the same directory, and the input file pattern is flexible enough to find all files. This is fine for the native ERA5 data in netCDF format (that we need to manually download and put into the corresponding directories). For other projects where files are stored in different directories, further changes are necessary (potentially in `local.py`). However, this PR is a **prerequisite** to make these other cases work.

By default, input cubes are grouped by filename for `fix_metadata` (i.e., each `fix_metadata` call operates only on a single file):

https://github.com/ESMValGroup/ESMValCore/blob/fd82b43e495cfe3a15650c369475feea0e9c54f8/esmvalcore/cmor/fix.py#L197-L204

Note that this is fully backwards-compatible since the new functionality needs to be explicitly enabled.

Closes #1806

Link to documentation: TBA

***

***


schlunma:
Here's a small recipe to test this for rsut (I also included rsdt to ensure that this still works fine):

```yml
# ESMValTool
---
documentation:
  title: test
  description: test
  authors:
    - schlund_manuel

datasets:
  - {project: native6, dataset: ERA5, type: reanaly, version: v1, tier: 3, timerange: 2000/2001}

diagnostics:

  test:
    variables:
      rsut:
        mip: Amon
      rsdt:
        mip: Amon
    scripts:
      null
```

Input files need to be arranged like this:

```
.
└── Tier3
    └── ERA5
        └── v1
            └── mon
                ├── rsdt
                │   ├── era5_toa_incident_solar_radiation_2000_monthly.nc
                │   └── era5_toa_incident_solar_radiation_2001_monthly.nc
                └── rsut
                    ├── era5_mean_top_net_short_wave_radiation_flux_2000_monthly.nc
                    ├── era5_mean_top_net_short_wave_radiation_flux_2001_monthly.nc
                    ├── era5_toa_incident_solar_radiation_2000_monthly.nc
                    └── era5_toa_incident_solar_radiation_2001_monthly.nc
```

@bouweandela do you think this approach is a reasonable solution to this problem? As mentioned in the description, it doesn't solve the problem for all problems, but a different grouping will be necessary for all cases. And it is fully sufficient for the ERA5 netCDF case.

bouweandela:
> do you think this approach is a reasonable solution to this problem? 

If it works for you, it should be fine

Issue #674, opened 2020-06-12
Make preprocessor lazy


bouweandela:
Overview issue with laziness status of preprocessor functions:

- [ ] fix_file - copies files on disk
- [x] load
- [ ] derive - lazy for most variables #42, but all data is saved to disk #377
- [ ] fix_metadata - depends on model/variable
- [x] concatenate
- [x] cmor_check_metadata
- [x] extract_time
- [x] extract_season
- [x] extract_month
- [ ] fix_data - ?
- [ ] extract_levels - not if interpolation is needed #35
- [x] weighting_landsea_fraction
- [ ] mask_landsea - no
- [ ] mask_glaciated - no
- [ ] mask_landseaice - no
- [ ] regrid - no https://github.com/SciTools/iris/issues/3700
- [ ] extract_point - ?
- [ ] mask_fillvalues - no
- [ ] mask_above_threshold - no
- [ ] mask_below_threshold - no
- [ ] mask_inside_range - no
- [ ] mask_outside_range - no
- [x] clip
- [x] extract_region
- [x] extract_shape
- [x] extract_volume
- [ ] extract_trajectory - ?
- [x] extract_transect
- [x] detrend - no non-lazy version available
- [ ] multi_model_statistics - no
- [x] extract_named_regions
- [ ] depth_integration - no
- [ ] area_statistics - ? #47
- [ ] volume_statistics - no #647
- [x] amplitude
- [x] zonal_statistics
- [x] meridional_statistics
- [x] daily_statistics
- [x] monthly_statistics
- [x] seasonal_statistics
- [x] annual_statistics
- [x] decadal_statistics
- [ ] climate_statistics - not if period='full' and operator='mean' or 'sum' #336
- [ ] anomalies - not if period='full' due to use of `climate_statistics`, see above
- [x] regrid_time
- [ ] timeseries_filter - ?
- [x] cmor_check_data
- [x] convert_units

It would be great if we could make more preprocessor functions lazy. The laziness status should also be indicated in the docstrings.

valeriupredoi:
good call!! I reckon this is a good first serious feature for v2.1 :beer:

bouweandela:
To get an idea of the priority of the preprocessor functions, here is a rough count of which number of recipes in the ESMValTool that they are used in:
```
regrid 60
extract_region 30
derive 26
extract_levels 24
mask_landsea 21
area_statistics 20
climate_statistics 18
multi_model_statistics 16
[...]

PR #2579, opened 2024-11-12
Allow `fix_file` to return dataset objects

schlunma:
## Description

Currently, if a file cannot be properly read with Iris, we use `fix_file` to create a copy of that file and modify it using `netCDF4.Dataset` ([example](https://github.com/ESMValGroup/ESMValCore/blob/c792a523569ea1ccc3fb49cdef84b0d6ccbbecc3/esmvalcore/cmor/_fixes/cmip6/cesm2_waccm.py#L15-L49)). This is very inefficient and slow.

A much better way to deal with this is to read the file with [ncdata](https://ncdata.readthedocs.io/en/latest/) or [xarray](https://docs.xarray.dev/en/stable/index.html) and then use ncdata to convert that object to an Iris object. However, for this to work, we need to allow `fix_file` to return dataset objects (instead of paths) and `load` to read dataset objects (instead of paths). This PR does that.

Closes https://github.com/ESMValGroup/ESMValCore/issues/2129
Related to https://github.com/ESMValGroup/ESMValCore/issues/674

Link to documentation:
- https://esmvaltool--2579.org.readthedocs.build/projects/ESMValCore/en/2579/develop/fixing_data.html#create-the-fix
- https://esmvaltool--2579.org.readthedocs.build/projects/ESMValCore/en/2579/quickstart/configure.html#filter-preprocessor-warnings
- https://esmvaltool--2579.org.readthedocs.build/projects/ESMValCore/en/2579/recipe/preprocessor.html#dataset-specific-fixes
- https://esmvaltool--2579.org.readthedocs.build/projects/ESMValCore/en/2579/api/esmvalcore.preprocessor.html#esmvalcore.preprocessor.fix_file

***

- [x] [🛠][1] ~All [checks below this pull request](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-checks) were successful~ Remaining Codacy/Codecov issues cannot be fixed

***


schlunma:
One more incentive why it would be great to have this feature:

```py
path = "/path/to/real_EMAC_file_with_855_variables.nc"

%%timeit
iris.load(path)
# 1min 47s ± 446 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
xr.open_dataset(path, chunks="auto")
# 277 ms ± 8.14 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
netCDF4.Dataset(path, mode="r")
# 5.2 ms ± 23.7 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

This uses the most up-to-date versions of all these packages (fresh install from 2025-05-09).

Since we sometimes need to read hundreds of those files (this example only contains a single time step), ESMValTool can spend hours just loading these datasets.

schlunma:
The missing tests here reported by codecov are in `esmvalcore/cmor/_fixes/ipslcm/ipsl_cm6.py`. In this file, I just renamed the argument `filepath` to `file`. Since adding tests for this is not trivial, I will skip that here.

tests/integration/preprocessor/_io/test_load.py
```diff
@@ -21,153 +22,175 @@ def _create_sample_cube():
...
+        category=UserWarning,
+        stacklevel=2,
+    )
+    return CubeList([Cube(0)])
+
+
+@mock.patch("iris.load_raw", autospec=True)
+def test_do_not_ignore_warnings(mock_load_raw):
+    """Test do not ignore specific warnings."""
+    mock_load_raw.side_effect = load_with_warning
```
bouweandela, line 173: Using the `mocker` fixture might be nice here:
```suggestion
def test_do_not_ignore_warnings(mocker):
    """Test do not ignore specific warnings."""
    mocker.patch.object(iris, "load_raw", autospec=True, side_effect=load_with_warning)
```

tests/integration/preprocessor/_io/test_load.py
```diff
@@ -21,153 +22,175 @@ def _create_sample_cube():
...
+    cube = _create_sample_cube()
+    temp_file = tmp_path / "cube.nc"
+    iris.save(cube, temp_file)
+
+    cubes = load(temp_file)
+
+    assert len(cubes) == 1
+    cube = cubes[0]
+    assert cube.attributes.globals["source_file"] == str(temp_file)
+    np.testing.assert_equal(cube.data, [1, 2])
```
bouweandela, line 36: Use `assert_array_equal` from the `tests` module here too?

tests/__init__.py
```diff
@@ -86,3 +95,62 @@ def __init__(self, cubes, filename, attributes, settings=None, **kwargs):
         self.copy_provenance = mock.Mock(return_value=self)
 
     group = PreprocessorFileBase.group
+
+
+def create_realistic_4d_cube():
```
bouweandela, line 100: It would be more in line with the pytest style of using fixtures to define this is a fixture in conftest.py

esmvalcore/preprocessor/_io.py
```diff
@@ -98,65 +108,63 @@ def load(
...
-        ignore_warnings = []
+    if not cubes:
+        raise ValueError(f"{file} does not contain any data")
 
-    # Avoid duplication of ignored warnings when load() is called more often
-    # than once
-    ignore_warnings = list(ignore_warnings)
+    for cube in cubes:
+        if "source_file" not in cube.attributes:
+            logger.warning(
```
bouweandela, line 135: Should we raise an exception for this? If it is only expected to happen when people contribute new fixes, I would raise an exception to make sure we don't bother users with many warnings. On the other hand, if it could happen to users, maybe it should be a debug log message instead?

bouweandela (approved):
Thanks @schlunma, I think this is a great first step towards supporting reading data from something else than posix filesystems.

tests/integration/preprocessor/_io/test_load.py
schlunma: Done in https://github.com/ESMValGroup/ESMValCore/pull/2579/commits/24bc1500089e9962c55d4a852b8fabfc3fc166e4.

esmvalcore/preprocessor/_io.py
schlunma: It might be better to raise an actual `ESMValCoreUserWarning` instead of using the logger since this problem can only occur when using the Python API. This makes it much more straightforward to ignore the warning if required. @bouweandela what do you think?

valeriupredoi:
I agree with @bouweandela but could you pls point me where you are loading the file with `ncdata.NcData()` please? That should be a well defined and exceptioned function in `_io.py` that then is called from `load` :beer:

esmvalcore/cmor/_fixes/cmip6/cesm2.py
```diff
@@ -22,25 +23,30 @@ class Cl(Fix):
 
     def _fix_formula_terms(
         self,
-        filepath,
-        output_dir,
-        add_unique_suffix=False,
-    ):
+        file: str | Path,
```
valeriupredoi, line 26: that's a URI or a Path - it's not a `File` ie an open file object, is it? If not, pls don't call it `file`, that may be misleading

esmvalcore/cmor/_fixes/cmip6/cesm2.py
```diff
@@ -54,7 +60,7 @@ def fix_file(self, filepath, output_dir, add_unique_suffix=False):
 
         Parameters
         ----------
-        filepath : str
+        file : str
```
valeriupredoi, line 63: same here - `file` is usually an open file object or handler

esmvalcore/iris_helpers.py
```diff
@@ -518,3 +527,126 @@ def safe_convert_units(cube: Cube, units: str | Unit) -> Cube:
...
+                if coord.var_name in ds_coords:
+                    ds_coord = ds_coords[coord.var_name]
+                    coord.units = _get_attribute(ds_coord, "units")
+
+            # If possible, add the source file as an attribute to support
+            # grouping by file when calling fix_metadata.
+            if filepath is not None:
+                cube.attributes.globals["source_file"] = str(filepath)
+
+    return cubes
```
valeriupredoi, line 652: why are you calling the ncdata-iris lock share here (ie in this script) if the open file call is not here? The lock should be created when the file is opened, otherwise it's not gonna work. Which leads me to my main question, where do you open the file with ncdata?

valeriupredoi (changes requested):
Hi Manu, sorry for being a party pooper, but I have some concerns wrt the implementation, could you please have a look at my questions :beer:

esmvalcore/cmor/_fixes/cmip6/cesm2.py
schlunma: It's called `file` to be consistent with its [parent method](https://github.com/ESMValGroup/ESMValCore/blob/f3e48c986a0a3b722b6311365c650b257b2bb1a1/esmvalcore/cmor/_fixes/fix.py#L86), in which `file` can have the types `str | Path | xr.Dataset | ncdata.NcData`. I agree that `file` is not optimal, but couldn't think of something else. Do you have a better suggestion?

esmvalcore/iris_helpers.py
schlunma: > why are you calling the ncdata-iris lock share here (ie in this script) if the open file call is not here?

As far as I understand the [doc](https://ncdata.readthedocs.io/en/latest/details/api/ncdata.threadlock_sharing.html#module-ncdata.threadlock_sharing) and the [code](https://github.com/pp-mo/ncdata/blob/16d60e5c147a469474503f2c2cf565351b3e20f9/lib/ncdata/threadlock_sharing.py#L64-L119), it doesn't matter where you call this function, as long as it's in the main thread. @pp-mo is that assumption correct?

Since the function [`dataset_to_iris`](https://github.com/ESMValGroup/ESMValCore/blob/f3e48c986a0a3b722b6311365c650b257b2bb1a1/esmvalcore/iris_helpers.py#L588-L652) lives in this module, I thought this is the best place for now. Would you have another suggestion? It should be in a module that is also loaded when using our Python API (maybe simply in `ESMValCore/esmvalcore/__init__.py`?).

> where do you open the file with ncdata?

This PR only implements the support for opening files with ncdata or xarray in the `fix_file` function, but doesn't actually use that. I plan to implement this in a future (dataset-specific) PR. There's no file-opening happening with ncdata or xarray in this PR.

valeriupredoi: OK, it makes sense, I was concerned about this previously thinking there was a call where the `ncdata.Dataset` is initialized and loaded, ie an actual open file call, there would have been the right place to create the lock, but since there is no such call, then this looks to be the right place (the to-iris bit justifies it)

valeriupredoi (approved):
> There's no file-opening happening with ncdata or xarray in this PR.

Bingo! Thanks, Manu! All clear to me now, sorry if I sounded like an old man yelling at the clouds :cloud:

Issue #2584, opened 2024-11-22
New data sources


bouweandela:
Design for improving the tool so new data sources can be added easily
----------------------------------------------------------------------

[...]

Issue #1078, opened 2021-04-22
Support for user- or dataset-defined regex patterns for dates


bsolino:
**Is your feature request related to a problem? Please describe.**
I'm currently working on an on-the-fly cmorizer for ICON data. The dataset files follow an unrecognized pattern for the date, so I have to extend the data_finder to add support to it.

I find a few flaws to this approach:
* Every new regex pattern could introduce a missmatch with an existing pattern
* If the pattern is not standard, this is basically hardcoding dataset information into the core
* Furthermore, developers for new datasets should be aware of maybe having to modify the data_finder

My suggestion is to include support for custom regex patterns that could be extracted from a configuration file (config-developer seems a good place). If such pattern is present, it could bypass the default checks, as we know for sure it applies to the dataset. The existing data_finder checks could then be left to the more general regex expressions that would work in most cases.

**Would you be able to help out?**
Would you have the time and skills to implement the solution yourself?

senesis:
I also had a problem with the standard date finding, because of an experiment name like 'expe-1950' just before the two dates. 

Albeit I was able to design an (hopefully) upward-compatible improvement of the code (by enforcing date pairs to have the same length), I agree that the ultimate solution is to allow for configuration, at the project or drs level 

A further improvement to the current date_finding scheme could be to request a place-holder in the filename pattern for indicating where the date info is to be found (e.g. {date}=

zklaus:
@bsolino, could you please add an example of what the ICON format looks like? To collect the examples here, @senesis mentioned in #1127 a similar issue with IPSLCM files that are named like
```
CM61-LR-hist-03.1950_18500101_19491231_1M_concbc.nc
```
where the data finder finds as range `1950_18500101`.

I think the data finder should not accept different resolutions for start and end, ie both should be "YYYY" or both "YYYYMM" or both "YYYYMMDD", etc., but not start like this and end like that.

But there may well be other problems.

bsolino:
@zklaus Here is an example of one of the outputs we currently have
```
icon-2.6.1_atm_amip_R2B5_r1v1i1p1l1f1_phy_3d_ml_20150101T000000Z.nc
```

I'm unsure what the numbers before Z mean.

zklaus:
[...]

PR #2601, opened 2024-12-02
Add ocean variable to the ACCESS Live CMORiser

rbeucher:
## Description

Adds Ocean variable to the ACCESS Live CMORIser

***

***


rbeucher:
@rhaegar325 please check that I have not introduced any issues.
@flicj191. This is originally @rhaegar325's work. Please give it a try.

rhaegar325:
HI, @valeriupredoi , this pr is about a new version of access on-the-fly cmoriser. we add some extra function to make it compatible with two ocean variable. could you please review and merge it if there is no conflict.

valeriupredoi:
many thanks folks, since @rbeucher assigned @flicj191 then I'll wait for her expert review then I'll have an overview looks and merge :+1:

doc/quickstart/find_data.rst
```diff
@@ -593,7 +593,7 @@ Key                  Description                            Default value if not
 ``modeling_realm``   Realm attribute include `atm`, `ice`   No default (needs to be
                      and `oce`                              specified in extra facets or
                                                             recipe if default DRS is used)
-```special_attr``    A special attribute in the filename    No default
+``freq_attribute``   A special attribute in the filename    No default
```
flicj191, line 596: As we are changing `special_attr` to freq`_attribute`  we need to edit above as well in lines 576 and 562.  Maybe also add for ocean directories and and files like line 561 and 562 
`'{dataset}/{sub_dataset}/{exp}/{modeling_realm}'`
`'ocean_{freq_attribute}.nc-*'`
and also add to the list of supported variables for ocean (ln 556/557)

yz9299_ESMValTool.yml
flicj191, line 1: this empty file can be removed

flicj191 (changes requested):
Thanks @rhaegar325 I was able to run a simple recipe and find the data. I've added comments to documentation.
Does still run very slow, it looks like it has to read from each file the date range as the file name (eg.`ocean_month.nc-20121231`) doesn't have the date surrounded by '_', '-', '.' as in this dataset part:
https://github.com/ESMValGroup/ESMValCore/blob/ea1734b16899881e7436da83f1eadadb0fc9a788/esmvalcore/local.py#L77C1-L82C34
maybe can add to here? @valeriupredoi @rbeucher 
not sure where's best to constrain reading specific variable from the file if that would improve? as each file might have like ~200cubes

rhaegar325:
Thanks @flicj191 for your review, I will remove the empty file. for the file-reading part, it is a problem and bother me for a long time, I think your suggestion could be a solution for this.

rhaegar325:
Hi, @valeriupredoi, it seems that I don't have write access to this repo since I can not push my local branch, could you please give me the write permission so that I could manage the branch of this pr?

flicj191:
> Thanks @rhaegar325 I was able to run a simple recipe and find the data. I've added comments to documentation. Does still run very slow, it looks like it has to read from each file the date range as the file name (eg.`ocean_month.nc-20121231`) doesn't have the date surrounded by '_', '-', '.' as in this dataset part: https://github.com/ESMValGroup/ESMValCore/blob/ea1734b16899881e7436da83f1eadadb0fc9a788/esmvalcore/local.py#L77C1-L82C34 maybe can add to here? @valeriupredoi @rbeucher not sure where's best to constrain reading specific variable from the file if that would improve? as each file might have like ~200cubes

I have edited the `local.py` so it can find the right files by date without loading, though not sure if this is the best was and the affect on other files. Will have to find the ruff format error it added, sorry!

Having a look at the test dataset (`- {project: ACCESS, dataset: ACCESS-ESM1-5, mip: Omon, sub_dataset: HI-CN-05, exp: history, 
    freq_attribute: month, start_year: 2000, end_year: 2005}`)
The raw data has latitude and longitude as both dimension coordinate and auxiliary coordinate, so both 1D and 2D. This one could be a regular grid so maybe to fix we can remove the auxiliary coordinates? we may need to check with the modellers? 
Otherwise, we should also add the `areacello` variable to be added as a 'supplementary variable' so that it can do area_statistics.

rbeucher:
Thanks for your help with this @flicj191

rhaegar325:
Really appreciate for your help and your advice @flicj191

rhaegar325:
Hi @valeriupredoi, I have a question about the load the  supplementary data in the live-cmoriser. I am making the live-cmoriser compatiable with ocean variables, and ocean variable have a special grid for the latitude and longitude, so I may need to load it while cmorising. I was wondering if there is any method allow me to load data store in `.nc` file while the live-cmoriser is working?

valeriupredoi:
> Hi @valeriupredoi, I have a question about the load the supplementary data in the live-cmoriser. I am making the live-cmoriser compatiable with ocean variables, and ocean variable have a special grid for the latitude and longitude, so I may need to load it while cmorising. I was wondering if there is any method allow me to load data store in `.nc` file while the live-cmoriser is working?

best person to answer that question is Manuel @schlunma mate - I have very little knowledge of the on-the-fly cmorizers, unfortunately. Manu, could you assist here, when you have a minute, please buds? :beer:

schlunma:
I am not sure I understand what you mean. Are you referring to supplementary variables (like grid cell areas, land/sea fractions, etc.)? See [here](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/recipe/preprocessor.html#supplementary-variables-ancillary-variables-and-cell-measures) for details on that. If you're referring to other netcdf files that contain other variables, have a look at the [ICON fix](https://github.com/ESMValGroup/ESMValCore/blob/8f94879b76ae25b3be8af2a9071ecf7d7683023d/esmvalcore/cmor/_fixes/icon/_base_fixes.py#L384), where we also load another file that describes the horizontal grid during the fix.

Also, I am very hesitant about the change in `local.py`. This has the potential to break data finding in other circumstances. I think we should fix this properly during the implementation of https://github.com/ESMValGroup/ESMValCore/issues/2584. There is also an open issue about this: https://github.com/ESMValGroup/ESMValCore/issues/1078

rhaegar325:
Hi, @valeriupredoi @schlunma , really appreciate for your suggestion about our pr, that's really helpful. Now we have solved all the issues and this pr is ready to be reviewed and merged. Please review it when it's convenient for you. Thanks again for all the suggestions and help.

rbeucher:
@charles-turner-1 do you think you can review that one for us please?

esmvalcore/cmor/_fixes/access/_base_fix.py
```diff
@@ -21,8 +26,103 @@ def fix_coord_system(self, cube):
     def get_cubes_from_multivar(self, cubes):
         """Get cube before calculate from multiple variables."""
         name_list = self.extra_facets.get("raw_name", self.vardef.short_name)
+        name_list = self.extra_facets.get("raw_name", self.vardef.short_name)
```
charles-turner-1, line 29: Duplicate line here

esmvalcore/cmor/_fixes/access/_base_fix.py
```diff
@@ -21,8 +26,103 @@ def fix_coord_system(self, cube):
...
 
         data_list = []
         for name in name_list:
             data_list.append(self.get_cube(cubes, name))
         return CubeList(data_list)
+
+    def fix_ocean_dim_coords(self, cube):
+        """Fix dim coords of ocean variables."""
+        cube.dim_coords[-2].points = np.array([int(i) for i in range(300)])
```
charles-turner-1, line 38: `np.arange(300)` here will give the same result as `np.array([int(i) for i in range(300)])` should be much faster.

esmvalcore/cmor/_fixes/access/_base_fix.py
```diff
@@ -21,8 +26,103 @@ def fix_coord_system(self, cube):
...
+        cube.dim_coords[-2].standard_name = None
+        cube.dim_coords[-2].var_name = "j"
+        cube.dim_coords[-2].long_name = "cell index along second dimension"
+        cube.dim_coords[-2].attributes = None
+
+        cube.dim_coords[-1].points = np.array([int(i) for i in range(360)])
```
charles-turner-1, line 44: As above, => `np.arange(360)`

esmvalcore/cmor/_fixes/access/_base_fix.py
```diff
@@ -21,8 +26,103 @@ def fix_coord_system(self, cube):
...
+        cube.dim_coords[-1].attributes = None
+        cube.dim_coords[-1].units = Unit(1)
+
+    def fix_ocean_aux_coords(self, cube):
+        """Fix aux coords of ocean variables."""
+        lat_bounds, lon_bounds = self.load_ocean_grid_data("ocean_grid_path")
+        temp_points = []
+        for i in cube.aux_coords[-1].points:
+            temp_points.append(
+                [j + 360 for j in i if j < 0] + [j for j in i if j >= 0]
```
charles-turner-1, line 57: Assuming this is a numpy array, you can write this as `temp_points.append( j % 360)` which is (IMO) easier to read, and won't run into any potential ordering issues. It'll also be substantially faster.

esmvalcore/cmor/_fixes/access/_base_fix.py
```diff
@@ -21,8 +26,103 @@ def fix_coord_system(self, cube):
...
+        cube.aux_coords[-1].standard_name = "longitude"
+        cube.aux_coords[-1].long_name = "longitude"
+        cube.aux_coords[-1].var_name = "longitude"
+        cube.aux_coords[-1].attributes = None
+        cube.aux_coords[-1].units = "degrees"
+        cube.aux_coords[-1].bounds = lon_bounds
+
+        temp_points = []
+        for i in cube.aux_coords[-2].points:
+            temp_points.append([j.astype(np.float64) for j in i])
```
charles-turner-1, line 69: I think this whole type conversion operation can be done in a single instruction, which again should be quite a lot faster:
`np.array(i, dtype=np.float64)`

Probably possible to remove the whole loop depending on the type of `cube.aux_coords[-2].points`, eg:
```python 
temp_points = np.array(cube.aux_coords[-2].points,dtype=np.float64)
```

tests/integration/cmor/_fixes/access/test_access_esm1_5.py
```diff
@@ -22,6 +24,80 @@
     units=Unit("days since 1851-01-01", calendar="noleap"),
     attributes={"test": 1, "time_origin": "will_be_removed"},
 )
+
+time_ocn_coord = DimCoord(
+    [int(i) for i in range(1, 13)],
```
charles-turner-1, line 29: Probably easier to just write `list(range(1,13))` here

tests/integration/cmor/_fixes/access/test_access_esm1_5.py
```diff
@@ -149,6 +251,43 @@ def check_heightxm(cube, height_value):
     assert height.bounds is None
 
 
+def check_ocean_dim_coords(cube):
+    """Check dim_coords of ocean variables."""
+    assert (
+        cube.dim_coords[-2].points == np.array([int(i) for i in range(300)])
```
charles-turner-1, line 257: `np.array([int(i) for i in range(300)])` => `np.arange(300)`

tests/integration/cmor/_fixes/access/test_access_esm1_5.py
```diff
@@ -149,6 +251,43 @@ def check_heightxm(cube, height_value):
...
+    ).all()
+    assert cube.dim_coords[-2].standard_name is None
+    assert cube.dim_coords[-2].var_name == "j"
+    assert cube.dim_coords[-2].long_name == "cell index along second dimension"
+    assert cube.dim_coords[-2].attributes == {}
+
+    assert (
+        cube.dim_coords[-1].points == np.array([int(i) for i in range(360)])
```
charles-turner-1, line 265: `np.array([int(i) for i in range(360)])` => `np.arange(360)`

charles-turner-1 (changes requested):
Couple of thing related to numpy arrays that I think would be useful to change for performance/readability.

Otherwise looks good - although I don't know enough about the logic of the CMORIsers so my opinion isn't worth much there.

esmvalcore/cmor/_fixes/access/_base_fix.py
charles-turner-1: Probably also the case that this whole loop can be written as 
```python
temp_points = cube.aux_coords[-1].points % 360
```
, similarly to what I suggested below - although I'd check this a little more carefully.

esmvalcore/cmor/_fixes/access/_base_fix.py
rhaegar325: Thanks charles for pointing that out, it might have been caused by the auto-merge

esmvalcore/cmor/_fixes/access/_base_fix.py
```diff
@@ -53,9 +52,7 @@ def fix_ocean_aux_coords(self, cube):
         lat_bounds, lon_bounds = self.load_ocean_grid_data("ocean_grid_path")
         temp_points = []
         for i in cube.aux_coords[-1].points:
-            temp_points.append(
-                [j + 360 for j in i if j < 0] + [j for j in i if j >= 0]
-            )
+            temp_points.append([j % 360 for j in i])
```
charles-turner-1, line 55: This doesn't need to be a list comprehension - you should be able to replace with either 
```python
temp_points = []       
for i in cube.aux_coords[-1].points:
    temp_points.append(i % 360)
``` 
or potentially - double check this is valid, I think it should be 
```python
temp_points = cube.aux_coords[-1].points % 360
```

charles-turner-1 (approved):

PR #2610, opened 2024-12-05
Only save data from one preprocessing task at a time with the Distributed scheduler

### LINKED ISSUE #2609: `max_parallel_tasks` and Dask
Issue reported by @schlunma 

> Since https://github.com/ESMValGroup/ESMValCore/pull/2522, `recipe_schlund20esd.yml` hangs indefinitely after a while when using the default threaded scheduler and `--max-parallel-tasks=16`. Another example is an ICON recipe that evaluates 3D data, which now hangs with the distributed scheduler and `--max-parallel-tasks=2`. Before https://github.com/ESMValGroup/ESMValCore/pull/2522, both setups run fine. With `--max-parallel-tasks=1`, the recipes also run fine with current `main` branch. So my guess is the changes in https://github.com/ESMValGroup/ESMValCore/pull/2522 do not play nicely with our own custom parallelization.

bouweandela:
## Description

When using the Distributed Dask scheduler, only save the files from one preprocessing task at a time to avoid running out of memory.

Further optimizations could be made by preparing all the task graphs for saving data and then running everything with a single call to `dask.compute`, as e.g. done in #2316. Such an approach would allow re-using loaded data if multiple preprocessing tasks load the same file and compute similar things. However, this would lead to an even larger Dask task graph, so we may want to make sure we do not create needlessly large task graphs anywhere in the preprocessor (see e.g. https://github.com/SciTools/iris/issues/5455) before implementing that.

Related to #2609

***

***


bouweandela:
This could really help to avoid issues with large memory use, but it would require some benchmarking to see how it affects the runtime of large recipes before merging.

schlunma (commented):
Code looks very good! I am running some tests at them moment, but it looks very promising. Thanks Bouwe 🚀

schlunma:
Here are some first test results with [`recipe_schlund20esd.yml`](https://github.com/ESMValGroup/ESMValTool/blob/main/esmvaltool/recipes/recipe_schlund20esd.yml) on one compute node of Levante. The distributed scheduler cases use 32 workers à 8 GiB (4 threads per worker).

| Code version | `--max-parallel-tasks` | Dask scheduler | run time (H:MM:SS)
| --- | --- | --- | --- |
| Before https://github.com/ESMValGroup/ESMValCore/pull/2522 (11a478781037d8d31b0a583f4cb4baa0f8341547)  | 1 | default  | ...running... |
| Before https://github.com/ESMValGroup/ESMValCore/pull/2522 (11a478781037d8d31b0a583f4cb4baa0f8341547)  | 16 | default  | ...running... |
| Before https://github.com/ESMValGroup/ESMValCore/pull/2522 (11a478781037d8d31b0a583f4cb4baa0f8341547)  | 1 | distributed | ...running... |
| Before https://github.com/ESMValGroup/ESMValCore/pull/2522 (11a478781037d8d31b0a583f4cb4baa0f8341547)  | 16 | distributed | ...running... |
| This PR  | 1 | default | 3:34:12 |
| This PR  | 16 | default | 1:30:28|
| This PR  | 1 | distributed | 2:33:38|
| This PR  | 16 | distributed | 0:32:23|

During the v2.11.0 testing, the recipe ran in [0:38:01](https://esmvaltool.dkrz.de/shared/esmvaltool/v2.11.0/recipe_schlund20esd_20240523_091921/).

bouweandela:
Thanks for all the testing @schlunma!

The NetCDF library used to read the data can only be used by a single thread at a time. I suspect that severely limits the amount of parallelism that can be achieved by the default, thread-based scheduler. On a compute node with many cores like the Levante compute nodes, this effect will be very notable. Observing CPU usage during a run of this recipe using the current `main` branch using `htop` suggests that in the best case almost 10 CPU cores are busy, while typically 1.2 are busy. Therefore we can set `num_workers: 10` without any loss of performance. With `max_parallel_tasks: 16`, that results in a runtime of 34 minutes with the current `main` branch.

Therefore this pull request only offers a performance advantage with the distributed scheduler. An advantage for the default scheduler is that it prevents the tool from crashing if the number of Dask workers is not configured or not correctly configured, but it comes at a considerable computational performance cost so it is probably not worth it. Better documentation seems a more attractive solution.

schlunma:
Thanks Bouwe, I didn't know the number of workers for the threaded scheduler has such a big impact! But considering that each process spawns its own workers, this probably makes sense. Thus, I agree that for the threaded scheduler, this approach here is not a good solution, since we actually want the different schedulers to run in parallel.

> Therefore this pull request only offers a performance advantage with the distributed scheduler.

This is also not the case (see my latest numbers; comparison of main vs. this PR).

What I really wonder here is why the run times from main and this PR are so vastly different when only 1 process is used. The lock shouldn't be active in this case, should it? I actually quite like the idea of a lock, at least for the distributed scheduler.

valeriupredoi:
Manu, by "vastly different" you mean this row?

1 	distributed 	3:04 	1:28 	2:20

schlunma:
Yeah, basically the difference in the columns main vs. this PR with 1 process, i.e., 2:23 -> 3:34 for the threaded scheduler and 1:28 -> 2:20 for the distributed one.

valeriupredoi:
Threaded is Dreaded: as much as possible we shouldn't use Threaded bc that is heavily dependent on the lock; you'd also expect this PR to make things wait a wee while longer bc it's just one process/task at a time vs a pool of processes all writing concurrently (to different files of course); best bet to test would be a recipe that outputs a lot of preprocesses files - there you should see the memory going down, while the times they may be longer, and by quite a bit

valeriupredoi:
...which I think you've already found and are using (ie your tests here) since I just noticed those are HH:MM and not MM:SS as I initially thought :grin:

schlunma:
The thing is - the lock should only for parallel processing. For serial processing, the corresponding variable is just `None`. So if I understand the code correctly, there shouldn't be ANY difference in run times for `--max-parallel-tasks=1` between the current main and this PR...

valeriupredoi:
no because the process (when you use Threaded scheduler) uses threading even if max parallel tasks = 1 ie it's one process/task that is distributed on the threads of the core where it's running

schlunma:
But that's already the case for the current main, right? This PR doesn't change anything in the `_run_sequential()` code, and because of this the new lock is `None` and in this case [not used](https://github.com/ESMValGroup/ESMValCore/pull/2610/files#diff-4c5468e8eebd3bebe05236ed962fa98483d27a9374d17c0ad43ba19287435a5dR741). So I really don't understand why there is a difference in run time of more than an hour between main and this PR.

bouweandela:
> For serial processing, the corresponding variable is just None. So if I understand the code correctly, there shouldn't be ANY difference in run times for --max-parallel-tasks=1 between the current main and this PR...

Indeed, I would not expect any difference. Is the difference you're seeing reproducible? Maybe it was just a fluke

schlunma:
Is the difference you're seeing reproducible? Maybe it was just a fluke

> Yes, I ran the recipes twice. So unless I screwed up big time by looking at the wrong logs, I'm pretty sure it's there.

valeriupredoi:
it'd be good to measure how long it takes this bit to run:
```
        try:
            logger.info(
                "Computing and saving data for preprocessing task %s",
                self.name,
            )
            _compute_with_progress(delayeds, description=self.name)
 ```
ie from lock acquiring to lock release. The problem you are seeing here is that there are two different locks: a scheduler's lock that Bouwe turns on, lets the task run and finish, then turns it off; but there's a second lock, inside netCDF4/HDF5 that stops a process spread on multiple threads to IO synchronously, so it basically waits until one thread has finished IOing, to let the next thread do IO. Bouwe's lock could be None for that since that's a process' lock, while the threads' lock is still active. Now, the question is, why is this implementation slower for nprocs=1 ie Bouwe's lock is indeed None, and the work is done over one process that threads as before?

valeriupredoi:
I'd make 100% sure Bouwe's lock is indeed None for the nprocs=1 case, otherwise the scheduler is doing something funky; print that out and measure the times between acquire/release, I can't think of anything else here, why the difference in time; unless you ran on a shared node, and not on your own machine

valeriupredoi:
proves out I was waffling about the scheduler lock, that is not set unless in `_run_parallel` which is not called unles max parallel tasks > 1. So for a small recipe, with only three sequential tasks, ie the Python example, all looks good:

with Bouwe
-----------
COMPUTE_WITH_PROGRESS 0.30625486373901367
COMPUTE_WITH_PROGRESS 0.10400819778442383
COMPUTE_WITH_PROGRESS 0.7586956024169922
Time for running the recipe was: 0:00:16.538996

COMPUTE_WITH_PROGRESS 0.305743932723999
COMPUTE_WITH_PROGRESS 0.10391664505004883
COMPUTE_WITH_PROGRESS 0.8586757183074951
Time for running the recipe was: 0:00:17.099344

without Bouwe
-------------
COMPUTE_WITH_PROGRESS 0.30536580085754395
COMPUTE_WITH_PROGRESS 0.10322737693786621
COMPUTE_WITH_PROGRESS 0.764275312423706
Time for running the recipe was: 0:00:15.638896

COMPUTE_WITH_PROGRESS 0.30529236793518066
COMPUTE_WITH_PROGRESS 0.10352373123168945
COMPUTE_WITH_PROGRESS 0.7534520626068115
Time for running the recipe was: 0:00:16.489703

It's also crazy to see that the actual 3 preproc tasks take 1.2s but the entire preproc takes about 11s (crazy overheads on IO)

Manu, I am wondering about those times you posted, I am thinking something funny is going on in IO land, for some bizarre reason, could you maybe measure the times it takes `_compute_with_progress` like I did here?

@bouweandela am loving that progress bar, man! :beer:

schlunma:
Okay, I retested this with `--max-parallel-tasks=1`. Turns out that the run times vary A LOT. All of my tests use the same dask distributed settings and an entire compute node on Levante (which is just used by me, no other uses). Here are the results:

- current main: 
  - run 1: 2:18:42
  - run 2: 2:56:50
  - run mentioned above: 1:28:00
- this PR:
  - run 1: 2:05:45
  - run mentioned above: 2:20:00

I am starting to believe that this is actually a Levante issue...

valeriupredoi:
yeh that's why I'd measure just the compute times, Manu - all sorts of issues with IO via a busy FS can happen, even if you are the sole user of the system :beer:

valeriupredoi:
> after 9 PM or on weekends

maybe, but those are the best times to watch Top Gun Maverick too, or sleep, if one had a few too many pints on Friday :grin:

bouweandela:
> will open a new pull request to try to improve on the issue reported in https://github.com/ESMValGroup/ESMValCore/issues/2609 with the default scheduler.

Done in #2626

schlunma (commented):
Code looks good! I'm running some additional tests at the moment, but these look good so far.

One question: currently the lock is only used for preprocessing tasks. Would it make sense to also use it for diagnostic tasks (at least for the Python ones)?

bouweandela:
It is a possibility, but I'm not sure if it would make things faster or avoid overloading the scheduler. As far as I'm aware, typical diagnostic scripts only do smaller computations, with the main data reduction already happening during preprocessing, and spend most of the time creating plots.

schlunma (approved):
Tests ran fine, code looks good, cheers @bouweandela 🚀

PR #2626, opened 2024-12-20
Use better defaults when using ``max_parallel_tasks`` with an unconfigured threaded scheduler

bouweandela:
## Description

Sets a reasonable default for `num_workers` when the user has not configured it and is running with `max_parallel_tasks > 1` and add more documentation on how to configure for this situation.

Fixes #2609.

Link to documentation: https://esmvaltool--2626.org.readthedocs.build/projects/ESMValCore/en/2626/quickstart/configure.html#f5

***

## [Before you get started](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#getting-started)

- [x] [☝ Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do

## [Checklist](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#checklist-for-pull-requests)

It is the responsibility of the author to make sure the pull request is ready to review. The icons indicate whether the item will be subject to the [🛠 Technical][1] or [🧪 Scientific][2] review.

[1]: https://docs.esmvaltool.org/en/latest/community/review.html#technical-review
[2]: https://docs.esmvaltool.org/en/latest/community/review.html#scientific-review

- [x] [🧪][2] The new functionality is [relevant and scientifically sound](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#scientific-relevance)
- [x] [🛠][1] This pull request has a [descriptive title and labels](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-title-and-label)
- [x] [🛠][1] Code is written according to the [code quality guidelines](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#code-quality)
- [x] [🧪][2] and [🛠][1] [Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available
- [x] [🛠][1] [Unit tests](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#tests) have been added
- [x] [🛠][1] Changes are [backward compatible](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#backward-compatibility)
- [x] [🛠][1] The [list of authors](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#list-of-authors) is up to date
[...]

PR #2636, opened 2025-01-17
Fix OSX compatibility

bouweandela:
## Description

In #2626 the use of a function that is not available in OSX was introduced. This implements a fallback mechanism to restore OSX compatibility.

***

***


esmvalcore/_task.py
```diff
@@ -785,6 +785,15 @@ def __repr__(self):
         return string
 
 
+def available_cpu_count() -> int:
+    """Return the number of available CPU cores."""
+    if hasattr(os, "sched_getaffinity"):
+        return len(os.sched_getaffinity(0))
+    if count := os.cpu_count():
+        return count
+    return 1
```
schlunma, line 794: From the docs, it looks to me like [`len(os.sched_getaffinity(0))`](https://docs.python.org/3/library/os.html#os.sched_getaffinity) is the same as [`os.process_cpu_count()`](https://docs.python.org/3/library/os.html#os.process_cpu_count) (the latter only available since Python 3.13. For the first, the docs say that this can be `None`, too. Thus, it might make sense to include the `if count` condition for both (Unix and OSX) cases.

It would also be great to add a tiny comment why this distinction is necessary.

schlunma (commented):
Thanks Bouwe, looks good! Just one comment on the code.

valeriupredoi:
good call, gents! I was getting puzzlated why our OSX GA's are failing :beer:

esmvalcore/_task.py
bouweandela: > From the docs, it looks to me like [len(os.sched_getaffinity(0))](https://docs.python.org/3/library/os.html#os.sched_getaffinity) is the same as [os.process_cpu_count()](https://docs.python.org/3/library/os.html#os.process_cpu_count) (the latter only available since Python 3.13). For os.process_cpu_count(), the docs say that this can return None, too. Thus, it might make sense to include the if count condition for both (Unix and OSX) cases

Are you sure? The `len` function seems unlikely to return `None`. I wouldn't know how to make it do that.

bouweandela: > It would also be great to add a tiny comment why this distinction is necessary (Unix vs. OSX).

Done in 90f0cd9

schlunma: Yes, agreed, but it can return `0`, which is might be happen if `os.process_cpu_count()` returns `None` (but this is just speculation). It will probably never happen (I have no idea under what circumstances), but given that you factored that in for `cpu_count`, I think it can't hurt to do the same for `sched_getaffinity`, too.

bouweandela: > Yes, agreed, but it can return 0, which might happen if os.process_cpu_count() returns None

> From the docs, it looks to me like [len(os.sched_getaffinity(0))](https://docs.python.org/3/library/os.html#os.sched_getaffinity) is the same as [os.process_cpu_count()](https://docs.python.org/3/library/os.html#os.process_cpu_count)

Maybe it helps to look at the code: https://github.com/python/cpython/blob/fafc618e2f535dc972f9502d8547e089d4c7908d/Lib/os.py#L1178-L1189. As you can see, it is the other way around: `process_cpu_count` is the same as `len(os.sched_getaffinity(0))` in some cases.

Now, I tried to understand what the return types are when calling `os.sched_getaffinity` if not just a set, so I looked at [the code for os.sched_getaffinity](https://github.com/python/cpython/blob/80189ff6475d8ab667de1199e659d5070f457ab9/Modules/posixmodule.c#L8528-L8597) but it isn't immediately obvious what `sched_getaffinity` returns if not a set. It appears to return various things in various error cases, but I assume these are somehow handled by the Python code and I'm not familiar enough with that.

My feeling is that this is starting to take more time than I would like to spend on it. If you feel that additional changes are needed here, you're welcome to make them and push them directly to this branch. Would that work for you?

schlunma: Ok, let's leave it like that. To me it's just confusing to treat those cases differently. What was your motivation to add the `if count := os.cpu_count():` line? You could have also simply done `return os.cpu_count()`.

bouweandela: Because it can return `None`: 

> Return the number of logical CPUs in the system. Returns `None` if undetermined.

from https://docs.python.org/3/library/os.html#os.cpu_count and if it returns None, the mathematics for calculating the number of threads to use for the Dask workers will crash because that only works with numbers.

schlunma: Yes, that's what I thought. My motivation for my comment was the same: because the docs for `process_cpu_count` (which seems to be very similar to `sched_getaffinity`) also says 

> Returns None if undetermined

I though it's likely that `len(os.sched_getaffinity(0))` can return `0`. But as I said, it's fine, we can just leave, this will probably never be an issue.

schlunma (approved):
Thanks @bouweandela!

valeriupredoi (approved):
thanks a lot gents :beer:

PR #2637, opened 2025-01-20
Make sure that supplementary variables and weights have same chunks as parent cube

schlunma:
## Description

This PR ensures that supplementary variables and weights used for statistical operations use the same chunking as their parent cube. This improves computational performance.

***

***


esmvalcore/preprocessor/_supplementary_vars.py
```diff
@@ -65,47 +78,62 @@ def add_cell_measure(cube, cell_measure_cube, measure):
...
             f"measure name must be 'area' or 'volume', got {measure} instead"
         )
-    measure = iris.coords.CellMeasure(
-        cell_measure_cube.core_data(),
+    coord_dims = tuple(
+        range(cube.ndim - len(cell_measure_cube.shape), cube.ndim)
+    )
+    cell_measure_data = cell_measure_cube.core_data()
+    if cell_measure_cube.has_lazy_data():
+        cube_chunks = tuple(cube.lazy_data().chunks[d] for d in coord_dims)
```
valeriupredoi, line 86: we should really ask Iris folks to make the `chunks` object tuple available a lot easier than that :grin:

esmvalcore/preprocessor/_supplementary_vars.py
```diff
@@ -65,47 +78,62 @@ def add_cell_measure(cube, cell_measure_cube, measure):
...
+        cell_measure_data = cell_measure_data.rechunk(cube_chunks)
+    cell_measure = iris.coords.CellMeasure(
+        cell_measure_data,
```
valeriupredoi, line 89: what if the data is not lazy? `cell_measure_data` will not be defined for that. EDIT: bugger, am blind - it's above as `core_data()` - but pls remind me, is `core_data()` defaults to simply realized `data` if the data is not lazy?

esmvalcore/preprocessor/_volume.py
```diff
@@ -104,7 +104,7 @@ def extract_volume(
     return cube.extract(z_constraint)
 
 
-def calculate_volume(cube: Cube) -> da.core.Array:
+def calculate_volume(cube: Cube) -> np.ndarray | da.Array:
```
valeriupredoi, line 107:
```suggestion
def calculate_volume(cube: Cube) -> np.ndarray | dask.array.Array:
```

valeriupredoi (approved):
looking very good, many thanks, Manu!  A couple of very minor bits from me, with one question here: what `chunks` were the suppl vars and weights on before this? ie what was the mechanism dictating their chunking - just a regular predefined chunks constructed by Iris?

esmvalcore/preprocessor/_supplementary_vars.py
schlunma: I don't think it's that bad TBH, it's just one `lazy_data()` or `core_data()` more in the call. I am not sure how useful a `cube.chunks` would be, given that data can be non-lazy (chunks are not defined in this case) or that multiple arrays can be assigned with a cube (coordinates, cell measures, ancillary variables), each with their own chunking...

schlunma:
> A couple of very minor bits from me, with one question here: what `chunks` were the suppl vars and weights on before this? ie what was the mechanism dictating their chunking - just a regular predefined chunks constructed by Iris?

Yes, it used the predefined chunks from Iris. In many cases this is probably fine, but to be safe I think it's a good approach to include the rechunking here.

esmvalcore/preprocessor/_supplementary_vars.py
valeriupredoi: it's be good to have it like `netCDF4.Dataset.chunks()` works - also a good comparison to that 0 the original HDF5 storage chunks vs iris rechunked data, but obv not in scope here. Thanks, Manu :beer:

esmvalcore/preprocessor/_supplementary_vars.py
valeriupredoi: fab, my memory is not yet failing me :grin:

esmvalcore/preprocessor/_volume.py
valeriupredoi: egh those should be full names of modules not `import numpy as cow; obj: cow.ndarray` :grin:

valeriupredoi:
> > A couple of very minor bits from me, with one question here: what `chunks` were the suppl vars and weights on before this? ie what was the mechanism dictating their chunking - just a regular predefined chunks constructed by Iris?
> 
> Yes, it used the predefined chunks from Iris. In many cases this is probably fine, but to be safe I think it's a good approach to include the rechunking here.

cheers, bud, makes sense - @bouweandela you have a minute have one look at and merge this, pls bud :beer:

esmvalcore/preprocessor/_volume.py
```diff
@@ -158,7 +158,9 @@ def calculate_volume(cube: Cube) -> da.core.Array:
     # Calculate Z-direction thickness
     thickness = depth.core_bounds()[..., 1] - depth.core_bounds()[..., 0]
     if cube.has_lazy_data():
-        thickness = da.array(thickness)
+        thickness = da.array(thickness).rechunk(
+            tuple(cube.lazy_data().chunks[d] for d in z_dim)
+        )
```
bouweandela, line 163: It is better to avoid rechunking if not necessary. Therefore I would recommend doing something like:
```suggestion
        chunks = tuple(cube.lazy_data().chunks[d] for d in z_dim)
        if isinstance(thickness, da.Array):
            thickness = thickness.rechunk(chunks)
        else:
            thickness = da.asarray(thickness, chunks=chunks)
        )
```

esmvalcore/preprocessor/_volume.py
```diff
@@ -182,6 +184,8 @@ def calculate_volume(cube: Cube) -> da.core.Array:
         thickness, cube.shape, z_dim, chunks=chunks
     )
     grid_volume = area_arr * thickness_arr
+    if cube.has_lazy_data():
+        grid_volume = grid_volume.rechunk(chunks)
```
bouweandela, line 188: Does this do anything? I would expect chunks to be preserved on multiplication.

bouweandela (commented):
Out of curiosity, did you measure performance and find some cases where this helps?

While it is a nice addition to have the cell measures and ancillary variables chunked the same way as the cube on load, I thought we were already broadcasting them to the cube shape and chunks before actually using them whenever that is needed.

esmvalcore/preprocessor/_volume.py
schlunma: I would expect so too, but this probably can't hurt? It looks like `rechunk` is a no-op if the chunks already match:

```py
import dask.array as da
x = da.ones((12, 24, 48), chunks=(6, 12, 24))
y = da.ones((12, 24, 48), chunks=(6, 12, 24)).rechunk((6, 12, 24))
print(y.dask == x.dask)  # True
```

schlunma:
> Out of curiosity, did you measure performance and find some cases where this helps?

No, but from experience it seems to be a good idea to align chunks of different arrays when they are used in a common calculation (see, e.g., https://github.com/ESMValGroup/ESMValCore/pull/2404).

 
> While it is a nice addition to have the cell measures and ancillary variables chunked the same way as the cube on load, I thought we were already broadcasting them to the cube shape and chunks before actually using them whenever that is needed.

Yes, as far as I am aware this is the case. However, (1) I am not 100% if this is true for all cases, and (2) broadcasting [cannot alter chunks of existing dimensions](https://docs.dask.org/en/stable/generated/dask.array.broadcast_to.html#dask.array.broadcast_to). Hence, we should make sure that the chunks of existing dimensions also align with those of the data.

bouweandela:
> broadcasting [cannot alter chunks of existing dimensions](https://docs.dask.org/en/stable/generated/dask.array.broadcast_to.html#dask.array.broadcast_to)

That's a very point, this should definitely help with improving performance then.

esmvalcore/preprocessor/_volume.py
bouweandela: I thought this would always be the case since we're broadcasting with the chunks specified, but as you pointed out, this will only rechunk dimensions that are being broadcast, so you're right and it does make sense to have this here.

schlunma: Technically, the non-broadcasted arrays `thickness` and `area_array` should already have the correct chunks (this is ensured in our code/Iris code earlier), but better safe than sorry.

PR #2641, opened 2025-01-21
Add preprocessor to extract surface values from 3D atmospheric variables

### LINKED ISSUE #2611: Add derivation of surface concentrations for additional trace gases (CH4 and N2O)
The variables `ch4` and `n2o` are 3D variables providing trace gas concentrations on pressure levels. The related surface quantities are often useful in order to compare to surface measurements (see cmorizer and diagnostics in development in ESMValTool repo [Issue 3838](https://github.com/ESMValGroup/ESMValTool/issues/3838) and [Issue 3839](https://github.com/ESMValGroup/ESMValTool/issues/3839) ). In the spirit of what was done for `co2s` #587 , a dedicated preprocessor could be added to create the corresponding surface variables.

Is there a way to make this surface derivation method somewhat standard or potentially used as a standalone preprocessor ? As it currently looks like, a dedicated preprocessing file would be needed in the [esmvalcore//preprocessor/_derive](https://github.com/ESMValGroup/ESMValCore/tree/main/esmvalcore/preprocessor/_derive) folder for each variable `ch4s` and `n2os`.

jlenh:
## Description

This pull request introduces a new preprocessor function to extract the surface values from 3D atmospheric variables and corresponding surface pressure `ps`. The preprocessor function is largely based on the implementation done for the derivation of surface CO2 `co2s` at https://github.com/ESMValGroup/ESMValCore/blob/main/esmvalcore/preprocessor/_derive/co2s.py. It thus avoids having to create either at standalone derive preprocessor or a new entry in the relevant CMOR tables. The attributes of the 3D atmospheric variables are transferred to the newly created cube and the variable name is changed to `f'{var_name}s'`.

Note: so far, the variable name is required as input of the preprocessor and needs to be used in a recipe as
```
preprocessors:
  extract_co2s:
    extract_surface_from_atm:
      var: 'co2'
```

Closes #2611 

***

***

### DIFF
--- a/esmvalcore/preprocessor/_supplementary_vars.py
+++ b/esmvalcore/preprocessor/_supplementary_vars.py
@@ -5,7 +5,6 @@ from typing import Iterable
 
 import iris.coords
 import iris.cube
-import numpy as np
 
 logger = logging.getLogger(__name__)
 
--- a/esmvalcore/preprocessor/_supplementary_vars.py
+++ b/esmvalcore/preprocessor/_supplementary_vars.py
@@ -5,6 +5,7 @@ from typing import Iterable
 
 import iris.coords
 import iris.cube
+import numpy as np
 
 logger = logging.getLogger(__name__)
 
@@ -105,13 +106,21 @@ def add_ancillary_variable(cube, ancillary_cube):
         var_name=ancillary_cube.var_name,
         attributes=ancillary_cube.attributes,
     )
-    start_dim = cube.ndim - len(ancillary_var.shape)
-    cube.add_ancillary_variable(ancillary_var, range(start_dim, cube.ndim))
-    logger.debug(
-        "Added %s as ancillary variable in cube of %s.",
-        ancillary_cube.var_name,
-        cube.var_name,
-    )
+    data_dims = [None] * ancillary_cube.ndim
+    for coord in ancillary_cube.coords():
+        for ancillary_dim, cube_dim in zip(
+            ancillary_cube.coord_dims(coord), cube.coord_dims(coord)):
+            data_dims[ancillary_dim] = cube_dim
+    if None in data_dims:
+        none_dims = ", ".join(
+            str(i) for i, d in enumerate(data_dims) if d is None)
+        msg = (
+            f"Failed to add {ancillary_cube} to {cube} as ancillary variable. "
+            f"No coordinate associated with ancillary cube dimensions {none_dims}"
+        )
+        raise ValueError(msg)
+
+    cube.add_ancillary_variable(ancillary_var, data_dims)
 
 
 def add_supplementary_variables(
--- a/esmvalcore/preprocessor/_volume.py
+++ b/esmvalcore/preprocessor/_volume.py
@@ -16,6 +16,7 @@ import numpy as np
 from iris.coords import AuxCoord, CellMeasure
 from iris.cube import Cube
 from iris.util import broadcast_to_shape
+import stratify
 
 from ._shared import (
     get_iris_aggregator,
@@ -603,3 +604,115 @@ def extract_trajectory(
     points = [("latitude", latitudes), ("longitude", longitudes)]
     interpolated_cube = interpolate(cube, points)  # Very slow!
     return interpolated_cube
+
+
+def _get_first_unmasked_data(array, axis):
+    """Get first unmasked value of an array along an axis."""
+    mask = da.ma.getmaskarray(array)
+    numerical_mask = da.where(mask, -1.0, 1.0)
+    indices_first_positive = da.argmax(numerical_mask, axis=axis)
+    indices = da.meshgrid(
+        *[da.arange(array.shape[i]) for i in range(array.ndim) if i != axis],
+        indexing="ij",
+    )
+    indices = list(indices)
+    indices.insert(axis, indices_first_positive)
+    first_unmasked_data = np.array(array)[tuple(indices)]
+    return first_unmasked_data
+
+
+@register_supplementaries(
+        variables=['ps'],
+        required='require_at_least_one'
+)
+def extract_surface_from_atm(
+    cube: Cube,
+    var: str
+) -> Cube:
+    """Extract data from the lowest unmasked height level.
+
+    Parameters
+    ----------
+    cube:
+        Input cube.
+    var:
+        Variable to extract at the surface.
+
+    Returns
+    -------
+    iris.cube.Cube
+        Collapsed cube.
+    """
+    # Declare required variables
+    #   - 3D atmo variable to extract at the surface
+    #   - surface_air_pressure (ps)
+    ps_cube = None
+    try:
+        ps_cube = cube.ancillary_variable("surface_air_pressure")
+    except iris.exceptions.AncillaryVariableNotFoundError:
+        logger.debug(
+            "Ancillary variable surface air pressure not found in cube. "
+            "Check file availability."
+        )
+    if ps_cube:
+        var_cube = cube.extract(
+            iris.Constraint(name=var)
+        )
+
+        # Fill masked data if necessary (interpolation fails with masked data)
+        (z_axis,) = var_cube.coord_dims(
+            var_cube.coord(axis="Z", dim_coords=True)
+        )
+        mask = da.ma.getmaskarray(var_cube.core_data())
+        if mask.any():
+            first_unmasked_data = _get_first_unmasked_data(
+                var_cube.core_data(), axis=z_axis
+            )
+            dim_map = [dim for dim in range(var_cube.ndim) if dim != z_axis]
+            first_unmasked_data = iris.util.broadcast_to_shape(
+                first_unmasked_data, var_cube.shape, dim_map
+            )
+            var_cube.data = da.where(
+                mask, first_unmasked_data, var_cube.core_data()
+            )
+
+        # Interpolation (not supported for dask arrays)
+        air_pressure_coord = var_cube.coord("air_pressure")
+        original_levels = iris.util.broadcast_to_shape(
+            air_pressure_coord.points,
+            var_cube.shape,
+            var_cube.coord_dims(air_pressure_coord),
+        )
+        target_levels = np.expand_dims(ps_cube.data, axis=z_axis)
+        sfc_data = stratify.interpolate(
+            target_levels,
+            original_levels,
+            var_cube.data,
+            axis=z_axis,
+            interpolation="linear",
+            extrapolation="linear",
+        )
+        sfc_data = np.squeeze(sfc_data, axis=z_axis)
+
+        # Construct surface var cube
+        indices = [slice(None)] * var_cube.ndim
+        indices[z_axis] = 0
+        var_cube = var_cube[tuple(indices)]
+        var_cube.data = sfc_data
+        if var_cube.coords("air_pressure"):
+            var_cube.remove_coord("air_pressure")
+        ps_coord = iris.coords.AuxCoord(
+            ps_cube.data,
+            var_name="plev",
+            standard_name="air_pressure",
+            long_name="pressure",
+            units=ps_cube.units,
+        )
+        var_cube.add_aux_coord(ps_coord, np.arange(var_cube.ndim))
+        var_cube.var_name = var + 's'
+        logger.debug("Extracting surface using surface air pressure.")
+
+    else:
+        raise ValueError("Surface air pressure could not be found. Stopping.")
+
+    return var_cube
--- a/esmvalcore/preprocessor/__init__.py
+++ b/esmvalcore/preprocessor/__init__.py
@@ -90,6 +90,7 @@ from ._volume import (
     extract_transect,
     extract_volume,
     volume_statistics,
+    extract_surface_from_atm,
 )
 from ._weighting import weighting_landsea_fraction
 
@@ -122,6 +123,8 @@ __all__ = [
     "resample_time",
     # Level extraction
     "extract_levels",
+    # Extract surface
+    "extract_surface_from_atm",
     # Weighting
     "weighting_landsea_fraction",
     # Mask landsea (fx or Natural Earth)


jlenh:
@bouweandela if you could point me to where in the documentation I should add relevant information ? Thanks :slightly_smiling_face:

bouweandela:
> @bouweandela if you could point me to where in the documentation I should add relevant information 

In this file, please: https://github.com/ESMValGroup/ESMValCore/blob/main/doc/recipe/preprocessor.rst

You can find this, and much more useful information on contributing, under the links in the checklist above and in our contribution guidelines: https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html

esmvalcore/preprocessor/_volume.py
```diff
@@ -603,3 +604,115 @@ def extract_trajectory(
...
+
+        # Interpolation (not supported for dask arrays)
+        air_pressure_coord = var_cube.coord("air_pressure")
+        original_levels = iris.util.broadcast_to_shape(
+            air_pressure_coord.points,
+            var_cube.shape,
+            var_cube.coord_dims(air_pressure_coord),
+        )
+        target_levels = np.expand_dims(ps_cube.data, axis=z_axis)
+        sfc_data = stratify.interpolate(
```
bouweandela, line 687: Have you looked into using the [`esmvalcore.preprocessor.extract_levels`](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/api/esmvalcore.preprocessor.html#esmvalcore.preprocessor.extract_levels) function?

esmvalcore/preprocessor/_volume.py
```diff
@@ -603,3 +604,115 @@ def extract_trajectory(
     points = [("latitude", latitudes), ("longitude", longitudes)]
     interpolated_cube = interpolate(cube, points)  # Very slow!
     return interpolated_cube
+
+
+def _get_first_unmasked_data(array, axis):
```
bouweandela, line 609: Naively, I would expect the first unmasked data to represent the surface. Is that not correct? Or not as accurate as using the surface pressure and interpolating/extrapolating?

esmvalcore/preprocessor/_volume.py
```diff
@@ -603,3 +604,115 @@ def extract_trajectory(
...
+        ps_cube = cube.ancillary_variable("surface_air_pressure")
+    except iris.exceptions.AncillaryVariableNotFoundError:
+        logger.debug(
+            "Ancillary variable surface air pressure not found in cube. "
+            "Check file availability."
+        )
+    if ps_cube:
+        var_cube = cube.extract(
+            iris.Constraint(name=var)
+        )
```
bouweandela, line 660: Iris Cube objects only contain a single variable, so this isn't needed. You could just use `cube.var_name` to retrieve `var`.

bouweandela:
Here is an interesting example of doing something similar for ocean data: https://cosima-recipes.readthedocs.io/en/latest/Recipes/Extract_Variables_at_Ocean_Bottom.html

I guess the ocean equivalent of `ps` would be `pbo`.

esmvalcore/preprocessor/_volume.py
jlenh: Ah I see, I was wondering how to access the variable name without including it as an argument so using `cube.var_name` should solve that too!

esmvalcore/preprocessor/_volume.py
jlenh: Yes basically it would not be as accurate as interpolating to the actual surface pressure value from the model outputs.

esmvalcore/preprocessor/_volume.py
jlenh: I tried to make it work but couldn't manage to. To replace the interpolation step I tried with
```
sfc_data = extract_levels(
            cube=cube,
            levels=ps_cube.data,
            scheme="linear",
            coordinate="air_pressure",
            rtol=1e-7,
            atol=None
            )
```
but I always get the following error `TypeError: extract_levels() missing 1 required positional argument: 'data'` which I don't really understand where it comes from. In case you have any insights on that.

bouweandela: This seems to be caused by the `preserve_float_dtype` decorator around `extract_levels`. Using
```python
sfc_data = extract_levels(
    cube,
    levels=ps_cube.data,
    scheme="linear",
    coordinate="air_pressure",
    rtol=1e-7,
    atol=None
)
```
avoids this problem. @schlunma Since you introduced the wrapper, would you have time to take a look at improving it to avoid this puzzling error?

jlenh: I started again looking at the `extract_levels` function with the fix you mentioned but I cannot manage to make it work in this case. Even though the parameter `levels` is expected to be of type `np.typing.ArrayLike | da.Array`, I encounter the error `TypeError: unhashable type: 'numpy.ndarray'` at https://github.com/ESMValGroup/ESMValCore/blob/e3eb9ff8797c500dda33a4469fc38a5893994884/esmvalcore/preprocessor/_regrid.py#L1273 which stems from the `set(levels)` part. Could the problem stem from the pressure levels at which to interpolate that are both time- and space-dependent so that the expected levels array does not fit some criteria to be used in the tests included in `extract_levels`? 

In general, I am not entirely sure how to go forward. I would suggest that using the code as is can be the most straightforward solution as it was already reviewed and tested as part of https://github.com/ESMValGroup/ESMValCore/pull/594 and this PR just offers to make it usable for any atmospheric 3D variable using the surface pressure in an exactly similar way.

bouweandela: You could add a check there that levels is a 1D numpy array to avoid the error. I would highly recommend re-using the existing extract_levels preprocessor because it avoids re-implementing all the same features here. Note that it is important for scalability that all new preprocessors are lazy, see #674 for all the work we've been doing to achieve that.

jlenh: I see, I completely understand! I will adapt the code so that the new preprocessor builds on the `extract_levels` preprocessor :slightly_smiling_face:

jlenh:
@bouweandela I think I managed to update the code accordingly with the call to `extract_levels` alongside adding tests for the new functions and error handling.
I am not sure how to go for solving the conflicts for the merging though (relatively large changes for `add_ancillary_variable` in `_supplementary_vars.py`, only some import statements for `_volume.py`)

jlenh:
I solved the pre-commit error for the last commit but now I run into different errors pinned by `mypy`:
- Various `error: No overload variant of "zip" [call-overload]` in 4 different places (`esmvalcore/preprocessor/_supplementary_vars.py:151`, `esmvalcore/preprocessor/_time.py:689`, `esmvalcore/_recipe/from_datasets.py:225`, `esmvalcore/preprocessor/_volume.py:514`)
- Various `error: X | Y syntax for unions requires Python 3.10  [syntax]` in `esmvalcore/config/_dask.py:158/177/178`
If there are any ideas as to what caused these?

schlunma:
I think the second problem can be solved by adding 

```py
from __future__ import annotations
```

to the top of your file (see also [here](https://docs.python.org/3/library/__future__.html#id1)).

No idea about the first though..

jlenh:
@schlunma Only a Codacy error left `Call to untyped function "add_ancillary_variable" in typed context. (no-untyped-call)` which I am not sure how to solve because the function works in-place. Should we get rid of the `-> None` in the function definition?

schlunma:
Hmm...no idea what's going on there. The function has type hints, so it should be fine. You can probably just ignore it.

jlenh:
> Hmm...no idea what's going on there. The function has type hints, so it should be fine. You can probably just ignore it.

Then I think it's good to go on my side!

valeriupredoi:
> > Hmm...no idea what's going on there. The function has type hints, so it should be fine. You can probably just ignore it.
> 
> Then I think it's good to go on my side!

same mypy issue we saw yesterday in your PR, Manu

valeriupredoi (approved):
very nice, many thanks, folks! :beers:

esmvalcore/preprocessor/_regrid.py
```diff
@@ -1288,8 +1308,10 @@ def extract_levels(
         result = cube
         # Set the levels to the requested values
         src_levels.points = levels
-    elif len(src_levels.shape) == 1 and set(levels).issubset(
-        set(src_levels.points)
+    elif len(src_levels.shape) == 1 and (
+        set(levels).issubset(set(src_levels.points))
+        if len(levels.shape) == 1
+        else False
```
schlunma, line 1314:
```suggestion
    elif len(src_levels.shape) == 1 and 
        len(levels.shape) == 1 and
        set(levels).issubset(set(src_levels.points))
```

probably easier to read.

esmvalcore/preprocessor/_supplementary_vars.py
```diff
@@ -101,39 +101,72 @@ def add_cell_measure(
     )
 
 
-def add_ancillary_variable(cube: Cube, ancillary_cube: Cube) -> None:
+def add_ancillary_variable(
+    cube: Cube, ancillary_cube: Cube | iris.coords.AncillaryVariable
```
schlunma, line 105: Could you please do the same for the `add_cell_measure` method?

esmvalcore/preprocessor/_supplementary_vars.py
```diff
@@ -101,39 +101,72 @@ def add_cell_measure(
...
     """
-    coord_dims = tuple(range(cube.ndim - len(ancillary_cube.shape), cube.ndim))
-    ancillary_data = ancillary_cube.core_data()
+    try:
+        ancillary_var = iris.coords.AncillaryVariable(
+            ancillary_cube.core_data(),
+            standard_name=ancillary_cube.standard_name,
+            units=ancillary_cube.units,
+            var_name=ancillary_cube.var_name,
+            attributes=ancillary_cube.attributes,
```
schlunma, line 127:
```suggestion
            attributes=ancillary_cube.attributes,
            long_name=ancillary_cube.long_name,
```

esmvalcore/preprocessor/_supplementary_vars.py
```diff
@@ -101,39 +101,72 @@ def add_cell_measure(
...
+            f"Failed to add {ancillary_cube} to {cube} as ancillary var."
+            "ancillary_cube should be either an iris.cube.Cube or an "
+            "iris.coords.AncillaryVariable object."
+        )
+        raise ValueError(msg) from err
+    data_dims: list[int | None] = []
+    if isinstance(ancillary_cube, iris.coords.AncillaryVariable):
+        start_dim = cube.ndim - len(ancillary_var.shape)
+        data_dims = list(range(start_dim, cube.ndim))
+    else:
```
schlunma, line 140: Could you add a little comment here what this code does? E.g., match coordinates of ancillary cube to dimensions of data cube to determine dimensions of ancillary data.

esmvalcore/preprocessor/_supplementary_vars.py
```diff
@@ -101,39 +101,72 @@ def add_cell_measure(
...
         ancillary_data = ancillary_data.rechunk(cube_chunks)
-    ancillary_var = iris.coords.AncillaryVariable(
-        ancillary_data,
-        standard_name=ancillary_cube.standard_name,
-        units=ancillary_cube.units,
-        var_name=ancillary_cube.var_name,
-        attributes=ancillary_cube.attributes,
-    )
-    cube.add_ancillary_variable(ancillary_var, coord_dims)
+        ancillary_var.data = ancillary_data
```
schlunma, line 168:
```suggestion
        ancillary_var.data = ancillary_cube.lazy_data().rechunk(cube_chunks)
```

Simpler, you can also remove the lines above (cannot do this here somehow...)

esmvalcore/preprocessor/_volume.py
```diff
@@ -603,3 +605,85 @@ def extract_trajectory(
...
+        *[da.arange(array.shape[i]) for i in range(array.ndim) if i != axis],
+        indexing="ij",
+    )
+    indices = list(indices)
+    indices.insert(axis, indices_first_positive)
+    first_unmasked_data = np.array(array)[tuple(indices)]
+    return first_unmasked_data
+
+
+@register_supplementaries(variables=["ps"], required="require_at_least_one")
```
schlunma, line 625:
```suggestion
@register_supplementaries(variables=["ps"], required="require_at_least_one")
@preserve_float_dtype
```

Just to be safe

esmvalcore/preprocessor/_volume.py
```diff
@@ -603,3 +605,85 @@ def extract_trajectory(
...
+    #   - 3D atmo variable to extract at the surface
+    #   - surface_air_pressure (ps)
+    ps_cube = None
+    try:
+        ps_cube = cube.ancillary_variable("surface_air_pressure")
+    except iris.exceptions.AncillaryVariableNotFoundError:
+        logger.debug(
+            "Ancillary variable surface air pressure not found in cube. "
+            "Check file availability."
+        )
```
schlunma, line 651:
```suggestion
    try:
        ps_cube = cube.ancillary_variable("surface_air_pressure")
    except iris.exceptions.AncillaryVariableNotFoundError as exc:
        raise ValueError("Surface air pressure could not be found") from exc
```

esmvalcore/preprocessor/_volume.py
```diff
@@ -603,3 +605,85 @@ def extract_trajectory(
...
+        target_levels = da.expand_dims(ps_cube.data, axis=z_axis)
+        var_cube = extract_levels(
+            cube,
+            levels=target_levels,
+            scheme="linear_extrapolate",
+            coordinate="air_pressure",
+            rtol=1e-7,
+            atol=None,
+        )
+        var_cube.var_name = cube.var_name + "s"
```
schlunma, line 676:
```suggestion
        if cube.var_name is not None:
            var_cube.var_name = cube.var_name + "s"
```

tests/integration/preprocessor/_supplementary_vars/test_add_supplementary_variables.py
```diff
@@ -252,3 +252,31 @@ def test_remove_supplementary_vars(self):
...
+            dim_coords_and_dims=[
+                (self.depth, 0),
+                (self.lats, 1),
+                (self.lons, 2),
+            ],
+        )
+        # Ancillary var not an iris.cube.Cube or iris.coords.AncillaryVariable
+        with pytest.raises(ValueError) as err:
+            add_ancillary_variable(cube, np.ones(self.new_cube_3D_data.shape))
+        assert "ancillary_cube should be either an iris" in str(err)
```
schlunma, line 269:
```suggestion
        msg = "ancillary_cube should be either an iris"
        with pytest.raises(ValueError, match=msg) as err:
            add_ancillary_variable(cube, np.ones(self.new_cube_3D_data.shape))
```

tests/integration/preprocessor/_supplementary_vars/test_add_supplementary_variables.py
```diff
@@ -252,3 +252,31 @@ def test_remove_supplementary_vars(self):
...
+            standard_name="air_pressure",
+            bounds=[[0, 1], [1, 2], [2, 3]],
+            units="Pa",
+        )
+        ancillary_cube = iris.cube.Cube(
+            np.ones((3)), dim_coords_and_dims=[(plev_dim, 0)]
+        )
+        with pytest.raises(ValueError) as err:
+            add_ancillary_variable(cube, ancillary_cube)
+        assert "No coordinate associated with ancillary" in str(err)
```
schlunma, line 282: Use `match` (see above)

tests/unit/preprocessor/_volume/test_volume.py
```diff
@@ -776,6 +802,45 @@ def test_extract_trajectory(self):
...
+        ps_ancillary = iris.coords.AncillaryVariable(
+            np.broadcast_to([[[0.25]], [[0.5]], [[5.0]], [[27.5]]], (4, 2, 2)),
+            standard_name="surface_air_pressure",
+            units="Pa",
+            var_name="ps",
+        )
+        # Test without air pressure ancillary variable
+        with self.assertRaises(ValueError) as err:
+            extract_surface_from_atm(self.grid_4d_ps)
+        assert "Surface air pressure could not be found" in str(err.exception)
```
schlunma, line 828: Use `match`

schlunma (commented):
Thanks Julien, looks great already! I just have a couple of minor comments. Cheers!

doc/recipe/preprocessor.rst
```diff
@@ -2118,7 +2119,7 @@ See also :func:`esmvalcore.preprocessor.extract_location`.
 
 ``zonal_statistics``
 --------------------
-
+preprocessor.rst:315: WARNING: u
```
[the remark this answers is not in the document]
jlenh: Ouh yes my bad!

esmvalcore/preprocessor/_supplementary_vars.py
jlenh: In this case would it be a `iris.coords.CellMeasure` object then?

tests/unit/preprocessor/_volume/test_volume.py
jlenh: In this case it is another type of object method so I adapted it to fit the other tests in the class using `self.assertEqual(msg, str(err.exception))`

esmvalcore/preprocessor/_regrid.py
jlenh: I wrote it like that because in the case of `levels` being an `np.array` of dimension > 1, the `set(levels)` call will return an error (`TypeError: unhashable type: 'numpy.ndarray'`). The inline `if ... else` statement allowed to prevent that.
However, to keep it clean I will commit your suggestion but change it to `set(levels.flatten())` to avoid any dimension problem.

esmvalcore/preprocessor/_supplementary_vars.py
schlunma: Yes!

doc/recipe/preprocessor.rst
```diff
@@ -2406,6 +2407,20 @@ Note that this function uses the expensive ``interpolate`` method from
 See also :func:`esmvalcore.preprocessor.extract_trajectory`.
 
 
+``extract_surface_from_atm``
```
schlunma, line 2410:
```suggestion
.. _extract_surface_from_atm:

``extract_surface_from_atm``
```

This should fix the doc build

esmvalcore/preprocessor/_regrid.py
schlunma: Actually that should work without the `flatten`. If one of the statements evaluates to False in this `and` cascade, the remaining ones are not checked. Thus, the `set(levels).issubset(set(src_levels.points))` will only be run if `len(levels.shape) == 1` is `True`. But that doesn't really matter, you can leave it like that!

esmvalcore/preprocessor/_volume.py
```diff
@@ -603,3 +605,80 @@ def extract_trajectory(
...
+    numerical_mask = da.where(mask, -1.0, 1.0)
+    indices_first_positive = da.argmax(numerical_mask, axis=axis)
+    indices = da.meshgrid(
+        *[da.arange(array.shape[i]) for i in range(array.ndim) if i != axis],
+        indexing="ij",
+    )
+    indices = list(indices)
+    indices.insert(axis, indices_first_positive)
+    first_unmasked_data = np.array(array)[tuple(indices)]
+    return first_unmasked_data
```
schlunma, line 622:
```suggestion
def _get_first_unmasked_data(
    array: np.ndarray | da.Array,
    axis: int,
) -> np.ndarray | da.Array:
    """Get first unmasked value of an array along an axis."""
    npx = get_array_module(array)
    mask = npx.ma.getmaskarray(array)
    numerical_mask = npx.where(mask, -1.0, 1.0)
    indices_first_positive = npx.argmax(numerical_mask, axis=axis)
    indices = npx.meshgrid(
        *[npx.arange(array.shape[i]) for i in range(array.ndim) if i != axis],
        indexing="ij",
    )
    indices = list(indices)
    indices.insert(axis, indices_first_positive)
    first_unmasked_data = npx.array(array)[tuple(indices)]
    return first_unmasked_data
```

To make it work with numpy and dask arrays. This needs `from esmvalcore.preprocessor._shared import get_array_module`.

esmvalcore/preprocessor/_volume.py
```diff
@@ -603,3 +605,80 @@ def extract_trajectory(
...
+    mask = da.ma.getmaskarray(cube.core_data())
+    if mask.any():
+        first_unmasked_data = _get_first_unmasked_data(
+            cube.core_data(), axis=z_axis
+        )
+        dim_map = [dim for dim in range(cube.ndim) if dim != z_axis]
+        first_unmasked_data = iris.util.broadcast_to_shape(
+            first_unmasked_data, cube.shape, dim_map
+        )
+        cube.data = da.where(mask, first_unmasked_data, cube.core_data())
```
schlunma, line 661:
```suggestion
    npx = get_array_module(array)
    if iris.util.is_masked(cube.core_data()):
        mask = npx.ma.getmaskarray(cube.core_data())
        first_unmasked_data = _get_first_unmasked_data(
            cube.core_data(), axis=z_axis
        )
        dim_map = [dim for dim in range(cube.ndim) if dim != z_axis]
        first_unmasked_data = iris.util.broadcast_to_shape(
            first_unmasked_data, cube.shape, dim_map
        )
        cube.data = npx.where(mask, first_unmasked_data, cube.core_data())
```

To make it work with numpy and dask arrays. This needs `import iris.util`.

esmvalcore/preprocessor/_volume.py
```diff
@@ -603,3 +605,80 @@ def extract_trajectory(
...
+
+    # Interpolation
+    target_levels = da.expand_dims(ps_cube.data, axis=z_axis)
```
schlunma, line 664:
```suggestion
    target_levels = npx.expand_dims(ps_cube.core_data(), axis=z_axis)
```

schlunma (commented):
I just realized that it would be nice if the preprocessor worked well with numpy and dask arrays. Should be very straightforward to achieve that (see my suggestions). Thanks 😊

esmvalcore/preprocessor/_regrid.py
jlenh: I see, I only had the issue because of the previous order of the statements then.

jlenh:
> I just realized that it would be nice if the preprocessor worked well with numpy and dask arrays. Should be very straightforward to achieve that (see my suggestions). Thanks 😊

I get an error now which seems to be due to sub-sampling/slicing of a dask array not supporting doing it with dask array indices `dask/array/slicing.py:944: NotImplementedError` arising when running the CO2 derivation test.
```
esmvalcore/preprocessor/_volume.py:634: in _get_first_unmasked_data
    first_unmasked_data = npx.array(array)[tuple(indices)]
lib/python3.12/site-packages/dask/array/core.py:1998: in __getitem__
    self, index2 = slice_with_int_dask_array(self, index2)
```
Any idea on that? I am not much of a dask expert..

esmvalcore/preprocessor/_volume.py
```diff
@@ -603,3 +607,85 @@ def extract_trajectory(
...
+    numerical_mask = npx.where(mask, -1.0, 1.0)
+    indices_first_positive = npx.argmax(numerical_mask, axis=axis)
+    indices = npx.meshgrid(
+        *[npx.arange(array.shape[i]) for i in range(array.ndim) if i != axis],
+        indexing="ij",
+    )
+    indices = list(indices)
+    indices.insert(axis, indices_first_positive)
+    first_unmasked_data = npx.array(array)[tuple(indices)]
+    return first_unmasked_data
```
schlunma, line 628:
```suggestion
    """Get first unmasked value of an array along an axis.
    
    Note: this uses fancy indexing, which is not supported by Dask (yet).
    
    """
    npx = get_array_module(array)
    
    # Use identity indices for axes != axis
    indices = npx.meshgrid(
        *[npx.arange(array.shape[i]) for i in range(array.ndim) if i != axis],
        indexing="ij",
    )
    indices = list(indices)
    
    # Use index of first unmasked data for selected axis
    mask = npx.ma.getmaskarray(array)
    numerical_mask = npx.where(mask, -1.0, 1.0)
    indices_first_positive = npx.argmax(numerical_mask, axis=axis)
    indices.insert(axis, indices_first_positive)
    
    # Compute Dask arrays to enable fancy indexing
    indices, np_array = dask.compute(indices, array)
    first_unmasked_data = np_array[tuple(indices)]
    
    # Ensure that new array uses same chunks as original array
    if isinstance(array, da.Array):
        chunks = tuple(array.chunks[i] for i in range(array.ndim) if i != axis)
        first_unmasked_data = da.array(first_unmasked_data).rechunk(chunks)

    return first_unmasked_data
```

esmvalcore/preprocessor/_volume.py
```diff
@@ -603,3 +607,85 @@ def extract_trajectory(
...
+    npx = get_array_module(cube.core_data())
+    if iris.util.is_masked(cube.core_data()):
+        mask = npx.ma.getmaskarray(cube.core_data())
+        first_unmasked_data = _get_first_unmasked_data(
+            cube.core_data(), axis=z_axis
+        )
+        dim_map = [dim for dim in range(cube.ndim) if dim != z_axis]
+        first_unmasked_data = iris.util.broadcast_to_shape(
+            first_unmasked_data, cube.shape, dim_map
+        )
```
schlunma, line 667:
```suggestion
        first_unmasked_data = iris.util.broadcast_to_shape(
            first_unmasked_data,
            cube.shape,
            dim_map,
            chunks=cube.lazy_data().chunks if cube.has_lazy_data() else None,
        )
```

esmvalcore/preprocessor/_volume.py
```diff
@@ -11,13 +11,15 @@
 
 import dask.array as da
```
schlunma, line 12:
```suggestion
import dask
import dask.array as da
```

schlunma (commented):
Ah, damn, completely forgot about that! I made some suggestion which should be able to fix that, hopefully there's no type.

If it doesn't work, I can directly push to your branch if you like!

Sorry for all this extra work, I though this would be easier 😩

schlunma:
Fantastic! One `pre-commit run --all` and I think this is ready to go!

esmvalcore/preprocessor/_regrid.py
```diff
@@ -1068,7 +1069,10 @@ def _create_cube(src_cube, data, src_levels, levels):
         result.add_dim_coord(coord, z_dim)
     except ValueError:
         coord = iris.coords.AuxCoord(levels, **kwargs)
-        result.add_aux_coord(coord, z_dim)
+        result.add_aux_coord(
+            coord,
+            z_dim if len(levels.shape) == 1 else np.arange(len(coord.shape)),
```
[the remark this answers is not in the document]
jlenh: Coming back to that, I think I added that for the case of the new preprocessor where the `levels` array is basically a surface pressure cube that covers all dimensions (see `_volume.py:692`), hence the `np.arange(len(coord.shape))` I added. If you think it is right then I will close this comment!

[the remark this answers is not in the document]
schlunma: Yes, we're good! If the levels are not 1D, they have to [have the same shape as the source data](https://github.com/SciTools/python-stratify/blob/b57e45ddf85a87773441885dc190d55157a2bf2b/src/stratify/_vinterp.pyx#L470-L476) along all axes apart from the Z dimensions for interpolation 👍

schlunma (approved):
Looks great, thanks so much @jlenh !!

PR #2684, opened 2025-03-05
test codacy ruff

valeriupredoi:
Test PR where we can test all the bits and bobs related to the new codacy-ruff integration and fix things all along. Ruff is now integrated and configured in our Codacy, but there may be issues.

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -12,8 +12,10 @@ engines:
         enabled: true
     duplication:
         enabled: true
+    # introduced 5 March 2025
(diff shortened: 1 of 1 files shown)


valeriupredoi:
first hurdle - it runs Ruff inside Codacy static analysis! And it picks up the correct configuration file too :partying_face:

bouweandela:
Nice! Could you try introducing some issues so we can see if they're picked up? e.g. add some random spaces somewhere

Issue #2695, opened 2025-03-26
ICON-XPP CMORizer: Add land variables


schlunma:
### Variables

- tasmin
- tasmax
- tsl
- mrsos
- sm
- gpp
- cVeg
- cSoil
- baresoilFrac
- grassFrac
- treeFrac
- schrubFrac
- cropFrac
- nbp
- lai

See also https://nc.uni-bremen.de/index.php/s/d9DYD5o4ndYqWoL

bettina-gier:
Added LAI and GPP. Since it's an untuned run, just comparing general behaviour and order of magnitude to see if it's fine!

LAI:
![Image](https://github.com/user-attachments/assets/961d160e-54fe-4d35-a49e-06da8c8e2e07)


GPP:
![Image](https://github.com/user-attachments/assets/38b82da3-cc69-4a61-8f32-7d24d5679314)

mabouman:
> should the timeseries be globally or for certain locations/regions/latituderanges?

As of now globally, but I would potentially also want to look at some regions later on. Probably defined by climate regions and/or Whittaker biomes.

PR #2700, opened 2025-04-02
[NOT FOR CHANGELOG; PR NOT IN MAIN] adding lai and gpp to icon xpp

bettina-gier:
Adding LAI and GPP to variables for ICON XPP. 

More details see #2695


schlunma (approved):
Thanks Tina, looks great! 🚀

PR #2728, opened 2025-05-12
Use `netCDF4.Dataset` to read start and end date from files

schlunma:
## Description

If start and end dates cannot be read from a file's name, we currently use [Iris to load the file and determine the dates from the time coordinate](https://github.com/ESMValGroup/ESMValCore/blob/0533aa1fcf9ba7610870ce6fe28627d0e19dc258/esmvalcore/local.py#L132-L154). This is particular relevant for native model output where the date in the filename can have arbitrary formats.

This native model output (in my particular case data from ICON and EMAC) usually contains lots of variables (sometimes hundreds) in a single netCDF file, and also consists of hundreds of netCDF files per dataset (sometimes one file only contains a single time step). Since reading netCDF files with many variables can be [very slow with Iris](https://github.com/SciTools/iris/issues/6223), this can lead to enormous run times (I've see logs where ESMValTool spent hours to read the dates from the files).

To make this process MUCH faster, this PR here uses `netCDF4.Dataset` to extract the dates from the files. Since those files are loaded just once for this one particular purpose, we don't need any fancy features of Iris but can simply rely on netCDF4. Here's an overview of the different load times (the file, which is an actual EMAC output file, contains 855 variables):

```py
path = "/path/to/real_EMAC_file_with_855_variables.nc"

%%timeit
iris.load(path)
# 1min 47s ± 446 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
xr.open_dataset(path, chunks="auto")
# 277 ms ± 8.14 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
netCDF4.Dataset(path, mode="r")
# 5.2 ms ± 23.7 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

As you can see, using netCDF4.Dataset instead of Iris is over 21000 times faster!

***

***

### DIFF
--- a/esmvalcore/local.py
+++ b/esmvalcore/local.py
@@ -13,7 +13,7 @@ from typing import TYPE_CHECKING, Any, Union
 
 import isodate
 from cf_units import Unit
-from netCDF4 import Dataset
+from netCDF4 import Dataset, Variable
 
 from .config import CFG
 from .config._config import get_project_config
@@ -64,6 +64,14 @@ def _get_from_pattern(pattern, date_range_pattern, stem, group):
                 start_point = end_point = end.group(group)
 
     return start_point, end_point
+
+
+def _get_var_name(variable: Variable) -> str:
+    """Get variable name (following Iris' Cube.name())."""
+    for attr in ("standard_name", "long_name"):
+        if attr in variable.ncattrs():
+            return variable.getncattr(attr)
+    return variable.name
 
 
 def _get_start_end_date(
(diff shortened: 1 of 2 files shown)


esmvalcore/local.py
```diff
@@ -136,22 +137,21 @@ def _get_start_end_date(
...
-
-        for cube in cubes:
-            logger.debug(cube)
-            try:
-                time = cube.coord("time")
-            except iris.exceptions.CoordinateNotFoundError:
-                continue
+        dataset = Dataset(file)
+        if (
+            "time" in dataset.variables
```
bouweandela, line 142: This may not be as comprehensive as `iris.cube.Cube.coord("time")`. Looking through the Iris code a bit, I suspect that function checks also if one of several variable attributes matches "time": https://github.com/SciTools/iris/blob/1285efbea1d852d6d902abd8a9a1cc55272684c8/lib/iris/common/metadata.py#L1191-L1211

bouweandela:
Out of curiosity, which version of the libraries (especially Iris) did you use? There have been various improvements in the speed with which Iris loads files with many variables.

I would also be quite interested in seeing the timings for `iris.load_raw`, [which we use in ESMValCore](https://github.com/ESMValGroup/ESMValCore/blob/0533aa1fcf9ba7610870ce6fe28627d0e19dc258/esmvalcore/preprocessor/_io.py#L150), as that can be considerably faster than `iris.load` as it does not attempt to merge the cubes it is loading.

From my experience with making `iris.load` faster on files with many variables, I suspect there is something not very optimized in the Iris merge logic that causes merge to do more work than necessary on things that obviously cannot be merged, but did not spend time on it since we do not use it.

schlunma:
Sorry, only included this sentence in https://github.com/ESMValGroup/ESMValCore/pull/2579:

> This uses the most up-to-date versions of all these packages (fresh install from 2025-05-09).

In particular, this means:

- Iris: 3.12.1 (not the most up-to-date, but it seems that 3.12.2 only adds support for the latest Dask version)
- Xarray: 2025.4.0
- netCDF4: 1.7.2

Here is the run time for `iris.load_raw`:

```py
%%timeit
iris.load_raw(path)
# 1.98 s ± 7.38 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

This is much faster than `iris.load`, but still a lot (~400 times) slower than `netCDF4.Dataset`.

valeriupredoi:
@schlunma be careful how you time things: `netcdf4.Dataset()` just creates a file handler which is very very cheap (and fast), but once you start calling attributes of it, then things get slower since it needs to start loading all manners of metadata. I am, however, completely against iris loading the *entire* file just to extract a single variable :beer:

valeriupredoi (approved):

schlunma:
Here is the example file on Levante:

```
/work/bd0854/b309141/example_with_855_vars.nc
```

esmvalcore/local.py
schlunma: Good point, will change that!

PR #2736, opened 2025-05-20
Allow forcing the usage of new configuration system with environment variable

schlunma:
## Description

This PR makes it possible to force using our [new configuration system](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/quickstart/configure.html#configuration) via the environment variable `ESMVALTOOL_USE_NEW_CONFIG`.

This is important feature for "encapsulated" ESMValTool runs (e.g., when used within the REF) that should work independently from the user's personal ESMValTool configuration. 

***

***

### DIFF
--- a/esmvalcore/_main.py
+++ b/esmvalcore/_main.py
@@ -399,7 +399,9 @@ class ESMValTool:
         # https://github.com/ESMValGroup/ESMValCore/issues/2280), but no error
         # has been raised if the file does not exist. Thus, reload the file
         # here with `load_from_file` to make sure a proper error is raised.
-        if "config_file" in kwargs:
+        if "config_file" in kwargs and not os.environ.get(
+            "ESMVALTOOL_USE_NEW_CONFIG"
+        ):
             cli_config_dir = kwargs["config_file"]
             CFG.load_from_file(kwargs["config_file"])
 
--- a/esmvalcore/config/_config_object.py
+++ b/esmvalcore/config/_config_object.py
@@ -388,16 +388,20 @@ class Config(ValidatedConfig):
         # TODO: remove in v2.14.0
         self.clear()
         _deprecated_config_user_path = Config._get_config_user_path()
-        if _deprecated_config_user_path.is_file():
+        if _deprecated_config_user_path.is_file() and not os.environ.get(
+            "ESMVALTOOL_USE_NEW_CONFIG"
+        ):
             deprecation_msg = (
                 f"Usage of the single configuration file "
                 f"~/.esmvaltool/config-user.yml or specifying it via CLI "
                 f"argument `--config_file` has been deprecated in ESMValCore "
                 f"version 2.12.0 and is scheduled for removal in version "
-                f"2.14.0. Please run `mkdir -p ~/.config/esmvaltool && mv "
+                f"2.14.0. To switch to the new configuration system, (1) run "
+                f"`mkdir -p ~/.config/esmvaltool && mv "
                 f"{_deprecated_config_user_path} ~/.config/esmvaltool` (or "
                 f"alternatively use a custom `--config_dir`) and omit "
-                f"`--config_file`."
+                f"`--config_file`, or (2) set the environment variable "
+                f"ESMVALTOOL_USE_NEW_CONFIG=1."
             )
             warnings.warn(
                 deprecation_msg, ESMValCoreDeprecationWarning, stacklevel=2
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -9,8 +9,9 @@ from esmvalcore.config._config_object import _get_global_config
 
 
 @pytest.fixture
-def cfg_default():
+def cfg_default(monkeypatch):
     """Create a configuration object with default values."""
+    monkeypatch.setenv("ESMVALTOOL_USE_NEW_CONFIG", "1")
     cfg = _get_global_config()
     cfg.load_from_dirs([])
     return cfg
--- a/tests/unit/config/test_config_object.py
+++ b/tests/unit/config/test_config_object.py
@@ -1,4 +1,5 @@
 import os
+import warnings
 from collections.abc import MutableMapping
 from pathlib import Path
 from textwrap import dedent
@@ -395,8 +396,30 @@ def test_get_user_config_dir_with_env_fail(tmp_path, monkeypatch):
 
 
 # TODO: remove in v2.14.0
-def test_get_global_config_deprecated(mocker, tmp_path):
+def test_get_global_config_force_new_config(mocker, tmp_path, monkeypatch):
     """Test ``_get_global_config``."""
+    monkeypatch.setenv("ESMVALTOOL_USE_NEW_CONFIG", "1")
+
+    # Create invalid old config file to ensure that this is not used
+    config_file = tmp_path / "old_config_user.yml"
+    config_file.write_text("invalid_option: /new/output/dir")
+    mocker.patch.object(
+        esmvalcore.config._config_object.Config,
+        "_get_config_user_path",
+        return_value=config_file,
+    )
+
+    # No deprecation message should be raised
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
+        esmvalcore.config._config_object._get_global_config()
+
+
+# TODO: remove in v2.14.0
+def test_get_global_config_deprecated(mocker, tmp_path, monkeypatch):
+    """Test ``_get_global_config``."""
+    monkeypatch.delenv("ESMVALTOOL_USE_NEW_CONFIG", raising=False)
+
     config_file = tmp_path / "old_config_user.yml"
     config_file.write_text("output_dir: /new/output/dir")
     mocker.patch.object(
@@ -472,9 +495,7 @@ def _setup_config_dirs(tmp_path):
         ),
     ],
 )
-def test_load_from_dirs_always_default(
-    dirs, output_file_type, rootpath, tmp_path
-):
+def test_load_from_dirs(dirs, output_file_type, rootpath, tmp_path):
     """Test `Config.load_from_dirs`."""
     _setup_config_dirs(tmp_path)
 


ehogan (changes requested):
Would it be possible to update the documentation at https://docs.esmvaltool.org/projects/ESMValCore/en/latest/quickstart/configure.html#yaml-files to be very explicit that:

- Using `--config_dir` does not override the configuration in the default location (`~/.config/esmvaltool`)
- Configuration files in `~/.esmvaltool` are also included

please? 😊

ehogan:
> > Using `--config_dir` does not override the configuration in the default location (`~/.config/esmvaltool`)
> 
> Would you have a suggestion for that?

When I was reading before, I didn't quite make it to the end of this sentence:

> ESMValCore searches for all YAML files within each of these directories and **merges them together** [...]

Would it be possible to add this to the first sentence in the [Specify configuration for `esmvaltool` command line tool](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/quickstart/configure.html#specify-configuration-for-esmvaltool-command-line-tool) section, please? For example:

> When running recipes via the [command line](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/quickstart/run.html#running), configuration options can be specified via YAML files and command line arguments, which are merged together to create a single configuration object.

Also, I'm wondering if it would be clearer to moving the "but this can be changed with the `ESMVALTOOL_CONFIG_DIR` environment variable." when listing the location of the configuration files. Especially now that this option completely overrides all configuration rather than merges.

Also, the documentation states:

> Preference follows the order in the list above (i.e., the directory specified via command line argument is preferred over the user configuration directory). 

I don't believe this is true. In the RTW we had provided a configuration file via `--config_dir`, but the configuration in my home area took precedence.

> > Configuration files in `~/.esmvaltool` are also included
> 
> Not all files in there are included, only `~/.esmvaltool/config-user.yml` (if present). The only reason for this is backwards-compatibility; we will remove support for this in v2.14.0. You'll also get lots of deprecation warnings if you use that. 

Ah, I can confirm that removing the `~/.esmvaltool/config-user.yml` file does mean all the deprecation warnings disappear.

> So we certainly don't want to advertise this "feature". In my opinion, our documentation should not mention that, but we should rather rely on the warnings.

I understand not wanting to advertise this, but I feel it may cause issues if people don't realise that the old configuration is merged with the new configuration (the RTW is now failing because of this!) 😭 How would you feel about adding another "Warning" admonition (or extending the current one) to explain this? It would be less of an advertisement if it's shown as a warning! 😊

schlunma:
I tried to improve this in https://github.com/ESMValGroup/ESMValCore/pull/2736/commits/988539377bff719428d0b40b0f8209bc5229628e, could you please have a look? 😊 [Here](https://esmvaltool--2736.org.readthedocs.build/projects/ESMValCore/en/2736/quickstart/configure.html#specify-configuration-for-esmvaltool-command-line-tool) is the rendered version.

> Also, I'm wondering if it would be clearer to moving the "but this can be changed with the ESMVALTOOL_CONFIG_DIR environment variable." when listing the location of the configuration files. Especially now that this option completely overrides all configuration rather than merges.

No, this will NOT override all configuration. Setting this variable just changes the location where ESMValCore looks for the "user configuration". By default this is `~/.config/esmvaltool/`. If `ESMVALTOOL_CONFIG_DIR` is set, ESMValCore [changes this location to the value of the env variable](https://github.com/ESMValGroup/ESMValCore/blob/82d795ecbd295de740f27e134d64e24e278cd562/esmvalcore/config/_config_object.py#L63), but will still use the exact same procedure to determine the final configuration object (i.e., it will merge the content of all YAML files into one object).

The behavior that changes with this PR is that the old configuration files are completely ignored if `ESMVALTOOL_CONFIG_DIR` is set. Before, this was not the case (see below).

> I don't believe this is true. In the RTW we had provided a configuration file via --config_dir, but the configuration in my home area took precedence.

I would certainly hope this is true, otherwise our tests would not be set up correctly 😬 Could you please post your configuration files and the problem you're having? Sometimes the merge can be a bit counterintuitive. For example, I had an issue where I specified a CMIP6 `rootpath` in my user config file, and only a default `rootpath` in the config file given by `--config-dir`. Due to the nested update, both keys ended up in the final config, and the CMIP6 entry is always preferred over the default entry. That way the option in the user config file "took precedence" over the `--config-dir` entry.

> I understand not wanting to advertise this, but I feel it may cause issues if people don't realise that the old configuration is merged with the new configuration (the RTW is now failing because of this!) 😭 How would you feel about adding another "Warning" admonition (or extending the current one) to explain this? It would be less of an advertisement if it's shown as a warning! 😊

If an old configuration file is found (i.e., either the file `~/.esmvaltool/config-user.yml` exists or the command line argument `--config-file` is given), this file will be read just like before our configuration overhaul to be fully backwards-compatible. There is NO merging going on in this case. All potentially available new configuration files will be IGNORED.

With this PR, setting `ESMVALTOOL_CONFIG_DIR` disables this search for the old configuration and ESMValCore will always use the new system, regardless if an old configuration is present or not.

schlunma:
> Just to check my understanding, if I don't provide anything via the `--config_dir` command line argument, then set `ESMVALTOOL_CONFIG_DIR`, it will only look in this directory for configuration files (and merges any it finds in this directory). Correct? 🤞

Yes!

> OK, so if I'm finally catching up now, this is what my example above is demonstrating: if an `~/.esmvaltool/config-user.yml` file exists, all new configuration options (e.g. `--config_dir`) will be ignored. If this is the case then I really do feel this should be documented! 😊

I added some text in https://github.com/ESMValGroup/ESMValCore/pull/2736/commits/a4aa51ac6444b58f49370a5ad1123404bbf586b3, is that okay for you? Rendered text [here](https://esmvaltool--2736.org.readthedocs.build/projects/ESMValCore/en/2736/quickstart/configure.html#yaml-files).

ehogan:
> > Just to check my understanding, if I don't provide anything via the `--config_dir` command line argument, then set `ESMVALTOOL_CONFIG_DIR`, it will only look in this directory for configuration files (and merges any it finds in this directory). Correct? 🤞
> 
> Yes!
> 
> > OK, so if I'm finally catching up now, this is what my example above is demonstrating: if an `~/.esmvaltool/config-user.yml` file exists, all new configuration options (e.g. `--config_dir`) will be ignored. If this is the case then I really do feel this should be documented! 😊
> 
> I added some text in [a4aa51a](https://github.com/ESMValGroup/ESMValCore/commit/a4aa51ac6444b58f49370a5ad1123404bbf586b3), is that okay for you? Rendered text [here](https://esmvaltool--2736.org.readthedocs.build/projects/ESMValCore/en/2736/quickstart/configure.html#yaml-files).

This is perfect, thank you so much @schlunma 🤩

ehogan:
I then tried:

```
% mkdir test_config_2
% cat > test_config_2/config-user.yml << EOF
max_parallel_tasks: 6
EOF
% echo $ESMVALTOOL_CONFIG_DIR
test_config
% esmvaltool run recipe_radiation_budget.yml --config_dir=test_config_2
```
which returned:
```
2025-06-16 15:47:24,048 UTC [253707] INFO    Reading configuration files from:
~/software/ESMValTool/ESMValCore/esmvalcore/config/configurations/defaults (defaults)
~/test_config (ESMVALTOOL_CONFIG_DIR environment variable)
~/test_config_2 (command line argument)
[...]
2025-06-16 15:47:24,957 UTC [253707] INFO    Running tasks using at most 6 processes
```
So, the configuration file specified via `--config_dir` was used preferentially to the configuration file specified by `ESMVALTOOL_CONFIG_DIR`. The documentation states "Preference follows the order in the list above", which is the the `--config_dir` command line argument, then the user configuration directory / `ESMVALTOOL_CONFIG_DIR`. So all good! 🥳

ehogan:
There are no `ESMValCoreDeprecationWarning` when I use the `ESMVALTOOL_CONFIG_DIR` environment variable 👍

tests/integration/test_main.py
```diff
@@ -66,13 +66,52 @@ def test_run():
...
+        with pytest.raises(RecipeError, match=msg):
+            ESMValTool().run(recipe_file, config_dir=config_dir)
+    log_file = os.path.join(
+        log_dir,
+        os.listdir(log_dir)[0],
+        "run",
+        "main_log.txt",
     )
+    filled_recipe = os.path.exists(
+        log_dir + "/" + os.listdir(log_dir)[0] + "/run/recipe_filled.yml",
```
ehogan, line 104: Would it be worth using `os.path.join` here too?

tests/integration/test_main.py
```diff
@@ -94,9 +133,71 @@ def test_empty_run(tmp_path, monkeypatch):
...
+        with pytest.warns(ESMValCoreDeprecationWarning, match=warn_msg):
+            ESMValTool().run(recipe_file, config_file=config_file)
+    log_file = os.path.join(
+        log_dir,
+        os.listdir(log_dir)[0],
+        "run",
+        "main_log.txt",
+    )
+    filled_recipe = os.path.exists(
+        log_dir + "/" + os.listdir(log_dir)[0] + "/run/recipe_filled.yml",
```
ehogan, line 148: Would it be worth using `os.path.join` here too?

ehogan (approved):
Many thanks for your hard work on this @schlunma! 🎉 

Two very minor comments, which you are welcome to ignore 😊

tests/integration/test_main.py
schlunma: Good point! I think the logic in that test was also flawed (it called `shutil.rmtree` and then asserts that a file does not exist within the removed dir?). Fixed that in https://github.com/ESMValGroup/ESMValCore/pull/2736/commits/078b8b303a507397afb46f90d212d8665529520f

tests/integration/test_main.py
schlunma: Good point! I think the logic in that test was also flawed (it called `shutil.rmtree` and then asserts that a file does not exist within the removed dir?). Fixed that in https://github.com/ESMValGroup/ESMValCore/pull/2736/commits/078b8b303a507397afb46f90d212d8665529520f

PR #2767, opened 2025-07-04
Avoid reading configuration from disk for every single test

bouweandela:
## Description

Avoid reading configuration from disk for every single test, which was introduced in #2736, and caused a slowdown in the tests.

***

***

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -44,7 +44,7 @@ def cfg_default():
 @pytest.fixture(autouse=True)
 def ignore_existing_user_config(monkeypatch, cfg_default):
     """Ignore user's configuration when running tests."""
-    monkeypatch.setattr(CFG, "_mapping", cfg_default)
+    monkeypatch.setattr(CFG, "_mapping", cfg_default._mapping)
 
 
 @pytest.fixture
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -44,7 +44,10 @@ def cfg_default():
 @pytest.fixture(autouse=True)
 def ignore_existing_user_config(monkeypatch, cfg_default):
     """Ignore user's configuration when running tests."""
-    monkeypatch.setattr(esmvalcore.config, "CFG", cfg_default)
+    for key in CFG:
+        monkeypatch.delitem(CFG, key)
+    for key, value in cfg_default.items():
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -33,20 +34,22 @@ def cfg_default():
     return cfg
 
 
+@pytest.fixture
```
schlunma, line 37:
```suggestion
@pytest.fixture(scope="session")
```

Have you tried this? If that works, it's simpler than adding `lru_cache`

bouweandela: Yes, it doesn't work because apparently `monkeypatch` is function scoped and therefore cannot be combined with a session scoped fixture

valeriupredoi:
let's turn on GHA and see how the times are for all de Pythons we have :snake:

tests/conftest.py
bouweandela: Oh wait, I tried adding `scope="session"` to the `ignore_existing_user_config` fixture, not this one. Adding it here would solve the problem of reloading the data, but then there is the risk of modifying. I can try using a mappingproxy instead of copying though

schlunma: As far as I understand, the output of the fixture is cached when a wider scope is set, and then this output is reused. I don't think that this output can be modified. I might be wrong though..

bouweandela: I forgot that CFG is not a plain dict but a custom object, so the mappingproxy thing doesn't work. The output of the fixture is an `esmvalcore.config.Config` object, which can be modified, so reusing it is not safe because then one test could break another test depending on the order in which you run them.

bouweandela:
I restored the previous code in [324fc2e](https://github.com/ESMValGroup/ESMValCore/pull/2767/commits/324fc2e41bcf1a3d6b034396b0f307f65fd26043), does that help? That does make the tests run twice as slow..

I can get back to this on Tuesday at the earliest, so feel free to implement something better if you have time before that.

tests/conftest.py
schlunma: Good point, agreed.

schlunma (approved):
Works fine now with existing config files and timing looks great as well, thanks so much Bouwe! 🚀

PR #2794, opened 2025-08-07
Pass `chunks={}` to Xarray dataset loader for Zarr stores

### LINKED ISSUE #2790: Optimizations for Zarr loading and processing in ESMValTool
Since https://github.com/ESMValGroup/ESMValCore/pull/2785 we now have backend support for Zarr loading in ESMValTool ie one can simply:

```python
    from esmvalcore.preprocessor._io import load


    zarr_path = (
        "https://uor-aces-o.s3-ext.jc.rl.ac.uk/"
        "esmvaltool-zarr/example_field_0.zarr3"
    )
    cubes = load(zarr_path)
```

and they get Iris cubes.

Front facing (ie passing instructions via recipe) will be in conjunction with #2765 - we want to load these files based on what catalogs they appear in etc. But before then, there are number of investigations still to be done:

- optimal memory handling: it appears that `ncdata` needs quite a bit of memory, when all things should be lazy, see https://github.com/pp-mo/ncdata/issues/139 - we are working with Patrick on that one;
- possible issues with locking, see https://github.com/pp-mo/ncdata/issues/138 - again, we are orking with Patrick and Martin on this one as well;
- it'd also be good for folks to start having a crack at their favourite Zarr data and see if they can spot any optimization issues

🍻

valeriupredoi:
## Description

Closes #2790 

There was quite a bit of investigation down about this in https://github.com/pp-mo/ncdata/issues/139 and in https://github.com/pydata/xarray/issues/10612 and in my own GH repo at  https://github.com/valeriupredoi/esmvaltool_zarr_tests and, with help from @pp-mo and @kmuehlbauer we managed to get the performance of Zarr loading go up by a big notch though keeping all data lazy, both for remote bucket and local Zarr stores. So far, a generic `chunks={}` should work OK, maybe we want to have smart chunking done in the future, but at this stage, this setting helps a LOT with memory consumption and runtime. Cheers very much, gents! :beers: 

EDIT (thanks, @schlunma ) `chunks="auto"` does not work for some files `NotImplementedError: Can not use auto rechunking with object dtype. We are unable to estimate the size in bytes of object data`

***

***

### DIFF
--- a/esmvalcore/preprocessor/_io.py
+++ b/esmvalcore/preprocessor/_io.py
@@ -163,6 +163,18 @@ def _load_zarr(
     ignore_warnings: list[dict[str, Any]] | None = None,
     backend_kwargs: dict[str, Any] | None = None,
 ) -> CubeList:
+    # note on ``chunks`` kwarg to ``xr.open_dataset()``
+    # docs.xarray.dev/en/stable/generated/xarray.open_dataset.html
+    # this is very important because with ``chunks=None`` (default)
+    # data will be realized as Numpy arrays and transferred in memory
+    # both for remote and local Zarr stores;
+    # ``chunks={}`` loads the data with dask using the engine preferred
+    # chunk size, generally identical to the formats chunk size. If not
+    # available, a single chunk for all arrays; testing shows this is the
+    # "best guess" compromise for typically CMIP-like chunked data.
+    # see https://github.com/pydata/xarray/issues/10612 and
+    # https://github.com/pp-mo/ncdata/issues/139
+
     # case 1: Zarr store is on remote object store
     # file's URI will always be either http or https
     if urlparse(str(file)).scheme in ["http", "https"]:
@@ -191,6 +203,7 @@ def _load_zarr(
             consolidated=True,
             decode_times=time_coder,
             engine="zarr",
+            chunks={},
             backend_kwargs=backend_kwargs,
         )
     # case 2: Zarr store is local to the file system
@@ -199,6 +212,7 @@ def _load_zarr(
             file,
             consolidated=False,
             engine="zarr",
+            chunks={},
             backend_kwargs=backend_kwargs,
         )
 
--- a/esmvalcore/preprocessor/_io.py
+++ b/esmvalcore/preprocessor/_io.py
@@ -166,8 +166,7 @@ def _load_zarr(
     # note on ``chunks`` kwarg to ``xr.open_dataset()``
     # docs.xarray.dev/en/stable/generated/xarray.open_dataset.html
     # this is very important because with ``chunks=None`` (default)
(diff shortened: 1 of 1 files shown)


bettina-gier (approved):
looks good to me, following the links that work ;)

esmvalcore/preprocessor/_io.py
```diff
@@ -163,6 +163,18 @@ def _load_zarr(
     ignore_warnings: list[dict[str, Any]] | None = None,
     backend_kwargs: dict[str, Any] | None = None,
 ) -> CubeList:
+    # note on ``chunks`` kwarg to ``xr.open_dataset()``
+    # docs.xarray.dev/en/stable/generated/xarray.open_dataset.html
+    # this is very important because with ``chunks=None`` (default)
+    # data will be realized as Numpy arrays and transferred in memory
+    # both for remote and local Zarr stores;
```
schlunma, line 170:
```suggestion
    # data will be realized as Numpy arrays and transferred in memory;
```

This is true for all data types (not only Zarr), so I suggest removing this bit

esmvalcore/preprocessor/_io.py
```diff
@@ -191,6 +203,7 @@ def _load_zarr(
             consolidated=True,
             decode_times=time_coder,
             engine="zarr",
+            chunks={},
```
schlunma, line 206: One thing you could do to avoid duplicating these keyword arguments is to define

```py
open_kwargs = {
    "consolidated": False,
    "engine": "zarr",
    "chunks": {},
    "backend_kwargs": backend_kwargs,
}
```

at the top and then update those for the remote store. Then you can simply do a

```py
zarr_xr = xr.open_dataset(file, **open_kwargs)
```

after the if/else

schlunma (approved):
Thanks V! Tiny comments to avoid duplication. 🍻

esmvalcore/preprocessor/_io.py
valeriupredoi: a very good idea, thanks, Manu! Done in d6289481447bbf77dc51834f2f40a51e32d004c6

valeriupredoi:
very many thanks @bettina-gier and @schlunma :beers:

pp-mo:
FYI
re: 
> chunks="auto" does not work for some files NotImplementedError: Can not use auto rechunking with object dtype. We are unable to estimate the size in bytes of object data

We've seen a similar message with data which uses the variable-length string datatype.
(In netcdf thats "string" type variables, as opposed to "char" arrays with string-length dimension).
Same thing : variable-length types mean that dask refuses to apply "auto" chunking, because it doesn't know how large data is. 

This happens especially in data from the ecmf CDS. 
cf. our issue [here](https://github.com/SciTools/iris/issues/6149)

PR #2850, opened 2025-09-23
Support for Python 3.14.0

valeriupredoi:
Support for Python 3.14.0 that will be released October 7, 2025.
What's new in 3.14 https://docs.python.org/3.14/whatsnew/3.14.html

**Past issues**

As of 8 October 2025, we have a number of non-supporting packages (the others do support Python 3.14):
- we need esmf-regrid to unpin iris https://github.com/conda-forge/iris-esmf-regrid-feedstock/pull/23
- At the moment, iris=3.14 is restricting to Python<3.14 too, expecting to have that relaxed with iris=3.14.1 -> need to pin to iris here
- Fiona is at odds, all else is now fixed: https://github.com/ESMValGroup/ESMValCore/actions/runs/18654737590/job/53180940262
-  - dask >=2025  # github.com/ESMValGroup/ESMValCore/issues/2503
-  - dask-jobqueue
-  - distributed
-  - fiona
-   - intake-esm
-  - iris >=3.12.2  # https://github.com/SciTools/iris/issues/6417
-  - iris-esmf-regrid >=0.11.0
-  - iris-grib >=0.20.0  # github.com/ESMValGroup/ESMValCore/issues/2535
-  - ncdata
-  - shapely >=2.0.0

### DIFF
--- a/doc/recipe/overview.rst
+++ b/doc/recipe/overview.rst
@@ -78,7 +78,7 @@ The ``datasets`` section includes dictionaries that, via key-value pairs or
 
 - dataset name (key ``dataset``, value e.g. ``MPI-ESM-LR`` or ``UKESM1-0-LL``).
 - project (key ``project``, value ``CMIP5`` or ``CMIP6`` for CMIP data,
-  ``OBS`` for observational data, ``ana4mips`` for ana4mips data,
+  ``OBS`` for observational data, ``ana4MIPs`` for ana4MIPs data,
   ``obs4MIPs`` for obs4MIPs data, ``ICON`` for ICON data).
 - experiment (key ``exp``, value e.g. ``historical``, ``amip``, ``piControl``,
   ``rcp85``).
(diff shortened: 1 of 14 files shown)


environment.yml
```diff
@@ -21,7 +21,7 @@ dependencies:
   - humanfriendly
   - intake-esgf >=2025.10.22
   - intake-esm
-  - iris >=3.12.2,<3.14.1 # https://github.com/SciTools/iris/issues/6417, https://github.com/ESMValGroup/ESMValCore/issues/2921
+  - iris >=3.14.1
```
bouweandela: Do we need this pin? Just removing the upper limit may be sufficient and it gives users some room to choose versions

valeriupredoi: problem is 3.14 was a bit iffy: they momentarily removed Python 3.14 support in that, maybe we can just `!=3.14`?

PR #2944, opened 2026-01-15
Remove an outdated file that is part of the ESMValTool tests

bouweandela:
It looks like I forgot to delete this file when I separated the ESMValCore package from the ESMValTool package.


valeriupredoi (approved):
Cheers  buds 🍺

PR #2981, opened 2026-02-16
Avoid hardcoded time coordinate names in CMOR check

bouweandela:
## Description

Avoid hardcoded time coordinate names in CMOR check. The CMIP7 CMOR tables contain a coordinate called "time4", which results in a crash of the CMOR checker because it only recognizes times up to "time3". This affects the use of the variables `sfcWind` with branding suffix `tmaxavg-h10m-hxy-u`, `tas` with branding `tmaxavg-h2m-hxy-u` (aka `tasmax` in CMIP6), and `tas` with branding suffix `tminavg-h2m-hxy-u` (aka `tasmin` in CMIP6).

This prevents the recipe `ref/recipe_ref_fire.yml` from running with CMIP7 data.

***

***

### DIFF
--- a/esmvalcore/cmor/_fixes/fix.py
+++ b/esmvalcore/cmor/_fixes/fix.py
@@ -882,18 +882,18 @@ class GenericFix(Fix):
 
     def _fix_time_bounds(self, cube: Cube, cube_coord: Coord) -> None:
         """Fix time bounds."""
-        times = {"time", "time1", "time2", "time3"}
-        key = times.intersection(self.vardef.coordinates)
-        if not key:  # cube has time, but CMOR variable does not
-            return
-        cmor = self.vardef.coordinates[" ".join(key)]
-        if cmor.must_have_bounds == "yes" and not cube_coord.has_bounds():
-            cube_coord.bounds = get_time_bounds(cube_coord, self.frequency)
-            self._warning_msg(
-                cube,
-                "Added guessed bounds to coordinate %s",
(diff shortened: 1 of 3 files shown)


schlunma:
I am slightly scared that this will break the CMOR checks. In the past, we often found that `coord.axis` is not very reliable. For the time axis, Iris [units as the only criterion](https://scitools-iris.readthedocs.io/en/stable/_modules/iris/util.html#guess_coord_axis) when guessing the axis type. I wonder if [`VariableInfo.has_coord_with_standard_name`](https://github.com/ESMValGroup/ESMValCore/blob/main/esmvalcore/cmor/table.py#L1093) would be a better choice, but not really sure about that.

The existing code is certainly also not optimal (I think it would break if there were multiple time coordinates due to the `" ".join(key)` that is used to access the coordinate.

In any case, you would also need to adapt the corresponding fix here:

https://github.com/ESMValGroup/ESMValCore/blob/1e53ae95608441ce6a7e36abe6c2b1bec6112eff/esmvalcore/cmor/_fixes/fix.py#L883

bouweandela:
I checked the `axis` attribute for the `time` coordinates in the CMOR tables that are shipped with ESMValCore, there aren't that many, and it is correctly set. But I would be happy to switch to using the standard name if that feels more reliable to you?

bouweandela:
I've updated the code for the automatic fix too now.

schlunma (approved):
Thanks, looks good to me!

PR #3021, opened 2026-03-19
Gracefully handle the case of missing facets required by ``esmvalcore.io.local.LocalDataSource``

### LINKED ISSUE #2995: All facets for all configured data sources of a project are always required
Not sure if a bug or feature.

If multiple data sources are defined for a single project, all facets are always required. For example, if two sources for dataset A (with facet `a`) and dataset B (with facet `b`) exist, and only dataset B is used in the recipe, facet `a`  would still need to be given in the recipe. This can be very confusing (see also discussion here: https://github.com/ESMValGroup/ESMValCore/pull/2946#discussion_r2781751597).

This is particularly relevant for the native6 project, which is supposed to be shared among many different datasets. Another example case is the ERA5 GRIB data source, whose facets are very different to its netcdf counterpart:

https://github.com/ESMValGroup/ESMValCore/blob/974ff8c5dc6ded56421346de08163ed18227e3b6/esmvalcore/config/configurations/data-hpc-dkrz.yml#L70-L83

As a consequence, all recipes we have currently published will fail when this default DKRZ configuration is used.

Solutions to this could be:

- Avoid using different facets in different data sources (that also means no sharing of the native6 project among different datasets)
- Make sure that sensible defaults exist for the differing facets (defined as extra facets)

I will open a PR which comments out the GRIB data source for ERA5 on Levante (this causes recipes to break), but this is only a temporary solution.

bouweandela:
## Description

Do not raise an exception if a facet required by `esmvalcore.io.local.LocalDataSource` is missing, but log the problem and continue. Maybe the data can be found using some other data source and if not, the user will see the log message.

Closes #2995

Link to documentation:

***

***

### DIFF
--- a/esmvalcore/cmor/_fixes/icon/_base_fixes.py
+++ b/esmvalcore/cmor/_fixes/icon/_base_fixes.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import copy
 import logging
 import os
 import shutil
@@ -326,17 +327,14 @@ class IconFix(NativeDatasetFix):
 
     def _get_grid_from_rootpath(self, grid_name: str) -> CubeList | None:
         """Try to get grid from the ICON rootpath."""
-        glob_patterns: list[Path] = []
         for data_source in _get_data_sources(self.session, "ICON"):  # type: ignore[arg-type]
             if isinstance(data_source, esmvalcore.io.local.LocalDataSource):
-                glob_patterns.extend(
-                    data_source._get_glob_patterns(**self.extra_facets),  # noqa: SLF001
-                )
-        possible_grid_paths = [d.parent / grid_name for d in glob_patterns]
-        for grid_path in possible_grid_paths:
-            if grid_path.is_file():
-                logger.debug("Using ICON grid file '%s'", grid_path)
-                return self._load_cubes(grid_path)
+                ds = copy.deepcopy(data_source)
+                ds.filename_template = grid_name
+                files = ds.find_data(**self.extra_facets)
+                if files:
+                    logger.debug("Using ICON grid file '%s'", files[0])
+                    return files[0].to_iris()
         return None
 
     def _get_downloaded_grid(self, grid_url: str, grid_name: str) -> CubeList:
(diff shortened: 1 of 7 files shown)


esmvalcore/io/local.py
```diff
@@ -421,6 +420,30 @@ def _select_files(
     return selection
 
 
+class MissingFacetError(KeyError):
+    """Error raised when a facet required for filling the template is missing."""
```
schlunma: Any particular reason why you're not putting this into `esmvalcore.exceptions`?

schlunma (commented):
Great, thanks Bouwe! Works exactly as expected.

I just have some minor comments on the code.

esmvalcore/io/local.py
```diff
@@ -421,6 +420,30 @@ def _select_files(
...
+        The collection of items to format.
+
+    Returns
+    -------
+    :
+        The formatted string.
+    """
+    items = [f"'{item}'" for item in sorted(collection)]
+    if len(items) > 1:
+        items[-1] = f"and {items[-1]}"
```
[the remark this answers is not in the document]
bouweandela: Done in c8d9225

esmvalcore/cmor/_fixes/icon/_base_fixes.py
```diff
@@ -326,17 +327,14 @@ def _get_grid_from_cube_attr(self, cube: Cube) -> Cube:
...
         for data_source in _get_data_sources(self.session, "ICON"):  # type: ignore[arg-type]
             if isinstance(data_source, esmvalcore.io.local.LocalDataSource):
-                glob_patterns.extend(
-                    data_source._get_glob_patterns(**self.extra_facets),  # noqa: SLF001
-                )
-        possible_grid_paths = [d.parent / grid_name for d in glob_patterns]
-        for grid_path in possible_grid_paths:
-            if grid_path.is_file():
-                logger.debug("Using ICON grid file '%s'", grid_path)
-                return self._load_cubes(grid_path)
```
schlunma: For some reason, this does not work (ICON grid files are not found anymore). Code looks ok, don't really know why.

Maybe just `try/except` the exception similar to the other changes? It's probably not worth it to investigate this in detail..

schlunma (changes requested):

esmvalcore/cmor/_fixes/icon/_base_fixes.py
bouweandela: But then the issue would not be solved for the ICON project. Do you have a recipe I can run on Levante to test?

schlunma (approved):
Looks good now, thanks Bouwe 🚀

PR #3044, opened 2026-04-17
Use pixi to manage development conda environments

bouweandela:
## Description

Switch to [pixi](https://pixi.prefix.dev/) for managing the development conda environments.

See https://github.com/ESMValGroup/ESMValTool/pull/4407 for an extensive description and https://github.com/ESMValGroup/ESMValTool/issues/3997 for further background.

Links to documentation:
- https://esmvaltool--3044.org.readthedocs.build/projects/ESMValCore/en/3044/quickstart/install.html
- https://esmvaltool--3044.org.readthedocs.build/projects/ESMValCore/en/3044/contributing.html#dependencies

***

***


.github/workflows/run-tests.yml
```diff
@@ -1,116 +1,58 @@
...
+          - test-py314
+          - test-r
+    name: ${{ matrix.os }} ${{ matrix.environment }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           fetch-depth: 0
           persist-credentials: false
-      - uses: conda-incubator/setup-miniconda@fc2d68f6413eb2d87b895e92f8584b5b94a10167 # v3.3.0
+      - uses: prefix-dev/setup-pixi@a0af7a228712d6121d37aba47adf55c1332c9c2e # v0.9.5
```
valeriupredoi, line 39: is this gonna be updated by dependabot? conda-incubator (I think) was not supported by the dependabot, so if pixi is then that'd be great

.github/workflows/run-tests.yml
```diff
@@ -1,116 +1,58 @@
...
-      - run: python -V 2>&1 | tee test_osx_artifacts_python_${{ matrix.python-version }}/python_version.txt
-      - run: mamba install -c conda-forge git
-      - run: conda list
-      - run: pip install --no-deps -e .[develop] 2>&1 | tee test_osx_artifacts_python_${{ matrix.python-version }}/install.txt
-      - run: conda list
-      - run: |
-          pre-commit install
-          pre-commit run -a
-      - run: pytest -n 2 -m "not installation" 2>&1 | tee test_osx_artifacts_python_${{ matrix.python-version }}/test_report.txt
+          frozen: true
```
valeriupredoi, line 41: what's this do?

doc/quickstart/install.rst
```diff
@@ -3,71 +3,68 @@
...
-    conda activate esmvalcore
-
-and install ESMValCore with the command:
+After installing the dependencies that are not available from PyPI_, install
+ESMValCore and any remaining Python dependencies with the command:
 
 .. code-block:: bash
 
-    pip install --no-deps esmvalcore
+    pip install esmvalcore
```
valeriupredoi, line 67: why are we dropping the --no-deps flag?

doc/conf.py
```diff
@@ -25,25 +25,6 @@
...
-    # are not set. Here, we hardcode them instead.
-    # In a normal environment, i.e. a local build of the documentation, the
-    # normal environment activation takes care of this.
-    rtd_version = os.environ.get("READTHEDOCS_VERSION", "latest")
-    rtd_project = os.environ.get("READTHEDOCS_PROJECT")
-    rtd_conda_prefix = f"/home/docs/checkouts/readthedocs.org/user_builds/{rtd_project}/conda/{rtd_version}"
-    os.environ["ESMFMKFILE"] = f"{rtd_conda_prefix}/lib/esmf.mk"
-    os.environ["PROJ_DATA"] = f"{rtd_conda_prefix}/share/proj"
-    os.environ["PROJ_NETWORK"] = "OFF"
-
```
valeriupredoi, line 46: I see the docs build nicely, so these can indeed be removed, but why are we removing them now - is it something to do with pixi?

pyproject.toml
```diff
@@ -230,3 +205,113 @@ known-first-party = ["esmvalcore"]
...
+jinja2 = "*"
+libnetcdf = "!=4.9.1"
+nc-time-axis = "*"
+ncdata = "*"
+nested-lookup = "*"
+netcdf4 = "*"
+numpy = "!=1.24.3"
+packaging = "*"
+pandas = "*"
+pillow = "*"
```
valeriupredoi, line 239: can we not just use "pillow" without the all (*) marker pin? ChatGPT says yes, but also no 😆

valeriupredoi (commented):
will be back 😁

valeriupredoi:
hey @bouweandela a few points:

- installing pixi from source via curl and install may be tricky for some systems, I think best bet is to point users to install it from conda-forge: though, having said that, my install + shell activation doesn't work:

```bash
(base) valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ pixi --version
pixi 0.67.0
(base) valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ pixi shell
 WARN Skipped running the post-link scripts because `run-post-link-scripts` = `false`
        - bin/.librsvg-pre-unlink.sh

To enable them, run:
	pixi config set --local run-post-link-scripts insecure

More info:
	https://pixi.sh/latest/reference/pixi_configuration/#run-post-link-scripts

WARNING: Did not detect successful shell initialization within 3 second(s).
         Please check on https://pixi.sh/latest/advanced/pixi_shell/#issues-with-pixi-shell for more tips.
er passphrase for /home/valeriu/.ssh/id_rsa_jasmin: 
Identity added: /home/valeriu/.ssh/id_rsa_jasmin (/home/valeriu/.ssh/id_rsa_jasmin)
Agent pid 963160
Enter passphrase for /home/valeriu/.ssh/id_rsa_cloud: 
valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ pixi config set --local run-post-link-scripts insecure
✅ Updated config at /home/valeriu/ESMValCore/.pixi/config.toml
valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ git status
On branch pixi
Your branch is up to date with 'origin/pixi'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	.pixi/

nothing added to commit but untracked files present (use "git add" to track)
valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ pixi shell
WARNING: Did not detect successful shell initialization within 3 second(s).
         Please check on https://pixi.sh/latest/advanced/pixi_shell/#issues-with-pixi-shell for more tips.
er passphrase for /home/valeriu/.ssh/id_rsa_jasmin: 
Agent pid ...
```

Why is it putting ` .pixi` dir in the source dir? I don't like that - can we add `.pixi` to gitignore then? Ideally I'd like it somewhere else anyway. And it starts sourcing my `.bashrc` - not liking that either. But more importantly, that shell never kicks in for me...

valeriupredoi:
> hey @bouweandela a few points:
> 
>     * installing pixi from source via curl and install may be tricky for some systems, I think best bet is to point users to install it from conda-forge: though, having said that, my install + shell activation doesn't work:
> 
> 
> ```shell
> (base) valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ pixi --version
> pixi 0.67.0
> (base) valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ pixi shell
>  WARN Skipped running the post-link scripts because `run-post-link-scripts` = `false`
>         - bin/.librsvg-pre-unlink.sh
> 
> To enable them, run:
> 	pixi config set --local run-post-link-scripts insecure
> 
> More info:
> 	https://pixi.sh/latest/reference/pixi_configuration/#run-post-link-scripts
> 
> WARNING: Did not detect successful shell initialization within 3 second(s).
>          Please check on https://pixi.sh/latest/advanced/pixi_shell/#issues-with-pixi-shell for more tips.
> er passphrase for /home/valeriu/.ssh/id_rsa_jasmin: 
> Identity added: /home/valeriu/.ssh/id_rsa_jasmin (/home/valeriu/.ssh/id_rsa_jasmin)
> Agent pid 963160
> Enter passphrase for /home/valeriu/.ssh/id_rsa_cloud: 
> valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ pixi config set --local run-post-link-scripts insecure
> ✅ Updated config at /home/valeriu/ESMValCore/.pixi/config.toml
> valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ git status
> On branch pixi
> Your branch is up to date with 'origin/pixi'.
> 
> Untracked files:
>   (use "git add <file>..." to include in what will be committed)
> 	.pixi/
> 
> nothing added to commit but untracked files present (use "git add" to track)
> valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ pixi shell
> WARNING: Did not detect successful shell initialization within 3 second(s).
>          Please check on https://pixi.sh/latest/advanced/pixi_shell/#issues-with-pixi-shell for more tips.
> er passphrase for /home/valeriu/.ssh/id_rsa_jasmin: 
> Agent pid ...
> ```
> 
> Why is it putting ` .pixi` dir in the source dir? I don't like that - can we add `.pixi` to gitignore then? Ideally I'd like it somewhere else anyway. And it starts sourcing my `.bashrc` - not liking that either. But more importantly, that shell never kicks in for me...

ah! It works now:

```
(ESMValCore) valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ esmvaltool version
Running esmvaltool executable from ESMValCore. No other command line utilities are available until ESMValTool is installed.
ESMValCore: 2.15.0.dev37+gb3905f59d
```

You gotta tell folks to not have anything that waits for a password in their `.bashrc` - that makes the shell init process wait more than the max 3 seconds.

.github/workflows/run-tests.yml
bouweandela: I think so, dependabot supports github actions and this is a github action: https://github.com/marketplace/actions/setup-pixi

.github/workflows/run-tests.yml
bouweandela: Install the environment as defined in the lockfile: https://pixi.prefix.dev/latest/reference/cli/pixi/install/#update-options

doc/quickstart/install.rst
bouweandela: Because this section explains how to install (as much as possible) from PyPI

doc/conf.py
bouweandela: Yes, as explained in the comment above, the conda environment used by the previous way of installing the environment on readthedocs was not activated. Pixi does not have this problem, so this code can be removed.

pyproject.toml
bouweandela: I tried, but it doesn't look like it:
```
$ pixi shell
Error:   × expected an equals, found a newline
     ╭─[/home/bandela/src/esmvalgroup/esmvalcore/pyproject.toml:210:8]
 209 │     [tool.pixi.dependencies]
 210 │ ╭─▶ aiohttp
 211 │ ╰─▶ cartopy
 212 │     cf-units = "*"
 213 │     cftime = "*"
     ╰────
```

bouweandela:
Thanks for reviewing V!

> ah! It works now:

I suspect you may have been on a different branch than the one from this pull request the first time you ran that command, because `.pixi` is added to `.gitignore` in this pull request. Does it work OK now?

> And it starts sourcing my .bashrc - not liking that either.

As the name suggests, the command `pixi shell` starts a new shell. You could probably get around that by using the command `pixi run` with the `--clean-env` flag instead, e.g. `pixi run --clean-env esmvaltool`.

valeriupredoi:
```
(ESMValCore) valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ git status
On branch pixi
Your branch is up to date with 'origin/pixi'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	.pixi/

nothing added to commit but untracked files present (use "git add" to track)
(ESMValCore) valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ vim .gitignore 
(ESMValCore) valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ git status
On branch pixi
Your branch is up to date with 'origin/pixi'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   .gitignore

no changes added to commit (use "git add" and/or "git commit -a")
```

I put it in `.gitignore` correctly at the second call :grin:

valeriupredoi:
note (am not sure how important it is) that a `pip install -e .[develop]` still works fine in a pixi shell - we still want to do that, for proper local development, but it does install some interesting packages that I've not seen before:
```
Successfully installed ESMValCore-2.15.0.dev38+gf8554ad53 httptools-0.7.1 uvloop-0.22.1 watchfiles-1.1.1 websockets-16.0
```
that don't show up in the pixi env:
```
(ESMValCore) valeriu@valeriu-PORTEGE-Z30-C:~/ESMValCore$ pixi list websockets
Error:   × No packages found in 'default' environment for 'linux-64' platform.
```

bouweandela:
It was already there, now you've added it a second time. Maybe it is because you ran `pixi config set --local run-post-link-scripts insecure` and that changed the file `.pixi/config.toml` which is versioned?

valeriupredoi:
> It was already there, now you've added it a second time. Maybe it is because you ran `pixi config set --local run-post-link-scripts insecure` and that changed the file `.pixi/config.toml` which is versioned?

nope that was showing up as ungit before - leave it like that - twice is better :grin:

bouweandela:
> pip install -e .[develop] still works fine in a pixi shell - we still want to do that

There should be no need to do that

valeriupredoi:
> As the name suggests, the command pixi shell starts a new shell. You could probably get around that by using the command pixi run with the --clean-env flag instead, e.g. pixi run --clean-env esmvaltool.

my point is that we need a bit of doc with this what you told me, or to tell folks not to have any shell init with passwords in their `.bashrc` otherwise pixi will moan

valeriupredoi:
> > pip install -e .[develop] still works fine in a pixi shell - we still want to do that
> 
> There should be no need to do that

it absolutely should be a need to do that when one wants to test if the package still builds well after changes to eg `pyproject.toml` or if you want to install the package in an eg module that will not load a pixi shell - think a central installation on an HPC

doc/contributing.rst
```diff
@@ -636,23 +645,31 @@ See this `statement <https://www.apache.org/licenses/GPL-compatibility.html>`__
...
-the following files:
+the following locations in ``pyproject.toml``:
+
+- ``dependencies``
+  contains dependencies that can be installed from `PyPI <https://pypi.org/>`_
+- ``[tool.pixi.dependencies]``
+  contains dependencies that can be installed from `conda-forge <https://conda-forge.org/>`_
+
+it is strongly preferred that those two lists are kept in sync, apart from
+differences in how packages are named. Run the command ``pixi lock --no-install``
```
bouweandela:
```suggestion
differences in how packages are named. Run the command ``pixi lock``
```

.github/workflows/run-tests.yml
valeriupredoi: yay

.github/workflows/run-tests.yml
valeriupredoi: ah frozen, duh - sorry, bud, me being dumber than Elsa :laughing:

bouweandela:
After changes to pyproject.toml, one can test that things work as expected by e.g. running `pixi lock` or any of the other pixi commands, such as `pixi shell` or `pixi run`.

To create a module for an HPC, you will probably want to do the same as in the Dockerfile: clone the repo in the location where the module software should be, checkout the released version, create a pixi environment, and put the activation script (entrypoint without the `esmvaltool` command in the Dockerfile) somewhere it will be executed when the module is loaded.

doc/quickstart/install.rst
valeriupredoi: as much as possible from PyPI is bad but hey, I checked the env contents and (most) all is from conda-forge, so am happy

doc/conf.py
valeriupredoi: I'd keep it somewhere somehow, just in case. I think the whole purge of conda-forge way of installing is a bit premature at this point in time, even if we have pixi in for end users- which is nice, I can already think of a few cases where pixi won't work, so I'd argue we still keep the old conda stashed somewhere, in case we have to use it. How, I need to think of that

pyproject.toml
valeriupredoi: blergh, thanks for trying

valeriupredoi:
> To create a module for an HPC, you will probably want to do the same as in the Dockerfile: clone the repo in the location where the module software should be, checkout the released version, create a pixi environment, and put the activation script (entrypoint without the `esmvaltool` command in the Dockerfile) somewhere it will be executed when the module is loaded.

Not as straightforward as that: for one, Docker is not allowed on an HPC, two, Singularity comes with all manners of permission issues, and three -most important - a user will never have access to the pixi shell that an admin creates. forget about working with shells there, it's all paths to various bin dirs

bouweandela:
> Not as straightforward as that: for one, Docker is not allowed on an HPC, two, Singularity comes with all manners of permission issues, and three -most important - 

I meant: "follow the same steps", not "use the container".. The steps are here: https://github.com/ESMValGroup/ESMValCore/blob/e0507eca93eca59ccda3bdec65c9a4265de32b04/docker/Dockerfile#L9-L12

doc/quickstart/install.rst
```diff
@@ -172,31 +170,18 @@ Installation from source
...
--  Activate the esmvalcore environment: ``conda activate esmvalcore``
--  Install in development mode: ``pip install --no-deps -e '.[develop]'``. If you
-   are installing behind a proxy that does not trust the usual pip-urls
-   you can declare them with the option ``--trusted-host``,
-   e.g. ``pip install --no-deps --trusted-host=pypi.python.org --trusted-host=pypi.org --trusted-host=files.pythonhosted.org -e .[develop]``
--  Test that your installation was successful by running
-   ``esmvaltool -h``.
--  Install the :ref:`esmvaltool:pre-commit` hooks by running:
-   ``pre-commit install``.
+-  Create and activate the development environment by running: ``pixi shell``
```
bouweandela:
```suggestion
-  Create the development environment and start a shell in this environment by running: ``pixi shell``
```

bouweandela:
>>    As the name suggests, the command pixi shell starts a new shell. You could probably get around that by using the command pixi run with the --clean-env flag instead, e.g. pixi run --clean-env esmvaltool.

> my point is that we need a bit of doc with this what you told me

Added in 94f10e0

valeriupredoi:
> > Not as straightforward as that: for one, Docker is not allowed on an HPC, two, Singularity comes with all manners of permission issues, and three -most important -
> 
> I meant: "follow the same steps", not "use the container".. The steps are here:
> 
> https://github.com/ESMValGroup/ESMValCore/blob/e0507eca93eca59ccda3bdec65c9a4265de32b04/docker/Dockerfile#L9-L12

just like Carroll Shelby told his chief mechanic that said she can come apart: one way to find out :grin: https://www.youtube.com/watch?v=5OfjrMWjrhs

valeriupredoi:
One thing I think you should mention is that people should not have multiple ESMValXXX dirs - and they have - lots of eg `ESMValTool_something_somethingelse`: since pixi is now local to the source code, and it's hefty ie ~2G for Core (prob 5G for Tool), disk space will soon start disappearing

valeriupredoi:
> One thing I think you should mention is that people should not have multiple ESMValXXX dirs - and they have - lots of eg `ESMValTool_something_somethingelse`: since pixi is now local to the source code, and it's hefty ie ~2G for Core (prob 5G for Tool), disk space will soon start disappearing

in this vein, how the heck does one create new named environments without duplicating the project ie the repo? Say I need an environment `esmvalcore1` and another one `esmvalcore2` that use the same base list of dependencies, but I want to further test in each of these?

bouweandela:
> One thing I think you should mention is that people should not have multiple ESMValXXX dirs - and they have - lots of eg ESMValTool_something_somethingelse: since pixi is now local to the source code, and it's hefty ie ~2G for Core (prob 5G for Tool), disk space will soon start disappearing

Pixi avoids the problem by using hardlinks, so it only keeps one copy of a file, even if it's used in multiple places: https://pixi.prefix.dev/latest/workspace/environment/#de-duplication

bouweandela:
> how the heck does one create new named environments without duplicating the project ie the repo?

Edit pyproject.toml, add the extra environment (plus any needed features) there and run `pixi install -e name_of_environment`: https://pixi.prefix.dev/latest/tutorials/multi_environment/

valeriupredoi:
> > One thing I think you should mention is that people should not have multiple ESMValXXX dirs - and they have - lots of eg `ESMValTool_something_somethingelse`: since pixi is now local to the source code, and it's hefty ie ~2G for Core (prob 5G for Tool), disk space will soon start disappearing
> 
> in this vein, how the heck does one create new named environments without duplicating the project ie the repo? Say I need an environment `esmvalcore1` and another one `esmvalcore2` that use the same base list of dependencies, but I want to further test in each of these?

ah! Found the solution - in pyproject.toml

```
[tool.pixi.environments]
default = { features = ["dev", "doc", "test"], solve-group = "default" }
esmvalcore = { solve-group = "default" }
# Environments for running tests with specific Python versions.
test-py312 = { features = ["py312", "test"], solve-group = "py312" }
test-py313 = { features = ["py313", "test"], solve-group = "py313" }
test-py314 = { features = ["py314", "test"], solve-group = "default" }
# Environment for testing R support
test-r = { features = ["test-r"], solve-group = "test-r" }
```

valeriupredoi:
Ow man, you gonna get so many users (not technical ones, though) so confused with all these new procedures - they are still confused about using conda/mamba - pixi takes this one notch up the level of confusability. Pixi is a lot more under the hood technical than conda - I don't think our ESMValTool users will get used to it easily if they want to do the slightest bit of development, more than `pixi shell` and get cracking with diags after that

valeriupredoi:
we should think along these lines when we explain Pixi to our users (prob best to structure the docs too, or the UET in their Tutorial) - lifted from ChatGPT that makes a very solid point:

pixi is a lot more technical than conda, and I think non-technical users can struggle with it more than with conda, what do you think?

You’re not wrong—but it’s a bit more nuanced than “Pixi is harder.”

### 🧠 The core difference

* **Conda** is designed to feel *imperative*:
  “run commands, get environment”

* **Pixi** is more *declarative*:
  “define environment in a file, then use it”

That shift alone makes Pixi feel more “technical,” especially at first.

---

## 👍 Where Conda is easier for non-technical users

* Very straightforward mental model:

  ```bash
  conda create -n myenv python=3.11
  conda activate myenv
  ```
* No config file required
* Immediate feedback, less abstraction
* Widely taught (lots of tutorials, StackOverflow answers)

👉 For beginners or casual users, Conda is usually easier to *start* with.

---

## 🤔 Where Pixi feels harder

* Requires understanding:

  * `pixi.toml`
  * features / environments
  * project-based workflow
* Less “just run a command and go”
* Errors can feel more “build-system-like” than user-friendly

👉 That *does* raise the barrier for non-technical users.

---

## ⚖️ But Pixi isn’t harder for no reason

Pixi is optimizing for things Conda struggles with:

* **Reproducibility** (lockfiles that actually work reliably)
* **Project isolation** (no mystery global envs)
* **Speed** (much faster solves/installs)
* **Team workflows** (commit `pixi.toml` and it just works)

So:

* Conda = easier *initial UX*
* Pixi = better *long-term consistency*

---

## 🎯 My honest take

* For **solo, non-technical users** → Conda is usually the better fit
* For **teams, CI, reproducible research, dev workflows** → Pixi quickly becomes nicer despite the learning curve

---

## 💡 Why Pixi *feels* more technical

It’s not just complexity—it’s *where the complexity lives*:

* Conda hides complexity in commands and global state
* Pixi exposes it in a config file

That’s great for engineers, but unfamiliar to beginners.

---

## 🧭 Practical compromise

A lot of people end up doing:

* Use Conda for:

  * quick experiments
  * teaching
* Use Pixi for:

  * projects that need to be shared/reproduced

---

bouweandela:
Learning something new is always hard, and learning a new technical skill is more annoying if it's not your main interest but rather something you want to get over with as quickly as possible with a minimal time investment. So yes, this will require some initial time investment and that may be frustrating to some, but I believe it will save our contributors time in the long run. No more waiting for solving environments will be an immediate time gain, and the much better reproducibility is very helpful when doing science.

While the observations above by ChatGPT above may be true in general, they are not for our project. Our contributors, at least the ones that want to change the dependencies, already need to understand two or more different configuration files: pyproject.toml for the Python dependencies and (multiple in the case of ESMValTool) environment.yml files for the conda-forge dependencies. Pixi simplifies that by making it possible to do all the configuration in a single file (pyproject.toml) with a single syntax.

One thing that we could do to make the transition easier is to do it at the [upcoming workshop](https://github.com/ESMValGroup/Community/discussions/273). Like that, we will be able to help out anyone who is there, either in person or online, immediately.

valeriupredoi (approved):
I agree with you, bud!

> One thing that we could do to make the transition easier is to do it at the https://github.com/ESMValGroup/Community/discussions/273. Like that, we will be able to help out anyone who is there, either in person or online, immediately.

I think that's a good idea too! At any rate, very many thanks to you for doing this work, bud - excuse my constant nagging during the review process, I am always trying to find holes in the plot so we can fix them when we review, rather than when we have users barking at us 🐶

valeriupredoi (approved):
I realized I've not formally approved, many thanks @bouweandela 🍺🍺

schlunma:
Did you run the changed GitHub actions workflows? Sorry, there's so many commits, I can't find them.

doc/quickstart/install.rst
```diff
@@ -3,71 +3,68 @@
...
 
 Of course it is also possible to choose a different name than ``esmvalcore`` for the environment.
 
 .. note::
 
 	  Creating a new Conda environment is often much faster and more reliable than trying to update an existing Conda environment.
 
+
 Pip installation
 -----------------
```
schlunma, line 50: What's the reason we mention this here? Is there any advantage of doing that over the mamba installation? Just saying "installing dependencies that are not available on PyPI in some other way" is also not really helpful.

I would argue we should remove that, especially since this is the section targeted to regular users (not developers).

doc/quickstart/install.rst
```diff
@@ -172,31 +170,18 @@ Installation from source
...
--  Activate the esmvalcore environment: ``conda activate esmvalcore``
--  Install in development mode: ``pip install --no-deps -e '.[develop]'``. If you
-   are installing behind a proxy that does not trust the usual pip-urls
-   you can declare them with the option ``--trusted-host``,
-   e.g. ``pip install --no-deps --trusted-host=pypi.python.org --trusted-host=pypi.org --trusted-host=files.pythonhosted.org -e .[develop]``
--  Test that your installation was successful by running
-   ``esmvaltool -h``.
--  Install the :ref:`esmvaltool:pre-commit` hooks by running:
-   ``pre-commit install``.
+-  Create the development environment and start a shell in this environment by running: ``pixi shell``
```
schlunma, line 177: If I understand correctly, `pixi shell` will solve the dependencies as defined in the manifest file and update the `pixi.lock` file. Since you mentioned that one advantage of Pixi is that dependencies don't need to be solved, we should mention here that one can also use `pixi shell --frozen` for a much faster installation with slightly outdated dependencies.

`pixi shell` can take a long time too run (much longer than `mamba activate` before), so we might recommend this as the default.

doc/contributing.rst
```diff
@@ -410,8 +418,8 @@ is cloned and run the command
 
    pytest
 
-Optionally you can skip tests which require additional dependencies for
-supported diagnostic script languages by adding ``-m 'not installation'`` to the
+or ``pixi run test``. Optionally you can skip tests which require additional
```
schlunma: Put ``pixi run test`` into a code block to make it more prominent?

pyproject.toml
```diff
@@ -230,3 +205,113 @@ known-first-party = ["esmvalcore"]
...
+xarray = "*"
+yamale = "*"
+zarr = ">3"
+
+[tool.pixi.workspace]
+channels = ["conda-forge"]
+platforms = ["linux-64", "osx-arm64"]
+
+[tool.pixi.pypi-dependencies]
+ESMValCore = { path = ".", editable = true }
```
schlunma, line 261:
```suggestion
esmvalcore = { path = ".", editable = true }
```

Shouldn't this refer to the package name?

schlunma (commented):
Thanks Bouwe, that's very nice! I was a bit skeptical of Pixi at first, but after using it for some time (in other projects) I really like it now.

The installation works very well. The only issue I found is that `esmvaltool config show` stopped working (see https://github.com/ESMValGroup/ESMValCore/issues/3062), but I am pretty sure this is not related to Pixi.

I just have a couple of minor comments on the code. Will look at the ESMValTool PR now.

valeriupredoi:
Hi @schlunma - it's great you're also looking at this, and already found a couple bits that I missed! Both Bouwe and I ran the GHAs so those should be solid. 👍🍺

schlunma:
Can you point me to the logs? I want to have a look at them 😁

valeriupredoi:
> Can you point me to the logs? I want to have a look at them 😁

If you go to Actions you can see the Branch tab there then you can select pixi to show the GHAs that ran only on the pixi branch, buds 🍺

schlunma:
Ah, perfect! I've never noticed the "branches" column before 😅 

One more question: After installing the environment, I get this warning:

```
WARN Skipped running the post-link scripts because `run-post-link-scripts` = `false`
        - bin/.librsvg-pre-unlink.sh

To enable them, run:
        pixi config set --local run-post-link-scripts insecure

More info:
        https://pixi.sh/latest/reference/pixi_configuration/#run-post-link-scripts
```

Do we need to run these post-link scripts? It seems to work without it. In this case, I wonder if we could silence this warning by explicitly setting `run-post-link-scripts = false` in a config.toml file.

valeriupredoi:
> Ah, perfect! I've never noticed the "branches" column before 😅
> 
> One more question: After installing the environment, I get this warning:
> 
> ```
> WARN Skipped running the post-link scripts because `run-post-link-scripts` = `false`
>         - bin/.librsvg-pre-unlink.sh
> 
> To enable them, run:
>         pixi config set --local run-post-link-scripts insecure
> 
> More info:
>         https://pixi.sh/latest/reference/pixi_configuration/#run-post-link-scripts
> ```
> 
> Do we need to run these post-link scripts? It seems to work without it. In this case, I wonder if we could silence this warning by explicitly setting `run-post-link-scripts = false` in a config.toml file.

Nothing to worry, bud. Quoting ChatGPT (diuble checked by yours truly)
That warning is coming from **Pixi**, and it’s usually harmless.

### What it means

Pixi is telling you it **skipped a post-link script** (`.librsvg-pre-unlink.sh`) because your configuration has:

```
run-post-link-scripts = false
```

Post-link scripts are small setup/cleanup scripts that some packages (like **librsvg**) try to run after installation or before removal.

### Should you worry?

Most of the time: **no** 👍

* It’s just informational, not an error
* Many environments disable these scripts for **security or reproducibility reasons**
* Things will usually work fine without them

### When it *might* matter

You might run into issues if:

* The package relies on the script to finish setup (rare)
* You see missing functionality related to **librsvg** (SVG rendering issues, etc.)

### How to enable scripts (if needed)

If you do need those scripts, you can turn them back on in your Pixi config:

```toml
run-post-link-scripts = true
```

Or via CLI (depending on your setup):

```bash
pixi config set run-post-link-scripts true
```

### Bottom line

If everything is working → ignore it.
If something breaks → enable scripts and try again

doc/quickstart/install.rst
bouweandela: Added this as a tip in [22fae20](https://github.com/ESMValGroup/ESMValCore/pull/3044/commits/22fae20728e79021cd8689b72ad002d27d52fd1a). I wouldn't expect it to take very long for ESMValCore though, since it doesn't have nearly as many dependencies as ESMValTool.

doc/quickstart/install.rst
bouweandela: Some people really don't like installing things with conda, so I think it is important that we mention that we have a PyPI package. Of course, we could do that in a different way. Do you have ideas on how that could be improved?

bouweandela: For example, for xarray they explain it like this: https://docs.xarray.dev/en/stable/getting-started-guide/installing.html#instructions

pyproject.toml
bouweandela: ESMValCore is the package name: https://github.com/ESMValGroup/ESMValCore/blob/45d2dcc991010fbc49813d22790a0d81768199d1/pyproject.toml#L76

doc/contributing.rst
bouweandela: Done in 1ad1f6f

bouweandela:
> Do we need to run these post-link scripts? It seems to work without it. In this case, I wonder if we could silence this warning by explicitly setting `run-post-link-scripts = false` in a config.toml file.

The post link script mentioned is https://github.com/conda-forge/librsvg-feedstock/blob/main/recipe/post-link.sh. Some googling of what it does suggests it's related to rendering icons in gnome, so it's probably not relevant. The only way to silence the warning appears to be to set `run-post-link-scripts = insecure`, and if the pixi folks think that is insecure we may not want to do that.

pyproject.toml
schlunma: Ah, I always thought the package name is the string that you need to `import` (`import esmvalcore`).

doc/quickstart/install.rst
schlunma: I guess with the new warning that we recommend mamba instead of pip this should be fine. Since I do not know how to install the dependencies without mamba/conda, I don't have a better suggestion.

schlunma (approved):
Looks good to me now (apart from the trivial merge conflict)! Thanks so much 🚀

pyproject.toml
bouweandela: `ESMValCore` is the distribution package and it provides the `esmvalcore` import package, you can read more about here: https://packaging.python.org/en/latest/discussions/distribution-package-vs-import-package/

bouweandela:
Thanks for reviewing @valeriupredoi and @schlunma!

PR #2777, opened 2025-07-16
Expand `Dataset.from_files` so it works properly with derived variables

schlunma:
## Description

This PR expands [`Dataset.from_files`](https://esmvaltool--2777.org.readthedocs.build/projects/ESMValCore/en/2777/api/esmvalcore.dataset.html#esmvalcore.dataset.Dataset.from_files) so it works properly with derived variables. In addition, a new attribute [`Dataset.input_datasets`](https://esmvaltool--2777.org.readthedocs.build/projects/ESMValCore/en/2777/api/esmvalcore.dataset.html#esmvalcore.dataset.Dataset.input_datasets) is available which returns the datasets necessary for derivation (or simply the dataset itself is no derivation is required). This can also be used within the [`derive`](https://esmvaltool--2777.org.readthedocs.build/projects/ESMValCore/en/2777/api/esmvalcore.preprocessor.html#esmvalcore.preprocessor.derive) preprocessor function.

This PR is the second step to make `Dataset.load` work with derived variables.

## Example

```py
dataset_template = Dataset(
    short_name="lwcre",
    mip="Amon",
    project="CMIP6",
    exp="historical",
    dataset="*",
    institute="*",
    ensemble="r1i1p1f1",
    grid="gn",
    derive=True,
    force_derivation=True,
)

datasets = list(dataset_template.from_files())
print(f"Found {len(datasets)} datasets")  # Found 36 datasets

dataset = datasets[0]
dataset.files  # []

for d in dataset.input_datasets:
    print(d["short_name"])
    print(d.files)

# rlut
# [ESGFFile:CMIP6/CMIP/AS-RCEC/TaiESM1/historical/r1i1p1f1/Amon/rlut/gn/v20200623/rlut_Amon_TaiESM1_historical_r1i1p1f1_gn_185001-201412.nc on hosts ['esgf.ceda.ac.uk', 'esgf.rcec.sinica.edu.tw', 'esgf3.dkrz.de', 'esgf3.dkrz.de']]
# rlutcs
# [ESGFFile:CMIP6/CMIP/AS-RCEC/TaiESM1/historical/r1i1p1f1/Amon/rlutcs/gn/v20200623/rlutcs_Amon_TaiESM1_historical_r1i1p1f1_gn_185001-201412.nc on hosts ['esgf.ceda.ac.uk', 'esgf.rcec.sinica.edu.tw', 'esgf3.dkrz.de']]
```

Related to https://github.com/ESMValGroup/ESMValCore/issues/2769.

[...]

PR #3053, opened 2026-04-24
Yet another alternative implementation for working with derived variables

bouweandela:
## Description

Yet another alternative implementation to #2777 and #3051 for finding input datasets for derived variables and loading them. Adds a new class method `esmvalcore.dataset.DerivedDataset that can be used to work with derived variables.

Note that derived variables with optional input datasets are not yet supported. The only derived variable that uses this feature is `amoc` at the moment, so the impact of that is limited.

Link to documentation: https://esmvaltool--3053.org.readthedocs.build/projects/ESMValCore/en/3053/notebooks/derived-variables.html

***

***


esmvalcore/dataset.py
```diff
@@ -1008,3 +1011,113 @@ def _update_timerange(self) -> None:
...
+    facets:
+        The facets describing the variable to be derived. This should at least
+        include the ``'short_name'``, ``'mip'``, and ``'project'`` facets, but
+        may include other facets as well.
+
+    """
+
+    def __init__(self, **facets: FacetValue) -> None:
+        self.facets: Facets = facets
+        self._input_datasets: tuple[Dataset, ...] = ()
```
schlunma: Looks like this is not used anywhere

esmvalcore/dataset.py
```diff
@@ -1008,3 +1011,113 @@ def _update_timerange(self) -> None:
...
+        )
+        return esmvalcore.preprocessor.derive(
+            cubes=CubeList(cubes),
+            short_name=var_info.short_name,  # type: ignore[union-attr]
+            long_name=var_info.long_name,  # type: ignore[union-attr]
+            units=var_info.units,  # type: ignore[union-attr]
+            standard_name=var_info.standard_name,  # type: ignore[union-attr]
+        )
+
+    def from_files(self) -> Iterator[Self]:
```
schlunma, line 1092:
```suggestion
    def from_files(self) -> Generator[Self]:
```

I think a `Generator` is more accurate (every Generator is an Iterator, but not vice versa).

schlunma (commented):
Thanks Bouwe, looks good! I prefer this approach over https://github.com/ESMValGroup/ESMValCore/pull/3051. This will certainly be very helpful when using the ESMValCore Python API.

I am not a very big fan of adding a feature to the Dataset class but not using it when parsing recipes, but in this case it's much easier to separate this. Incorporating this cleanly into the existing code was by far the hardest problem I had in https://github.com/ESMValGroup/ESMValCore/pull/3051.

Will look into this in more detail once the PR is ready for review! Thanks so much!

esmvalcore/dataset.py
bouweandela: Removed in 0c6b607

bouweandela:
Thanks for taking a look @schlunma!

> I am not a very big fan of adding a feature to the Dataset class but not using it when parsing recipes, but in this case it's much easier to separate this.

We could probably use this feature when parsing the recipes, but let's think about that later. It may not be as relevant there because starting out from a recipe with wildcards to develop a complete recipe doesn't work for all but the simplest cases because the recipe doesn't allow specifying the relations between datasets that you typically want, e.g. give me all datasets that provide these 3 variables for these two experiments and don't have data quality issues is not something you can easily express in a recipe, but you can do in code.

> Incorporating this cleanly into the existing code was by far the hardest problem I had in https://github.com/ESMValGroup/ESMValCore/pull/3051.

Yes, my take on it was that it just doesn't fit and it's easier to add a new dataset class for derived variables because it needs slightly different features from the existing dataset class, e.g there is no straightforward way to interpret what the `.files` attribute would mean for derived variables.

> Will look into this in more detail once the PR is ready for review! Thanks so much!

Awesome, if you like the concept, I will continue working on this one.

esmvalcore/dataset.py
bouweandela: The typing documentation [here](https://docs.python.org/3/library/typing.html#annotating-generators-and-coroutines) suggests:

> Simple generators that only ever yield values can also be annotated as having a return type of either [Iterable[YieldType]](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable) or [Iterator[YieldType]](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterator):
> ```python
> def infinite_stream(start: int) -> Iterator[int]:
>     while True:
>         yield start
>         start += 1
> ```

which to me at least, seems to match the use case here, but happy to hear your thoughts

schlunma: If typing suggests that, then I guess it's fine. Naively, I would have gone with the lowest class in the type hierarchy (`Generator` [inherits](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes) from `Iterator`), so that we won't get typing errors when the return type of this is used by an object that expects a `Generator`.

schlunma:
> It may not be as relevant there because starting out from a recipe with wildcards to develop a complete recipe doesn't work for all but the simplest cases because the recipe doesn't allow specifying the relations between datasets that you typically want, e.g. give me all datasets that provide these 3 variables for these two experiments and don't have data quality issues is not something you can easily express in a recipe, but you can do in code.

I would argue that it's certainly relevant to have wildcards in dataset definitions in a recipe also work for derived variables. The meaning of this should be straightforward.

But I agree that this would be something for another PR.

Thanks!!

Issue #3069, opened 2026-05-10
Flaky tests because CMOR tables have not been loaded


bouweandela:
These tests occasionally fail because the CMOR tables have not been loaded before they are run:
```
FAILED tests/unit/preprocessor/_derive/test_soz.py::test_soz_calculate_hybrid_plevs - KeyError: 'plev19'
FAILED tests/unit/preprocessor/_derive/test_toz.py::test_toz_calculate_masked_cubes_hybrid_plevs - KeyError: 'plev19'
FAILED tests/unit/preprocessor/_derive/test_troz.py::test_troz_calculate_hybrid_plevs - KeyError: 'plev19'
FAILED tests/unit/preprocessor/_other/test_other.py::test_align_metadata - ValueError: Unable to convert from 'Unit('celsius')' to 'Unit('unknown')'.
FAILED tests/unit/preprocessor/_other/test_other.py::test_align_metadata_alt_names[sic] - ValueError: Unable to convert from 'Unit('%')' to 'Unit('unknown')'.
FAILED tests/unit/preprocessor/_other/test_other.py::test_align_metadata_alt_names[siconc] - ValueError: Unable to convert from 'Unit('%')' to 'Unit('unknown')'.
FAILED tests/unit/preprocessor/_other/test_other.py::test_align_metadata_invalid_short_name_strict - AssertionError: Regex pattern did not match.
  Expected regex: "Variable 'zzz' not available for table 'Amon' of project 'CMIP6'"
  Actual message: "Unable to convert from 'Unit('1')' to 'Unit('unknown')'."
FAILED tests/unit/preprocessor/_other/test_other.py::test_align_metadata_invalid_short_name_not_strict - ValueError: Unable to convert from 'Unit('1')' to 'Unit('unknown')'.
```

Example runs:
- https://app.circleci.com/pipelines/github/ESMValGroup/ESMValCore/14951/workflows/445e2fc0-5fa3-4d21-84bc-da299a18b623/jobs/60905/tests
- https://app.circleci.com/pipelines/github/ESMValGroup/ESMValCore/14952/workflows/0bb0ef3a-f12c-40ce-a265-e1c8b839360f/jobs/60908/tests

bouweandela:
We do automatically set up the default configuration for the tests: https://github.com/ESMValGroup/ESMValCore/blob/e232092d59d468ab8053b81cbe0ef7f1629486e3/tests/conftest.py#L28-L55

[...]

PR #3084, opened 2026-05-19
[DNM] Investigate CircleCI Flaky tests

valeriupredoi:
As reported by @bouweandela in #3069 

- my take is there is a caching issue on CircleCI since we don't see those tests fail on GHA - but alas, the test with upstream dev pkgs is not run there (on GHA)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,6 +118,7 @@ workflows:
   commit:
     jobs:
       - test
+      - test_with_upstream_developments
 
   nightly:
     triggers:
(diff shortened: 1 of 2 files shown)


valeriupredoi:
hey @bouweandela - it IS a caching issue on Circle, I ran this here a bazillion times (OK real runs: 13-14 maybe 15 I reckon), and no flakes. Problem is, the caching issue might have been sorted out on Circle in the meantime (maybe just a bad SSD or configuration), so if we adopt this one we are taking a hit in runtime (about 20-30%) with no reason. What do you reckon?

valeriupredoi:
patched by @bouweandela in https://github.com/ESMValGroup/ESMValCore/pull/3086

Issue #3117, opened 2026-06-22
Tests failing - github actions pre-commit, mypy - error: "TypeAliasType" not callable


flicj191:
Tests have started failing on 22 June with:
```
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

esmvalcore/preprocessor/_time.py:1773: error: "TypeAliasType" not callable  [operator]
tests/unit/preprocessor/test_shared.py:442: error: "TypeAliasType" not callable  [operator]
tests/unit/preprocessor/test_shared.py:459: error: "TypeAliasType" not callable  [operator]
Found 3 errors in 2 files (checked 625 source files)
```
see [here](https://github.com/ESMValGroup/ESMValCore/actions/runs/27923711194)
for [run-tests](https://github.com/ESMValGroup/ESMValCore/actions/workflows/run-tests.yml)

For `np.ma.masked_array` in affected files, additional dependencies for `mypy` have `numpy`

I can update the identified files.

bouweandela:
This issue should be solved in the next numpy release https://github.com/numpy/numpy/issues/31737

PR #3124, opened 2026-06-22
Use mypy and ruff from pixi environment

bouweandela:
## Description

Use mypy and ruff from the pixi environment instead of installing them with pre-commit.

## Pros

- More accurate mypy checks because all dependencies are installed
- Manage only one environment (pixi) instead of two (pixi and pre-commit)
- No unexpected test breaking because pre-commit does not use lock files for its environments as in #3117
- `mypy` and `ruff` can be used without pre-commit. Useful if you want to run these tools with additional options, e.g. `ruff check --fix --unsafe-fixes`

## Cons

- Does not work with pre-commit.ci
- Runs slower on HPC, I suspect this is because mypy is accessing more files

We could decide to move away from pre-commit.ci and instead just run the pre-commit hooks on CircleCI as part of the test suite. That would also have the advantage that it has one less CI service, so easier to understand for new contributors.

Link to documentation: https://esmvaltool--3124.org.readthedocs.build/projects/ESMValCore/en/3124/contributing.html#code-quality

***

***


valeriupredoi (approved):
thanks bud - a very welcome approach indeed! Approving preemptively (don't pull a Klaus on me since I approve drafts 😆 ) since I'll be away most next week

valeriupredoi:
gotta be a solution to have pre-commit.ci work - let me dig around for a bit tomorrow

valeriupredoi:
hey @bouweandela we has pixi in the pre-commit CI environment! Unfortunately that pixi is newer than the pixi we have in the box, so it says we should `pixi migrate` - I'll leave it to you if you want to accept this solution, or to revert to the non-pixi in CI env way. It took me a while to get to this, the pre-commit CI funcs are not nicely documented, but I think we have a good workaround now :beer: Enjoy your holidays :beach_umbrella:

bouweandela:
Thanks for having a go V! It looks like this installs the unrelated [pixi](https://pypi.org/project/pixi/) package from PyPI. While it is probably possible to update the pre-commit hooks so they silently install pixi if it is not available locally by writing some script, I'm not convinced this would be doing our users a service. It may be more clear to just fail with the message that pixi is not installed, as they'll need to install it anyway in order to run ESMValTool.

.github/workflows/zizmor.yml
schlunma, line 1: Just out of curiosity: what is the advantage of having zizmor run via GH actions vs. pre-commit? As far as I can tell, zizmor does not require installing ESMValTool, does it?

schlunma (commented):
Thanks Bouwe, I like the idea!

Tested this locally and on Levante.

#### Locally

First run: ~38s
Subsequent runs: ~4s

#### Levante

First run: ~2.25 min
Subsequent runs: ~26s

Locally, that waiting time is totally fine, but on Levante it feels very slow. I am not 100% sure if this is still acceptable or not. Not sure if this is related to the current maintenance though, but as far as I am aware `/home` should not be affected by it.

Do you know how long the cache is valid when running `pre-commit` manually? As far as I can tell, on CircleCI you configured it so that it's valid for a week.

.github/dependabot.yml
```diff
@@ -14,6 +14,24 @@ updates:
...
+      interval: "weekly"
+    cooldown:
+      default-days: 7
+    groups:
+      pre-commit:
+        patterns:
+          - "*"
     labels:
       - dependencies
       - github_actions
```
schlunma: Not sure if this label is correct here

.github/workflows/zizmor.yml
bouweandela: Most of our contributors never make modifications to the GitHub Actions files, so it's faster to only run this when relevant instead of on everyone's computer at every commit.

.github/dependabot.yml
bouweandela:
```suggestion
```

bouweandela: Removed in cfa1546

bouweandela:
> Do you know how long the cache is valid when running pre-commit manually? As far as I can tell, on CircleCI you configured it so that it's valid for a week.

With the current setup, pre-commit creates its own virtual environments and those are stored in `~/.cache/pre-commit/`. I'm not entirely sure how pre-commit computes the cache keys, but I imagine it's based on the package name and version number from .pre-commit-config.yaml. With the changes here, the pixi environment will be used, which needs to be installed only when pixi.lock changed w.r.t. the previous run. There is also a `.mypy_cache` directory, which contains cached mypy stuff, I'm not sure how that is invalidated.

.github/dependabot.yml
```diff
@@ -14,7 +14,23 @@ updates:
       interval: "weekly"
     cooldown:
       default-days: 7
+    groups:
+      gha:
```
schlunma:
```suggestion
      github-actions:
```

Just to be a little bit more verbose; this string will appear in the PR title.

schlunma (commented):
I am pretty sure that this is related to mypy, which is by far the slowest step.

bouweandela:
I agree, it takes about 18 seconds to run mypy on Levante. There is the option of using the daemon to speed things up, I enabled that now, but only works for reruns.

We could also consider switching to another type checker, see https://pydevtools.com/handbook/explanation/how-do-mypy-pyright-and-ty-compare/ for an overview. `pyrefly` seems an attractive option, it is a lot faster and has nice additional features such as integration in VSCode, so the type hint errors you see there are the same as what you see on the command line, and a command line tool to automatically add type hints.

schlunma:
Yes, pyrefly looks very nice! I think we should give it a try.

bouweandela:
@schlunma I added a `pyrefly` configuration and started fixing some of the issues it reported. However, it does seem to struggle a bit with dask. What do you reckon? Do we accept that it is noisy and produces quite a few false positives? Or stick with mypy?

Issue #2984, opened 2026-02-18
Update required input variables for CMIP7 derived variables


bouweandela:
Derived variables specify which input variables they need for the derivation. In many cases this will work just fine for CMIP7, but in some cases, this will need to be updated.

For example, the CMIP7 project may not be one of the available options:
https://github.com/ESMValGroup/ESMValCore/blob/da81d5f67158f3d2603831b56ab6b4fb8a388d86/esmvalcore/preprocessor/_derive/amoc.py#L15-L26

Or the `mip` needs to be updated:
https://github.com/ESMValGroup/ESMValCore/blob/1e53ae95608441ce6a7e36abe6c2b1bec6112eff/esmvalcore/preprocessor/_derive/et.py#L16-L18
(would need to be `atmos` for CMIP7)

Or if, if some of the input variables have a different branding suffix, a special entry for CMIP7 with a branding suffix may need to be added:
https://github.com/ESMValGroup/ESMValCore/blob/1e53ae95608441ce6a7e36abe6c2b1bec6112eff/esmvalcore/preprocessor/_derive/vegfrac.py#L25-L34
(would need to be `sftlf: Facets = {"short_name": "sftlf", "branding_suffix": "ti-u-hxy-u", "mip": "atmos"}` for CMIP7)

PR #3160, opened 2026-07-10
CMIP6Plus support for ESGF download

ghossh:
## Description
This pull request adds initial support for the CMIP6Plus project in ESMValCore. CMIP6Plus is the WCRP project used for climate model data published on ESGF after the formal closure of CMIP6. This PR enables ESMValCore to recognise CMIP6Plus as a project and to search/download data from ESGF where supported.

Changes in this pull request:
- Add CMIP6Plus to the ESGF facet configuration
- Add CMIP6Plus to the example ESGF data source configuration

Further CMIP6Plus-related improvements may be added in follow-up commits on
this branch.

Documentation Link:
https://esmvaltool--3160.org.readthedocs.build/projects/ESMValCore/en/3160/reference/facets.html#cmip6plus


bouweandela (changes requested):
Great addition @ghossh! It would be nice to add a bit more before merging:
- Add a documentation page, probably a copy of or addition to https://docs.esmvaltool.org/projects/ESMValCore/en/latest/reference/facets.html#cmip6
- Enable CMOR checks, here is the example for CMIP6: https://github.com/ESMValGroup/ESMValCore/blob/1acf6029efe8a3d20657da6e5f005ca4698e022b/esmvalcore/config/configurations/defaults/cmor_tables.yml#L10-L15. Are there special tables for CMIP6Plus or do they just use the CMIP6 CMOR tables?
- Add an entry to the configuration file for intake-esgf, as we recommend that people switch over to that because it supports the new ESGF search interface. It can be a copy of https://github.com/ESMValGroup/ESMValCore/blob/1acf6029efe8a3d20657da6e5f005ca4698e022b/esmvalcore/config/configurations/data-intake-esgf.yml#L3-L16
- Add an entry to the local data source so people have a good example of how to organize the directory structure according to the official DRS, can be a copy of https://github.com/ESMValGroup/ESMValCore/blob/1acf6029efe8a3d20657da6e5f005ca4698e022b/esmvalcore/config/configurations/data-local.yml#L11-L17

It is also good to be aware that derived variables may need additional work in some cases, just like for CMIP7: #2984

bouweandela:
Based on the information at https://wcrp-cmip.org/cmip-phases/cmip6plus/, it looks like the CMIP6Plus CMOR tables live at https://github.com/PCMDI/mip-cmor-tables/tree/main/Tables and the CV file would need to be copied from https://github.com/WCRP-CMIP/CMIP6Plus_CVs/blob/main/CVs/CMIP6Plus_CV.json

bouweandela (commented):
Thanks, looks almost complete now. Could you please add a `README.md` file in the directory `esmvalcore/cmor/tables/cmip6plus` in which you mention what version of the tables is copied there? e.g. like this: https://github.com/ESMValGroup/ESMValCore/blob/config_CMIP6Plus/esmvalcore/cmor/tables/cmip7/README.md

bouweandela (approved):
Thanks!

PR #3120, opened 2026-06-22
Release 2.15 draft release notes

flicj191:
add release notes to changelog. and citation version

Closes #

Link to documentation:

***

## [Before you get started](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#getting-started)

- [ ] [☝ Create an issue](https://github.com/ESMValGroup/ESMValCore/issues) to discuss what you are going to do

## [Checklist](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#checklist-for-pull-requests)

It is the responsibility of the author to make sure the pull request is ready to review. The icons indicate whether the item will be subject to the [🛠 Technical][1] or [🧪 Scientific][2] review.

[1]: https://docs.esmvaltool.org/en/latest/community/review.html#technical-review
[2]: https://docs.esmvaltool.org/en/latest/community/review.html#scientific-review

- [ ] [🧪][2] The new functionality is [relevant and scientifically sound](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#scientific-relevance)
- [ ] [🛠][1] This pull request has a [descriptive title and labels](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-title-and-label)
- [ ] [🛠][1] Code is written according to the [code quality guidelines](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#code-quality)
- [ ] [🧪][2] and [🛠][1] [Documentation](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#documentation) is available
- [ ] [🛠][1] [Unit tests](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#tests) have been added
- [ ] [🛠][1] Changes are [backward compatible](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#backward-compatibility)
- [ ] [🛠][1] Any changed [dependencies have been added or removed](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#dependencies) correctly
- [ ] [🛠][1] The [list of authors](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#list-of-authors) is up to date
- [ ] [🛠][1] All [checks below this pull request](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/contributing.html#pull-request-checks) were successful

***

To help with the number pull requests:

[...]

PR #3171, opened 2026-07-16
Merge tags from `v2.15.x` to `main` after 2.15.0 release

flicj191:
Merge tags from `v2.15.x` to `main` after 2.15.0 release
- Don't have permissions to push for [step in docs](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/how-to/how-to-make-a-release.html#mark-the-release-in-the-main-branch)
- commits should already be in `main`


bouweandela:
What is the purpose of this pull request?

> Don't have permissions to push for [step in docs](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/how-to/how-to-make-a-release.html#mark-the-release-in-the-main-branch)

That was already done in #3120. There is no need to do it again.

flicj191:
Ok, the way the docs sounds is that we want the all the tags in `main` as well so I thought the final one `v2.15.0` should be in there even if we said we didn't need the other release candidate tags. That in #3120 just has the first release candidate tag and not the others. But happy to close this as not needed.

bouweandela:
Thank you for explaining. The only reason we want a tag v2.15.something tag that points to a commit in `main` is to get the automatic versioning with [setuptools-scm](https://setuptools-scm.readthedocs.io) right. The first release candidate is enough for that. I will update the docs on making a release sometime soon. The Met Office said they would do that after I spent a lot of time explaining the release process to them back when they did the v2.11 release in 2024, but sadly it never happened.

schlunma:
I personally think it wouldn't be the worst idea to ensure that all tags are in the main branch. For example, this would ensure that tools like [`git describe`](https://git-scm.com/docs/git-describe) actually work. But that's probably just a matter of taste.

flicj191:
Thanks! I'll try to update the docs and send through a PR

PR #3195, opened 2026-08-11
add GHA workflow that runs tests and coverage only on fork PRs

### LINKED ISSUE #3181: Codecov is not run in forks
Looks like Codecov is not run in forks, e.g., https://github.com/ESMValGroup/ESMValCore/pull/3178. This might be problematic since it could allow pushing code without tests.

I didn't find a fix for this unfortunately after briefly searching the web.

valeriupredoi:
## Description

Closes #3181 

This is in response to @schlunma 's suggestion to have coverage reporting for forked PRs - which is in essence a very useful thing to have; given the token limitations described in the issue, this may be the only viable solution on the back of our current infrastructure, and not using some dodgy GH action from some bloke's repo; the only caveat is that this means the developer will have tests running in their fork PR, but I minimized the amount of testing done to only the latest Python. Note that we explicitly turn off tests in forked PRs in our regular GHAs so this is a bit of a U-turn.

Live testing via dummy forked PR https://github.com/ESMValGroup/ESMValCore/pull/3196 and this action behaves as per specs, flagging `_ducky.py` as missing coverage :duck: 

***

***

### DIFF
--- a/.github/workflows/run-tests-fork.yml
+++ b/.github/workflows/run-tests-fork.yml
@@ -15,7 +15,7 @@ defaults:
 
 jobs:
   test:
-    runs-on: ${{ matrix.os }}
+    runs-on: ubuntu-latest
     # run action ONLY on a fork PR
     if: github.event.pull_request.head.repo.full_name != github.repository
     strategy:
--- a/.github/workflows/run-tests-fork.yml
+++ b/.github/workflows/run-tests-fork.yml
@@ -34,7 +34,7 @@ jobs:
       - uses: prefix-dev/setup-pixi@a09b6247153796b190642a2b53fac4241043cf6f # v0.10.0
         with:
           frozen: true
-          environments: ${{ matrix.environment }}
+          environments: test-py314
       - name: Check dependencies are compatible
         run: pip check
       - name: Run tests
--- a/.github/workflows/run-tests-fork.yml
+++ b/.github/workflows/run-tests-fork.yml
@@ -25,7 +25,7 @@ jobs:
           - ubuntu-latest
         environment:
           - test-py314
-    name: ${{ matrix.os }} ${{ matrix.environment }}
+    name: ubuntu-latest test-py314
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
--- a/.github/workflows/run-tests-fork.yml
+++ b/.github/workflows/run-tests-fork.yml
@@ -20,11 +20,6 @@ jobs:
     if: github.event.pull_request.head.repo.full_name != github.repository
     strategy:
       fail-fast: false
-      matrix:
-        os:
-          - ubuntu-latest
-        environment:
-          - test-py314
     name: ubuntu-latest test-py314
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
--- a/.github/workflows/run-tests-fork.yml
+++ b/.github/workflows/run-tests-fork.yml
@@ -1,0 +1,52 @@
+---
+name: Test and Cov Fork PR
+
+permissions: {}
+
+# runs only on a forked PR targeting ESMValCore main
+on:
+  pull_request:
+    branches:
+      - main
+
+defaults:
+  run:
+    shell: pixi run bash -e {0}
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    # run action ONLY on a fork PR
+    if: github.event.pull_request.head.repo.full_name != github.repository
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+          - macos-latest
+        environment:
+          # perhaps add test-r to be complete?
+          - test-py314
+    name: ${{ matrix.os }} ${{ matrix.environment }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+      - uses: prefix-dev/setup-pixi@a09b6247153796b190642a2b53fac4241043cf6f # v0.10.0
+        with:
+          frozen: true
+          environments: ${{ matrix.environment }}
+      - name: Check dependencies are compatible
+        run: pip check
+      - name: Check code quality
+        run: pre-commit run -a
+      - name: Run tests
+        run: pytest -n 2 --durations=10 --cov --cov-branch --cov-report=xml
+      - uses: codecov/codecov-action@v7.0.0
+        with:
+          files: ./coverage.xml
+      - name: Check the command line interface
+        run: |
+          esmvaltool version
+          esmvaltool -- --help
--- a/.github/workflows/run-tests-fork.yml
+++ b/.github/workflows/run-tests-fork.yml
@@ -18,8 +18,6 @@ jobs:
     runs-on: ubuntu-latest
     # run action ONLY on a fork PR
     if: github.event.pull_request.head.repo.full_name != github.repository
-    strategy:
-      fail-fast: false
     name: ubuntu-latest test-py314
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
--- a/.github/workflows/run-tests-fork.yml
+++ b/.github/workflows/run-tests-fork.yml
@@ -23,9 +23,7 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
-          - macos-latest
         environment:
-          # perhaps add test-r to be complete?
           - test-py314
     name: ${{ matrix.os }} ${{ matrix.environment }}
     steps:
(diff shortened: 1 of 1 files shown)


valeriupredoi:
@schlunma you have a full solution here, bud. Though I'd still like this to be approved by @bouweandela also if you don't mind, this is a U-turn from our decision to have minimal or no testing done in forks. :beer:

.github/workflows/run-tests-fork.yml
```diff
@@ -0,0 +1,52 @@
...
+
+defaults:
+  run:
+    shell: pixi run bash -e {0}
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    # run action ONLY on a fork PR
+    if: github.event.pull_request.head.repo.full_name != github.repository
```
schlunma, line 20: Could we also use this workflow for PRs from our repo instead of the Codecov orb on CircleCI? This would make it more consistent

.github/workflows/run-tests-fork.yml
```diff
@@ -0,0 +1,52 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+          - macos-latest
```
schlunma: Which OS is used to evaluate the coverage? Or does it somehow show the union of missing lines for both?

.github/workflows/run-tests-fork.yml
```diff
@@ -0,0 +1,52 @@
...
+          fetch-depth: 0
+          persist-credentials: false
+      - uses: prefix-dev/setup-pixi@a09b6247153796b190642a2b53fac4241043cf6f # v0.10.0
+        with:
+          frozen: true
+          environments: ${{ matrix.environment }}
+      - name: Check dependencies are compatible
+        run: pip check
+      - name: Check code quality
+        run: pre-commit run -a
```
schlunma:
```suggestion
```

We could omit that

.github/workflows/run-tests-fork.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Run tests
+        run: pytest -n 2 --durations=10 --cov --cov-branch --cov-report=xml
+      - uses: codecov/codecov-action@fb8b3582c8e4def4969c97caa2f19720cb33a72f # v7.0.0
+        with:
+          files: ./coverage.xml
+      - name: Check the command line interface
+        run: |
+          esmvaltool version
+          esmvaltool -- --help
```
schlunma:
```suggestion
```

Also not really necessary

schlunma (commented):
Thanks V, that's awesome!! Got a couple of comments, but I think this is the way to go! 🚀

.github/workflows/run-tests-fork.yml
valeriupredoi: yes indeed we could, @bouweandela initially configured Codecov through CircleCI but we can always switch to GHAs instead, I have Codecov running through GHAs in other repos and it's doing very well :smiley:

.github/workflows/run-tests-fork.yml
valeriupredoi: well, it's running the Test and Cov Fork PR GHA workflow under both Linux/Ubuntu and OSX  https://github.com/ESMValGroup/ESMValCore/actions/runs/31499584821/job/93805790387?pr=3196

That is done that way since:
- we don't distribute the regular Test GHA and allow it to run in forks
- we don't know if the proposed change in the fork PR is compatible with OSX

.github/workflows/run-tests-fork.yml
valeriupredoi: well, we could, but then if pre-commit.CI is off for some reason (see what Bouwe is proposing to have it off since it's not using *our* pixi) we'll have no way to check until we merge the forked PR.

This, together with your previous comment about OSX, and the following comment about having an `esmvaltool --help` check on: I made this workflow as complete as possible in terms of the checks we run, since the regular Test GHA is not permitted to run on forks, so it's meant to catch as much as possible before merging the fork PR. On the other hand, one can make the argument that. since we allow Circle CI tests on forks, we should just keep the workflow only to report coverage diffs, so then we can remove all the other checks. A matter of opinion, I should think. But note that we don't run any testing on OSX in Circle.

.github/workflows/run-tests-fork.yml
valeriupredoi: see me comment above, Manu :beer:

.github/workflows/run-tests-fork.yml
schlunma: As far as I understand this, both workflows (Linux + OSX) create a `coverage.xml` file, which is then uploaded to Codecov. Which of those coverage report is used and shown at Codecov? Is this determined randomly? Or somehow combined?

If it's the first, I would rather just run this test in Linux.

.github/workflows/run-tests-fork.yml
schlunma: Yeah, my main point would have been that we run those checks in CircleCI on forks. If we want to have a more or less complete GitHub action test workflow running on forks, we should keep it, but if the aim is just coverage reporting, we can drop it.

.github/workflows/run-tests-fork.yml
valeriupredoi: ah what a good point, totally missed it, thanks, bud! Short answer: the OSX one - OSX one takes longer to complete, and I also looked at the timestamps: 14h-08m for Linux, 14h-09m for OSX.

.github/workflows/run-tests-fork.yml
valeriupredoi: sounds good to me! Then I can also remove the OSX bit as per the comment above, we can deal with any OSX-specific issues once we merged the PR

valeriupredoi:
@schlunma your comments encapsulated in cb36d7b539b3b7940729082b3d45dd4eaea6f983

.github/workflows/run-tests-fork.yml
```diff
@@ -0,0 +1,44 @@
...
+    branches:
+      - main
+
+defaults:
+  run:
+    shell: pixi run bash -e {0}
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
```
schlunma:
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/run-tests-fork.yml
```diff
@@ -0,0 +1,44 @@
...
+    # run action ONLY on a fork PR
+    if: github.event.pull_request.head.repo.full_name != github.repository
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+        environment:
+          - test-py314
```
schlunma:
```suggestion
```

.github/workflows/run-tests-fork.yml
```diff
@@ -0,0 +1,44 @@
...
+    name: ${{ matrix.os }} ${{ matrix.environment }}
```
schlunma:
```suggestion
    name: ubuntu-latest test-py314
```

.github/workflows/run-tests-fork.yml
```diff
@@ -0,0 +1,44 @@
...
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+      - uses: prefix-dev/setup-pixi@a09b6247153796b190642a2b53fac4241043cf6f # v0.10.0
+        with:
+          frozen: true
+          environments: ${{ matrix.environment }}
```
schlunma:
```suggestion
          environments: test-py314
```

schlunma (approved):
Great, thanks V! 🚀 

Just a tiny comment left: there's no need to use a matrix here.

.github/workflows/run-tests-fork.yml
valeriupredoi: what no, we need those, I need to clean up the way they are defined though

schlunma: Don't think we do - no need to define a `matrix` if you don't use it

valeriupredoi: ah yes no - matrix went in the latest commit, fail-fast is needed tho. pre-commit complaining about thw workflow, it's really none of its business :rofl:

schlunma: `fail-fast` can go too, it only applies to workflows that use a matrix: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstrategyfail-fast

Removing this will probably also silence pre-commit.

valeriupredoi: it did! Good man, I hate that son of a gun :grinning:

valeriupredoi:
very many thanks for the review Manu :beers:

valeriupredoi:
BTW I decided to merge now and not ask for Bouwe's review anymore since we decided to make the action very very light, so it's not really a contradiction of our decision to not have stuff run in forks

schlunma:
Didn't you wanna wait for @bouweandela's approval? 😄

valeriupredoi:
> Didn't you wanna wait for @bouweandela's approval? 😄

shh :shushing_face: He don't know about this :grin:

